Issue #215, opened 2014-05-20
KPM and KVM will not work when behind a proxy


glennc:
I've got errors when i use kvm and kpm without proxy

I changed kvm.ps1 for using webProxy

``` powershell
 $wc = New-Object System.Net.WebClient
 $proxy = New-Object System.Net.WebProxy
 $proxy.Address="http://myProxy:8888"
 $proxy.UseDefaultCredentials=true
 $wc.Proxy=$proxy
```

I'm not so familar with Powershell, but it works for me.

But I don't understand how setup proxy settings to packageManager.
kpm restore return

Error: FindPackageById: System.Console
An error occured while sending request

I hope you help me with this problem

PR #324, opened 2014-06-11
Add proxy support in kpm

ChengTian:
- Add proxy option to "kpm restore" subcommand
- "kpm restore" uses global proxy (http_proxy env var) by default

parent #215


sleemer:
@ChengTian in my enviroment it doesn't work.
What I found out is when using UriBuilder we couldn't specify domain user name (for example 'company.com\username:password' or 'username@company.com:password')

ChengTian:
@sleemer , thank you for reporting the problem.

I created an issue at https://github.com/aspnet/DNX/issues/1364. Next time when you want to report any problem, please create an issue at https://github.com/aspnet/DNX/issues. Thank you for your contribution!

Issue #538, opened 2014-08-17
Rename --port flag to --server flag and KRE_COMPILATION_SERVER_PORT to KRE_COMPILATION_SERVER


davidfowl:
Today we support the runtime passing a flag that allows it to get bits from the design time host via the --port flag and KRE_COMPILATION_SERVER_PORT. We should change the names of these things from:

--port to --server
KRE_COMPILATION_SERVER_PORT to KRE_COMPILATION_SERVER

We'll likely experiment/support different transports (tcp/named pipes etc)

muratg:
No

PR #630, opened 2014-09-12
Rename option and env var for DTH port

ChengTian:
- Rename '--port' to '--dth-port'
- Rename 'KRE_COMPILATION_SERVER_PORT' to 'KRE_DTH_SERVER'

parent #538 

@davidfowl @loudej , please help to review at your earliest convenience.


ChengTian:
We don't want to take this. Close dead PR.

Issue #393, opened 2014-06-26
"kpm /?" doesn't print out help for kpm


maumar:
This is a common way for people to ask for help (and works for kvm, for instance). However kpm doesn't accept this format. 

This is what gets printed at the moment:

"Calls one batch program from another.
CALL [drive:][path]filename [batch-parameters]
  batch-parameters   Specifies any command-line information required by the
                     batch program.
If Command Extensions are enabled CALL changes as follows:
CALL command now accepts labels as the target of the CALL.  The syntax
is:
    CALL :label arguments
...
"

ChengTian:
This is not true any more after we reduce scripts in this commit https://github.com/aspnet/DNX/commit/29e6e8a10f4dfff004da9510d4f634070a6d6430

PR #798, opened 2014-10-22
Make sure '/?' doesn't take over control in top level scripts

ChengTian:
parent #393 

Did thorough tests with this fix.


ChengTian:
Not needed any more, discarding

Issue #807, opened 2014-10-23
We should not enumerate files excluded in packExclude


rustd:
We enumerate folders in packExclude. This breaks when we try to enumerate node_modules folder since the file path can be long due to the deep nesting structure of node_modules

Repro steps
1. Create a starter web.
2. Delete the node_module
3. Close the App.
4. Reopen the app to force a restore. This will restore the node_module folder.
5. Do Kpm pack and you will run into the following exception

System.IO.PathTooLongException: The specified path, file name, or both are too long. The fully qualified file name must
be less than 260 characters, and the directory name must be less than 248 characters.

PR #810, opened 2014-10-24
Use non-recursive copy operation to fix path too long exception during "kpm pack"

ChengTian:
parent #807


gamicklisafisher:
For deletion or handling of long name i check that "Long Path Tool" software should be try, many articles refer this for such solutions.

PR #881, opened 2014-11-11
Enable Background GC

alinapopa:
This change enables background GC.
Without this flag klr.exe runs with non-concurrent GC.


davidfowl:
This should be a flag on the exe, not on by default

davidfowl:
We need to enable this via a flag. Closing this for now.

src/klr.core45/klr.core45.cpp
```diff
@@ -390,7 +390,8 @@ extern "C" __declspec(dllexport) bool __stdcall CallApplicationMain(PCALL_APPLIC
     STARTUP_FLAGS dwStartupFlags = (STARTUP_FLAGS)(
         STARTUP_FLAGS::STARTUP_LOADER_OPTIMIZATION_SINGLE_DOMAIN |
         STARTUP_FLAGS::STARTUP_SINGLE_APPDOMAIN |
-        STARTUP_FLAGS::STARTUP_SERVER_GC
+        STARTUP_FLAGS::STARTUP_SERVER_GC |
```
moozzyk: note - this flag is not supported on ARM CoreCLR

Issue #931, opened 2014-11-24
Support getting latest behavior of quirks for binary compatibility


AlexGhiondea:
Enable the hosts to set the right TFM or compatibility flag to enable getting the latest behavior from a binary compatibility perspective.

For the desktop host, please set the TFM when creating the AppDomain that will host the application
For the coreCLR host, please set the 'UseLatestBehaviorWhenTFMMissing' compatibility flag when setting up the host.

PR #953, opened 2014-12-02
Configure KRE to enable the latest behavior for back compat

AlexGhiondea:
- Enable the CoreCLR host to pass the right compat switch
 - Set the .Net 4.6 TFM for the desktop host

#931


AlexGhiondea:
Ping.

davidfowl:
Going to merge this in

AlexGhiondea:
We can test the before/after behavior for a compat path (one that is guarded by BinaryCompatibility). I don't have one in mind right now, but I will look into it.

AlexGhiondea:
Here is a repro you can use to test this change:

https://gist.github.com/AlexGhiondea/8c87590d93616c8b26de

The expected behavior is that we throw an exception.

glennc:
@davidfowl Should we still merge this given our latest discussions?

troydai:
https://github.com/aspnet/dnx/pull/1398

712fb54

Issue #987, opened 2014-12-12
kpm ignores errors of project.json event scripts


cherrydev:
If you set, for example, a "prebuild" script in project.json, and the command fails (returns a non-zero exit code) kpm completely ignores it.  Likewise, if a "postbuild" script fails, "kpm build" still reports "Build Succeeded".

```json
...
"scripts": {
        "prebuild": "foo"
    }
...
```

and then run ```> kpm build``` you will receive the message
```
'foo' is not recognized as an internal or external command,
operable program or batch file.
```

The remainder of the build will then continue.  I've yet to think of reason why this would be the desired behaviour.
Running KRE-CLR-x86.1.0.0-beta1

Praburaj:
Fixed https://github.com/aspnet/KRuntime/pull/1016

PR #1016, opened 2014-12-29
Return the exit code in ScriptExecutor

troydai:
Solving issue #987

Now the exit code from the script in project.json need to be zero to ensure a successful build/pack/restore.


troydai:
Rebased, squashed and verified. Going to merge in a few minutes.

PR #1021, opened 2015-01-02
Capitalize the special folders

Tragetaschen:
As a followup of [MVC#1679](https://github.com/aspnet/Mvc/pull/1679), this PR makes the default patterns for the Compiler/* folders Uppercase.

There is a second commit that simplifies opening the file stream for embedded resources. I can omit/squash if necessary.

Nuget packages also have a 'shared' folder, this was kept as is.

`build.cmd` could not find System.Collections.Immutable, System.Reflection.Metadata as well as the beta5 pieces of xunit, so I could not `verify` the changes.


davidfowl:
ugh

Tragetaschen:
Now, [MVC#2277](https://github.com/aspnet/Mvc/pull/2277) and [MusicStore#410](https://github.com/aspnet/MusicStore/pull/410) introduce the lower-case variant. Can this be closed then?

troydai:
@Tragetaschen we're adopting lower cases search patterns now (https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Framework.Runtime.Hosting/ProjectFilesCollection.cs#L18) 

This can be closed. Thanks,

PR #1022, opened 2015-01-02
kpm pack --native is broken without Microsoft.Framework.Project

shhsu:
We need to include Microsoft.Framework.Project for crossgen.
Also, do we need to fix the .sh file? Is crossgen by any chance supported in Linux?


ChengTian:
Looks good to me.

However, I have no idea whether we want crossgen on *nix. /CC @davidfowl .
One hint is that, if we don't want a feature on *nix, we should hide the option/command with
```c#
if (!PlatformHelper.IsMono)
```

shhsu:
This is an interesting point. I'd like to keep this PR strictly as a bug fix though.
We'll await @davidfowl 's response

shhsu:
Actually this solution doesn't work anymore.
It would cause this line to throw
"System.NotSupportedException: The given path's format is not supported."
https://github.com/aspnet/KRuntime/blob/dev/src/dotnet.host/Bootstrapper.cs#L59

davidfowl:
Yah this is all sorts of busted.

shhsu:
Fixed with the proper syntax.
Tested, it was able to pack the sample web app in the mvc branch

davidfowl:
It's time to get rid of the Project assembly and merge it into kpm.

shhsu:
Project is gone

borgdylan:
On Mono you can implement the --native option by calling Mono with AOT flags like this:

```
mono --aot MyPackage.dll
```

which will create accompanying native image files.

davidfowl:
@shhsu can you rebase this?

shhsu:
So I am still unclear what is the solution we are going with. Do we still going with removing Project and have the feature "kpm  --crossgen" exposed? (kpm --crossgen crossgens kre)

shhsu:
any updates?

davidfowl:
@shhsu Can you file a bug for this? /cc @muratg We'll do this properly in beta6.

PR #1027, opened 2015-01-04
Introduce ICompilerOptionsReader to allow specifying a service that can parse compiler options from a project.json

### LINKED ISSUE #993: Refactor parsing project.json so that parsing of compilation options can be pluggable
Today, the parsing of project.json is entirely managed inside Project.cs which assumes the schema of `compilationOptions` is fixed. This works well for C#\ Roslyn, but does not lend well to custom languages \ project resolvers.

## Proposal
* Introduce `ICompilationOptionsReader` than can parse `compilationOptions` node in project.json. The specific type of this interface is specified via a `compilationOptionsReader` node under the languages section.
* Introduce `IProjectReader` that parses project.json. This type will read the languages node to determine the reader to use for `compilationOptions`.
* Make the `Project` type a POCO type. The `CompilationOptions` property on `Project` would be a marker interface. `ProjectReferenceProvider`s would cast down this property to the appropriate implementation type that was created by `ICompilationOptionsReader`. 

Considerations:
Maybe merge `IProjectReferenceProvider` and `ICompilationOptionsReader` since it's the only consumer? It does seem odd though since one has to parse json and the other does compilation.

pranavkm:
Fixes #993


pranavkm:
I'll update the docs and such once we settle on the design.

src/Microsoft.Framework.PackageManager/Commands/AddCommand.cs
```diff
@@ -33,8 +35,9 @@ public bool ExecuteCommand()
 
             ProjectDir = ProjectDir ?? Directory.GetCurrentDirectory();
 
+            var reader = new ProjectReader(LoadContextAccessor.Instance.Default);
```
davidfowl: Passing in this static doesn't look right. Why not flow it properly?

PR #1031, opened 2015-01-05
Propagate build errors during pack to a proper exit code

Tragetaschen:
When building during packing fails, this is only visible through the log messages---the pack command returns a 0 success code. The end result is a folder with everything including the script files except the actual package folder with actual project.

I was blinkered during these changes, but all in all, the error handling seemed quite messy :-/

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Bundle/BundleProject.cs
+++ b/src/Microsoft.Framework.PackageManager/Bundle/BundleProject.cs
@@ -37,23 +37,27 @@ namespace Microsoft.Framework.PackageManager.Bundle
         public string WwwRoot { get; set; }
         public string WwwRootOut { get; set; }
 
-        public void Emit(BundleRoot root)
+        public bool Emit(BundleRoot root)
         {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Framework.PackageManager/Bundle/BundleProject.cs
```diff
@@ -82,9 +86,11 @@ private void EmitSource(BundleRoot root)
             UpdateWebRoot(root, TargetPath);
 
             _applicationBase = Path.Combine("..", BundleRoot.AppRootName, "src", project.Name);
+
+            return true;
```
victorhurdugaci: Why is this method returning `bool` when it is always returning `true`?

davidfowl:
/cc @ChengTian 

I think we should get this in for beta4 /cc @glennc @muratg

davidfowl:
@Tragetaschen @ChengTian rebased your work for you :smile:.Closing out this PR

muratg:
@davidfowl actually closing this one.

Issue #845, opened 2014-10-31
KPM blows up if there is a nonexistent local package source


halter73:
If you have a local package source active in NuGet.Config pointing to a directory that does not currently exist, any call to `kpm restore` fails with an error similar to the following:

    Loaded Module: klr.core45.dll
    Found DLL Export: CallApplicationMain
    Information: [LoaderContainer]: Load name=Microsoft.Framework.PackageManager
    Information: [PathBasedAssemblyLoader]: Loaded name=Microsoft.Framework.PackageManager in 5ms
    Information: [LoaderContainer]: Load name=Microsoft.Framework.Runtime
    Information: [PathBasedAssemblyLoader]: Loaded name=Microsoft.Framework.Runtime in 0ms
    Information: [LoaderContainer]: Load name=Newtonsoft.Json
    Information: [PathBasedAssemblyLoader]: Loaded name=Newtonsoft.Json in 1ms
    Restoring packages for F:\dev\Universe\SignalR-Server\samples\SignalRSample.Web\project.json
    Information: [LoaderContainer]: Load name=Microsoft.Net.Http.Client
    Information: [PathBasedAssemblyLoader]: Loaded name=Microsoft.Net.Http.Client in 0ms
    ----------
    System.IO.DirectoryNotFoundException: Could not find a part of the path 'F:\dev\SignalR\artifacts\Debug\packages'.
       at System.IO.FileSystemEnumerableIterator`1.CommonInit()
       at System.IO.FileSystemEnumerableIterator`1..ctor(String path, String originalUserPath, String searchPattern, SearchOption searchOption, SearchResultHandler`1 resultHandler)
       at System.IO.Directory.EnumerateFiles(String path, String searchPattern)
       at Microsoft.Framework.PackageManager.PackageFolderFactory.<>c__DisplayClass1.<CreatePackageFolderFromPath>b__2(String dir)
       at Microsoft.Framework.PackageManager.PackageFolderFactory.CreatePackageFolderFromPath(String path, IReport report)
       at Microsoft.Framework.PackageManager.RestoreCommand.AddRemoteProvidersFromSources(List`1 remoteProviders, List`1 effectiveSources)
       at Microsoft.Framework.PackageManager.RestoreCommand.<RestoreForProject>d__1.MoveNext()
    --- End of stack trace from previous location where exception was thrown ---
       at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

Issue #1018, opened 2014-12-29
kpm restore fails to proceed to fallback sources when there is an error in the first listed feed


Praburaj:
_From @itsananderson on May 13, 2014 15:51_

I was walking through the getting started section, and I noticed an exception being thrown from the `kpm restore` step.

```
Attempting to resolve dependency System.Console >= 4.0.0.0
  OK https://my-feed.cloudapp.net/api/v2/FindPackagesById()?Id='System.Console' 448ms
Warning: FindPackagesById: System.Console
  '=' is an unexpected token. The expected token is ';'. Line 21, position 78.
```

(I've sanitized the actual feed URL, but I'm a Microsoft employee if that helps)

My guess is that one of three things is happening:

1. kpm doesn't know how to negotiate authentication with the private NuGet feed
1. The private feed needs an update to work with kpm, and kpm doesn't gracefully handle communicating with old feeds
1. The feed simply doesn't contain the "System.Console" package, and kpm doesn't gracefully handle missing packages.

Other details:

KRE builds tested (both reproduced issue)

```
0.1-alpha-build-0421 svr50   x86          C:\Users\me\.kre\packages
0.1-alpha-build-0425 svr50   x86          C:\Users\me\.kre\packages
```

Pretty sure these aren't relevant, but:

* Windows 8.1 update
* IIS 8.5

Here's  a full dump of console output, and the actual response I get from the NuGet feed when I run the `Id='System.Console'` query in the browser: https://gist.github.com/itsananderson/3f6c8ec3fb1bb8699a91

_Copied from original issue: aspnet/Home#15_

davidfowl:
That's by design. Use --ignore-failed-sources to ignore sources that may not be reachable or failing etc

PR #1032, opened 2015-01-05
"kpm restore" ignores nonexistent local package feeds and gives warnings

ChengTian:
parent #845 

`kpm restore` gives warning instead of blowing up now.
untitled

Also extracted some shared code to have DRYer code.

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Restore/RemoteWalkProvider.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/RemoteWalkProvider.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
@@ -16,10 +17,10 @@ namespace Microsoft.Framework.PackageManager
     {
         private readonly IPackageFeed _source;
 
-        public RemoteWalkProvider(IPackageFeed source, bool isHttp)
+        public RemoteWalkProvider(IPackageFeed source)
         {
             _source = source;
-            IsHttp = isHttp;
+            IsHttp = IsHttpSource(source);
         }
 
         public bool IsHttp { get; private set; }
@@ -86,6 +87,12 @@ namespace Microsoft.Framework.PackageManager
             {
                 await nupkgStream.CopyToAsync(stream);
             }
+        }
+
+        private static bool IsHttpSource(IPackageFeed source)
+        {
+            return source.Source.StartsWith("http://", StringComparison.OrdinalIgnoreCase) ||
+                source.Source.StartsWith("https://", StringComparison.OrdinalIgnoreCase);
         }
     }
 }
--- a/src/Microsoft.Framework.PackageManager/Commands/InstallCommand.cs
+++ b/src/Microsoft.Framework.PackageManager/Commands/InstallCommand.cs
@@ -55,15 +55,14 @@ namespace Microsoft.Framework.PackageManager
             var packageFeeds = new List<IPackageFeed>();
             foreach (var source in effectiveSources)
             {
-                if (new Uri(source.Source).IsFile)
+                var feed = PackageSourceUtils.CreatePackageFeed(
+                    source,
+                    _restoreCommand.NoCache,
+                    _restoreCommand.IgnoreFailedSources,
+                    Reports);
+                if (feed != null)
                 {
-                    packageFeeds.Add(PackageFolderFactory.CreatePackageFolderFromPath(source.Source, Reports.Quiet));
-                }
-                else
-                {
-                    packageFeeds.Add(new NuGetv2Feed(
-                        source.Source, source.UserName, source.Password, _restoreCommand.NoCache, Reports,
-                        _restoreCommand.IgnoreFailedSources));
+                    packageFeeds.Add(feed);
                 }
             }
 
--- a/src/Microsoft.Framework.PackageManager/Restore/NuGet/IPackageFeed.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/NuGet/IPackageFeed.cs
@@ -10,6 +10,7 @@ namespace Microsoft.Framework.PackageManager.Restore.NuGet
 {
     public interface IPackageFeed
     {
+        string Source { get; }
         Task<IEnumerable<PackageInfo>> FindPackagesByIdAsync(string id);
         Task<Stream> OpenNupkgStreamAsync(PackageInfo package);
         Task<Stream> OpenNuspecStreamAsync(PackageInfo package);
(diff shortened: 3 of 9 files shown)


src/Microsoft.Framework.PackageManager/Restore/RemoteWalkProvider.cs
```diff
@@ -16,10 +16,10 @@ public class RemoteWalkProvider : IWalkProvider
     {
         private readonly IPackageFeed _source;
 
-        public RemoteWalkProvider(IPackageFeed source, bool isHttp)
+        public RemoteWalkProvider(IPackageFeed source)
         {
             _source = source;
-            IsHttp = isHttp;
+            IsHttp = IsHttpSource(source);
```
ChengTian: Moving initialization of `IsHttp` here can simplify extraction of shared code (`PackageSourceUtils.CreatePackageFeed()`). Moreover, the `source` argument already contains `IsHttp` information, so passing `isHttp` introduces duplicated info.

src/Microsoft.Framework.PackageManager/Restore/RemoteWalkProvider.cs
```diff
@@ -87,6 +87,11 @@ public RemoteWalkProvider(IPackageFeed source, bool isHttp)
                 await nupkgStream.CopyToAsync(stream);
             }
         }
+
+        private static bool IsHttpSource(IPackageFeed source)
+        {
+            return (source is NuGetv2Feed);
```
davidfowl: This is going to need to change when we support v3 feeds

src/klr.core45/tpa.cpp
```diff
@@ -6,78 +6,82 @@
 
 BOOL CreateTpaBase(LPWSTR** ppNames, size_t* pcNames, bool bNative)
 {
-    const size_t count = 32;
+    const size_t count = 34;
```
davidfowl: Why is this part of this change? Can somebody check this in separately?

src/Microsoft.Framework.PackageManager/Restore/RemoteWalkProvider.cs
ChengTian: Yeah. I'll change it to `return (source is NuGetv2Feed) || (source is NuGetv3Feed)` when we have that change. Will add TODO comment here.

davidfowl: Maybe IPackageFeed should have a similar property?

```C#
public interface IPackageFeed
{
     bool IsHttp { get; }
     ...
}
```

src/klr.core45/tpa.cpp
ChengTian: I didn't add any new reference in this change but `tpa.cpp` is modified by build script. I can put this change in a separate commit. cc @troydai to confirm there is nothing wrong with this part.

troydai: @ChengTian Yes, you can check this file separately. The dpa running in the second round of the build and find these new dependencies for klr.core45.managed. It may be introduced by you but by CoreCLR.

src/Microsoft.Framework.PackageManager/Restore/RemoteWalkProvider.cs
ChengTian: Nice suggestion! Followed the second approach.

src/klr.core45/tpa.cpp
ChengTian: Thanks for the clarification. The update of TPA list is checked in with a separate commit.

ChengTian:
@davidfowl , thanks for the feedbacks. Please check the latest commits.

Praburaj:
Is this PR related to this bug? https://github.com/aspnet/KRuntime/issues/1018 ?

Praburaj:
The issue addressed with this PR and the bug #1018 seems to be similar/same except the difference of local or remote source. If `--ignore-failed-sources` needs to be used for `http` sources to ignore on failure it will be good to have the same behavior for local sources.

ChengTian:
Talked to @Praburaj and here is a summary:
If we take the current PR, local feeds and remote feeds will have inconsistent behaviors.
Local feeds: if not accessible, always show warning and skip it
Remote feeds: if not accessible, `kpm restore` exits with errors; if not accessible but `--ignore-failed-sources` is given, show warning and skip it

I think @Praburaj is right. It's ideal if local feeds and remote feeds have consistent behaviors. However, with a nonexistent folder, we cannot determine type of the local feed (`KpmPackageFolder` or `NuGetPackageFolder`) so we cannot proceed to the next step of `kpm restore`. To have the same behavior as remote feeds, we can create a class `NonexistentPackageFolder : IPackageFeed`, in which `FindPackagesByIdAsync` throws exception when ``--ignore-failed-sources` is not given.

src/klr.core45/tpa.cpp
davidfowl: Does this PR need an update then?

ChengTian: Removed the TPA list change in this PR

ChengTian:
Made behaviors of local feeds more consistent with remote feeds:
* When `kpm restore` from nonexistent local feed:
capture2
capture3


* When `kpm restore --ignore-failed-sources` from nonexistent local feed, we give a warning and ignore the failed feed:
capture1

Praburaj:
Can we highlight the warning message with a different color code. The message looks buried a bit.

ChengTian:
@Praburaj , the output above is using the same color code as remote feed error messages. If we want to change the color code, we should change both local and remote feeds in a separate PR. Please create an issue for it if you feel strongly. Thanks.

davidfowl:
:shipit:

src/Microsoft.Framework.PackageManager/Restore/NuGet/NonexistentPackageFolder.cs
```diff
@@ -0,0 +1,65 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Microsoft.Framework.PackageManager.Restore.NuGet
+{
+    public class NonExistentPackageFolder : IPackageFeed
```
davidfowl: I don't like this ehhhh

ChengTian: If the structure of a folder doesn't conform with `NuGetPackageFolder`, we always assume it is a `KpmPackageFolder`. So I remove `NonExistentPackageFolder` and move the logic to `KpmPackageFolder`. Do you like this approach? https://github.com/aspnet/KRuntime/commit/b34cda0ac6924c8e911133bb6198bd36fa89b855

Praburaj:
@ChengTian sure I filed https://github.com/aspnet/KRuntime/issues/1051

davidfowl:
I guess it's fine, :shipit:

PR #1034, opened 2015-01-06
Fix build break

troydai:


pranavkm:
:shipit:

troydai:
Will push in seconds.

Issue #1035, opened 2015-01-06
Term change in kpm: pack -> bundle, build -> pack


ChengTian:
Reactions to #1037 

Impacted elements:
1. In `project.json`: `packExclude`, `prepack`, `postpack`, `prebuild`, `postbuild`
2. In KoreBuild: `k-build` -> `k-pack`

ChengTian:
Will do all these change at once under the issue: #967

Issue #967, opened 2014-12-05
Rename kpm commands


davidfowl:
kpm pack should produce nuget packages
kpm build should produce assemblies
kpm bundle should produce a deployable bundle

PR #1037, opened 2015-01-06
Rename kpm commands

ChengTian:
parent #967 

Follow-up reaction: #1035

Impacted units: Tooling, test scripts, build scripts, wiki pages


Praburaj:
Just a note : https://github.com/aspnet/MusicStore/blob/dev/test/E2ETests/Common/DeploymentUtility.cs#L313 needs to change as well with this change.

src/Microsoft.Framework.PackageManager/Program.cs
```diff
@@ -142,9 +142,9 @@ public int Main(string[] args)
                 });
             });
 
-            app.Command("build", c =>
+            app.Command("pack", c =>
             {
-                c.Description = "Build NuGet packages for the project in given directory";
+                c.Description = "Pack NuGet packages for the project in given directory";
```
davidfowl: Build is still a better verb to use.

ChengTian:
Changed script names in the latest commit:
packExclude -> bundleExclude
prepack -> prebundle
postpack -> postbundle
prebuild -> prepack
postbuild -> postpack

We also have `prebuild` and `postbuild` that serves for assembly-only scenarios.

src/Microsoft.Framework.PackageManager/Program.cs
```diff
@@ -142,7 +142,7 @@ public int Main(string[] args)
                 });
             });
 
-            app.Command("build", c =>
+            app.Command("pack", c =>
             {
                 c.Description = "Build NuGet packages for the project in given directory";
```
dougbu: ? "Produce NuGet packages..."

dougbu:
:shipit: from my perspective but I did not check thoroughly enough to be the primary reviewer.

will this lead to additional changes in the tooling schema?  other (new) downstream changes?

ChengTian:
Need a rebase after https://github.com/aspnet/XRE/pull/1117 is checked in.

ChengTian:
We will do this in release branch. Please check the new PR: #1124

Issue #563, opened 2014-08-22
Use AppDomain.ApplyPolicy(assemblyName) before loading assemblies on CLR


davidfowl:
Today we use this hack to load retarget able assemblies:

https://github.com/aspnet/KRuntime/blob/4fbf1fbfff08e131bb4d5cd737f1f4be75702996/src/klr.hosting.shared/RuntimeBootstrapper.cs#L174

http://msdn.microsoft.com/en-us/library/system.appdomain.applypolicy(v=vs.110).aspx

We're supposed to apply this policy before calling into the loaders.

troydai:
Fixed in https://github.com/aspnet/KRuntime/commit/bb3cdbdb67ca04366fcb9a04e56d0dc05752debb

PR #1038, opened 2015-01-06
Use AppDomain.ApplyPolicy before loading assemblies on CLR (#563)

troydai:
#563

### DIFF
--- a/src/klr.hosting.shared/RuntimeBootstrapper.cs
+++ b/src/klr.hosting.shared/RuntimeBootstrapper.cs
@@ -177,13 +177,14 @@ namespace klr.hosting
 
             ResolveEventHandler handler = (sender, a) =>
             {
-                // Special case for retargetable assemblies on desktop
-                if (a.Name.EndsWith("Retargetable=Yes"))
+                var appDomain = (AppDomain)sender;
+                var afterPolicy = appDomain.ApplyPolicy(a.Name);
+                if (afterPolicy != a.Name)
                 {
-                    return Assembly.Load(a.Name);
+                    return Assembly.Load(afterPolicy);
                 }
 
-                return loaderCallback(new AssemblyName(a.Name));
+                return loaderCallback(new AssemblyName(afterPolicy));
             };
 
             AppDomain.CurrentDomain.AssemblyResolve += handler;


troydai:
@davidfowl

davidfowl:
Can you verify something that was broken before and fixed after to make sure this is actually effective?

troydai:
@davidfowl sure, we should do that. can you point me to one example? I can work out a scenario of loading retargetable assembly but not sure if it covers the actual issue.

src/klr.hosting.shared/RuntimeBootstrapper.cs
```diff
@@ -180,7 +180,8 @@ public static Task<int> ExecuteAsync(string[] args)
                 // Special case for retargetable assemblies on desktop
                 if (a.Name.EndsWith("Retargetable=Yes"))
                 {
-                    return Assembly.Load(a.Name);
+                    var appDomain = (AppDomain)sender;
+                    return Assembly.Load(appDomain.ApplyPolicy(a.Name));
                 }
```
troydai: I think we should remove the `Assembly.Load` here.

I can't find a case the special treatment of the `Retargetable` is needed. Both Autofac 3.3.0 and 3.5.2 seems to work fine. And their assemblies are not retargetable according to ILSpy and their assembly full name.

And the logic here will cause stack overflow. Invoke `Assembly.Load` in `AssemblyResolve` event is not recommended according to [MSDN] (http://msdn.microsoft.com/en-us/library/ff527268(v=vs.110).aspx). And in reality if a library is built to be retargetable the stack overflow will happen because there is no chance the code reaches `loaderCallback` in next lines.

Should we have issue with `retargetable` assemblies, we need to find another way to solve.

troydai: @davidfowl

davidfowl: Sure, as long as we don't regress the existing behavior I'm fine with that.

troydai:
:watch: @davidfowl 

Verified, tested on Autofac and "retargetable" class library.

davidfowl:
:shipit:

Issue #750, opened 2014-10-13
Desktop CLR silently fails if .NET 451 is missing


victorhurdugaci:
I had a Win7 machine that doesn't have .NET 45. It only had 35. 
When I run klr, it silently fails and there is no indication about what is wrong.

Could we detect that .NET is missing and give an error message?

troydai:
rebase on latest dev

troydai:
Fixed in a30e304

PR #1044, opened 2015-01-07
Print error message when it fails to initialize CLR runtime (#750)

troydai:
Addressing issue #750 

After the fixing the output is:
```
C:\Users\trdai>klr
Failed to initialize runtime (-2147467263).
```

### DIFF
--- a/src/klr.net45/klr.net45.cpp
+++ b/src/klr.net45/klr.net45.cpp
@@ -8,20 +8,25 @@
 
 IKatanaManagerPtr g_katanaManager;
 
-
 extern "C" __declspec(dllexport) HRESULT __stdcall CallApplicationMain(PCALL_APPLICATION_MAIN_DATA data)
 {
     HRESULT hr = S_OK;
-    
+
     IKatanaManagerPtr manager = new ComObject<KatanaManager, IKatanaManager>();
 
     g_katanaManager = manager;
-    _HR(manager->InitializeRuntime(data->applicationBase));
-    g_katanaManager = NULL;
 
-    _HR(manager->CallApplicationMain(data->argc, data->argv));
-
-    data->exitcode = hr;
+    hr = manager->InitializeRuntime(data->applicationBase);
+    if (SUCCEEDED(hr))
+    {
+        g_katanaManager = NULL;
+        data->exitcode = manager->CallApplicationMain(data->argc, data->argv);
+    }
+    else 
+    {
+        printf_s("Failed to initalize runtime (%x)", hr);
+        data->exitcode = hr;
+    }
 
     return hr;
 }
--- a/src/klr.core45/klr.core45.cpp
+++ b/src/klr.core45/klr.core45.cpp
@@ -498,7 +498,7 @@ extern "C" __declspec(dllexport) HRESULT __stdcall CallApplicationMain(PCALL_APP
     {
         wprintf_s(L"TPA      %d %S\n", wcslen(pwszTrustedPlatformAssemblies), pwszTrustedPlatformAssemblies);
         wprintf_s(L"AppPaths %S\n", wszAppPaths);
-        printf_s("Failed to create app domain (%d).\n", hr);
+        printf_s("Failed to create app domain (%x).\n", hr);
         return hr;
     }
 
@@ -513,7 +513,7 @@ extern "C" __declspec(dllexport) HRESULT __stdcall CallApplicationMain(PCALL_APP
 
     if (FAILED(hr))
     {
-        printf_s("Failed to create main delegate (%d).\n", hr);
+        printf_s("Failed to create main delegate (%x).\n", hr);
         return hr;
     }
 


troydai:
Do we need to clean up this piece of code as well?

https://github.com/aspnet/KRuntime/blob/dev/src/klr.net45/stdafx.h#L40

src/klr.net45/klr.net45.cpp
```diff
@@ -16,13 +16,24 @@ extern "C" __declspec(dllexport) HRESULT __stdcall CallApplicationMain(PCALL_APP
...
 
-    data->exitcode = hr;
+        return hr;
+    }
 
+    g_katanaManager = NULL;
+    hr = manager->CallApplicationMain(data->argc, data->argv);
+    if (FAILED(hr))
+    {
+        printf_s("Application returns result indicating failure.");
```
Praburaj: why not print the return code here as well?

Praburaj: And can this be `Application` => `CallApplicationMain`?

troydai: For question regarding why not print the return code here:
The message of result status will be handle by outer codes.

For question regarding the message:
Not sure if we want to expose the function name in the error message here. The name may change.

Praburaj: KRE traces print the function name any way. So it's easier to correlate than a generic error in my opinion.

troydai: @davidfowl @Eilon @muratg  We need some guideline in terms of how much information we disclose in the error message.

src/klr.net45/klr.net45.cpp
```diff
@@ -8,20 +8,25 @@
...
 
-    data->exitcode = hr;
+    hr = manager->InitializeRuntime(data->applicationBase);
+    if (SUCCEEDED(hr))
+    {
+        g_katanaManager = NULL;
+        data->exitcode = manager->CallApplicationMain(data->argc, data->argv);
+    }
+    else 
+    {
```
troydai: When the `hr` indicates failure here, it means the runtime initialization failed. Application is not executed at all.

src/klr.net45/klr.net45.cpp
```diff
@@ -8,20 +8,25 @@
...
-    g_katanaManager = NULL;
-
-    _HR(manager->CallApplicationMain(data->argc, data->argv));
 
-    data->exitcode = hr;
+    hr = manager->InitializeRuntime(data->applicationBase);
+    if (SUCCEEDED(hr))
+    {
+        g_katanaManager = NULL;
+        data->exitcode = manager->CallApplicationMain(data->argc, data->argv);
```
troydai: The return value here is not an actually `HRESULT`, but a `int` returns from the application. It can't be judge by `SUCCEEDED(hr)`. No message should be print no matter what the result is. Output should be controlled by the application itself.

src/klr.net45/klr.net45.cpp
```diff
@@ -8,20 +8,25 @@
...
+    }
+    else 
+    {
+        printf_s("Failed to initalize runtime (%x)", hr);
```
troydai: After consult the code in `klr.core45`, I choose to print the message here instead of outer `klr` caller. For two reasons:

1. [`klr.core45.cpp`](https://github.com/aspnet/KRuntime/blob/dev/src/klr.core45/klr.core45.cpp) already prints all kind of errors.
2. There is advantage to print the error message here. Since it is close to the crash site the error can contain meaningful information. Returning `HRESULT` to caller doesn't necessary explain everything.

troydai: The `hr` is print in the hexadecimal format.

troydai:
New look on CLR:
```
C:\Users\trdai\Desktop>klr
Failed to initalize runtime (80004001)
```

troydai:
:watch:

src/klr.net45/klr.net45.cpp
davidfowl: The only problem is that hosts like helios can't decipher anything if we don't return specific exit codes. Printing to the console is useless in that scenario.

troydai: Allow me to clarify all the return values here. There are two values representing the exit codes here:

1. Return value from the method, in the form of `HRESULT`. It is always the exit code of the __runtime initialization__.
2. `data->exitcode`, in the form of `int`. When __runtime is initialized successfully__ it is the return value of __application main__ (which is defined by user code); on the other hand when __runtime initialization failed__ it is assigned the returned `HRESULT`. (I do think this part is a little bit odd, maybe we should leave the `data->exitcode` unassigned in this situation.)

So any caller should do:
1. Verify the return `HRESULT` first to ensure runtime is successfully initialized.
2. If so read the `data->exitcode` to determine the program exit status.

Unless **Helios** operate differently or I miss anything, the caller should have the chance to decipher the runtime status.

Did I missing anything?

davidfowl:
:shipit:

troydai:
Rebase and squashed. Will merge after last round of testing.

src/klr.net45/klr.net45.cpp
Eilon: There's a typo in the error message :smile:

PR #1045, opened 2015-01-07
fixed the issue when PackManager gets runtime path which ends with a slash

### LINKED ISSUE #1005: kpm pack with --runtime path ending with a slash throws null reference exception
```
kpm pack --runtime C:\Users\testuser\.kre\packages\KRE-CoreCLR-x86.1.0.0-rc1-10815\ -o c:\Test
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Framework.PackageManager.Packing.PackManager.Package()
   at Microsoft.Framework.PackageManager.Program.<>c__DisplayClass2.<Main>b__11()
   at Microsoft.Framework.Runtime.Common.CommandLine.CommandLineApplication.Execute(String[] args)
   at Microsoft.Framework.PackageManager.Program.Main(String[] args)
```

vkairys:
Fix #1005 .  Use DirectoryInfo class to parse runtime name from path.

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Packing/PackManager.cs
+++ b/src/Microsoft.Framework.PackageManager/Packing/PackManager.cs
@@ -109,7 +109,7 @@ namespace Microsoft.Framework.PackageManager.Packing
 
             foreach (var runtime in _options.Runtimes)
             {
-                var frameworkName = DependencyContext.GetFrameworkNameForRuntime(Path.GetFileName(runtime));
+                var frameworkName = DependencyContext.GetFrameworkNameForRuntime(new DirectoryInfo(runtime).Name);
                 var runtimeLocated = TryAddRuntime(root, frameworkName, runtime);
                 List<string> runtimeProbePaths = null;
 


Praburaj:
Change looks good to me. /CC @ChengTian

davidfowl:
@vkairys did you sign a CLA?

vkairys:
@davidfowl  yes

troydai:
The fix looks good to me as well. But my hint will be the problem should be addressed at the place where `PackOptions.Runtimes` property are assigned (https://github.com/vkairys/KRuntime/blob/dev/src/Microsoft.Framework.PackageManager/Program.cs#L130)

``` C#
Runtimes = optionRuntime.HasValue() ?
    string.Join(";", optionRuntime.Values).
        Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries) :
    new string[0],
```

vkairys:
Should I rewrite my code and move runtime name parsing to the Program class?

troydai:
After took another look, I think we better off with your solution for now since it is simpler. 

My first instinct was that since this is about normalize the user input, it should be place close to the inputs' origin. However since the `runtime` value could be either a name or a path to the KRE, they cannot be normalized just simply by `new DirectoryInfo(runtime).Name` at `Program`. It will require to combine checking directory's existence and normalizing the path. To accomplish that more than a line of code are needed at the `Program`, which is already big. It doesn't worth it given we're tackling a minor issue.

vkairys:
Should I just close this PR or create a new one with the same change in https://github.com/aspnet/XRE/blob/dev/src/Microsoft.Framework.PackageManager/Bundle/BundleManager.cs#L112  ?

Issue #925, opened 2014-11-23
Rename 'sources' in global.json to something else (e.g. 'projectSearchPaths')


akoeplinger:
(based on today's discussion on JabbR)

In global.json the `sources` key is used to specify paths where the K runtime is looking for projects. This is easily confused with the `sources` section in project.json and also doesn't make it clear that you should be specifying the folder **above** project.json there.

Proposal: rename `sources` to `projectSearchPaths` in global.json, to clear up the confusion.

davidfowl:
@akoeplinger we've renamed it to projects

PR #1056, opened 2015-01-10
Rename sources to projectSearchPaths in global.json

davidfowl:
- Allow sources to stay for now while projects transition

#925


muratg:
projectSearchPath --> projects

Issue #442, opened 2014-07-16
Support version ranges in project.json


davidfowl:
Today we only support minimum. There should be a way in the future to lock dependency ranges.

davidfowl:
Initial work is done.

davidfowl:
In the RC we'll look at implementing syntax similar to https://github.com/npm/node-semver#semver1----the-semantic-versioner-for-npm

jasonwilliams200OK:
@davidfowl, does this mean we will have support for tilde (`~`) and circumflex accent (`^`) etc. like npm in dnx world sometime in future? Is there an issue tracking it?

From npm docs - https://docs.npmjs.com/files/package.json#dependencies:

>See semver(7) for more details about specifying version ranges.
* version Must match version exactly
* >version Must be greater than version
* >=version etc
* <version
* <=version
* ~version "Approximately equivalent to version" See semver(7)
* ^version "Compatible with version" See semver(7)
* 1.2.x 1.2.0, 1.2.1, etc., but not 1.3.0
* http://... See 'URLs as Dependencies' below
* * Matches any version
* "" (just an empty string) Same as *
* version1 - version2 Same as >=version1 <=version2.
* range1 || range2 Passes if either range1 or range2 are satisfied.
* git... See 'Git URLs as Dependencies' below
* user/repo See 'GitHub URLs' below
* tag A specific version tagged and published as tag See npm-tag(1)
* path/path/path See Local Paths below
>
For example, these are all valid:
>
```json
{ "dependencies" :
  { "foo" : "1.0.0 - 2.9999.9999"
  , "bar" : ">=1.0.2 <2.1.2"
  , "baz" : ">1.0.2 <=2.3.4"
  , "boo" : "2.0.1"
  , "qux" : "<1.0.0 || >=2.3.1 <2.4.5 || >=2.5.2 <3.0.0"
  , "asd" : "http://asdf.com/asdf.tar.gz"
  , "til" : "~1.2"
  , "elf" : "~1.2.3"
  , "two" : "2.x"
  , "thr" : "3.3.x"
  , "lat" : "latest"
  , "dyl" : "file:../dyl"
  }
}
```
#### URLs as Dependencies
>
You may specify a tarball URL in place of a version range.
>
This tarball will be downloaded and installed locally to your package at install time.
>
#### Git URLs as Dependencies
>
Git urls can be of the form:
>>
git://github.com/user/project.git#commit-ish
[...]

PR #1057, opened 2015-01-10
Initial plumbing for version ranges

davidfowl:
- Added RequestedVersion to Library as IVersionSpec
- Removed Snapshots from SemanticVersion and added them to
VersionSpec instead.

Things left to implement:
- Adding support for maximums to the runtime and restore algorithms.
- Conflict resolution support
- Possible support for ~1.0 syntax

#442

/cc @lodejard @loudej

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Restore/RestoreOperations.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/RestoreOperations.cs
@@ -20,28 +20,27 @@ namespace Microsoft.Framework.PackageManager
             _report = report;
         }
 
-        public async Task<GraphNode> CreateGraphNode(RestoreContext context, Library library, Func<string, bool> predicate)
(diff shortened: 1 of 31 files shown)


src/Microsoft.Framework.PackageManager/Restore/RestoreOperations.cs
```diff
@@ -33,20 +33,6 @@ public RestoreOperations(IReport report)
 
             if (node.Item != null)
             {
-                if (node.Library != null &&
```
davidfowl: @lodejard can you remember why we have this? With the new code, any resolved library can't have a snapshot as a version. Only RequestedVersions (IVersionSpec) can be a snapshot.

davidfowl: Figured it out

PR #1059, opened 2015-01-11
Fix kpm wrap null ref and clean up a few issues

davidfowl:
- Handle nuget package references that don't specify target framework.
- Special case mscorlib since it doesn't have any of the other flags in some cases.
I ran kpm wrap on the corefx repository and got some failures, these fix them.

/cc @ChengTian


ChengTian:
:shipit: Thanks.

Issue #1052, opened 2015-01-10
Specify the ErrorSource, LineNumber and Column number when errors occur in DTH


davidfowl:
If the exception that occurs is related to parsing a file, then return that information so that the IDE can show that in the error list.

PR #1065, opened 2015-01-13
Added ProjectFormatException and more detailed errors to DTH

davidfowl:
- ProjectFormatException has the project path, line and Column
(1 based), where the exception ocurred in the project.json file.
- Give more specific errors when they aren't just JsonReaderExceptions,
but semantic exceptions within the project format itself (beyond just JSON).
- Added Path, Line and Column to ErrorMessage in DTH

#1052

/cc @BillHiebert

### DIFF
--- a/src/Microsoft.Framework.Runtime/Project.cs
+++ b/src/Microsoft.Framework.Runtime/Project.cs
@@ -6,6 +6,8 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Runtime.Versioning;
+using System.Text;
+using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using NuGet;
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.Framework.Runtime/Project.cs
```diff
@@ -436,9 +476,13 @@ private static IEnumerable<string> GetSourcesSplit(string sourceDescription)
             {
                 foreach (var dependency in dependencies)
                 {
-                    if (String.IsNullOrEmpty(dependency.Key))
+                    if (string.IsNullOrEmpty(dependency.Key))
                     {
-                        throw new InvalidDataException("Unable to resolve dependency ''.");
+
+                        throw ProjectFormatException.Create(
+                            "Unable to resolve dependency ''.",
```
Eilon: Kind of a crappy error. Should it say like "empty dependency" or something?

Eilon: Or say that the dependency isn't specified? Or that dependency names can't be empty?

PR #1069, opened 2015-01-14
Moving ICompilationException to Runtime.Interfaces

pranavkm:


src/Microsoft.Framework.Runtime.Roslyn/RoslynCompilationException.cs
```diff
@@ -1,23 +1,57 @@
...
+                                                     .Select(d => new RoslynCompilationFailure(d));
+                }
+
+                return _compilationFailure;
+            }
+        }
+        private static string GetErrorMessage(string assemblyName)
         {
-            return RoslynDiagnosticUtilities.Convert(diagnostics.Where(RoslynDiagnosticUtilities.IsError)).ToList();
+            return string.Format(CultureInfo.CurrentCulture,
```
davidfowl: Did we have a message before?

src/Microsoft.Framework.Runtime.Roslyn/RoslynCompilationFailure.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        /// <inheritdoc />
+        public IEnumerable<ICompilationMessage> Messages { get; }
+
+        /// <inheritdoc />
+        public string SourceFilePath { get; }
+
+        /// <inheritdoc />
+        public string SourceFileContent
```
davidfowl: Why not make the caller do this? If you have a source file can't you just get the text?

src/Microsoft.Framework.Runtime/Loader/DesignTimeCompilationException.cs
```diff
@@ -0,0 +1,51 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Framework.Runtime
+{
+    internal class DesignTimeCompilationException : Exception, ICompilationException
+    {
+        public DesignTimeCompilationException(IList<string> errors)
```
davidfowl: Maybe we should remote the unformatted messages? This means in VS right now we won't get the same nice error page in helios

src/Microsoft.Framework.Runtime.Roslyn/RoslynCompilationException.cs
pranavkm: We did. But it would combine the individual diagnostic messages. We were doing something similar in Mvc but this was causing the diagnostic page to  repeat the individual diagnostics twice - once at the top with the exception message and once with the individual error displays.

We could alternatively chose to not display the error message in Helios and change this back to old behavior.

src/Microsoft.Framework.Runtime.Roslyn/RoslynCompilationFailure.cs
pranavkm: For Mvc, this could be content from an embedded file system since you can set `RazorViewEngineOptions.FileSystem`. Helios \ Diagnostics wouldn't be able to read it unless we pass the `IFileSystem` along. Consequently since Mvc already has the file content in memory, we assign the property value.

src/Microsoft.Framework.Runtime/Loader/DesignTimeCompilationException.cs
pranavkm: ok. Could we wait until we have Helios support for the exception type to start with? Would make it easier to debug

pranavkm:
Updated.

src/Microsoft.Framework.Runtime/Infrastructure/JsonExtensions.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Framework.Runtime.Infrastructure
+{
+    internal static class JsonExtensions
+    {
+        public static string[] ValueAsStringArray(this JObject obj, string key)
```
davidfowl: I think this can go away now as there's a new method that @Praburaj added that handles this for generic <T>

src/Microsoft.Framework.DesignTimeHost/ApplicationContext.cs
```diff
@@ -242,15 +259,8 @@ private bool ProcessMessage()
                     break;
                 case "GetCompiledAssembly":
                     {
-                        var libraryKey = new RemoteLibraryKey
```
davidfowl: Revert this, @Praburaj made this change for performance reasons.

Issue #1013, opened 2014-12-28
Make it easier to run klr


davidfowl:
### Scenario 1

`klr /path/to/MyApp.exe`
`klr /path/to/MyApp.dll`

### Scenario 2
`klr /path/to/folder/project.json`
`klr /path/to/folder`

### Expansions

`klr /path/to/MyApplication.dll` --->  `klr --appbase /path/to/ --lib /path/to/ MyApplication`
`klr /path/to/folder run` ---> `klr --appbase /path/to/folder Microsoft.Framework.ApplicationHost run`

This makes k.cmd less important possibly extinct.

PR #1070, opened 2015-01-14
Improve klr

ChengTian:
parent #1013 #1071 

* `RuntimeBootstrapper` already does the following expansion:
`klr /path/to/MyApplication.dll` --->  `klr --lib /path/to/ MyApplication`.
So we only need to generate `--appbase /path/to/` part in this case

(See https://github.com/aspnet/KRuntime/blob/dev/src/klr.hosting.shared/RuntimeBootstrapper.cs#L298)

* Modifying `src/klr.mono.managed/EntryPoint.cs` to do the expansion on Mono is much more easier. We can even add `--appbase` for klr on Mono in `EntryPoint.cs` if we want it.

### DIFF
--- a/src/klr.hosting.shared/RuntimeBootstrapper.cs
+++ b/src/klr.hosting.shared/RuntimeBootstrapper.cs
@@ -84,10 +84,27 @@ namespace klr.hosting
 #endif
             var app = new CommandLineApplication(throwOnUnexpectedArg: false);
             app.Name = "klr";
+
+            // RuntimeBootstrapper doesn't need to consume '--appbase' option because
+            // klr/klr.cpp consumes the option value before invoking RuntimeBootstrapper
+            // This is only for showing help info and swallowing useless '--appbase' option
+            var optionAppbase = app.Option("--appbase <PATH>", "Application base directory path",
+                CommandOptionType.SingleValue);
             var optionLib = app.Option("--lib <LIB_PATHS>", "Paths used for library look-up",
                 CommandOptionType.MultipleValue);
             app.HelpOption("-?|-h|--help");
             app.VersionOption("--version", GetVersion());
+
+            // Options below are only for help info display
+            // They will be forwarded to Microsoft.Framework.ApplicationHost
+            var optionsToForward = new[]
+            {
+                app.Option("--watch", "Watch file changes", CommandOptionType.NoValue),
+                app.Option("--packages <PACKAGE_DIR>", "Directory containing packages", CommandOptionType.SingleValue),
+                app.Option("--configuration <CONFIGURATION>", "The configuration to run under", CommandOptionType.SingleValue),
+                app.Option("--port <PORT>", "The port to the compilation server", CommandOptionType.SingleValue)
+            };
+
             app.Execute(args);
 
             // Help information was already shown because help option was specified
@@ -101,6 +118,40 @@ namespace klr.hosting
             {
                 app.ShowHelp();
                 return Task.FromResult(2);
+            }
+
+            // Some options should be forwarded to Microsoft.Framework.ApplicationHost
+            var appHostName = "Microsoft.Framework.ApplicationHost";
+            var appHostIndex = app.RemainingArguments.FindIndex(s =>
+                string.Equals(s, appHostName, StringComparison.OrdinalIgnoreCase));
+            foreach (var option in optionsToForward)
+            {
+                if (option.HasValue())
+                {
+                    if (appHostIndex < 0)
+                    {
+                        Console.WriteLine("The option '--{0}' can only be used with {1}", option.LongName, appHostName);
+                        return Task.FromResult(1);
+                    }
+
+                    if (option.OptionType == CommandOptionType.NoValue)
+                    {
+                        app.RemainingArguments.Insert(appHostIndex + 1, "--" + option.LongName);
+                    }
+                    else if (option.OptionType == CommandOptionType.SingleValue)
+                    {
+                        app.RemainingArguments.Insert(appHostIndex + 1, "--" + option.LongName);
+                        app.RemainingArguments.Insert(appHostIndex + 2, option.Value());
+                    }
+                    else if (option.OptionType == CommandOptionType.MultipleValue)
+                    {
+                        foreach (var value in option.Values)
+                        {
+                            app.RemainingArguments.Insert(appHostIndex + 1, "--" + option.LongName);
+                            app.RemainingArguments.Insert(appHostIndex + 2, value);
+                        }
+                    }
+                }
             }
 
             // Resolve the lib paths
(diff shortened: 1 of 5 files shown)


src/klr/klr.cpp
```diff
@@ -7,7 +7,143 @@ void GetModuleDirectory(HMODULE module, LPWSTR szPath)
...
+    int nLastBackSlashIndex = LastIndexOfChar(pszPath, L'\\');
+    return max(nLastSlashIndex, nLastBackSlashIndex);
+}
+
+void GetParentDir(LPCWSTR const pszPath, LPWSTR const pszParentDir)
+{
+    int nLastSeparatorIndex = LastPathSeparatorIndex(pszPath);
+    if (nLastSeparatorIndex < 0)
+    {
+        StringCchCopyW(pszParentDir, MAX_PATH, L".");
```
ChengTian: strcpy_s() works for char arrays but we need the version for wide char (WCHAR) arrays. The secure version of lstrcpy() is `StringCchCopyW()`:
http://msdn.microsoft.com/en-us/library/windows/desktop/ms647490(v=vs.85).aspx
http://msdn.microsoft.com/en-us/library/windows/desktop/ms647527(v=vs.85).aspx

src/klr/klr.cpp
```diff
@@ -7,7 +7,143 @@ void GetModuleDirectory(HMODULE module, LPWSTR szPath)
...
+            break;
+        }
+    } while (--nIndex >= 0);
+
+    return nIndex;
+}
+
+bool StringsEqual(LPCWSTR const pszStrA, LPCWSTR const pszStrB)
+{
+    return ::_wcsicmp(pszStrA, pszStrB) == 0;
```
davidfowl: Is there a secure equivalent here?

src/klr/klr.cpp
```diff
@@ -7,7 +7,143 @@ void GetModuleDirectory(HMODULE module, LPWSTR szPath)
...
+        StringCchCopyW(ppszExpandedArgv[1], MAX_PATH, szParentDir);
+
+        return true;
+    }
+
+    // "klr /path/project.json run" --> "klr --appbase /path/ Microsoft.Framework.ApplicationHost run"
+    // "klr /path/ run" --> "klr --appbase /path/ Microsoft.Framework.ApplicationHost run"
+    WCHAR szFileName[MAX_PATH];
+    GetFileName(ppszArgv[0], szFileName);
+    if (StringsEqual(szFileName, L"project.json"))
```
davidfowl: What happens when I do `klr . `

ChengTian: capture

ChengTian: This case is captured by:
https://github.com/aspnet/KRuntime/pull/1070/files#diff-ce00d5b3fa27ae892561d404f3ae9c9fR81

src/klr/klr.cpp
ChengTian: Nice catch! We can use `_wcsnicmp(str1, str2, max_count)` here.

src/klr.hosting.shared/RuntimeBootstrapper.cs
```diff
@@ -300,7 +300,7 @@ private static string[] ResolveSearchPaths(IEnumerable<string> libPaths, List<st
                     extension.Equals(".exe", StringComparison.OrdinalIgnoreCase))
                 {
                     // Add the directory to the list of search paths
-                    searchPaths.Add(Path.GetDirectoryName(application));
```
davidfowl: Do we even need this logic anymore?

ChengTian: Since `RuntimeBootstrapper` already does the following expansion:
`klr /path/to/MyApplication.dll` --->  `klr --lib /path/to/ MyApplication`.
So we only need to generate `--appbase /path/to/` part in this case.
Keeping logic here helps to reduce C++ code.

davidfowl: Awesome!

ChengTian:
Found a bug:
capture1

Fixed in the latest commit. After fix:
capture2

src/klr/klr.cpp
ChengTian: Sorry I didn't notice the dot. Yeah, it reveals a bug and I just fixed it in the latest commit. After fix:
capture

davidfowl:
One thing we need to do is merge the help between the lower layer and apphost,

klr -> Shows help for the klr.exe
klr /path/to/project.json -> Shows help for klr.exe Microsoft.Framework.ApplicationHost

The latter is kind of weird so we should try to make this transparent so that klr --help just shows all the options.

src/klr.host/Bootstrapper.cs
```diff
@@ -59,6 +59,7 @@ public Task<int> Main(string[] args)
 #else
             string applicationBaseDirectory = AppContext.BaseDirectory;
 #endif
+            Console.WriteLine("!!! AppBase: " + applicationBaseDirectory);
```
troydai: Err, we really want to print three `!`?

ChengTian: It was a temporary backup and that line was for testing & debugging. Please check the latest commit.

ChengTian:
Merged the help info in the latest commit:
capture

Now running ApplicationHost without any argument doesn't give help info:
capture1

Note that some options only applies to ApplicationHost case. If you are running an assembly with an incompatible option:
capture2

davidfowl:
Nice!

ChengTian:
I would suggest that we do https://github.com/aspnet/KRuntime/issues/1071 in this PR because help info of klr on Mono now is
```
Usage: klr [options]

Options:
  --appbase <PATH>                 Application base directory path
  --lib <LIB_PATHS>                Paths used for library look-up
  -?|-h|--help                     Show help information
  --version                        Show version information
  --watch                          Watch file changes
  --packages <PACKAGE_DIR>         Directory containing packages
  --configuration <CONFIGURATION>  The configuration to run under
  --port <PORT>                    The port to the compilation server
```
`--appbase` should be available on Mono, as the help info suggested.

ChengTian:
Added --appbase for klr on Mono.
Tested with the following test cases, on both Windows and *nix:
```
klr
klr foo
klr --foo
klr --port
klr --port 8888
klr .\ConsoleApp
klr .\ConsoleApp run
klr --appbase .\ConsoleApp .\ConsoleApp run
klr --port 8888 .\ConsoleApp run
klr --watch .\ConsoleApp run
klr --port 8888 --watch .\ConsoleApp run
klr .\ConsoleApp\bin\Debug\aspnet50\ConsoleApp.dll
klr --appbase .\ConsoleApp .\ConsoleApp\bin\Debug\aspnet50\ConsoleApp.dll
klr --watch .\ConsoleApp\bin\Debug\aspnet50\ConsoleApp.dll
```
@davidfowl , please check the latest commit and sign off.

Issue #982, opened 2014-12-09
klr and managed code has inconsistent behaviors when handling KRE_TRACE


ChengTian:
klr thinks verbose flag is on when `KRE_TRACE` has a value, no matter what the value is:
https://github.com/aspnet/KRuntime/blob/dev/src/klr/klr.cpp#L16

`RuntimeBootstrapper` thinks verbose flag is on only when value of `KRE_TRACE` is `1`:
https://github.com/aspnet/KRuntime/blob/dev/src/klr.hosting.shared/RuntimeBootstrapper.cs#L75

So if you set `KRE_TRACE` to `0`. Managed code doesn't print any trace information, while klr still prints verbose information like:
```
Loaded Module: k
Found DLL Export: CallApplicationMain
```

PR #1076, opened 2015-01-16
KLR print trace only when KRE_TRACE=1

troydai:
To make it consistent with managed code.

#982

### DIFF
--- a/src/klr/klr.cpp
+++ b/src/klr/klr.cpp
@@ -12,8 +12,15 @@ void GetModuleDirectory(HMODULE module, LPWSTR szPath)
 
 int CallFirmwareProcessMain(int argc, wchar_t* argv[])
 {
-    TCHAR szKreTrace[1];
-    bool m_fVerboseTrace = GetEnvironmentVariableW(L"KRE_TRACE", szKreTrace, 1) > 0;
+    TCHAR szKreTrace[2];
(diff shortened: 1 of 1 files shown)


src/klr/klr.cpp
```diff
@@ -12,8 +12,15 @@ void GetModuleDirectory(HMODULE module, LPWSTR szPath)
...
 int CallFirmwareProcessMain(int argc, wchar_t* argv[])
 {
-    TCHAR szKreTrace[1];
-    bool m_fVerboseTrace = GetEnvironmentVariableW(L"KRE_TRACE", szKreTrace, 1) > 0;
+	TCHAR szKreTrace[2];
+	DWORD nEnvTraceSize = GetEnvironmentVariableW(L"KRE_TRACE", szKreTrace, 2);
+	bool m_fVerboseTrace = (nEnvTraceSize == 1);
+	if (m_fVerboseTrace)
+	{
+		szKreTrace[1] = L'\0';
```
davidfowl: Why is this tabbed?

davidfowl:
When the tabs are fixed, then ship it :shipit:

src/klr/klr.cpp
troydai: Newly installed VS is dumb. I'll fix it.

troydai:
Updated

PR #1078, opened 2015-01-16
Breadcrumbs for servicing

victorhurdugaci:
It still has a TODO for what should be the marker for the serviceable package and that check returns true all the time.
Currently there are two files created (with and without version).

@glennc @lodejard @loudej @GrabYourPitchforks @davidfowl


src/Microsoft.Framework.Runtime/DependencyManagement/NuGetDependencyResolver.cs
```diff
@@ -144,6 +144,8 @@ public void Initialize(IEnumerable<LibraryDescription> packages, FrameworkName t
             var cacheResolvers = GetCacheResolvers();
             var defaultResolver = new DefaultPackagePathResolver(_repository.RepositoryRoot);
 
+            Servicing.ServicingBreadcrumbs breadcrumbs = new Servicing.ServicingBreadcrumbs();
```
glennc: Servicing.Breadcrumbs instead?

src/Microsoft.Framework.Runtime/Servicing/ServicingBreadcrumbs.cs
```diff
@@ -0,0 +1,80 @@
...
+        public ServicingBreadcrumbs()
+            : this(ResolveBreadcrumbsFolder())
+        {
+        }
+
+        internal ServicingBreadcrumbs(string breadcrumbsFolder)
+        {
+            // If the directory doesn't exist, don't create it because it
+            // needs special permissions on it
+            _breadcrumbsFolder = Directory.Exists(breadcrumbsFolder) ? breadcrumbsFolder : null; ;
```
glennc: Double ;

src/Microsoft.Framework.Runtime/Servicing/ServicingBreadcrumbs.cs
```diff
@@ -0,0 +1,80 @@
...
+            if (string.IsNullOrWhiteSpace(programDataFolder))
+            {
+                return null;
+            }
+
+            string breadcrumbsFolder = Path.Combine(
+                programDataFolder,
+                "Microsoft",
+                "KRE",
+                "breadcrumbs");
```
victorhurdugaci: This needs to be `BreadcrumbStore`, will update soon.

src/Microsoft.Framework.Runtime/Servicing/ServicingBreadcrumbs.cs
glennc: Also not 100% sure that we shouldn't create it. We could still try and throw if the permissions aren't correct. That would allow a tool to create the breadcrumbs folder to use this class.

src/Microsoft.Framework.Runtime/Servicing/ServicingBreadcrumbs.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            // If the directory doesn't exist, don't create it because it
+            // needs special permissions on it
+            _breadcrumbsFolder = Directory.Exists(breadcrumbsFolder) ? breadcrumbsFolder : null; ;
+        }
+
+        public static bool IsPackageServiceable(IPackage package)
+        {
+            // TODO: Figure out what makes a package serviceable
+            return true;
```
GrabYourPitchforks: return false by default until we have the correct logic hooked up here.

src/Microsoft.Framework.Runtime/Servicing/ServicingBreadcrumbs.cs
```diff
@@ -0,0 +1,80 @@
...
+            }
+
+            // Create both files for now until we get clear instructions about the format of the name
+            CreateBreadcrumbfile(packageId);
+            CreateBreadcrumbfile(packageId + "." + packageVersion);
+        }
+
+        private static string ResolveBreadcrumbsFolder()
+        {
+            var programDataFolder = Environment.GetEnvironmentVariable("ProgramData");
```
GrabYourPitchforks: Consider using SpecialFolder.CommonApplicationData instead so that we're not hardcoding this.

glennc: @loudej are you ok with this path?

src/Microsoft.Framework.Runtime/DependencyManagement/NuGetDependencyResolver.cs
victorhurdugaci: Sure

src/Microsoft.Framework.Runtime/Servicing/ServicingBreadcrumbs.cs
victorhurdugaci: I thought about that and I can definitely create it from code but I am worried about the permissions on it. If the code runs under a limited account, it cannot restrict the delete permissions to admins only. Also, in `%programdata%\Microsoft` only admins can write by default so the script that creates the folder must run elevated

src/Microsoft.Framework.Runtime/Servicing/ServicingBreadcrumbs.cs
victorhurdugaci: Okay

src/Microsoft.Framework.Runtime/Servicing/ServicingBreadcrumbs.cs
victorhurdugaci: SpecialFolder is not available on CoreCLR

test/Microsoft.Framework.Runtime.Tests/ServicingFacts.cs
```diff
@@ -0,0 +1,68 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.Security.AccessControl;
+using Xunit;
+
+namespace Microsoft.Framework.Runtime.Tests
+{
+    public class ServicingFacts
```
davidfowl: Unit tests shouldn't touch the disk, this looks like a functional test, put it in that assembly.

victorhurdugaci: Sure. Moved

src/Microsoft.Framework.Runtime/Properties/AssemblyInfo.cs
```diff
@@ -4,3 +4,4 @@
 using System.Runtime.CompilerServices;
 
 [assembly: InternalsVisibleTo("Microsoft.Framework.Runtime.Tests")]
+[assembly: InternalsVisibleTo("Microsoft.Framework.Runtime.FunctionalTests")]
```
Eilon: No IVT to functional tests, only to unit tests.

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
```diff
@@ -0,0 +1,80 @@
...
+            _breadcrumbsFolder = Directory.Exists(breadcrumbsFolder) ? breadcrumbsFolder : null;
+        }
+
+        public static bool IsPackageServiceable(IPackage package)
+        {
+            // TODO: Figure out what makes a package serviceable
+            return false;
+        }
+
+        public void LeaveBreadcrumb(string packageId, SemanticVersion packageVersion)
```
Eilon: `CreateBreadcrumb`?

victorhurdugaci:
Added breadcrumbs for the KRE itself

src/Microsoft.Framework.Runtime/Properties/AssemblyInfo.cs
victorhurdugaci: Done

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
```diff
@@ -0,0 +1,101 @@
...
+        public Breadcrumbs()
+            : this(ResolveBreadcrumbsFolder())
+        {
+        }
+
+        public Breadcrumbs(string breadcrumbsFolder)
+        {
+            // If the directory doesn't exist, don't create it because it
+            // needs special permissions on it
+            _breadcrumbsFolder = Directory.Exists(breadcrumbsFolder) ? breadcrumbsFolder : null;
```
Eilon: It's kind of a matter of style, but this could be a simple `if()` statement. If the directory doesn't exist no need to set the folder to null because it's already null. Might be easier to debug.

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        public static bool IsPackageServiceable(IPackage package)
+        {
+            // TODO: Figure out what makes a package serviceable
```
Eilon: Does there need to be a bug for this? (Or is there already?)

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        private void CreateBreadcrumbfile(string fileName)
+        {
+            string fullFilePath = Path.Combine(_breadcrumbsFolder, fileName);
+
+            try
+            {
+                if (!File.Exists(fullFilePath))
+                {
+                    File.Create(fullFilePath).Dispose();
```
Eilon: Should we do any logging here? On success? On failure?

test/Microsoft.Framework.Runtime.FunctionalTests/project.json
```diff
@@ -0,0 +1,14 @@
+{
+	"dependencies": {
+		"Microsoft.AspNet.Testing": "1.0.0-*",
```
Eilon: O. M. G.

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
victorhurdugaci: No. This is stuff only for us

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
victorhurdugaci: There will be one

test/Microsoft.Framework.Runtime.FunctionalTests/project.json
victorhurdugaci: I guess I can remove that package and let the tests run on mono even though the feature is not available on that platform. We should consider doing something that allows us to use test extensions in kruntime

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
Eilon: It's running on a customer machine, so it's definitely not "only for us." (Not that that implies we should log everything, but this strikes me as a somewhat interesting event.)

test/Microsoft.Framework.Runtime.FunctionalTests/project.json
Eilon: Was referring to the `\t` :smile:

victorhurdugaci: :speak_no_evil:

victorhurdugaci: Btw, it should be O.M.T. (t = tab)

victorhurdugaci:
@Eilon updated with logging. It uses Trace because that's what kruntime does instead of `ILogger`

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
blowdart: It'd be an interesting debug point if you can't log a breadcrumb. That would indicate something rather problematic as servicing would no be impossible. I'm not even sure a simple log is enough on failure.

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
```diff
@@ -0,0 +1,117 @@
...
+            }
+            catch (DirectoryNotFoundException exception)
+            {
+                LogBreadcrumbsCreationFailure(fileName, exception);
+            }
+        }
+
+        private static void LogBreadcrumbsCreationFailure(string fileName, Exception exception)
+        {
+            Trace.TraceInformation("Failed to write servicing breadcrumb for {0} because of {1}", fileName, exception);
```
Eilon: Change to `... because an exception was thrown: {1}`. Otherwise there's no way the logged text would make sense.

Eilon:
Yeah I'm not sure what happens if you use Trace - that doesn't go anywhere useful does it? Like, if I'm running an app, where would I ever see that?

victorhurdugaci:
@Eilon when you set `KRE_TRACE=1` you see the traces in the console because we are adding a console listener to `Trace`

Eilon:
Ah OK... but the customer won't ever see that... I wanted this to go the customer's log file so that they see what magic stuff we're doing behind the scenes.

blowdart:
As this is a potential security issue it needs to be up front and centre. If you can't breadcrumb you can't service.

victorhurdugaci:
The customer will see the message if they enable the trace. There is no information from `KRuntime` written in the console otherwise.

@blowdart if that's the case, I think more information should be surfaced that just what I did in this PR. For example, even when the servicing index is malformed, you cannot service.

blowdart:
Yea, I think this needs surfacing if it's the case, without a user having to add trace listeners or anything fancy. @GrabYourPitchforks @davidfowl

Eilon:
I don't want it written to the console, I want it written to the log file. Is logging not available from this call site?

blowdart:
I'd like it in the console, and in the event log if on Windows. Because seriously no servicing directory is a bad bad thing.

victorhurdugaci:
Nope. There is no logging (as in `ILogger`) in KRuntime

lodejard:
We should talk about what impact would be for moving ILoggerFactory from aspnet hosting into base program.main hosting.

Writing to console is probably bad, if we expect people to bundle an xcopy-able program that can be run without a machine-wide install. You can run it, but you get an error message nagging you to install dotnet in order to configure the machine-wide servicing directory.

blowdart:
We could, of course, provide a util that elevates so it can create and set the ACLs correctly on the servicing directory.

victorhurdugaci:
I don't think elevation is required to create the directory and set its permissions in `%ProgramData%` but it is not KRuntime's responsibility to create that directory and set the permissions.

davidfowl:
@lodejard YES! Lets move logger further down the stack!

victorhurdugaci:
How about we split this in two: the breadcrumbs as they are now and the ILogger? I'll open a bug for the later one and we'll address it after. Also, part of that will be to report in a more visible place the fact that servicing doesn't work.

blowdart:
Define "after". After what? :alarm_clock:

victorhurdugaci:
After I check this in. Basically have two different work items since they are unrelated

Eilon:
"After after." -- Mr. Miyagi to Danielsan

victorhurdugaci:
The "after after" is here: https://github.com/aspnet/XRE/issues/1118

victorhurdugaci:
Anymore comments for this PR? If someone signs off, then I will redo the PR against the Release branch...

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
```diff
@@ -0,0 +1,117 @@
...
+            string breadcrumbsFolder = Path.Combine(
+                programDataFolder,
+                "Microsoft",
+                ".NET XRE",
+                "BreadcrumbStore");
+
+            return breadcrumbsFolder;
+        }
+
+        private void CreateBreadcrumbfile(string fileName)
```
Eilon: `file` --> `File` :smile:

Eilon:
Overall looks good to me.

victorhurdugaci:
Closing this one for now. Will send a new one for the release branch

Issue #81, opened 2014-04-04
Support xml docs


davidfowl:
We should output this by default when building the nupkg

PR #1084, opened 2015-01-16
Fixes #1083 by adding version resources to native projects

### LINKED ISSUE #1083: klr.core45.dll and similar need an embedded file version info
The native launcher will also read the file version from klr.core45.dll to determine whether we're about to load a vulnerable version and kick in the servicing logic.

We should also take this time to add copyright info to the manifests.

anurse:
Fixes #1083 

Results:

image

klr.exe is marked as file type "VFT_APP" and the others are "VFT_DLL" (http://msdn.microsoft.com/en-us/library/windows/desktop/aa381058%28v=vs.85%29.aspx)

/cc @GrabYourPitchforks @davidfowl


Eilon:
Looks good to me. After this is merged can you check that the CI really does the right thing by doing the same Explorer / Properties check on the CI's *signed* build output from Coherence-Signed?

anurse:
Yep, can do.

davidfowl:
Can we avoid this on local builds and only do it on the CI? Its pretty annoying.

anurse:
Totally, good call.

anurse:
this was merged, and includes @davidfowl 's suggestion

moozzyk:
@davidfowl  @anurse - I am doing this for Helios - #81 in Helios - (and imagine redoing this as well since it should not modify the .rc file on each build) - do you remember what was the annoying part (as per the comment above)? Was it that the .rc file was always modified but could not be added to .gitignore?

davidfowl:
@anurse changed the code to only do it on the CI

moozzyk:
@davidfowl - I do understand this. I don't know if I understand the "why".

davidfowl:
Having files changed on disk after build sucks

moozzyk:
OK. This is what I thought. I believe we should use auto gen include that is on the .gitignore list - like this: https://github.com/aspnet/SignalR-Client-Cpp/blob/dev/src/signalrclientdll/Build/VS2013/signalrclientdll.vcxproj#L104

PR #1085, opened 2015-01-16
Fix #1082 by changing environment variables for finding servicing table

### LINKED ISSUE #1082: Servicing code doesn't read ProgramFiles location correctly
Currently the servicing code reads the index file from the %PROGRAMFILES% environment variable.  This could result in different paths being read between 32-bit and 64-bit executables.

The correct logic should be:
- First check Environment.GetFolderPath(Environment.SpecialFolder.ProgramFilesX86).
- If that is null or empty, then check Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles).
- If that is null or empty, then check %HOME%.

FYI, this code is in ServicingTable.LoadIndex().

anurse:
Fixes #1082


anurse:
this was merged into dev

PR #1087, opened 2015-01-17
Rename klr to dotnet

ChengTian:
parent #1086 

`./build verify` passed


davidfowl:
Keep the `KRE_` env vars but also support `DOTNET_` as well.

davidfowl:
@ChengTian as part of this change lets also make the "run" command implicit.

`klr /path/`
`klr project.json`
`klr /path/project.json`

Should just execute without requiring a command called run.

davidfowl:
We should also try out the new code path in the makefile:

- Change the test execution to use `dotnet /path/` etc

KRuntime.sln
```diff
@@ -79,6 +65,20 @@ Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.Framework.Runtime
 EndProject
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "DependencyAnalyzer", "src\DependencyAnalyzer\DependencyAnalyzer.kproj", "{C70D1E4C-D41A-4302-98B0-194B92E49043}"
 EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "dotnet.core45.managed", "src\dotnet.core45.managed\dotnet.core45.managed.kproj", "{1DEBD970-0B84-41F7-B247-D4E853432432}"
+EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "dotnet.host", "src\dotnet.host\dotnet.host.kproj", "{68048350-2996-4B4A-90BE-3405F088BD5A}"
+EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "dotnet.hosting.shared", "src\dotnet.hosting.shared\dotnet.hosting.shared.kproj", "{FE569E9D-AAAE-43CE-BE29-2C2197C0DF8F}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "dotnet.net45.managed", "src\dotnet.net45.managed\dotnet.net45.managed.csproj", "{48CC04A6-169E-4DF5-AC70-11DD6C4F0143}"
```
dougbu: curious: why can't this be a .kproj anymore?

KRuntime.sln
```diff
@@ -79,6 +65,20 @@ Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.Framework.Runtime
 EndProject
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "DependencyAnalyzer", "src\DependencyAnalyzer\DependencyAnalyzer.kproj", "{C70D1E4C-D41A-4302-98B0-194B92E49043}"
 EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "dotnet.core45.managed", "src\dotnet.core45.managed\dotnet.core45.managed.kproj", "{1DEBD970-0B84-41F7-B247-D4E853432432}"
```
dougbu: will we need a net45 -> net46 rename in the near future?  might be better to choose names w/o versions now

scripts/kpm.cmd
```diff
@@ -2,7 +2,7 @@
 SETLOCAL
 SET ERRORLEVEL=
 
-"%~dp0klr" --appbase "%CD%" %K_OPTIONS% --lib "%~dp0lib\Microsoft.Framework.PackageManager" Microsoft.Framework.PackageManager %*
+"%~dp0dotnet" --appbase "%CD%" %K_OPTIONS% --lib "%~dp0lib\Microsoft.Framework.PackageManager" Microsoft.Framework.PackageManager %*
```
dougbu: odd to use `K_OPTIONS` in the `kpm` script.  should this be `DOTNET_OPTIONS` since the env variable seems to be used for all `dotnet` invocations?

KRuntime.sln
dougbu: > why can't this ~~be a~~ use the .kproj anymore?

both exist but it's not clear why

KRuntime.sln
davidfowl: It's really just clr

KRuntime.sln
davidfowl: Looks like a bug. Anything managed should be kproj. Anything else is native

scripts/kpm.cmd
davidfowl: +1, `DOTNET_OPTIONS`

davidfowl: Also we should simplify this script now. It's:

`dotnet "%~dp0lib\Microsoft.Framework.PackageManager.dll" %*`

src/dotnet.core45/dotnet.core45.cpp
```diff
@@ -136,7 +136,13 @@ HMODULE LoadCoreClr()
     errno_t errno = 0;
     bool fSuccess = true;
     TCHAR szKreTrace[1] = {};
-    bool m_fVerboseTrace = GetEnvironmentVariableW(L"KRE_TRACE", szKreTrace, 1) > 0;
+    // TODO: remove KRE_ env var
+	DWORD dwRet = GetEnvironmentVariableW(L"DOTNET_TRACE", szKreTrace, 1);
```
davidfowl: extra space here

davidfowl:
One big thing missed is the renaming of the nuspec files that actually create the KRE. Replace KRE with dotnet in the nuspec files and ids.

makefile.shade
```diff
@@ -268,15 +269,15 @@ var RC_FILES = '${FindAllFiles("Resource.rc", "klr", "klr.net45", "klr.core45")}
             {
                 var kreName = Path.GetFileNameWithoutExtension(nupkgPath);
                 var krePath = Path.Combine(TEST_RESULTS, "KRE", kreName);
-                Environment.SetEnvironmentVariable("K_APPBASE", helloWorld);
+                Environment.SetEnvironmentVariable("DOTNET_APPBASE", helloWorld);
```
davidfowl: We shouldn't have to set this anymore

makefile.shade
```diff
@@ -268,15 +269,15 @@ var RC_FILES = '${FindAllFiles("Resource.rc", "klr", "klr.net45", "klr.core45")}
...
-                var kPath = Path.Combine(krePath, "bin", "k");
+                var dotnetPath = Path.Combine(krePath, "bin", "dotnet");
                 var kpmPath = Path.Combine(krePath, "bin", "kpm");
 
                 Log.Info("kreName = " + kreName);
-                Log.Info("K_APPBASE = " + helloWorld);
+                Log.Info("DOTNET_APPBASE = " + helloWorld);
 
-                Exec("cmd", "/C " + kPath + " run");
+                Exec("cmd", "/C " + dotnetPath + " " + helloWorld + " run");
```
davidfowl: ```Exec(dotnet, helloWorld + " run");```

ChengTian:
Rebased and fixed naming (using runtime instead of dotnet)

ChengTian:
Some variables are stilling using the term `dotnet` instead of `runtime`. For exmaple, `DotnetHomeDir`, this variable is related to the `DOTNET_HOME` environment variable and I want to keep some consistency.

ChengTian:
@davidfowl @muratg , please check the two new commits

KRuntime.sln
anurse: Can we just have `dotnet.coreclr.dll` and `dotnet.fullclr.dll` ? `core45` makes even less sense because we are trying to avoid referring to a specific version of the CoreCLR

nuspec/dotnet-mono.nuspec
```diff
@@ -1,13 +1,13 @@
 <?xml version="1.0" encoding="utf-8"?>
 <package>
     <metadata>
-        <id>KRE-CLR-x86</id>
+        <id>dotnet-mono</id>
         <version>0.0.1-pre</version>
         <authors>Microsoft</authors>
         <requireLicenseAcceptance>false</requireLicenseAcceptance>
-        <description>The K Runtime Environment for Desktop CLR x86</description>
+        <description>The .NET Runtime Environment for Mono on OSX</description>
```
anurse: Just "for Mono" I think. It's the same package on Linux/BSD/OSX/etc. because we don't have a native host for Mono, we boot straight to managed code.

src/Microsoft.Framework.PackageManager/Packing/NativeImageGenerator.cs
```diff
@@ -126,7 +126,7 @@ public static NativeImageGenerator Create(PackOptions options, PackRoot root, IE
                 if (VersionUtility.IsDesktop(frameworkName))
                 {
                     options.Reports.Information.WriteLine(
-                        "Native image generation is only supported for KLR Core flavors.".Yellow());
+                        "Native image generation is only supported for dotnet core flavors.".Yellow());
```
anurse: "CoreCLR flavors"? Sounds a bit better I think... or maybe ".NET Core".

IMHO "dotnet" is really only for cases where ".NET" is inappropriate (package IDs, file/folder names, etc.)

ChengTian:
Thanks for @anurse 's feedbacks. Fixed in the latest commit.

test/Microsoft.Framework.ApplicationHost.FunctionalTests/Microsoft.Framework.ApplicationHost.FunctionalTests.kproj
```diff
@@ -14,4 +14,9 @@
     <SchemaVersion>2.0</SchemaVersion>
   </PropertyGroup>
   <Import Project="$(VSToolsPath)\AspNet\Microsoft.Web.AspNet.targets" Condition="'$(VSToolsPath)' != ''" />
-</Project>
+  <ProjectExtensions>
```
pranavkm: Undo

ChengTian: Thanks. Fixed.

src/Microsoft.Framework.PackageManager/Packing/PackManager.cs
```diff
@@ -117,27 +117,24 @@ public bool Package()
                 {
                     runtimeProbePaths = new List<string>();
                     runtimeProbePaths.Add(runtime);
-                    var kreHome = Environment.GetEnvironmentVariable("KRE_HOME");
-                    if (string.IsNullOrEmpty(kreHome))
+                    // TODO: remove KRE_ env var
+                    var dotnetHome = Environment.GetEnvironmentVariable("DOTNET_HOME") ?? Environment.GetEnvironmentVariable("KRE_HOME");
+                    if (string.IsNullOrEmpty(dotnetHome))
                     {
+                        var dotnetGlobalPath = Environment.GetEnvironmentVariable("DOTNET_GLOBAL_PATH");
```
anurse: FYI, Right now the only place this environment variable is set is in the `kvm` tests. It's an override, but if it isn't specified, it needs to have a default value.

If `DOTNET_GLOBAL_PATH` isn't set, this code should fall back to `%ProgramFiles%\Microsoft .NET Cross-Platform Runtime` (or whatever we finalize on... maybe put it in a constant to make it easier to move around :))

anurse:
Looking ok to me :shipit:

PR #1109, opened 2015-01-20
Use full path for appbase in dotnet.exe

ChengTian:
@davidfowl

### DIFF
--- a/src/dotnet/dotnet.cpp
+++ b/src/dotnet/dotnet.cpp
@@ -294,6 +294,23 @@ int CallFirmwareProcessMain(int argc, wchar_t* argv[])
         data.applicationBase = szCurrentDirectory;
     }
 
+    // Prevent coreclr native bootstrapper from failing with relative appbase
+    WCHAR szFullAppBase[MAX_PATH];
+    DWORD dwFullAppBase = GetFullPathNameW(data.applicationBase, MAX_PATH, szFullAppBase, nullptr);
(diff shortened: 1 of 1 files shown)


davidfowl:
@anurse can you move those files to be changed only on the CI, it's pretty annoying

davidfowl:
@ChengTian add those files to teh gitignore

src/dotnet/dotnet.cpp
```diff
@@ -294,6 +294,21 @@ int CallFirmwareProcessMain(int argc, wchar_t* argv[])
...
+    DWORD dwFullAppBase = GetFullPathNameW(data.applicationBase, MAX_PATH, szFullAppBase, nullptr);
+    if (!dwFullAppBase)
+    {
+        ::wprintf_s(L"Failed to get full path of application base: %S\r\n", data.applicationBase);
+        goto Finished;
+    }
+    else if (dwFullAppBase > MAX_PATH)
+    {
+        ::wprintf_s(L"Full path of application base is too long\r\n");
+        goto Finished;
```
davidfowl: Whats the exit code in this case? non 0?

ChengTian: Fixed. Returns 1 in this case

ChengTian:
Rebased on top of @anurse 's change to get rid of annoying resource file changes. Also reverted resource files changes.

davidfowl:
:shipit:

PR #1111, opened 2015-01-20
Rename klr.sh to dotnet.sh

### LINKED ISSUE #1110: 'dotnet' command does not exist on mono
The mono package for the runtime still has the 'klr' executable (it was missed because it's an 'sh' file instead of a native executable). I have a fix for this, but wanted to capture it here. PR coming soon.

anurse:
Fixes #1110


davidfowl:
:shipit:

davidfowl:
Ummmmm what about the build script?

anurse:
It just copies "*.sh" and then strips the ".sh". I checked :). Running a build locally to be sure. Won't merge until I've confirmed there.

PR #1112, opened 2015-01-21
Fix ServicingTable on mono

anurse:
/cc @lodejard @ChengTian @davidfowl - Specifically: Is it OK to leave ServicingIndex uninitialized? As far as I can tell, yes. It just acts like an empty index.

For now, we don't provide a "default" value for Linux. When we have one, we could add it here, but until then the null value is causing dotnet-mono to crash on startup.


davidfowl:
:shipit:

davidfowl:
@anurse I merge this in for you

anurse:
Cool, thanks. I was just about to do that :)

Issue #1115, opened 2015-01-21
KPM Pack: Generated bash scripts are broken


borgdylan:
Currently the last line in a bash script for k run would be:

```
exec "dotnet" Microsoft.Framework.ApplicationHost run "$@"
```

But it should be:

```
exec "dotnet" --appbase $DOTNET_APPBASE Microsoft.Framework.ApplicationHost run "$@"
```

The DOTNET_APPBASE variable is already defined in the scripts.

davidfowl:
@ChengTian @muratg

muratg:
@ChengTian let's get this once the branching is done.

PR #1117, opened 2015-01-21
Bash scripts generated by "kpm pack" should use '--appbase' option

ChengTian:
parent #1115


davidfowl:
Why is this in dev and not release?

ChengTian:
I thought I might be able to check in before beta3 branching. Sent a new PR: #1123

PR #1120, opened 2015-01-22
Fixing preprocess code failing to load a class library

Praburaj:
This fixes bug : https://github.com/aspnet/XRE/issues/1048

On Mono while loading the !preprocess assembly the ResolveEventArgs.RequestingAssembly is not populated.
As a result when code in !preprocess references a class library the load fails on the class library.
Below is a work around to try populate the requesting assembly.

/CC @davidfowl

### DIFF
--- a/src/Microsoft.Framework.Runtime.Loader/LoadContext.cs
+++ b/src/Microsoft.Framework.Runtime.Loader/LoadContext.cs
@@ -2,6 +2,9 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Reflection;
+#if ASPNET50
+using System.Linq;
+#endif
 #if ASPNETCORE50
 using System.Runtime.Loader;
 #endif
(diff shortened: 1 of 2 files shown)


davidfowl:
I think this fix needs to go into the release branch.

src/Microsoft.Framework.Runtime.Loader/LoadContext.cs
```diff
@@ -224,10 +227,30 @@ private Assembly ResolveAssembly(object sender, ResolveEventArgs args)
...
+                // Below is a work around to try populate the requesting assembly.
+
+                if (Microsoft.Framework.Runtime.PlatformHelper.IsMono &&
+                    requestingAssembly == null &&
+                    !assemblyName.Name.Contains("!"))
+                {
+                    // See if this loading is for !preprocess.
+                    var domain = (AppDomain)sender;
+
+                    requestingAssembly = domain.GetAssemblies()
```
davidfowl: Did the call context thing not work?

Praburaj:
This is an rc1 bug. So the PR is to dev branch.

Praburaj:
Do you think this needs to go in beta3?

davidfowl:
Yes

src/Microsoft.Framework.Runtime.Loader/LoadContext.cs
Praburaj: I couldn't find a better place to hack this and this looked like the simplest fix with minimal impact. This code automatically no-ops if mono fixes the bug as `requestingAssembly` becomes non `null`.

davidfowl: I don't like it. I'm playing around with it now.

src/Microsoft.Framework.Runtime.Loader/LoadContext.cs
```diff
@@ -224,10 +227,30 @@ private Assembly ResolveAssembly(object sender, ResolveEventArgs args)
...
+                var requestingAssembly = args.RequestingAssembly;
+
+#if ASPNET50
+                // On Mono while loading the !preprocess assembly the ResolveEventArgs.RequestingAssembly is not populated. 
+                // As a result when code in !preprocess references a class library the load fails on the class library.
+                // Below is a work around to try populate the requesting assembly.
+
+                if (Microsoft.Framework.Runtime.PlatformHelper.IsMono &&
+                    requestingAssembly == null &&
+                    !assemblyName.Name.Contains("!"))
```
Praburaj: actually this can be `"!preprocess"`

Tragetaschen:
Yes, please

src/Microsoft.Framework.Runtime.Loader/LoadContext.cs
Praburaj: sure let me know if this PR should be closed.

davidfowl: not yet

borgdylan:
Is this also related to loading compiler libraries during kpm operations?

Praburaj:
Nope. This has a specific scope. 

Mono + kpm build + Preprocessor code referencing other class libraries.

This PR is going to be closed as this is a mono bug getting fixed in mono. Refer https://github.com/aspnet/XRE/issues/1048 for mono PR details.

davidfowl:
Prabu, it's the same bug. It will fix both issues

Praburaj:
I see. Good to know.

Praburaj:
@Tragetaschen's fix to mono has gone into mono right now : https://github.com/mono/mono/commit/46b5b849ae2e6e71326030e3a6806d59cdb0eb4a

This issue should disappear when running on an upcoming mono release or mono built from sources. Closing this PR.

PR #1124, opened 2015-01-22
Rename kpm commands in release

ChengTian:
parent #967 
Replacing #1037 because we are doing this in release branch now

@davidfowl , please help to review, we should checkout this in ASAP.


davidfowl:
:shipit:

PR #1128, opened 2015-01-23
Map aspnet50 to net451

davidfowl:
/cc @ChengTian @Eilon


ChengTian:
:shipit:

jskulavik:
Is this permanent?  Will there be additional mappings for 4.6+ or will aspnet50 be remapped to forthcoming framework releases?

davidfowl:
@jskulavik No, there'll be a way to target different full framework runtimes. Just not in beta3

Issue #979, opened 2014-12-09
"kpm pack" should consider wrapped projects as special cases


ChengTian:
@davidfowl found a bug when we `kpm pack` an app which references results of `kpm wrap`. Reproduce steps:
1. `kpm wrap` a csproj
2. Reference the wrapped csproj from a kproj
3. `kpm pack` the kproj
4. Run the packed app

The packed app cannot run because `kpm pack` only copies source code and it doesn't consider `assembly` & `pdb` in project.json of dependency projects.

Here is the solution @davidfowl gives:
`kpm pack --no-source` can resolve this problem. We can automatically "no source" the wrapped projects. That is, if you declare `assembly` and `pdb`, we treat the project like a package and build package out of it during `kpm pack`.

PR #1136, opened 2015-01-27
Special case wrapper projects during "kpm bundle"

ChengTian:
parent #979

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Bundle/BundleProject.cs
+++ b/src/Microsoft.Framework.PackageManager/Bundle/BundleProject.cs
@@ -38,13 +38,14 @@ namespace Microsoft.Framework.PackageManager.Bundle
         public string TargetPath { get; private set; }
         public string WwwRoot { get; set; }
         public string WwwRootOut { get; set; }
+        public bool IsPackage { get; private set; }
 
         public void Emit(BundleRoot root)
         {
             root.Reports.Quiet.WriteLine("Using {0} dependency {1} for {2}", _libraryDescription.Type,
                 _libraryDescription.Identity, _libraryDescription.Framework.ToString().Yellow().Bold());
 
-            if (root.NoSource)
+            if (root.NoSource || IsWrappingAssembly())
             {
                 EmitNupkg(root);
             }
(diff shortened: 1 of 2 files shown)


src/Microsoft.Framework.PackageManager/Bundle/BundleProject.cs
```diff
@@ -507,5 +501,63 @@ private string BasePath(string relativePath)
...
+            {
+              "frameworks": {
+                "net45": {
+                  "bin": {
+                    "assembly": "relative/path/to/ClassLibrary1.dll"
+                  }
+                }
+              }
+            } */
+            var rootObj = JObject.Parse(File.ReadAllText(project.ProjectFilePath));
```
davidfowl: Why aren't you using the project's object model to check for the bin path?

src/Microsoft.Framework.PackageManager/Bundle/BundleProject.cs
```diff
@@ -507,5 +501,63 @@ private string BasePath(string relativePath)
             var index2 = (relativePath + Path.AltDirectorySeparatorChar).IndexOf(Path.AltDirectorySeparatorChar);
             return relativePath.Substring(0, Math.Min(index1, index2));
         }
+
+        private Runtime.Project ResolveProject(string name)
+        {
+            Runtime.Project project;
+            if (!_projectResolver.TryResolveProject(name, out project))
+            {
+                throw new Exception("TODO: unable to resolve project named " + name);
```
victorhurdugaci: I know it is unrelated but I think we should start addressing these TODO error messages

src/Microsoft.Framework.PackageManager/Bundle/BundleProject.cs
```diff
@@ -507,5 +501,63 @@ private string BasePath(string relativePath)
...
+            } */
+            var rootObj = JObject.Parse(File.ReadAllText(project.ProjectFilePath));
+            var frameworksObj = rootObj["frameworks"] as JObject;
+
+            if (frameworksObj == null)
+            {
+                return false;
+            }
+
+            foreach (var frameworkProp in frameworksObj.Properties())
```
victorhurdugaci: This loop executes once at most in all cases (there is no break/continue in it and return is the last statement)... Why is it a loop?

ChengTian:
Thanks for the feedback. Please check the updated commit.

ChengTian:
@davidfowl , please check the updated commit. I verified both `kpm bundle` and `kpm bundle --no-source` with wrapped assembly

src/Microsoft.Framework.PackageManager/Bundle/BundleProject.cs
```diff
@@ -38,13 +38,14 @@ public class BundleProject
         public string TargetPath { get; private set; }
         public string WwwRoot { get; set; }
         public string WwwRootOut { get; set; }
+        public bool BecomesNupkg { get; private set; }
```
davidfowl: This name isn't good. IsPackage?

ChengTian: Fixed

davidfowl:
:shipit:

PR #1139, opened 2015-01-28
Install global

victorhurdugaci:
Change: 
- The install global command. Named temporary `install-g`.
- Tests for `install-g`
- Moved feed options in a centralized place

Will address uninstall in a different PR.

Please review: @davidfowl @loudej


borgdylan:
Will bash scripts be made like it is the case with kpm bundle?

victorhurdugaci:
@borgdylan yes but for now (as in the code checked in so far) is only cmd.

victorhurdugaci:
I said that uninstall with go in a later iteration but I have a working prototype of it and I checked in

Fixes https://github.com/aspnet/XRE/issues/1140

victorhurdugaci:
Closing for now because I have more changes to add

PR #1157, opened 2015-02-03
ProjectResolver to look only for global.json

Praburaj:
Currently it looks for both a global.json and .sln file. With this change any projects without
a global.json file but only a .sln file will break.

@davidfowl

### DIFF
--- a/src/Microsoft.Framework.Runtime/ProjectResolver.cs
+++ b/src/Microsoft.Framework.Runtime/ProjectResolver.cs
@@ -42,7 +42,7 @@ namespace Microsoft.Framework.Runtime
 
             return false;
         }
-        
+
         private IEnumerable<string> ResolveSearchPaths(string projectPath, string rootPath)
         {
             var paths = new List<string>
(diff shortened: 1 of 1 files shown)


src/Microsoft.Framework.Runtime/ProjectResolver.cs
```diff
@@ -69,8 +69,7 @@ public static string ResolveRootDirectory(string projectPath)
 
             while (di.Parent != null)
             {
-                if (di.EnumerateFiles(GlobalSettings.GlobalFileName).Any() ||
-                    di.EnumerateFiles("*.sln").Any())
+                if (di.EnumerateFiles(GlobalSettings.GlobalFileName).Any())
```
davidfowl: Do this in a more optimized way:

```C#
var candidate = Path.Combine(di.FullName, GlobalSettings.GlobalFileName);
if (File.Exists(candidate))
{
  return di.FullName;
}
```

Praburaj: Good point. I'll fix it.

Praburaj:
Sending a different PR.

PR #1163, opened 2015-02-04
Fix loaded module name in trace info

ChengTian:
parent #1164 
After upgrade of C++ compiler, the name of loaded module in KRE_TRACE info becomes incomplete. Seems the new C++ compiler is using a different spec for wide-char string formatted output.


muratg:
Do you need to update kre.coreclr.cpp as well?

ChengTian:
Thanks for the reminder. Fixed kre.coreclr.cpp and filed an issue.

ChengTian:
Closing this because we will have this in next milestone. New PR targeting `dev` branch is at: https://github.com/aspnet/XRE/pull/1165

Issue #1164, opened 2015-02-04
Loaded module name is incomplete in KRE_TRACE info


ChengTian:
If you turn on KRE_TRACE and then run `klr.exe`, you will see:
![capture1](https://cloud.githubusercontent.com/assets/1383883/6045275/c57b7e24-ac4d-11e4-9584-309e71028c5c.PNG)

Note that “Loaded Module” name is incomplete. It should be
![capture2](https://cloud.githubusercontent.com/assets/1383883/6045285/d02c45a6-ac4d-11e4-8945-8dc4d2dab39e.PNG)
or
![capture3](https://cloud.githubusercontent.com/assets/1383883/6045288/d4e78bc8-ac4d-11e4-8724-f87f75fdd1de.PNG)
based on the flavor of runtime in use.

PR #1165, opened 2015-02-04
Fix formatted output with wide-char strings in native bootstrappers

ChengTian:
parent #1164 

@muratg , this is the new PR targeting `dev` branch.


muratg:
:shipit:

PR #1166, opened 2015-02-04
ProjectResolver to look only for global.json

Praburaj:
Fixes : https://github.com/aspnet/XRE/issues/1153

Currently it looks for both a global.json and .sln file. With this change any projects without
a global.json file but only a .sln file will break.

@davidfowl


Praburaj:
@davidfowl - good to go in this PR?

davidfowl:
:shipit:

Praburaj:
Merged : https://github.com/aspnet/XRE/commit/b4a87b7669baacfe840e22e7040180bd78d14cdc

PR #1168, opened 2015-02-04
Use kpm list in makefile

troydai:


troydai:
@davidfowl

makefile.shade
```diff
@@ -628,85 +541,97 @@ var RC_FILES = '${FindAllFiles("Resource.rc", BOOTSTRAPPER_EXE_NAME, BOOTSTRAPPE
...
+                        if (runtimeAssemblies.Add(file))
+                        {
+                            var targetFilePath = Path.Combine(target.TargetFolder, Path.GetFileName(file));
+                            Log.Info("Copying " + file + " to " + targetFilePath);
+                            File.Copy(file, targetFilePath, true);
+                        }
                     }
                 }
-                else
+                catch (Exception ex)
```
davidfowl: When would this throw?

makefile.shade
```diff
@@ -751,6 +698,62 @@ functions @{
...
+        }
+
+        int result;
+        if (!IsMono)
+        {
+            result = ExecuteAndRedirectOutput("cmd", string.Format("/C kpm {0}", argument), tmpFile);
+        }
+        else 
+        {
+            result = ExecuteAndRedirectOutput("kpm", argument, tmpFile);
```
davidfowl: Call it ``ExecuteAndCaptureOutput `. What about exit codes? Should it return a tuple?

makefile.shade
```diff
@@ -628,85 +541,97 @@ var RC_FILES = '${FindAllFiles("Resource.rc", BOOTSTRAPPER_EXE_NAME, BOOTSTRAPPE
...
-            
-            var runtimeDeps = deps["Runtime"];
-            
-            foreach(var targetClr in RUNTIME_CORECLR_TARGETS)
+
+            // copy required CoreCLR assemblies
+            if (target.Framework == "aspnetcore50")
             {
-                foreach(var file in Directory.EnumerateFiles(targetClr, "System.*.dll"))
+                foreach (var file in Directory.GetFiles(coreclrFolder).Where(f => !Path.GetFileName(f).StartsWith("System")))
```
davidfowl: Why this change?

davidfowl:
:shipit:

makefile.shade
troydai: File copy failure.

makefile.shade
troydai: These are different. Originally it copies all files and then go delete `System.*.dll` if it doesn't exist in dependency list. Now it copy all files except `System.*.dll` which are copied in previous logic.

troydai:
rebase and verified. will merge in seconds

PR #1178, opened 2015-02-06
Convert functional tests in makefile to xUnit test cases

ChengTian:
parent #884 

Added tests for simplified `klr.exe` usage as well.


davidfowl:
:shipit: !

PR #1203, opened 2015-02-12
Remove ANIs and start splitting runtime

anurse:
* ANIs removed and converted to standard interfaces (this will break the world, @pranavkm and I will go around a fix it :))
* Split runtime at AppHost to prepare for some major refactoring - Right now the "new" runtime doesn't really exist :)

[There's a minor change coming to add something to add M.F.R.Interfaces to the dependencies copied in to the kre folder, stand by for that]


src/kre.clr.managed/kre.clr.managed.kproj
```diff
@@ -13,6 +13,9 @@
   <PropertyGroup>
     <SchemaVersion>2.0</SchemaVersion>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
```
pranavkm: Intentional?

pranavkm:
:shipit:

src/kre.clr.managed/kre.clr.managed.kproj
anurse: No... VS did it :(. I'll remove.

davidfowl:
:shipit:

PR #1205, opened 2015-02-12
Wait for debugger flag on KLR

victorhurdugaci:
Fixes https://github.com/aspnet/XRE/issues/1189

@davidfowl @troydai

### DIFF
--- a/src/klr/klr.cpp
+++ b/src/klr/klr.cpp
@@ -87,6 +87,7 @@ int BootstrapperOptionValueNum(LPCWSTR pszCandidate)
         return 1;
     }
     else if (StringsEqual(pszCandidate, L"--watch") ||
+        StringsEqual(pszCandidate, L"--debug") ||
         StringsEqual(pszCandidate, L"--help") ||
         StringsEqual(pszCandidate, L"-h") ||
         StringsEqual(pszCandidate, L"-?") ||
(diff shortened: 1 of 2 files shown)


davidfowl:
You need to so it for the mono host as well.

anurse:
:+1: DO WANT!

:shipit: with mono host code. I usually use something like this when doing the same thing in Managed code: https://github.com/NuGet/NuGet.Operations/blob/master/src/NuCmd/Program.cs#L38

victorhurdugaci:
@davidfowl - added mono support. 

Btw, who's debugging on mono? :trollface:

src/klr/klr.cpp
```diff
@@ -373,5 +374,25 @@ int CallFirmwareProcessMain(int argc, wchar_t* argv[])
...
+        {
+            if (!IsDebuggerPresent())
+            {
+                ::wprintf_s(L"Waiting for the debugger to attach...\r\n");
+
+                // Do not use getchar() like in corerun because it doesn't work well with remote sessions
+                while (!IsDebuggerPresent())
+                {
+                    Sleep(250);
+                }
```
troydai: Should we have set a time limitation, when beyond the limit the program should keep running without debugger?

troydai: And should be trigger a `DebugBreak` after the debugger is present so that the user get the chance to control the process?

victorhurdugaci: Time limitation: no. If you use the debug flag, it can be annoying if you leave it too much and then realize that the program ran. Also, it adds unnecessary complexity.

Debug break: no again. My initial code had that but then I realized that most of the times you set a breakpoint somewhere in the code and you expect that one to be hit first. You don't want every time you run the program to get a breakpoint before the one that you set, a breakpoint that you cannot disable.

troydai: Fair. We can revisit the discussion after a few weeks' experience.

davidfowl:
@victorhurdugaci can you print the process id as well? Since we have a single host, it will help identify which klr.exr we need to attach to

victorhurdugaci:
@davidfowl updated to display the process id

borgdylan:
@victorhurdugaci I do debugging on mono, since my compiler can emit debug symbols for mono

troydai:
:shipit:

PR #1216, opened 2015-02-15
Check source folder and files before copy

troydai:


davidfowl:
:shipit:

davidfowl:
Do we need this for the sh scripts?

troydai:
We'll know in a few minutes. I'll check in this first.

PR #1226, opened 2015-02-16
Issue1119: Fallback to runtime framework when project doesn't list any frameworks

troydai:
Addressing issue: https://github.com/aspnet/XRE/issues/1119

@Praburaj @glennc

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -518,7 +518,7 @@ var RC_FILES = '${FindAllFiles("Resource.rc", BOOTSTRAPPER_EXE_NAME, BOOTSTRAPPE
             var runtimeAssemblies = new HashSet<string>();
             foreach (var project in runtimeProjects)
             {
-                if (project == "dotnet.core45.managed" && target.Framework == "aspnet50")
+                if (project == BOOTSTRAPPER_CORECLR_MANAGED_NAME && target.Framework == "aspnet50")
                     continue;
 
                 var projectFolder = Path.Combine(ROOT, "src", project);
--- a/src/Microsoft.Framework.PackageManager/List/DependencyListCommand.cs
+++ b/src/Microsoft.Framework.PackageManager/List/DependencyListCommand.cs
@@ -1,40 +1,36 @@
 ﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.Collections.Generic;
-using System.Linq;
 using System.Runtime.Versioning;
+using Microsoft.Framework.PackageManager.Utils;
 
 namespace Microsoft.Framework.PackageManager.List
 {
     internal class DependencyListCommand
     {
         private readonly DependencyListOptions _options;
+        private readonly FrameworkName _fallbackFramework;
 
-        public DependencyListCommand(DependencyListOptions options)
+        public DependencyListCommand(DependencyListOptions options, FrameworkName fallbackFramewok)
         {
             _options = options;
+            _fallbackFramework = fallbackFramewok;
         }
 
         public int Execute()
         {
-            var result = 0;
             _options.Reports.Information.WriteLine("Listing dependencies for {0} ({1})", _options.Project.Name, _options.Project.ProjectFilePath);
 
-            var frameworks = new HashSet<FrameworkName>(_options.Project.GetTargetFrameworks().Select(f => f.FrameworkName));
-            if (_options.Framework != null)
+            string frameworkSelectionError;
+            var frameworks = FrameworkSelectionHelper.SelectFrameworks(_options.Project, 
+                                                                       _options.TargetFrameworks, 
+                                                                       _fallbackFramework, 
+                                                                       out frameworkSelectionError);
+            if (frameworks == null)
             {
-                if (frameworks.Contains(_options.Framework))
-                {
-                    frameworks.Clear();
-                    frameworks.Add(_options.Framework);
-                }
-                else
-                {
-                    _options.Reports.Error.WriteLine("Project doesn't support framework: {0}", _options.Framework.FullName);
-                    return 0;
-                }
-            }
+                _options.Reports.Error.WriteLine(frameworkSelectionError);
+                return 1;
+            } 
 
             foreach (var framework in frameworks)
             {
@@ -47,7 +43,7 @@ namespace Microsoft.Framework.PackageManager.List
                 }
             }
 
-            return result;
+            return 0;
         }
     }
 }
--- a/src/Microsoft.Framework.PackageManager/Utils/FrameworkSelectionHelper.cs
+++ b/src/Microsoft.Framework.PackageManager/Utils/FrameworkSelectionHelper.cs
@@ -1,0 +1,63 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.Versioning;
+using Microsoft.Framework.Runtime.Helpers;
+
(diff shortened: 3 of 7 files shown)


src/Microsoft.Framework.PackageManager/List/DependencyListCommand.cs
```diff
@@ -4,35 +4,50 @@
...
-            var result = 0;
             _options.Reports.Information.WriteLine("Listing dependencies for {0} ({1})", _options.Project.Name, _options.Project.ProjectFilePath);
 
             var frameworks = new HashSet<FrameworkName>(_options.Project.GetTargetFrameworks().Select(f => f.FrameworkName));
-            if (_options.Framework != null)
+
+            if (!frameworks.Any())
             {
-                if (frameworks.Contains(_options.Framework))
+                if (_options.Framework == null)
```
Praburaj: do you need this check if frameworks is empty list?

troydai: Here the `frameworks` must be an empty list.

troydai:
@davidfowl

troydai:
Ping

src/Microsoft.Framework.PackageManager/List/DependencyListCommand.cs
davidfowl: Get rid of this code and use shared code with kpm build.

troydai:
Ping @davidfowl

troydai:
Ping @davidfowl

src/Microsoft.Framework.PackageManager/Utils/FrameworkSelectionHelper.cs
```diff
@@ -0,0 +1,65 @@
...
+            {
+                if (!projectFrameworks.Contains(framework.Value))
+                {
+                    errorMessage = framework.Key + " is not specified in project.json";
+                    return false;
+                }
+            }
+
+            errorMessage = string.Empty;
+            return success;
```
Praburaj: nit: do you need this variable. Just return `true` or use it everywhere.

src/Microsoft.Framework.PackageManager/List/DependencyListCommand.cs
```diff
@@ -1,40 +1,36 @@
...
-            var result = 0;
             _options.Reports.Information.WriteLine("Listing dependencies for {0} ({1})", _options.Project.Name, _options.Project.ProjectFilePath);
 
-            var frameworks = new HashSet<FrameworkName>(_options.Project.GetTargetFrameworks().Select(f => f.FrameworkName));
-            if (_options.Framework != null)
+            string frameworkSelectionErrors;
+            var frameworks = FrameworkSelectionHelper.SelectFrameworks(_options.Project, 
+                                                                       _options.TargetFrameworks, 
+                                                                       _fallbackFramework, 
+                                                                       out frameworkSelectionErrors);
```
Praburaj: nit: `frameworkSelectionErrors` => `error` or `frameworkSelectionError`, If you always populate a single error.

Praburaj:
:shipit: after above comments.

src/Microsoft.Framework.PackageManager/Utils/FrameworkSelectionHelper.cs
troydai: will update.

src/Microsoft.Framework.PackageManager/List/DependencyListCommand.cs
troydai: will update to `frameworkSelectionError`

troydai:
Ping @davidfowl

troydai:
Bring in a few more pairs of eyes here: @victorhurdugaci @anurse @ChengTian

makefile.shade
```diff
@@ -508,7 +508,7 @@ var RC_FILES = '${FindAllFiles("Resource.rc", BOOTSTRAPPER_EXE_NAME, BOOTSTRAPPE
             var runtimeAssemblies = new HashSet<string>();
             foreach (var project in runtimeProjects)
             {
-                if (project == "dotnet.core45.managed" && target.Framework == "aspnet50")
+                if (project == BOOTSTRAPPER_CORECLR_MANAGED_NAME && target.Framework == "aspnet50")
```
victorhurdugaci: `aspnet50` -> constant?

src/Microsoft.Framework.PackageManager/Utils/FrameworkSelectionHelper.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.Versioning;
+using Microsoft.Framework.Runtime.Helpers;
+
+namespace Microsoft.Framework.PackageManager.Utils
+{
+    public static class FrameworkSelectionHelper
+    {
+        public static IEnumerable<FrameworkName> SelectFrameworks(Runtime.Project project,
```
victorhurdugaci: When there are no matches, return an empty collection, not null. It makes the consumer code simpler because instead of having to write something like:

```
var result = SelectFrameworks(....);
if (result != null)
{
     foreach(var fx in result) { ... }
}
```

You can write the for loop without the null check.

src/Microsoft.Framework.PackageManager/Utils/FrameworkSelectionHelper.cs
```diff
@@ -0,0 +1,63 @@
...
+        private static bool ValidateFrameworks(HashSet<FrameworkName> projectFrameworks,
+                                               IDictionary<string, FrameworkName> specifiedFrameworks,
+                                               out string errorMessage)
+        {
+            foreach (var framework in specifiedFrameworks)
+            {
+                if (!projectFrameworks.Contains(framework.Value))
+                {
+                    errorMessage = framework.Key + " is not specified in project.json";
+                    return false;
```
victorhurdugaci: Idea: rather than stopping at the first error, collect all possible errors and then return. Why? Because if you have 2 errors, you get only one of them, you fix it and only then you get the second one.

makefile.shade
troydai: so far we use the string in make file directly. no a good practice but I don't intent to tackle this issue in this PR so as to keep the fix coherent. #1244 is just filed to track this issue.

src/Microsoft.Framework.PackageManager/Utils/FrameworkSelectionHelper.cs
troydai: The function needs to indicate caller whether frameworks are correctly selected. Returns empty collection could mislead the caller.

src/Microsoft.Framework.PackageManager/Utils/FrameworkSelectionHelper.cs
troydai: Nice idea but I won't adopt here. The price for figure out all the issues is low: remove the string from the command line. But to collect the errors and print them in a readable format is harder doesn't seem to bring that much of value.

src/Microsoft.Framework.PackageManager/Utils/FrameworkSelectionHelper.cs
victorhurdugaci: That's why you have the error list. >0 errors = something is wrong

troydai:
Ping @davidfowl , updates addressed your review feedbacks. Please take another look.

src/Microsoft.Framework.PackageManager/Utils/FrameworkSelectionHelper.cs
troydai: Well, then in this case consumer needs to create a list to store error messages and also need to find a proper way to display the list of messages, which defeats the purpose of simplify consumer codes. 

Both pattern works, they're generally the equivalent. Just one is already adopted and I don't see a strong case to change to another pattern.

PR #1234, opened 2015-02-17
Ignore unlisted packages

troydai:
https://github.com/aspnet/XRE/issues/1208

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
@@ -21,6 +21,10 @@ namespace Microsoft.Framework.PackageManager.Restore.NuGet
         private static readonly XName _xnameProperties = XName.Get("properties", "http://schemas.microsoft.com/ado/2007/08/dataservices/metadata");
         private static readonly XName _xnameId = XName.Get("Id", "http://schemas.microsoft.com/ado/2007/08/dataservices");
         private static readonly XName _xnameVersion = XName.Get("Version", "http://schemas.microsoft.com/ado/2007/08/dataservices");
+        private static readonly XName _xnamePublish = XName.Get("Published", "http://schemas.microsoft.com/ado/2007/08/dataservices");
+
+        // An unlisted package's publish time must be 1900-01-01T00:00:00.
+        private static readonly DateTime _unlistedPublishedTime = new DateTime(1900, 1, 1, 0, 0, 0);
 
         private readonly string _baseUri;
         private readonly Reports _reports;
@@ -89,7 +93,7 @@ namespace Microsoft.Framework.PackageManager.Restore.NuGet
                     var page = 1;
                     while (true)
                     {
-                        // TODO: Pages for a package Id are cahced separately.
+                        // TODO: Pages for a package Id are cached separately.
                         // So we will get inaccurate data when a page shrinks.
                         // However, (1) In most cases the pages grow rather than shrink;
                         // (2) cache for pages is valid for only 30 min.
@@ -104,7 +108,8 @@ namespace Microsoft.Framework.PackageManager.Restore.NuGet
 
                                 var result = doc.Root
                                     .Elements(_xnameEntry)
-                                    .Select(x => BuildModel(id, x));
+                                    .Select(x => BuildModel(id, x))
+                                    .Where(x => x != null);
 
                                 results.AddRange(result);
 
@@ -169,6 +174,16 @@ namespace Microsoft.Framework.PackageManager.Restore.NuGet
             var properties = element.Element(_xnameProperties);
             var idElement = properties.Element(_xnameId);
             var titleElement = element.Element(_xnameTitle);
+
+            var publishElement = properties.Element(_xnamePublish);
+            if (publishElement != null)
+            {
+                DateTime publishDate; 
+                if (DateTime.TryParse(publishElement.Value, out publishDate) && (publishDate == _unlistedPublishedTime))
+                {
+                    return null; 
+                }
+            }
 
             return new PackageInfo
             {
--- a/src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
@@ -41,7 +41,7 @@ namespace Microsoft.Framework.PackageManager
         public string PackageFolder { get; set; }
         public string GlobalJsonFile { get; set; }
 
-        public string RestorePackageId { get; set; }
+        public string RestorePackageId { get; set; } 
         public string RestorePackageVersion { get; set; }
 
         public string AppInstallPath { get; private set; }
--- a/src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
@@ -155,7 +155,7 @@ namespace Microsoft.Framework.PackageManager.Restore.NuGet
                     }
                 }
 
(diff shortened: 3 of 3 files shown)


troydai:
@danroth27 @davidfowl @jeffhandley

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
```diff
@@ -30,7 +30,6 @@ public RestoreCommand(IApplicationEnvironment env)
         }
```
davidfowl: Can you not format unrelated files :smile: ?

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
```diff
@@ -21,6 +21,10 @@ public class NuGetv2Feed : IPackageFeed
         private static readonly XName _xnameProperties = XName.Get("properties", "http://schemas.microsoft.com/ado/2007/08/dataservices/metadata");
         private static readonly XName _xnameId = XName.Get("Id", "http://schemas.microsoft.com/ado/2007/08/dataservices");
         private static readonly XName _xnameVersion = XName.Get("Version", "http://schemas.microsoft.com/ado/2007/08/dataservices");
+        private static readonly XName _xnamePublish = XName.Get("Published", "http://schemas.microsoft.com/ado/2007/08/dataservices");
+
+        // An unlisted package's publish time mush be 1900-01-01T00:00:00.
+        private static readonly DateTime _unlistedPublishedTime = DateTime.Parse("1900-01-01T00:00:00");
```
davidfowl: Why are you parsing this instead of just newing it up?

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
```diff
@@ -170,15 +175,24 @@ public PackageInfo BuildModel(string id, XElement element)
...
+                {
+                    // If 'Id' element exist, use its value as accurate package Id
+                    // Otherwise, use the value of 'title' if it exist
+                    // Use the given Id as final fallback if all elements above don't exist
+                    Id = idElement?.Value ?? titleElement?.Value ?? id,
+                    Version = SemanticVersion.Parse(properties.Element(_xnameVersion).Value),
+                    ContentUri = element.Element(_xnameContent).Attribute("src").Value,
+                };
+            }
+            else
```
davidfowl: Remove the else

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
```diff
@@ -170,15 +175,24 @@ public PackageInfo BuildModel(string id, XElement element)
             var idElement = properties.Element(_xnameId);
             var titleElement = element.Element(_xnameTitle);
 
-            return new PackageInfo
+            var isUnlisted = DateTime.Parse(properties.Element(_xnamePublish).Value) == _unlistedPublishedTime;
+
+            if (!isUnlisted)
```
davidfowl: invert the logic, it makes it clearer.

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
```diff
@@ -170,15 +175,24 @@ public PackageInfo BuildModel(string id, XElement element)
             var idElement = properties.Element(_xnameId);
             var titleElement = element.Element(_xnameTitle);
 
-            return new PackageInfo
+            var isUnlisted = DateTime.Parse(properties.Element(_xnamePublish).Value) == _unlistedPublishedTime;
```
davidfowl: Code defensively here. Use TryParse and verify that the value isn't null.

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
troydai: I ... just ... can't ... help ... doing ... that ...

davidfowl: Do it in a stand alone pull request after.

troydai:
@jeffhandley @davidfowl 
A few questions:
1. Can `d:Published` missing?
2. In the case if it is missing, can I regard this package as listed?

troydai:
Ping @davidfowl @danroth27 @jeffhandley

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
```diff
@@ -21,6 +21,10 @@ public class NuGetv2Feed : IPackageFeed
         private static readonly XName _xnameProperties = XName.Get("properties", "http://schemas.microsoft.com/ado/2007/08/dataservices/metadata");
         private static readonly XName _xnameId = XName.Get("Id", "http://schemas.microsoft.com/ado/2007/08/dataservices");
         private static readonly XName _xnameVersion = XName.Get("Version", "http://schemas.microsoft.com/ado/2007/08/dataservices");
+        private static readonly XName _xnamePublish = XName.Get("Published", "http://schemas.microsoft.com/ado/2007/08/dataservices");
+
+        // An unlisted package's publish time mush be 1900-01-01T00:00:00.
```
BrennanConroy: mush -> must

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
```diff
@@ -170,6 +175,23 @@ public PackageInfo BuildModel(string id, XElement element)
             var idElement = properties.Element(_xnameId);
             var titleElement = element.Element(_xnameTitle);
```
BrennanConroy: Move these below the publishElement checks since they aren't used until after the checks?

jeffhandley:
I don't think Published can be missing, but it can't hurt to guard against it in case someone's custom NuGet feed doesn't have it.  I would interpret no Published field to Listed, as that would mean they cannot differentiate between listed and unlisted.

davidfowl:
+1 @jeffhandley Make the change @troydai and :shipit:

troydai:
Thanks for the advice. I'll update accordingly.

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
```diff
@@ -170,6 +175,16 @@ public PackageInfo BuildModel(string id, XElement element)
...
 
+            var publishElement = properties.Element(_xnamePublish);
+            if (publishElement != null)
+            {
+                DateTime publishDate; 
+                if (DateTime.TryParse(publishElement.Value, out publishDate) && (publishDate == _unlistedPublishedTime))
```
victorhurdugaci: Can you make `IsPublished` a property of the `PackageInfo` class? It might be useful for other purposes, not just for this check

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
```diff
@@ -170,6 +175,16 @@ public PackageInfo BuildModel(string id, XElement element)
...
+                {
+                    return null;
```
victorhurdugaci: If you do what I said above, then you can return the model always and then only select those that are published

troydai: I'm not sure about changing the `PackageInfo` interface. The `IsPublish` value maybe useful, but we don't know yet. We can get back to save the value once we have a case.

PR #1235, opened 2015-02-17
Servicing index use latest entry

victorhurdugaci:
Fixes https://github.com/aspnet/XRE/issues/1170

Please review @ChengTian @lodejard

### DIFF
--- a/src/Microsoft.Framework.Runtime/Servicing/ServicingIndex.cs
+++ b/src/Microsoft.Framework.Runtime/Servicing/ServicingIndex.cs
@@ -10,13 +10,13 @@ namespace Microsoft.Framework.Runtime.Servicing
 {
     public class ServicingIndex
     {
-        string _folderPath;
-        Dictionary<EntryKey, Entry> _entries = new Dictionary<EntryKey, Entry>();
+        private string _patchesRootFolder;
+        private Dictionary<EntryKey, Entry> _entries = new Dictionary<EntryKey, Entry>();
 
-        public void Initialize(string folderPath)
+        public void Initialize(string servicingFolder)
         {
-            _folderPath = folderPath;
-            var indexFilePath = Path.Combine(folderPath, "index.txt");
+            _patchesRootFolder = servicingFolder;
+            var indexFilePath = Path.Combine(servicingFolder, "index.txt");
             if (!File.Exists(indexFilePath))
             {
                 Logger.TraceInformation("[{0}]: Servicing index not found at {1}", GetType().Name, indexFilePath);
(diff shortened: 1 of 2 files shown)


src/Microsoft.Framework.Runtime/Servicing/ServicingIndex.cs
```diff
@@ -83,22 +90,30 @@ public void Initialize(string folderPath)
...
             if (_entries.TryGetValue(new EntryKey(packageId, packageVersion), out entry))
             {
                 var normalizedAssetPath = assetPath.Replace('\\', '/');
+
+                // Go through the entire list of mappings and use the latest one
                 foreach (var mapping in entry.Mappings)
                 {
                     if (string.Equals(normalizedAssetPath, mapping.AssetPath, StringComparison.OrdinalIgnoreCase))
                     {
-                        replacementPath = Path.Combine(_folderPath, mapping.ReplacementPath);
```
victorhurdugaci: Here is the actual fix

src/Microsoft.Framework.Runtime/Servicing/ServicingIndex.cs
```diff
@@ -28,54 +28,61 @@ public void Initialize(string folderPath)
...
-                        var parts = line.Split(new[] { '=' }, 2);
-                        if (parts.Length != 2)
+                    if (line.StartsWith("#") || string.IsNullOrWhiteSpace(line))
+                    {
+                        continue;
+                    }
+                    var parts = line.Split(new[] { '=' }, 2);
+                    if (parts.Length != 2)
+                    {
+                        Logger.TraceInformation("[{0}]. Line ({1}): malformed servicing file", GetType().Name, lineNumber);
```
troydai: Should the trace information to be `[{0}]: Line ....` ?

src/Microsoft.Framework.Runtime/Servicing/ServicingIndex.cs
```diff
@@ -28,54 +28,61 @@ public void Initialize(string folderPath)
...
+                        continue;
+                    }
+                    parts[0] = parts[0].Trim();
+                    parts[1] = parts[1].Trim();
+                    if (parts[0].Length == 0 || parts[1].Length == 0)
```
troydai: Can we add lines before and after `if-else`?

victorhurdugaci: This code was already there :) I only changed the log message and extracted a new method

src/Microsoft.Framework.Runtime/Servicing/ServicingIndex.cs
troydai: Just an idea. Will the entry mappings be big? If so, would it be faster if just reverse the sequence and return the first hit?

src/Microsoft.Framework.Runtime/Servicing/ServicingIndex.cs
troydai: Understand

troydai:
I have no comments except beside the trace formatting and potential performance impact change. But they're not essential. Looks good to me.

src/Microsoft.Framework.Runtime/Servicing/ServicingIndex.cs
victorhurdugaci: I like the idea. Stay tuned for the change :)

victorhurdugaci:
@troydai pushed a change that includes your suggestion for reverse looking in the match list and 2 more tests

ChengTian:
:shipit:

troydai:
:+1: :shipit:

Issue #735, opened 2014-10-10
Need better error when project.json is malformed


yishaigalatzer:
### Repro steps
Create a folder
create a `project.json` file that doesn't match the json spec (say just put hello world)
kpm build

Result  - Error (expected) see below, but no physical path to where it happens, so when kpm build runs inside a script you don't get enough input to where things happened.

Would like to get the full physical path to the project.json file (if available)

### Error example

```
Newtonsoft.Json.JsonReaderException: Unexpected character encountered while pars
ing value: E. Path '', line 0, position 0.
   at Newtonsoft.Json.JsonTextReader.ParseValue()
   at Newtonsoft.Json.JsonTextReader.ReadInternal()
   at Newtonsoft.Json.JsonTextReader.Read()
   at Newtonsoft.Json.Linq.JObject.Load(JsonReader reader)
   at Newtonsoft.Json.Linq.JObject.Parse(String json)
   at Microsoft.Framework.Runtime.Project.GetProject(String json, String project
Name, String projectPath)
   at Microsoft.Framework.Runtime.Project.TryGetProject(String path, Project& pr
oject)
   at Microsoft.Framework.PackageManager.BuildManager.Build()
   at Microsoft.Framework.PackageManager.Program.<>c__DisplayClass3.<Main>b__12(
)
   at Microsoft.Framework.Runtime.Common.CommandLine.CommandLineApplication.Exec
ute(String[] args)
   at Microsoft.Framework.PackageManager.Program.Main(String[] args)
```

BrennanConroy:
This should be fixed with 6b0dc2bfeb79a6490f7613c8086c2d6de13a95a6

PR #1236, opened 2015-02-17
Path to project.json now shown

BrennanConroy:
"Error in C:\{path}\project.json" is put in front of the exception text now

@ChengTian 
#735

### DIFF
--- a/src/Microsoft.Framework.Runtime/FileFormatException.cs
+++ b/src/Microsoft.Framework.Runtime/FileFormatException.cs
@@ -63,5 +63,10 @@ namespace Microsoft.Framework.Runtime
                 Line = exception.LineNumber
             };
         }
+
+        public override string ToString()
+        {
+            return String.Format("Error in {0}\n{1}", Path, base.ToString());
+        }
     }
 }
--- a/src/Microsoft.Framework.Runtime/Project.cs
+++ b/src/Microsoft.Framework.Runtime/Project.cs
@@ -128,7 +128,6 @@ namespace Microsoft.Framework.Runtime
             }
             catch (JsonReaderException ex)
             {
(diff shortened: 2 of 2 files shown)


troydai:
Should we use `Logger` instead? https://github.com/aspnet/XRE/blob/brecon/bug735/src/Microsoft.Framework.Runtime.Common/Impl/Logger.cs#L14

anurse:
Or maybe put the string in the exception Message itself?

ChengTian:
Yes, we should follow the original approach and put the string in exception message.

BrennanConroy:
Made the FileFormatException insert the path into the exception string

davidfowl:
What is the purpose of this change?

BrennanConroy:
To tell the developer where the project.Json file that failed is located. The exception message doesn't contain it

Sent from my HTC

----- Reply message -----

davidfowl:
Did you test this change in visual studio?

davidfowl:
I think you should let the exception get thrown and handle it in the apphost just like compilation exceptions are handled

src/Microsoft.Framework.Runtime/FileFormatException.cs
```diff
@@ -63,5 +63,10 @@ internal static FileFormatException Create(JsonReaderException exception, string
                 Line = exception.LineNumber
             };
         }
+
+        public override string ToString()
+        {
+            return String.Format("Error in {0}\n{1}", Path, base.ToString());
```
victorhurdugaci: `Environment.NewLine` instead of `\n` ?

BrennanConroy:
Running kpm restore/build propagates this exception to kre.host.shared where it calls PrintError(ex) so we can't handle it in apphost.

Issue #1238, opened 2015-02-18
Internalize FileSystemGlobbing into XRE


ChengTian:

ChengTian:
One more change is needed: when this change https://github.com/aspnet/XRE/commit/4a7994d226513397196928fd708780a307c9f5f5 passes Coherence-Signed, we need to update metaprogramming to use `IProjectContext`

PR #1239, opened 2015-02-18
Internalize FileSystemGlobbing

ChengTian:
parent #1238 

@davidfowl @loudej @troydai

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
+[submodule "submodules/FileSystem"]
+	path = submodules/FileSystem
+	url = https://github.com/aspnet/FileSystem.git
 [submodule "submodules/NuGet.Versioning"]
 	path = submodules/NuGet.Versioning
 	url = https://github.com/NuGet/NuGet.Versioning
--- a/global.json
+++ b/global.json
@@ -1,3 +1,8 @@
 {
-    "sources": ["src"]
+    "sources": [
+        "src",
+        "submodules/FileSystem/src",
+        "submodules/NuGet.Packaging/src",
+        "submodules/NuGet.Versioning/src"
+    ]
 }
--- a/src/Microsoft.Framework.Runtime/project.json
+++ b/src/Microsoft.Framework.Runtime/project.json
@@ -2,18 +2,28 @@
     "version": "1.0.0-*",
     "description": "ASP.NET 5 runtime infrastructure for loading NuGet packages and projects.",
     "compilationOptions": { "define": [ "LOADER", "TRACE" ], "allowUnsafe": true, "warningsAsErrors": true },
+    "code": [
+        "**/*.cs",
+        "../../submodules/FileSystem/src/Microsoft.Framework.FileSystemGlobbing/**/*.cs"
+    ],
+    "preprocess": [
+        "../../ext/compiler/preprocess/Internalization.cs"
+    ],
     "dependencies": {
         "System.Reflection.Metadata": "1.0.18-beta",
+        "Microsoft.CodeAnalysis.CSharp": "1.0.0-*",
+        "Microsoft.Framework.Runtime.Roslyn.Interfaces": "1.0.0-*",
         "Microsoft.Framework.Runtime.Loader": "1.0.0-*",
         "Microsoft.Framework.Runtime.Common": { "version": "1.0.0-*", "type": "build" },
         "Microsoft.Framework.Runtime.Interfaces": "1.0.0-*",
-        "Microsoft.Framework.FileSystemGlobbing": "1.0.0-*",
         "Newtonsoft.Json": "6.0.6"
     },
     "frameworks": {
         "aspnet50": {
             "frameworkAssemblies": {
                 "System.Collections": "",
+                "System.Text.Encoding": "",
+                "System.Threading.Tasks": "",
                 "System.IO": "",
                 "System.IO.Compression": "",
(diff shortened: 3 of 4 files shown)


ext/preprocess/FileSystemGlobbingInternalization.cs
```diff
@@ -0,0 +1,98 @@
+﻿using System;
```
Praburaj: license header

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "submodules/FileSystem"]
+	path = submodules/FileSystem
+	url = git@github.com:aspnet/FileSystem.git
```
Praburaj: May be a lame question. Is there a fallback for a `https` url if not `git@`?

src/Microsoft.Framework.Runtime/project.json
```diff
@@ -2,18 +2,28 @@
     "version": "1.0.0-*",
     "description": "ASP.NET 5 runtime infrastructure for loading NuGet packages and projects.",
     "compilationOptions": { "define": [ "LOADER", "TRACE" ], "allowUnsafe": true, "warningsAsErrors": true },
+    "code": [
+        "**/*.cs",
+        "../../submodules/FileSystem/src/Microsoft.Framework.FileSystemGlobbing/**/*.cs"
```
Praburaj: Just wondering if this is required if the globbing code is preprocessed?

.gitmodules
troydai: no. git and http are two transfer protocols for git. https is not considered as fall back. the url recorded here is the url used to clone the repo.

troydai: http://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols

src/Microsoft.Framework.Runtime/project.json
davidfowl: The preprocessor makes stuff internalized.

ext/preprocess/FileSystemGlobbingInternalization.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        public FileSystemGlobbingInternalization(IServiceProvider services)
+        {
+            var appEnv = services.GetService(typeof(IApplicationEnvironment)) as IApplicationEnvironment;
+            _appBase = appEnv.ApplicationBasePath;
+        }
+
+        public void BeforeCompile(IBeforeCompileContext context)
+        {
+            var fileSystemGlobbingDir = Path.Combine(_appBase, "..", "..", "submodules", "FileSystem", "src",
```
davidfowl: We need this PR (https://github.com/aspnet/XRE/pull/1240) to accurately get project information. Right now the application base gives you information about the environment the compile was run from:

A -> B

If you k run in A's directory and you end up compiling B, the IApplicationEnvironment will be for A instead of B.

global.json
```diff
@@ -1,3 +1,6 @@
 {
-    "sources": ["src"]
+    "sources": [
+        "src",
+        "submodules/FileSystem/src"
```
anurse: Is there any way to glob this? "submodules/*/src"?

anurse:
:shipit: with minor comments.

src/Microsoft.Framework.Runtime/project.json
ChengTian: Preprocessor only flips public to internal, we still need to include the code if we want to use them.

.gitmodules
ChengTian: Seems there is no way to add a fallback. We'll leave it for now and see whether it can lead to some inconvenience.

ChengTian: Thanks to @Praburaj 's follow-up. I just git cloned XRE via https and tried restore submodules. With git@github.com protocol, we do need to provide credentials. Using https urls for submodules is more convenient.

davidfowl: Uh...really? Ssh ftw?

ChengTian: Submodules are never cloned for dev purpose. We won't change anything inside them and we won't push to repos in submodules. So I guess using https is fine and it is convenient by allowing you to fetch submodules without credentials.

anurse: Yeah, we should use https. You only ever need creds if you push from the submodule. This is a known issue with submodules, users can't customize the protocol they use. It sucks but hey, you can always use a credential helper like [this fantastic specimen from some amazing developer](https://gitcredentialstore.codeplex.com/) :)

davidfowl: Ah then I don't care

davidfowl: @anurse it's on codeplex..... nuff said :stuck_out_tongue_closed_eyes:

ext/preprocess/FileSystemGlobbingInternalization.cs
davidfowl: Can we put this code in a project so that we get an editor experience around it?

ext/preprocess/FileSystemGlobbingInternalization.cs
ChengTian: Since Coherence-Signed didn't pass and your change didn't get in, I'll push the current implementation to unblock @anurse for now. I will change it to use more accurate info in another PR.

global.json
ChengTian: Tried both `submodules/*/src` and `submodules/**/src` but none of them works. I'll file an issue for `sources` supporting globbing.

PR #1240, opened 2015-02-18
Some ICompileModule improvements

davidfowl:
- Added some doc comments to some of the APIs
- Invoke AfterCompile
- Expanded IAfterCompileContext to offer the AssemblyStream, SymbolsStream
and XmlDocStream.
- Added IProjectContext on IBeforeCompileContext and IAfterCompileContext
- Fixed property names

/cc @troydai

### DIFF
--- a/samples/HelloWorld/compiler/preprocess/HelloMetaProgramming.cs
+++ b/samples/HelloWorld/compiler/preprocess/HelloMetaProgramming.cs
@@ -6,6 +6,7 @@ using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.Framework.Runtime;
+using Microsoft.Framework.Runtime.Roslyn;
 
 namespace HelloWorld.Compiler.Preprocess
 {
@@ -17,9 +18,10 @@ namespace HelloWorld.Compiler.Preprocess
 
         public void BeforeCompile(IBeforeCompileContext context)
         {
+            // TODO: Enable
             var options = CSharpParseOptions.Default
-                .WithLanguageVersion(context.CSharpCompilation.LanguageVersion);
-            context.CSharpCompilation = context.CSharpCompilation.AddSyntaxTrees(CSharpSyntaxTree.ParseText(@"
+                .WithLanguageVersion(context.Compilation.LanguageVersion);
+            context.Compilation = context.Compilation.AddSyntaxTrees(CSharpSyntaxTree.ParseText(@"
 public class Foo 
 {
     public string Message
--- a/src/Microsoft.Framework.Runtime.Roslyn.Interfaces/IProjectContext.cs
+++ b/src/Microsoft.Framework.Runtime.Roslyn.Interfaces/IProjectContext.cs
@@ -1,0 +1,36 @@
+﻿using System;
+using System.Runtime.Versioning;
+
+namespace Microsoft.Framework.Runtime.Roslyn
+{
+    /// <summary>
+    /// Stores information about the project
+    /// </summary>
+    public interface IProjectContext
+    {
+        /// <summary>
+        /// Path to the project directory
+        /// </summary>
+        string ProjectDirectory { get; }
+
+        /// <summary>
+        /// Path to the project file
+        /// </summary>
+        string ProjectFilePath { get; }
+
+        /// <summary>
+        /// The name of the project
+        /// </summary>
+        string Name { get; }
+
+        /// <summary>
+        /// The target framework of the project currently being compiled
+        /// </summary>
+        FrameworkName TargetFramework { get; }
+
+        /// <summary>
+        /// Project version
+        /// </summary>
+        string Version { get; }
+    }
+}
--- a/src/Microsoft.Framework.Runtime.Roslyn.Interfaces/project.json
+++ b/src/Microsoft.Framework.Runtime.Roslyn.Interfaces/project.json
@@ -15,5 +15,13 @@
                 "System.Runtime": "4.0.20-beta-*"
             }
         }
+    },
+
+    "scripts": {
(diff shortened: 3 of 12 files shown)


src/Microsoft.Framework.Runtime.Roslyn.Interfaces/project.json
```diff
@@ -15,5 +15,13 @@
                 "System.Runtime": "4.0.20-beta-*"
             }
         }
+    },
+
+    "scripts": {
+        "postbuild": [
+            "%project:Directory%/../../build/batchcopy %project:BuildOutputDir%/Debug/aspnet50/*.* %project:Directory%/../../artifacts/build/kre-clr-win-x86/bin",
```
anurse: I wonder if we should just have a "build/placebinaries" script that does all three of these commands for you. Also, it would be nice if we could also put the binaries into the right place for x64 builds. That's a side-issue though.

anurse:
:shipit:

samples/HelloWorld/compiler/preprocess/HelloMetaProgramming.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace HelloWorld.Compiler.Preprocess
 {
-    public class HelloMetaProgramming : ICompileModule
+    /*public class HelloMetaProgramming : ICompileModule
```
pranavkm: undo?

src/Microsoft.Framework.Runtime.Roslyn.Interfaces/IProjectContext.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
+using System.Runtime.Versioning;
+
+namespace Microsoft.Framework.Runtime.Roslyn
+{
+    /// <summary>
+    /// Stores information about the project
+    /// </summary>
+    public interface IProjectContext
```
pranavkm: If we add this, could you ensure `IApplicationEnvironment` is removed from precompilation? Also, this might need the equivalent of `IApplicationEnvironment.Configuration`. Mvc needs that value for compilation options parsing

src/Microsoft.Framework.Runtime.Roslyn/AfterCompileContext.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System;
```
pranavkm: License header

src/Microsoft.Framework.Runtime.Roslyn/RoslynProjectReference.cs
```diff
@@ -152,12 +169,12 @@ public IDiagnosticResult EmitAssembly(string outputPath)
 
                 var sw = Stopwatch.StartNew();
 
-                EmitResult result = null;
+                EmitResult emitResult = null;
```
pranavkm: Does this need to be initialized?

src/Microsoft.Framework.Runtime.Roslyn/AfterCompileContext.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        public IProjectContext ProjectContext
+        {
+            get
+            {
+                return _context.ProjectContext;
+            }
+        }
+
+        public Stream AssemblyStream { get; set; }
```
pranavkm: Any chance we get an assembly here instead? If the AfterCompileContext doesn't change it, you'll no-op. I could use this for Razor precompilation - but with the current approach, we'll need to emit the result twice - once with this stream to find interesting types and again when we hydrate it with resources.

samples/HelloWorld/compiler/preprocess/HelloMetaProgramming.cs
davidfowl: No, it needs to be left out for now.

src/Microsoft.Framework.Runtime.Roslyn.Interfaces/IProjectContext.cs
davidfowl: It wont be removed. I can add configurations

src/Microsoft.Framework.Runtime.Roslyn.Interfaces/project.json
davidfowl: Yar

src/Microsoft.Framework.Runtime.Roslyn/AfterCompileContext.cs
davidfowl: Nah. I want to give the user a chance to change it before loading. I don't see how you can avoid the double emit.

src/Microsoft.Framework.Runtime.Roslyn/RoslynProjectReference.cs
davidfowl: This variable was renamed. Nothing has changed

troydai:
I'm updating samples as well as test coverage.

troydai:
@pranavkm do you need me to coordinate any tests with MVC?

src/Microsoft.Framework.Runtime.Roslyn/RoslynProjectReference.cs
pranavkm: Ehh, it doesn't look like it needs to be initialized regardless. Just a code cleanup comment.

src/Microsoft.Framework.Runtime.Roslyn/AfterCompileContext.cs
pranavkm: It avoids us having to do the first emit. Just a thought since its quite likely you're going to have to emit it if the resulting Assembly is interesting. Quite unlikely you're going to look at the Stream does much.

PR #1242, opened 2015-02-18
Pull in NuGet code via submodule

anurse:
This doesn't actually internalize or integrate the code in any way, just pulls it in and adds it to the solution, like so:

capture

This should merge well with #1239 since it uses the same layout for submodules. Next step will be to start pulling this code in to the Microsoft.Framework.Runtime.Hosting project as source.

/cc @davidfowl @ChengTian


ChengTian:
Thanks for using the same layout :+1:

anurse:
Of course :). The .gitmodules file _should_ merge fine, but if it doesn't the conflict should be extremely easy to fix :P

troydai:
:shipit:

PR #1245, opened 2015-02-19
Set up new runtime projects

anurse:
This adds the new Runtime packages and configures AppHost2 to use them (not that AppHost2 does anything yet :))


ChengTian:
Looks good to me :+1:

anurse:
As per discussion with @davidfowl, I re-merged the interfaces. The new runtime will share interfaces with the old, but eventually the old one will go away. We just have to be sure to not use old interfaces in the new runtime :).

PR #1246, opened 2015-02-19
Making properties on RoslynCompilationFailure settable

pranavkm:


pranavkm:
Makes it easier to use this in Razor compilation since we have values for some of these properties.

davidfowl:
:shipit:

PR #1247, opened 2015-02-19
Exclude comments when getting JToken values as an array

Andorbal:
Comments are currently included in the list of patterns used by kpm, which causes build and restore operations to fail.  I noticed this when working with the [Azure storage](https://github.com/Azure/azure-storage-net/blob/master/Lib/AspNet/Microsoft.WindowsAzure.Storage/project.json) library, as their comment causes kpm to fail.

Putting this change in the JTokenExtensions.ValueAsArray method seemed a little heavy handed, but it didn't seem like any of the references could have made use of the comments anyway.

Please let me know if there should be more tests or if there are any other changes that I should make.

### DIFF
--- a/test/Microsoft.Framework.Runtime.Tests/ProjectFilesCollectionFacts.cs
+++ b/test/Microsoft.Framework.Runtime.Tests/ProjectFilesCollectionFacts.cs
@@ -59,6 +59,30 @@ namespace Microsoft.Framework.Runtime.Tests
             Assert.Equal(new[] { "a.cs", @"foo.js" }, target.ResourcesPatterns);
         }
 
+
+        [Fact]
+        public void FilePatternsDoNotIncludeComments()
+        {
+            var rawProject = Deserialize(@"
+{
+    ""code"": [""*.cs"", /* FOO */ ""../*.cs""],
+    ""exclude"": [""buggy/*.*"", /* FOO */],
+    ""bundleExclude"": [""no_pack/*.*"", /* FOO */],
+    ""preprocess"": [""other/**/*.cs"", /* FOO */ ""*.cs"", ""*.*""],
+    ""shared"": [""shared/**/*.cs;../../shared/*.cs"", /* FOO */],
+    ""resources"": [""a.cs"", /* FOO */ ""foo.js""]
+}");
+
+            var target = new ProjectFilesCollection(rawProject, string.Empty);
+
+            Assert.Equal(new[] { "*.cs", @"../*.cs" }, target.SourcePatterns);
+            Assert.Equal(new[] { @"buggy/*.*" }, target.ExcludePatterns);
+            Assert.Equal(new[] { @"no_pack/*.*" }, target.BundleExcludePatterns);
+            Assert.Equal(new[] { @"other/**/*.cs", "*.cs", "*.*" }, target.PreprocessPatterns);
+            Assert.Equal(new[] { @"shared/**/*.cs", @"../../shared/*.cs" }, target.SharedPatterns);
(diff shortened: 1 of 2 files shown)


msotclas:
Hi __@Andorbal__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>You've already signed the contribution license agreement. Thanks!</span>
        <p>The agreement was validated by Microsoft Open Technologies, Inc. and real humans are currently evaluating your PR.</p>

TTYL, MSOTBOT;

test/Microsoft.Framework.Runtime.Tests/ProjectFilesCollectionFacts.cs
```diff
@@ -59,6 +59,30 @@ public void FilePatternsWorkForArraysAreSet()
             Assert.Equal(new[] { "a.cs", @"foo.js" }, target.ResourcesPatterns);
         }
 
+
+        [Fact]
+        public void FilePatternsDoNotIncludeComments()
+        {
+            var rawProject = Deserialize(@"
+{
+    ""code"": [""*.cs"", /* FOO */ ""../*.cs""],
```
Eilon: What happens with a single-line comment, like this:

```
    ""morestuff"": [
        ""abc"", // comment
        ""xyz""],
```

test/Microsoft.Framework.Runtime.Tests/ProjectFilesCollectionFacts.cs
```diff
@@ -59,6 +59,30 @@ public void FilePatternsWorkForArraysAreSet()
...
+    ""exclude"": [""buggy/*.*"", /* FOO */],
```
Eilon: BTW there are a couple of these that aren't really valid - there's an `undefined` last element in the array because there's a trailing comma. There's another example below in `shared`.

test/Microsoft.Framework.Runtime.Tests/ProjectFilesCollectionFacts.cs
```diff
@@ -59,6 +59,30 @@ public void FilePatternsWorkForArraysAreSet()
...
+    ""bundleExclude"": [""no_pack/*.*"", /* FOO */],
+    ""preprocess"": [""other/**/*.cs"", /* FOO */ ""*.cs"", ""*.*""],
+    ""shared"": [""shared/**/*.cs;../../shared/*.cs"", /* FOO */],
+    ""resources"": [""a.cs"", /* FOO */ ""foo.js""]
+}");
+
+            var target = new ProjectFilesCollection(rawProject, string.Empty);
+
+            Assert.Equal(new[] { "*.cs", @"../*.cs" }, target.SourcePatterns);
```
Eilon: Kind of funny that these use `@"foo"` string literals even though there's nothing to escape here (I see it's just copied from other tests).

test/Microsoft.Framework.Runtime.Tests/ProjectFilesCollectionFacts.cs
Andorbal: Thanks for the feedback!  I can address the issues you've brought up.  Does the general concept look good, though or should I take a different approach?

Sent from my Windows Phone

-----Original Message-----

Eilon: Yeah the approach looks good to me. I saw @davidfowl looking at it and he didn't seem to freak out so I think this is fine.

davidfowl:
We have a new JSON parser that doesnt support comments https://github.com/aspnet/Announcements/issues/24

PR #1250, opened 2015-02-19
Use accurate project info in metaprogramming

ChengTian:
parent #1238 

@davidfowl


troydai:
:+1: Looks good to me.

PR #1232, opened 2015-02-17
Pull in new LibraryModel code for unified K+NuGet

anurse:
I integrated the LibraryModel code from @davidfowl 's prototypes.

As part of this, I renamed what was originally called `Library` to `LibraryIdentity` and what was originally `LibraryDescription` to `Library`. This seemed to be a little more appropriate naming. Now, a `LibraryIdentity` represents the name of a library and a `Library` represents the resolved library itself.

Next up: Add Project parsing. I plan to make a Project just a `Library` with additional properties (and probably a class wrapping that all up in some static types for convenience). After all, the `Project` class already shares `Name`, `Version`, `Dependencies` and `Path` with `Library`

PR #1251, opened 2015-02-20
Integrate NuGet.LibraryModel code

anurse:
Integrates the NuGet.LibraryModel code using the new source-included NuGet.Versioning package.

This should address most of the feedback from #1232 as well.


src/NuGet.LibraryModel/project.json
```diff
@@ -0,0 +1,20 @@
+﻿{
+    "version": "1.0.0-*",
+    "dependencies": {
+        "Newtonsoft.Json": "6.0.6",
+        "NuGet.Versioning": "1.0.0-*"
```
anurse: Note: When the runtime uses this, it will be pulling it and NuGet.Versioning in as source, so this project.json will be irrelevant.

src/Microsoft.Framework.Runtime.Common/Impl/Logger.cs
```diff
@@ -25,17 +34,41 @@ public static void TraceInformation(string message, params object[] args)
...
             get
             {
-                return Environment.GetEnvironmentVariable(EnvironmentNames.Trace) == "1";
+                if (_traceLevel == null)
+                {
+                    string envVar = Environment.GetEnvironmentVariable(EnvironmentNames.Trace);
+                    int newLevel;
+                    if (string.IsNullOrWhiteSpace(envVar) || !Int32.TryParse(envVar, out newLevel))
+                    {
+                        newLevel = 0;
```
ChengTian: newLevel = OffLevel?

anurse: Yes, good catch :), I introduced the constants after this code and forgot to put it in here.

src/Microsoft.Framework.ApplicationHost2/Program.cs
```diff
@@ -22,9 +26,111 @@ public Program(IAssemblyLoaderContainer container, IApplicationEnvironment envir
...
+            }
+            else
+            {
+                outArgs = remainingArgs.ToArray();
+            }
+
+            return true;
+        }
+
+        private static string GetVersion()
```
troydai: Can we cache the version?

src/NuGet.LibraryModel/LibraryDependencyType.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NuGet.LibraryModel
+{
+    public class LibraryDependencyType
+    {
+        private readonly LibraryDependencyTypeFlag[] _keywords;
+
+        public static LibraryDependencyType Default;
```
troydai: public static field? make it either readonly or property?

src/NuGet.LibraryModel/LibraryDependencyType.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            Default = new LibraryDependencyType(LibraryDependencyTypeKeyword.Default.FlagsToAdd as LibraryDependencyTypeFlag[]);
+        }
+
+        public LibraryDependencyType()
+        {
+            _keywords = new LibraryDependencyTypeFlag[0];
+        }
+
+        private LibraryDependencyType(LibraryDependencyTypeFlag[] flags)
```
troydai: put private members at the end.

src/NuGet.LibraryModel/LibraryDependencyTypeKeyword.cs
```diff
@@ -0,0 +1,138 @@
...
+{
+    public class LibraryDependencyTypeKeyword
+    {
+        private static ConcurrentDictionary<string, LibraryDependencyTypeKeyword> _keywords = new ConcurrentDictionary<string, LibraryDependencyTypeKeyword>();
+
+        public static LibraryDependencyTypeKeyword Default;
+        public static LibraryDependencyTypeKeyword Build;
+        public static LibraryDependencyTypeKeyword Preprocess;
+        public static LibraryDependencyTypeKeyword Private;
+        public static LibraryDependencyTypeKeyword Dev;
```
troydai: Errr, public static field?

src/NuGet.LibraryModel/NuGetVersionFloatBehavior.cs
```diff
@@ -0,0 +1,20 @@
...
+    public enum NuGetVersionFloatBehavior
+    {
+        None,
+        Prerelease,
+        Revision,
+        Build,
+        Minor,
+        Major
+    }
+
```
troydai: remove this line.

src/Microsoft.Framework.ApplicationHost2/Program.cs
anurse: Well, we call this method _exactly_ once, so I'm not sure it's worth allocating the memory :P

src/NuGet.LibraryModel/LibraryDependencyType.cs
anurse: Even when they are constructors? I think that organizing members by type (Constructor/Property/Method/etc) trumps organizing by visiblity. I think this should be right by the other (public) constructor.

Issue #878, opened 2014-11-10
References do not get added correctly when when there is a trailing forward slash in sources of global.json file


rmangalore:
REPRO STEPS:

1. Create a Starter web under c:\school. This gets created here-> C:\school\WebApplication2\src\WebApplication2
	2. Add a new solution folder by name Test. Add a class library to it. Say CalssLibrary1. This gets created here: C:\school\WebApplication2\ClassLibrary1
	3. Edit the global.json file to include this:
	"sources": [ "src", "test","C:/school/WebApplication2/"]
	4. Add reference to Classlibrary1 from WebApplication2.
ACTUAL: The project.sjon gets modified to add "ClassLibrary1". The references node in WebAPplication2 does not reflect the newly added reference.

This works if the trailing "/" is removed from the Global.json file. C:/school/WebApplication2"

EXPECTED: The references node to reflect the newly added reference.

PR #1255, opened 2015-02-20
Trim / from path ends

BrennanConroy:
#878


troydai:
Looks good to me.

davidfowl:
Did you look at all of the places that this class is used? This looks pretty dangerous to me...

BrennanConroy:
It is used in 1 place

davidfowl:
:shipit:

PR #1254, opened 2015-02-20
Update the readme about submodules

troydai:
For a clean cloned repository the build will fail if the submodules are not updated. The read me document needs to be updated to inform user to prepare the repo.

@Eilon @davidfowl @muratg Please help review.

PR #1256, opened 2015-02-20
Switch from Git Submodules to Git Subtree Merges

anurse:
**Don't be scared**! It looks like this PR has a big copy-pasta from the three modules, but it was done using "subtree" merges (read on for more). It is NOT copy-pasted!

After reviewing some of the info for submodules, I learned about "git subtree" from a post on the [Atlassian Blog](http://blogs.atlassian.com/2013/05/alternatives-to-git-submodule-git-subtree/). They seem to be a better approach for merging other code in. The differences are as follows:

* The individual files from the other repo are merged directly in to your repo, but there is still metadata that tracks the fact that the content came from another repo
* **UNLIKE** submodules, people who clone the repo don't need to do anything. Everything they need comes down in the initial clone.
* Developers who need to make changes to the subtree have to go through a slightly separate process to push to the other repo (there is a 'git subtree' command to help with this)
* GitHub shows the code as a raw copy, so we don't quite get the nice linking to additional repos

Given #1254, I feel like submodules may be confusing for people, and few developers actually need to _modify_ the sub-repos. Subtrees are a better fit for our "everyone needs to get the sub-repo, few need modify the sub-repo" scenario.

Thoughts? If people disagree, I'm happy to close this out, just wanted to show what it would look like. It just took three `git subtree add` commands to get here (plus removing the .gitmodules and old submodule dirs) so it's not like it was a lot of time :)


troydai:
New tricks every day. Quick question, looks like `subtree` doesn't manage the version of sub project, does it?

davidfowl:
I'm not sure I like it.

Tragetaschen:
I don't like it.

- The code in the submodules is in fact separate and the responsibility for its development lies somewhere else
- The way submodules handle history (log) is IMHO preferable
- Most of the friction regarding submodules being correctly checked out can be solved by the build script. Take Mono [for](https://github.com/mono/mono/blob/master/configure.ac#L3901) an [example](https://github.com/mono/mono/blob/master/scripts/update_submodules). I never had to even think about it other than seeing the automatic update code during the `configure` step.

- I always prefer referential integrity over a blunt copy :-)

davidfowl:
:-1:

anurse:
Fair enough, just wanted to throw it out there :). I've already had problems with forgetting `git submodule update` so I was concerned about others encountering the same thing.

PR #1257, opened 2015-02-20
Add the ef command to the pack ignore list

victorhurdugaci:
@troydai @ChengTian @lodejard


davidfowl:
This is pretty bad.

victorhurdugaci:
It is not elegant for sure but we had two options
1. Add a special section in project.json for exported commands
2. Use this black list

PR #1269, opened 2015-02-24
Fix kpm commands install folder

victorhurdugaci:
Fixes https://github.com/aspnet/XRE/issues/1268

@troydai @ChengTian can you please review

@tugberkugurlu here is the fix if you want to try it :)

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Install/AppCommandsFolderRepository.cs
+++ b/src/Microsoft.Framework.PackageManager/Install/AppCommandsFolderRepository.cs
@@ -135,12 +135,22 @@ namespace Microsoft.Framework.PackageManager
 
         public static AppCommandsFolderRepository CreateDefault()
         {
+            // TODO: use Environment.GetFolderPath(Environment.SpecialFolder.UserProfile) when it's available on CoreCLR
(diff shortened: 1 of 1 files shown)


davidfowl:
What does this do?

victorhurdugaci:
The bug: `kpm commands install` installs in  `%userprofile%\bin` rather than  `%userprofile%\.k\bin`.
This change: fixes that and installs in the right folder

src/Microsoft.Framework.PackageManager/Install/AppCommandsFolderRepository.cs
```diff
@@ -140,7 +140,16 @@ public static AppCommandsFolderRepository CreateDefault()
             {
                 userProfileFolder = Environment.GetEnvironmentVariable("HOME");
             }
-            var binFolder = Path.Combine(userProfileFolder, "bin");
+
```
ChengTian: Add TODO comments?
```c#
// TODO: use Environment.GetFolderPath(Environment.SpecialFolder.UserProfile) when it's available on CoreCLR
```

victorhurdugaci: Done

ChengTian:
:shipit:

PR #1276, opened 2015-02-24
Leave package dependencies unresolved when lockfile is missing

ChengTian:
Fixing a bug introduced by https://github.com/aspnet/XRE/pull/1260: when lockfiles are missing, VS is frozen after you open the solution.

@davidfowl , leave package dependencies unresolved when lockfile is missing.


davidfowl:
Did you try this in VS?

davidfowl:
:shipit:

PR #1291, opened 2015-02-26
More performance improvements

davidfowl:
- Do all project parsing via IProjectResolver
- Cache lookup of lock file libraries
- Get versions from the redist list
- Reduce IO at project lookup time by scanning global paths up front
- Use lock file to figure out contract path

/cc @shhsu @anurse @ChengTian

### DIFF
--- a/src/Microsoft.Framework.Runtime/NuGet/Repositories/PackageRepository.cs
+++ b/src/Microsoft.Framework.Runtime/NuGet/Repositories/PackageRepository.cs
@@ -14,7 +14,7 @@ namespace NuGet
         private readonly Dictionary<string, IEnumerable<PackageInfo>> _cache;
         private readonly IFileSystem _repositoryRoot;
         private readonly bool _checkPackageIdCase;
-        private LockFile _lockFile;
+        private ILookup<string, LockFileLibrary> _lockFileLibraries;
 
         public PackageRepository(string path, bool caseSensitivePackagesName = false)
             : this(new PhysicalFileSystem(path), caseSensitivePackagesName)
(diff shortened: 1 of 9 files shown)


src/Microsoft.Framework.Runtime/ProjectResolver.cs
```diff
@@ -30,25 +37,19 @@ public bool TryResolveProject(string name, out Project project)
...
+            if (_projects.TryGetValue(name, out projectInfo))
             {
-                var projectPath = Path.Combine(searchPath, name);
-
-                if (Project.TryGetProject(projectPath, out project))
-                {
-                    return true;
-                }
+                project = projectInfo.Project;
+                return true;
```
ChengTian: Should be
```c#
return project != null;
```
`_projects` contains potential project dirs that can be non-project dirs.

src/Microsoft.Framework.Runtime/NuGet/Repositories/PackageRepository.cs
```diff
@@ -53,7 +53,7 @@ public IFileSystem RepositoryRoot
 
         public void ApplyLockFile(LockFile lockFile)
         {
-            _lockFile = lockFile;
+            _lockFileLibraries = lockFile.Libraries.ToLookup(l => l.Name);
```
ChengTian: Should make the lookup based on `_checkPackageIdCase`:
```c#
var stringComparer = _checkPackageIdCase ? StringComparer.Ordinal : StringComparer.OrdinalIgnoreCase;
_lockFileLibraries = lockFile.Libraries.ToLookup(l => l.Name, stringComparer);
```

ChengTian:
Local tests passed. Pushing to dev.

Issue #1286, opened 2015-02-26
Lock file should use / instead of \


davidfowl:
Because having to escape `\\` in JSON files sucks.

PR #1299, opened 2015-02-27
Use '/' instead of '\' in lockfile

ChengTian:
parent #1286

### DIFF
--- a/src/Microsoft.Framework.Runtime/DependencyManagement/LockFileFormat.cs
+++ b/src/Microsoft.Framework.Runtime/DependencyManagement/LockFileFormat.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Framework.Runtime.DependencyManagement
 {
     public class LockFileFormat
     {
-        public static readonly int Version = -10000;
+        public const int Version = -9999;
         public const string LockFileName = "project.lock.json";
 
         public LockFile Read(string filePath)
@@ -99,7 +99,7 @@ namespace Microsoft.Framework.Runtime.DependencyManagement
             }
             library.Sha = ReadString(json["sha"]);
             library.FrameworkGroups = ReadObject(json["frameworks"] as JObject, ReadFrameworkGroup);
-            library.Files = ReadArray(json["files"] as JArray, ReadString);
+            library.Files = ReadPathArray(json["files"] as JArray, ReadString);
             return library;
         }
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.Framework.Runtime/DependencyManagement/LockFileFormat.cs
```diff
@@ -243,7 +243,7 @@ private IPackageFile ReadPackageFile(string property, JToken json)
         private JProperty WritePackageFile(IPackageFile item)
         {
             var json = new JObject();
-            return new JProperty(item.Path, new JObject());
+            return new JProperty(item.Path.Replace('\\', '/'), new JObject());
```
troydai: nit: we have at least two places have codes to unify the separator. should be have common code?

ChengTian: We do have several places using this
untitled

I can add a method to `NuGet.PathUtility`. Does `PathUtility.GetForwardSlashSeparatedPath(string path)` sound good?

davidfowl: GetPathWithForwardSlashes

davidfowl:
This will probably break things in the NuGet resolver. Make sure the translation to physical paths works on all OSes

ChengTian:
Convert path separators back to `Path.DirectorySeparator` when reading lockfile. This is the easiest way to resolve introduced bugs. Need to check startup performance impact.

ChengTian:
Verified with dotTrace on MusicStore project. This change doesn't have observable impact on performance of lockfile reading.

davidfowl:
Is this good to go ?

ChengTian:
Yeah, just rebased and did `build verify`. Please sign off.

troydai:
:shipit: after rebase.

ChengTian:
Thanks for the sign-off. It's better to rebase this PR after https://github.com/aspnet/DNX/pull/1404 is merged in.

ChengTian:
From @davidfowl : this is a good change to practice lock file versioning. We can push this PR after we have version check for lockfile

ChengTian:
Just verified, this PR **doesn't** fix the issue #1410 . If using path separator of current OS can cause an issue, it will only happen when we `kpm restore` on one OS and move the app to another OS to run it.

Need further investigation in issue #1410

PR #1314, opened 2015-03-04
integrating logging as a submodule instead of current low-level logger

anurse:
Now we have real logging instead of just a hacked up low-level logging facility!

* By default, logging is disabled and all loggers are no-op loggers (NullLogger)
* If KRE_TRACE=1, all Runtime Loggers will print INFO and up messages to console
* KRE_TRACE can also be a special string that specifies the loggers to enable and the levels to enable them at. The string is a `;`-delimited collection of filters, each of the form `[logger name]:[level]` (where the `:[level]` portion is optional and defaults to `:information`)
  * If a filter contains a Logger name, that logger is enabled at Info level (i.e. `RuntimeHost`)
  * If a filter contains a Logger name followed by `:off`, that logger is disabled (i.e. `RuntimeHost:off`)
  * If a filter contains a Logger name followed by `:[level]`, that logger is enabled at that level (i.e. `RuntimeHost:debug`)
  * The Logger Name may be `*`, in which case that filter will apply to all loggers (i.e. `*:warning`)
  * Logger Names are checked using Equals, StartsWith and EndsWith, so the name can be partial (i.e. 'Microsoft.Framework' or 'RuntimeHost' for `Microsoft.Framework.Runtime.Hosting.RuntimeHost`)
  * Most importantly, filters are processed from left to right and the first matching filter is applied.

Examples:
* ALL of these examples will disable EVERY logger:
  * `*:off`
  * `*:off;RuntimeHost:information`
  * `*:off;*:information`
* `*`
  * Turn on all loggers at Info level
* `RuntimeHost:debug;*`
  * Turn on all loggers at info level, and any logger starting or ending with `RuntimeHost` at debug level
* `Microsoft.Framework.Runtime.Hosting.RuntimeHost:information`
  * Turn on ONLY the `Microsoft.Framework.Runtime.Hosting.RuntimeHost` logger at info level

All this crazy parsing happens ONLY if the KRE_TRACE variable is non-empty, so the perf impact on a production system should be minimal.

/cc @davidfowl @ChengTian

Depends on aspnet/Logging#132

### DIFF
--- a/src/Microsoft.Framework.ApplicationHost2/project.json
+++ b/src/Microsoft.Framework.ApplicationHost2/project.json
@@ -6,7 +6,8 @@
         "Microsoft.Framework.Runtime.Common": { "version": "1.0.0-*", "type": "build" },
         "Newtonsoft.Json": "6.0.6",
         "Microsoft.Framework.Runtime.Roslyn.Interfaces": "1.0.0-*",
-        "Microsoft.Framework.Runtime.Interfaces": "1.0.0-*"
+        "Microsoft.Framework.Runtime.Interfaces": "1.0.0-*",
+        "Microsoft.Framework.NotNullAttribute.Internal": { "type": "build", "version": "1.0.0-*" }
     },
     "code": [
         "**/*.cs",
@@ -18,7 +19,10 @@
         "../../submodules/NuGet.Packaging/src/NuGet.Packaging.Core.Types/**/*.cs",
         "../../submodules/NuGet.Packaging/src/NuGet.Repositories/**/*.cs",
         "../../submodules/NuGet.Packaging/src/NuGet.DependencyResolver.Core/**/*.cs",
-        "../../submodules/NuGet.Packaging/src/NuGet.ProjectModel/**/*.cs"
+        "../../submodules/NuGet.Packaging/src/NuGet.ProjectModel/**/*.cs",
+        "../../submodules/Logging/src/Microsoft.Framework.Logging/**/*.cs",
+        "../../submodules/Logging/src/Microsoft.Framework.Logging.Console/**/*.cs",
+        "../../submodules/Logging/src/Microsoft.Framework.Logging.Interfaces/**/*.cs"
     ],
     "exclude": [
         "../../submodules/NuGet.Versioning/src/NuGet.Versioning/compiler/**/*.cs",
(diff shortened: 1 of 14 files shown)


src/Microsoft.Framework.Runtime.Hosting/RuntimeLogging.cs
```diff
@@ -0,0 +1,99 @@
...
+{
+    public static class RuntimeLogging
+    {
+        private static ILoggerFactory _loggerFactory;
+        private static Func<string, LogLevel, bool> _filter;
+
+        public static Func<string, LogLevel, bool> Filter { get { return _filter; } }
+        public static bool IsEnabled { get { return _filter != null; } }
+
+        public static void Initialize(Func<ILoggerFactory> loggerFactoryFactory)
```
anurse: Yep, I went there. Logger Factory Factory :). I don't want to run the delegate if the KRE_TRACE variable is empty, so that's what you get!

src/Microsoft.Framework.ApplicationHost2/project.json
```diff
@@ -51,12 +55,12 @@
                 "System.Collections": "4.0.10-beta-*",
                 "System.Diagnostics.Tools": "4.0.0-beta-*",
                 "System.Diagnostics.Debug": "4.0.10-beta-*",
+                "System.Diagnostics.TraceSource": "4.0.0-beta-*",
```
davidfowl: Ugh, noooo. We need to decoupled from the logging implementation and write our own.

anurse:
Confirmed with dotTrace that when KRE_TRACE is off, perf is unaffected. Logging methods are not visible in Sampling and when Tracing they show as taking "0ms" (which I interpret as: <1ms with rounding :))

src/Microsoft.Framework.ApplicationHost2/project.json
anurse: Are we moving TraceSource stuff out of M.F.Logging? It's not on the queue from the API review. Unfortunately, due to source-inclusion, we have to have this package here even if we don't use it, unless the TraceSource logger is moved out of M.F.Logging and into a separate package.

src/Microsoft.Framework.Runtime.Hosting/RuntimeHostBuilder.cs
```diff
@@ -37,13 +38,15 @@ public static RuntimeHostBuilder ForProjectDirectory(string projectDirectory, Nu
...
             }
 
+            var log = RuntimeLogging.Logger<RuntimeHostBuilder>();
             var hostBuilder = new RuntimeHostBuilder();
 
             // Load the Project
             var projectResolver = new PackageSpecResolver(projectDirectory);
             PackageSpec packageSpec;
-            if (projectResolver.TryResolvePackageSpec(GetProjectName(projectDirectory), out packageSpec))
+            if (projectResolver.TryResolvePackageSpec(GetProjectName(projectDirectory), out packageSpec) && packageSpec != null)
```
davidfowl: What? TryResovleSpec can return true and packageSpec can be null?

anurse: Yeah, the current code seems to do that if a folder with the specified name exists but doesn't contain a project.json. I think I can fix that.

Issue #1313, opened 2015-03-03
kpm bundle should generate lockfile when the original project doesn't have one


ChengTian:
Reported by @anurse :
Given a project that doesn't have any dependency, user should be able to run it without `kpm restore`. So the project may not have lockfile. In this case, `kpm bundle --no-source` should generate lockfile in the output.

PR #1318, opened 2015-03-04
Fix lockfile for no source bundle

ChengTian:
Fixing #1313 and https://github.com/aspnet/XRE/commit/9a1e70fcec3a94d7691666f1e42fc94570c491c7#commitcomment-10023178


davidfowl:
We need a test

anurse:
Looks good to me. :shipit: with some tests

PR #1320, opened 2015-03-05
[New Runtime] readjust some namespaces

anurse:
Microsoft.Framework.Runtime.Hosting was getting unnecessarily long. The assembly will continue to be called that but I'm removing the ".Hosting" subnamespace. It's just an unnecessary extra layer...

/cc @davidfowl @ChengTian


davidfowl:
:shipit:

PR #1324, opened 2015-03-06
Fix sporadic exception thrown during "kpm restore"

ChengTian:
The original approach is problematic: it creates the folder `<packages>/<PackageName>/<RandomName>` as temporary folder. However, some threads may treat `<RandomName>` as a version number and read nuspec from it. Directory.Move() will fail when a file in it is being read.


davidfowl:
:shipit:  Verify on mono

ChengTian:
Verified on Mono and it works well

davidfowl:
:shipit: !

ChengTian:
Thanks for signing off. It's not an urgent change, so I guess it's better to check in next Monday after renaming.

Issue #963, opened 2014-12-04
Ability to strong-name assemblies


bricelam:
`kpm buid` should allow the assemblies to be strong-named **at least** for non-ASP.NET frameworks if nothing else.

Eilon:
@glennc @davidfowl @bricelam @muratg @lodejard re-opening this based on our recent discussions.

muratg:
:+1:

davidfowl:
We can do this now everywhere thanks to https://github.com/dotnet/roslyn/commit/f591db040a09b0d958b628a4183a332bea6834c1

muratg:
YAY!

muratg:
@davidfowl still Beta5?

bricelam:
As far as the original report is concerned, this is resolved by #2004

muratg:
Yup, @moozzyk essentially took over this while he's working on servicing.

PR #1325, opened 2015-03-06
Strong name all the things!

davidfowl:
- Strong name everything by default because why not
- Automatically rewrite InternalsVisibleTo attributes

#963

### DIFF
--- a/src/Microsoft.Framework.Runtime.Roslyn.Common/CompilationDefaults.cs
+++ b/src/Microsoft.Framework.Runtime.Roslyn.Common/CompilationDefaults.cs
@@ -1,0 +1,20 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Immutable;
+
+namespace Microsoft.Framework.Runtime.Roslyn.Common
+{
+    public static class CompilationDefaults
+    {
+        public static readonly ImmutableArray<byte> PublicKey = GetPublicKey();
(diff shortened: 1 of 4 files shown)


GrabYourPitchforks:
Not all InternalsVisibleTo attributes should be rewritten. For instance, the one for Moq dynamic proxy should remain as-is.

davidfowl:
@GrabYourPitchforks good point. We could, just force you to add the public key yourself.

src/Microsoft.Framework.Runtime.Roslyn.Common/CompilationDefaults.cs
```diff
@@ -0,0 +1,20 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Immutable;
+
+namespace Microsoft.Framework.Runtime.Roslyn.Common
+{
+    public static class CompilationDefaults
+    {
+        public static readonly ImmutableArray<byte> PublicKey = GetPublicKey();
+        public static readonly string PublicKeyHex = @"0024000004800000940000000602000000240000525341310004000001000100bbe2884a1ce0432e3d6fb9225fd417a67a116c714bfec6a27ef678a1906374ddd6bc9c34c7c42c241b46005668b4aabbd1a5c80087895fc8329b2c14ee41b07e86275e03de27951630b3223f602f053d1d7d8fbc7e86e2f549f9b47bd0e4923918c537d86a9d3c4fa5ae5cf9f3ad15fd475aeba2dd3ad77fb56c9ad23939c9a2";
```
muratg: Can we generate this from GetPublicKey() as well?  Just in case we change this.

src/Microsoft.Framework.Runtime.Roslyn/RoslynCompiler.cs
```diff
@@ -226,6 +227,43 @@ private CompilationModules GetCompileModules(ILibraryKey target)
...
+                }
+            }
+
+            if (assemblyNames.Count > 0)
+            {
+                // TODO: Don't reparse here
+                var sb = new StringBuilder();
+                foreach (var assemblyName in assemblyNames)
+                {
+                    sb.AppendFormat(@"[assembly: System.Runtime.CompilerServices.InternalsVisibleTo(""{0}, PublicKey={1}"")]", assemblyName, CompilationDefaults.PublicKeyHex)
```
muratg: Can't we do this in the loop above?

PR #1328, opened 2015-03-06
Roughing in runtime.json preview

lodejard:


NuGet.Config
```diff
@@ -1,7 +1,8 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
```
davidfowl: Revert this

PR #1339, opened 2015-03-08
Updating to DNX names.

NTaylorMullen:
Prepping to be merged once ready.


davidfowl:
This change has to wait on @anurse's changes

NTaylorMullen:
si!

makefile.shade
```diff
@@ -153,16 +153,16 @@ var RC_FILES = '${FindAllFiles("Resource.rc", BOOTSTRAPPER_EXE_NAME, BOOTSTRAPPE
             var environmentRuntimeTargetOS = Environment.GetEnvironmentVariable("RUNTIME_TARGET_OS");
             var configRuntimeTargetOS = "";
         
-            Exec(MSBUILD, bootstrapperProj + " /p:Configuration=" + Configuration2 + ";Platform=Win32;TargetFramework=aspnet50");
```
davidfowl: These changes will break the C++ build. We need to update dnx.vcxproj. While we're at it, lets just call it what it really is:

- coreclr
- clr

We should change the property to `Runtime=CLR` and `Runtime=CoreCLR`

muratg:
This looks good to me.  Since @davidfowl is around, i'll let him put the shipit. :smile:

PR #1345, opened 2015-03-08
fix the cpp build

anurse:


davidfowl:
You need to update the cpp file as well since you changed the define.

davidfowl:
:shipit:

PR #1350, opened 2015-03-09
fix VersionUtility tests

anurse:


test/Microsoft.Framework.Runtime.Tests/VersionUtilityFacts.cs
```diff
@@ -7,48 +7,55 @@ namespace Microsoft.Framework.Runtime.Tests
     public class VersionUtilityFacts
     {
         [Theory]
-        [InlineData("net45", "dnx451", false)]
+        [InlineData("net45", "aspnet50", false)]
```
anurse: This is restoring the old test values. Eventually they can be dnxified but for now it's sufficient to add the below cases

test/Microsoft.Framework.Runtime.Tests/VersionUtilityFacts.cs
```diff
@@ -7,48 +7,55 @@ namespace Microsoft.Framework.Runtime.Tests
...
         [InlineData("dnxcore50", "portable-net45+win8", true)]
         [InlineData("dnxcore50", "portable-net451+win81", true)]
         [InlineData("dnxcore50", "portable-net40+sl5+win8", false)]
         [InlineData("dnxcore50", "portable-net45+win8", true)]
         [InlineData("dnxcore50", "portable-net451+win81", true)]
         [InlineData("dnxcore50", "portable-net40+sl5+win8", false)]
-
-        // Tests for aspnet -> dnx rename
-        // DNX packages installed into DNX451 projects? OK!
-        [InlineData("dnx451", "dnx452", true)]
```
davidfowl: bring this back

anurse: As discussed in JabbR, this was a bad test. They should be swapped. I'll add that in to the final merge.

davidfowl:
:shipit:

PR #1353, opened 2015-03-09
fix build scripts and remove some BOMs

### LINKED ISSUE #1346: Fix post-build scripts
Many still copy to "kre-..." instead of "dnx-...". Just a simple find-replace :)

anurse:
Fix #1346 by updating postbuild scripts.

This also removes some, but not all BOMs, only because I was touching those files.


davidfowl:
:shipit:

Issue #1358, opened 2015-03-10
Long coreclr home and application base path cause errors in CoreCLR


troydai:
https://github.com/aspnet/XRE/blob/dev/src/dnx.coreclr/dnx.coreclr.cpp#L463

The size of the buffer to store coreclr path and app path is set to `MAX_PATH`, which can be easily exceeded when both paths are long. It happens in the case of IBC build on CI.

Which makes it worse is that the error is swallowed. Not trace or exception to be captured.

BrennanConroy:
Should be [(MAX_PATH + 1) * 2] unless CreateAppDomainWithManager expects a certain sized string.

We should also print the value of errno in the CHECK_RETURN_VALUE_FAIL_EXIT_VIA_FINISHED macro and possibly include the line number __LINE__ and file just in case __FILE__

davidfowl:
Closing this as it should be part of @moozzyk's general native code clean up.

PR #1363, opened 2015-03-10
Fix max path bug for non-path

BrennanConroy:
#1358
@troydai @davidfowl 
Changed Klr function name and removed an unused variable too.

### DIFF
--- a/src/dnx.coreclr/dnx.coreclr.cpp
+++ b/src/dnx.coreclr/dnx.coreclr.cpp
@@ -7,9 +7,10 @@
 #include "dnx.coreclr.h"
 #include "tpa.h"
 
+#undef errno
 #define TRUSTED_PLATFORM_ASSEMBLIES_STRING_BUFFER_SIZE_CCH (63 * 1024) //32K WCHARs
-#define CHECK_RETURN_VALUE_FAIL_EXIT_VIA_FINISHED(errno) { if (errno) { printf_s("Error in file: %s at line: %d error: %d\n", __FILE__, __LINE__ - 1, errno); goto Finished;}}
-#define CHECK_RETURN_VALUE_FAIL_EXIT_VIA_FINISHED_SETSTATE(errno,SET_EXIT_STATE) { if (errno) { printf_s("Error in file: %s at line: %d error: %d\n", __FILE__, __LINE__ - 1, errno); SET_EXIT_STATE; goto Finished;}}
+#define CHECK_RETURN_VALUE_FAIL_EXIT_VIA_FINISHED(errno) { if (errno) { printf_s("Error in file: %s at line: %d error: %s\n", __FILE__, __LINE__ - 1, strerror(errno)); goto Finished;}}
+#define CHECK_RETURN_VALUE_FAIL_EXIT_VIA_FINISHED_SETSTATE(errno,SET_EXIT_STATE) { if (errno) { printf_s("Error in file: %s at line: %d error: %s\n", __FILE__, __LINE__ - 1, strerror(errno)); SET_EXIT_STATE; goto Finished;}}
 
 typedef int (STDMETHODCALLTYPE *HostMain)(
     const int argc,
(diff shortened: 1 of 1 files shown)


msotclas:
Hi __@BrennanConroy__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <p>
        It looks like you're a Microsoft contributor (Brennan Conroy). If you're full-time, we DON'T require a Contribution License Agreement. If you are a vendor, please DO sign the electronic Contribution License Agreement. It will take 2 minutes and there's no faxing! https://cla2.msopentech.com.
    </p>

TTYL, MSOTBOT;

src/dnx.coreclr/dnx.coreclr.cpp
```diff
@@ -8,8 +8,8 @@
 #include "tpa.h"
 
 #define TRUSTED_PLATFORM_ASSEMBLIES_STRING_BUFFER_SIZE_CCH (63 * 1024) //32K WCHARs
-#define CHECK_RETURN_VALUE_FAIL_EXIT_VIA_FINISHED(errno) { if (errno) { goto Finished;}}
-#define CHECK_RETURN_VALUE_FAIL_EXIT_VIA_FINISHED_SETSTATE(errno,SET_EXIT_STATE) { if (errno) { SET_EXIT_STATE; goto Finished;}}
+#define CHECK_RETURN_VALUE_FAIL_EXIT_VIA_FINISHED(errno) { if (errno) { printf_s("Error in file: %s at line: %d error: %d\n", __FILE__, __LINE__ - 1, errno); goto Finished;}}
+#define CHECK_RETURN_VALUE_FAIL_EXIT_VIA_FINISHED_SETSTATE(errno,SET_EXIT_STATE) { if (errno) { printf_s("Error in file: %s at line: %d error: %d\n", __FILE__, __LINE__ - 1, errno); SET_EXIT_STATE; goto Finished;}}
```
troydai: :+1: the tracing the error messages. Make sure it is tested.

src/dnx.coreclr/dnx.coreclr.cpp
```diff
@@ -451,7 +450,7 @@ extern "C" __declspec(dllexport) HRESULT __stdcall CallApplicationMain(PCALL_APP
     CHECK_RETURN_VALUE_FAIL_EXIT_VIA_FINISHED(errno);
 
     //wstring appPaths(szCurrentDirectory);
-    WCHAR wszAppPaths[MAX_PATH];
+    WCHAR wszAppPaths[(MAX_PATH + 1) * 2];
```
troydai: Add comments explaining how the size of the buffer is chosen/

troydai:
Codes look good to me. Please make sure you tested the scenarios it intents to cover.

BrennanConroy:
Updated

src/dnx.coreclr/dnx.coreclr.cpp
```diff
@@ -332,6 +349,9 @@ extern "C" __declspec(dllexport) HRESULT __stdcall CallApplicationMain(PCALL_APP
     size_t cchTrustedPlatformAssemblies = 0;
     LPWSTR pwszTrustedPlatformAssemblies = nullptr;
 
+    // set handler for security-enhanced CRT functions
+    _set_invalid_parameter_handler(invalid_parameter_handler);
```
troydai: Hum. It's not a bad idea to set a handler. However, the implementation of the function be passed in seems not to be inline with what expected. On [MSDN(https://msdn.microsoft.com/en-us/library/a9yf33zb.aspx)] it says 
> It should not return control to the main function unless you are confident that the error is recoverable.

My another concern is that replace the handler will change all the behavior down the roads, will this impact any existing error handling logic we have today? (Per discussion with @BrennanConroy the default handler will fast-fail and leave no trace. In that sense this could be better.)

KKhurin: Is it possible to find out the scope of ```_set_invalid_parameter_handler``` call and if it affects the thread or the whole process (as opposed to just current execution frame), restore the previous handler? We don't want to always print and swallow errors.

troydai:
/cc @BrennanConroy @moozzyk @davidfowl 

Any progress?

BrennanConroy:
I can make the scope of the parameter handler only for this function and then the PR should be final. But we were wanting to use the STL which would get rid of this issue entirely @moozzyk

moozzyk:
I think we should do just `WCHAR wszAppPaths[(MAX_PATH + 1) * 2];`  for now. I am starting looking at STL again but it will take some time to get there. 

WRT STL this ` LPWSTR lpwszModuleFileName = L"api-ms-win-core-processthreads-l1-1-1.dll";` scares me btw.

davidfowl:
Closing this for now. @moozzyk This should be part of your native code refactoring. Can you add this to that issue? Basically a set of bullets to clean up in the native code /cc @muratg

moozzyk:
#1937

moozzyk:
This is now fixed by using stl for string concatenation.

troydai:
STL! Way to go!

PR #1384, opened 2015-03-12
cleared local nuget sources

tugberkugurlu:
Clearing local NuGet sources as it sends unnecessary bunch of requests during restore when building project.

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
+    <clear />
     <add key="AspNetVNext" value="https://www.myget.org/F/aspnetvnext/api/v2" />
     <add key="NuGet" value="https://nuget.org/api/v2/" />
   </packageSources>


msotclas:
Hi __@tugberkugurlu__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>You've already signed the contribution license agreement. Thanks!</span>
        <p>The agreement was validated by Microsoft Open Technologies, Inc. and real humans are currently evaluating your PR.</p>

TTYL, MSOTBOT;

davidfowl:
We don't make inconsistent changes to our NuGet.config files. We use a template in aspnet/Universe and copy that to all of the repositories. That said, I don't know if this works for us. 
@pranavkm @Praburaj  Can you see if this breaks NuGet.redirect.config?

dougbu:
whether or not this breaks NuGet.redirect.config, it will definitely break local caches e.g. `$env:userprofile\.nuget`

davidfowl:
@dougbu good point. We shouldn't do this

tugberkugurlu:
@dougbu, @davidfowl So, does this mean that **we** shouldn't do this as well? :smile: 

Also, it was able to get the packages from the cache when I cleared the local sources. Is that a different cache?

davidfowl:
not a cache, a local packages folder that we use by convention.

pranavkm:
Is the `.nuget` still used? We install to the `.k` directory when you do `build install` which was the primarily use case for it. It's probably useful for testing `KoreBuild` but not a great deal outside of it?

davidfowl:
Ah you're right, we do kpm packages add now for build install.

dougbu:
well, as @pranavkm said and I made more definite
> It's ~~probably~~ useful for testing KoreBuild

is that no longer a valid scenario?

PR #1406, opened 2015-03-14
DNX on Linux bring-up.

ellismg:
This seems like a good checkpoint.

The following changes make DNX cross platform and add another second level unamanged loader which can load CoreCLR on Linux and jump into it.  These components are not part of the official build yet, so instead Makefiles are provided to build them on an ad-hoc basis.

With these changes, it's possible to run a pre-compiled console hello world application, using CoreCLR on Linux, via dnx <path-to-hello-world.dll>.

The next step is to get HelloWorld running when compiled dynamically via Roslyn.

There are still some TODOs in the code that will get addressed later, for now this is enough to start getting things working.

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -879,17 +879,17 @@ functions @{
         content.Add("#include \"stdafx.h\"");
         content.Add("#include \"tpa.h\"");
         content.Add(string.Empty);
-        content.Add("BOOL CreateTpaBase(LPWSTR** ppNames, size_t* pcNames, bool bNative)");
+        content.Add("BOOL CreateTpaBase(LPCTSTR** ppNames, size_t* pcNames, bool bNative)");
         content.Add("{");
         content.Add("    const size_t count = " + tpa.Length + ";");
-        content.Add("    LPWSTR* pArray = new LPWSTR[count];");
+        content.Add("    LPCTSTR* pArray = new LPCTSTR[count];");
         content.Add(string.Empty);
         content.Add("    if (bNative)");
         content.Add("    {");
 
         for (int i = 0; i < tpa.Length; ++i)
         {
-            content.Add(string.Format("        pArray[{0}] = _wcsdup(L\"{1}{2}\");", i, tpa[i], ".ni.dll"));
+            content.Add(string.Format("        pArray[{0}] = _T(\"{1}{2}\");", i, tpa[i], ".ni.dll"));
         }
 
         content.Add("    }");
@@ -898,7 +898,7 @@ functions @{
 
         for (int i = 0; i < tpa.Length; ++i)
         {
-            content.Add(string.Format("        pArray[{0}] = _wcsdup(L\"{1}{2}\");", i, tpa[i], ".dll"));
+            content.Add(string.Format("        pArray[{0}] = _T(\"{1}{2}\");", i, tpa[i], ".dll"));
         }
 
         content.Add("    }");
@@ -909,13 +909,8 @@ functions @{
         content.Add("    return true;");
         content.Add("}");
         content.Add(string.Empty);
-        content.Add("BOOL FreeTpaBase(const LPWSTR* values, const size_t count)");
+        content.Add("BOOL FreeTpaBase(const LPCTSTR* values, const size_t count)");
         content.Add("{");
-        content.Add("    for (size_t idx = 0; idx < count; ++idx)");
-        content.Add("    {");
-        content.Add("        delete[] values[idx];");
-        content.Add("    }");
-        content.Add(string.Empty);
         content.Add("    delete[] values;");
         content.Add(string.Empty);
         content.Add("    return true;");
--- a/makefile.shade
+++ b/makefile.shade
@@ -911,7 +911,7 @@ functions @{
         content.Add("    return true;");
         content.Add("}");
         content.Add(string.Empty);
-        content.Add("BOOL FreeTpaBase(const LPCTSTR* values, const size_t count)");
+        content.Add("BOOL FreeTpaBase(const LPCTSTR* values)");
         content.Add("{");
         content.Add("    delete[] values;");
         content.Add(string.Empty);
--- a/makefile.shade
+++ b/makefile.shade
@@ -628,6 +628,7 @@ var RC_FILES = '${FindAllFiles("Resource.rc", BOOTSTRAPPER_EXE_NAME, BOOTSTRAPPE
         {
             var dotnetcoreclrDepFile = Path.Combine(BUILD_DIR2, "dotnetcoreclr-dependencies.txt");
             var tpacppFile = Path.Combine(ROOT, "src", BOOTSTRAPPER_CORECLR_NAME, "tpa.cpp");
+            var tpacppFileUnix = Path.Combine(ROOT, "src", BOOTSTRAPPER_CORECLR_NAME + ".unix", "tpa.cpp");
             Environment.SetEnvironmentVariable("DNX_TRACE", "0");
             Exec("cmd", string.Format(@"/C kpm list {0} --assemblies --runtime {1} > {2}",
                 Path.Combine(ROOT, "src", BOOTSTRAPPER_HOST_NAME),
@@ -642,6 +643,7 @@ var RC_FILES = '${FindAllFiles("Resource.rc", BOOTSTRAPPER_EXE_NAME, BOOTSTRAPPE
             };
 
             UpdateTpa(dotnetcoreclrDepFile, tpaProjects, tpacppFile);
+            UpdateTpa(dotnetcoreclrDepFile, tpaProjects, tpacppFileUnix);
         }
         finally
         {
--- a/src/Microsoft.Framework.Runtime.Loader/LoadContext.cs
+++ b/src/Microsoft.Framework.Runtime.Loader/LoadContext.cs
@@ -60,7 +60,7 @@ namespace Microsoft.Framework.Runtime.Loader
         private string GetNativeImagePath(string ilPath)
         {
             var directory = Path.GetDirectoryName(ilPath);
-            var arch = Environment.GetEnvironmentVariable("PROCESSOR_ARCHITECTURE");
+            var arch = IntPtr.Size == 4 ? "x86" : "AMD64";
 
             var nativeImageName = Path.GetFileNameWithoutExtension(ilPath) + ".ni.dll";
             var nativePath = Path.Combine(directory, arch, nativeImageName);
--- a/src/dnx.coreclr.unix/dnx.coreclr.cpp
+++ b/src/dnx.coreclr.unix/dnx.coreclr.cpp
@@ -60,7 +60,7 @@ bool GetTrustedPlatformAssembliesList(const std::string& tpaDirectory, bool isNa
 
     if (ppszTpaAssemblyNames)
     {
-        FreeTpaBase(ppszTpaAssemblyNames, cTpaAssemblyNames);
+        FreeTpaBase(ppszTpaAssemblyNames);
     }
 
     return true;
--- a/src/dnx.coreclr.unix/dnx.coreclr.cpp
+++ b/src/dnx.coreclr.unix/dnx.coreclr.cpp
@@ -1,0 +1,231 @@
+#include "stdafx.h"
+#include "tpa.h"
+#include <assert.h>
+
+// Windows types used by the ExecuteAssembly function
+typedef int32_t HRESULT;
+typedef const char* LPCSTR;
+typedef uint32_t DWORD;
+
+typedef struct CALL_APPLICATION_MAIN_DATA
+{
+    char* applicationBase;
+    char* runtimeDirectory;
+    int argc;
+    char** argv;
+    int exitcode;
+} *PCALL_APPLICATION_MAIN_DATA;
+
+// Prototype of the ExecuteAssembly function from the libcoreclr.do
+typedef HRESULT (*ExecuteAssemblyFunction)(
+                    LPCSTR exePath,
+                    LPCSTR coreClrPath,
+                    LPCSTR appDomainFriendlyName,
+                    int propertyCount,
+                    LPCSTR* propertyKeys,
+                    LPCSTR* propertyValues,
+                    int argc,
+                    LPCSTR* argv,
+                    LPCSTR managedAssemblyPath,
+                    LPCSTR entryPointAssemblyName,
+                    LPCSTR entryPointTypeName,
+                    LPCSTR entryPointMethodsName,
+                    DWORD* exitCode);
+
+const HRESULT S_OK = 0;
+const HRESULT E_FAIL = -1;
+
+bool GetTrustedPlatformAssembliesList(const std::string& tpaDirectory, bool isNative, std::string& trustedPlatformAssemblies)
+{
+    size_t cTpaAssemblyNames = 0;
+    LPCTSTR* ppszTpaAssemblyNames = nullptr;
+
+    CreateTpaBase(&ppszTpaAssemblyNames, &cTpaAssemblyNames, isNative);
+
+    assert(ppszTpaAssemblyNames != nullptr || cTpaAssemblyNames == 0);
+
+    //TODO: The Windows version of this actaully ensures the files are present.  We just fail for native and assume MSIL is present
+    if (isNative)
+    {
+        return false;
+    }
+
+    for (size_t i = 0; i < cTpaAssemblyNames; i++)
+    {
+        trustedPlatformAssemblies.append(tpaDirectory);
+        trustedPlatformAssemblies.append("/");
+        trustedPlatformAssemblies.append(ppszTpaAssemblyNames[i]);
+        trustedPlatformAssemblies.append(":");
+    }
+
+    if (ppszTpaAssemblyNames)
+    {
+        FreeTpaBase(ppszTpaAssemblyNames, cTpaAssemblyNames);
+    }
+
+    return true;
+}
+
+// TODO: Figure out if runtimeDirectory should have a trailing "/".  Need to look at what happens on Windows.
+void* LoadCoreClr(std::string& runtimeDirectory)
+{
+    void* ret = nullptr;
+
+    char* coreClrEnvVar = getenv("CORECLR_DIR");
+
+    if (coreClrEnvVar)
+    {
+        runtimeDirectory = coreClrEnvVar;
+
+        std::string coreClrDllPath = runtimeDirectory;
+        coreClrDllPath.append("/");
+        coreClrDllPath.append("libcoreclr.so");
+
+        ret = dlopen(coreClrDllPath.c_str(), RTLD_NOW | RTLD_GLOBAL);
+    }
+
+    if (!ret)
+    {
+        // Try to load coreclr from application path.
+        char pathToBootstrapper[PATH_MAX];
+        ssize_t pathLen = readlink("/proc/self/exe", pathToBootstrapper, PATH_MAX - 1);
+
+        if (pathLen != -1)
+        {
+            pathToBootstrapper[pathLen] = '\0';
+            runtimeDirectory.assign(pathToBootstrapper);
+
+            size_t lastSlash = runtimeDirectory.rfind('/');
+
+            if (lastSlash != std::string::npos)
+            {
+                runtimeDirectory.erase(lastSlash);
+
+                std::string coreClrDllPath = runtimeDirectory;
+                coreClrDllPath.append("/");
+                coreClrDllPath.append("libcoreclr.so");
+
+                ret = dlopen(coreClrDllPath.c_str(), RTLD_NOW | RTLD_GLOBAL);
+            }
+        }
+    }
+
+    return ret;
+}
+
+extern "C" HRESULT CallApplicationMain(PCALL_APPLICATION_MAIN_DATA data)
+{
+    HRESULT hr = S_OK;
+    size_t cchTrustedPlatformAssemblies = 0;
+    std::string runtimeDirectory;
+
+    if (data->runtimeDirectory)
+    {
+        runtimeDirectory = data->runtimeDirectory;
+    }
+    else
+    {
+        // TODO: This should get the directory that this library is in, not the CWD.
+        char szCurrentDirectory[PATH_MAX];
+
+        if (!getcwd(szCurrentDirectory, PATH_MAX))
+        {
+            return E_FAIL;
+        }
+
+        runtimeDirectory = std::string(szCurrentDirectory);
+    }
+
+    std::string coreClrDirectory;
+    void* coreClr = LoadCoreClr(coreClrDirectory);
+
+    if (!coreClr)
+    {
+        char* error = dlerror();
+        fprintf(stderr, "failed to locate coreclr.dll with error %s\n", error);
+        return E_FAIL;
+    }
+
+    const char* property_keys[] =
+    {
+        "APPBASE",
+        "TRUSTED_PLATFORM_ASSEMBLIES",
+        "APP_PATHS",
+    };
+
+    std::string trustedPlatformAssemblies;
+
+    // Try native images first
+    if (!GetTrustedPlatformAssembliesList(coreClrDirectory.c_str(), true, trustedPlatformAssemblies))
+    {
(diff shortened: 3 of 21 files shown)


msotclas:
Hi __@ellismg__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <p>
        It looks like you're a Microsoft contributor (Matt Ellis). If you're full-time, we DON'T require a Contribution License Agreement. If you are a vendor, please DO sign the electronic Contribution License Agreement. It will take 2 minutes and there's no faxing! https://cla2.msopentech.com.
    </p>

TTYL, MSOTBOT;

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+    CreateTpaBase(&ppszTpaAssemblyNames, &cTpaAssemblyNames, bNative);
+
+    //TODO: The Windows version of this actaully ensures the files are present.  We just fail for native and assume MSIL is present
+    if (bNative)
+    {
+        return false;
+    }
+
+    trustedPlatformAssemblies = std::string();
```
BrennanConroy: .clear() instead

borgdylan:
Please do not remove mono support for now. CoreCLR is not ready for 32bit on Linux and EF7 is not mature enough for database stuff as of now.

davidfowl:
Nobody said we're removing support for mono.

src/dnx.coreclr.unix/tpa.cpp
```diff
@@ -0,0 +1,105 @@
+// TODO: Unify this file with the one generate at build time for
```
davidfowl: I was about to comment and saw this :smile:

src/dnx.coreclr.unix/tpa.cpp
```diff
@@ -0,0 +1,105 @@
...
+    *pcNames = count;
+
+    return true;
+}
+
+BOOL FreeTpaBase(const LPWSTR* values, const size_t count)
+{
+    for (size_t idx = 0; idx < count; ++idx)
+    {
+        delete[] values[idx];
```
rpetrich: Should perhaps use `free` instead of `delete[]`

erkinalp: rpetrich: `malloc` goes with `free`, `new` goes with `delete` and `delete[]`. They have different free lists and you will get a double free if you mix on the same object.

src/Microsoft.Framework.Runtime.Loader/LoadContext.cs
```diff
@@ -60,7 +60,7 @@ public static void InitializeDefaultContext(LoadContext loadContext)
         private string GetNativeImagePath(string ilPath)
         {
             var directory = Path.GetDirectoryName(ilPath);
-            var arch = Environment.GetEnvironmentVariable("PROCESSOR_ARCHITECTURE");
+            var arch = IntPtr.Size == 4 ? "x86" : "AMD64";
```
mmdriley: What about ARM?

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -0,0 +1,227 @@
...
+#include "tpa.h"
+
+// Windows types used by the ExecuteAssembly function
+typedef int32_t HRESULT;
+typedef const char* LPCSTR;
+typedef unsigned int DWORD;
+#define S_OK 0;
+#define E_FAIL -1;
+
+typedef struct CALL_APPLICTION_MAIN_DATA
```
mmdriley: 'APPLICATION'

src/Microsoft.Framework.Runtime.Loader/LoadContext.cs
ellismg: I asked @davidfowl about exactly this and he said they used this pattern elsewhere in the stack as well, so there's already debt.  If we actually end up needing to detect this, I would much prefer adding an API to the BCL and reading that vs environment variables.

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -0,0 +1,227 @@
+#include "stdafx.h"
+#include "tpa.h"
+
+// Windows types used by the ExecuteAssembly function
+typedef int32_t HRESULT;
+typedef const char* LPCSTR;
+typedef unsigned int DWORD;
```
mmdriley: more appropriately uint32_t

ellismg:
It would appear that I included the DNX binary itself in these commits.  That was unintentional and I will rebase to remove it.

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -0,0 +1,227 @@
...
+#define S_OK 0;
```
mmdriley: danger! semicolon in macro definition

let's just use const

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -0,0 +1,227 @@
...
+                    LPCSTR* propertyValues,
+                    int argc,
+                    LPCSTR* argv,
+                    LPCSTR managedAssemblyPath,
+                    LPCSTR entryPointAssemblyName,
+                    LPCSTR entryPointTypeName,
+                    LPCSTR entryPointMethodsName,
+                    DWORD* exitCode);
+
+bool GetTrustedPlatformAssembliesList(const std::string& szDirectory, bool bNative, std::string& trustedPlatformAssemblies)
```
mmdriley: seems like we should skip the Hungarian for szDirectory, at least. Maybe bNative too since it's a C++ bool.

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -0,0 +1,227 @@
...
+// TODO: Figure out if runtimeDirectory should have a trailing "/".  Need to look at what happens on Windows.
+void* LoadCoreClr(std::string& runtimeDirectory)
+{
+    void* ret = nullptr;
+
+    char* coreClrEnvVar = getenv("CORECLR_DIR");
+
+    if (coreClrEnvVar != nullptr)
+    {
+        runtimeDirectory = std::string("CORECLR_DIR");
```
mmdriley: can just assign the string literal.
runtimeDirectory = "CORECLR_DIR";

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+        std::string coreClrDllPath = runtimeDirectory;
+        coreClrDllPath.append("/");
+        coreClrDllPath.append("libcoreclr.so");
+
+        ret = dlopen(coreClrDllPath.c_str(), RTLD_NOW | RTLD_GLOBAL);
+    }
+
+    if (ret == nullptr)
```
mmdriley: if (!ret)

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -0,0 +1,227 @@
...
+                    LPCSTR managedAssemblyPath,
+                    LPCSTR entryPointAssemblyName,
+                    LPCSTR entryPointTypeName,
+                    LPCSTR entryPointMethodsName,
+                    DWORD* exitCode);
+
+bool GetTrustedPlatformAssembliesList(const std::string& szDirectory, bool bNative, std::string& trustedPlatformAssemblies)
+{
+    size_t cTpaAssemblyNames = 0;
+    char** ppszTpaAssemblyNames = NULL;
```
mmdriley: consistency, NULL vs. nullptr. Prefer the latter.

ellismg: Old habits die hard :-(.

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -0,0 +1,227 @@
...
+        char pathToBootstrapper[PATH_MAX];
+        ssize_t pathLen = readlink("/proc/self/exe", pathToBootstrapper, PATH_MAX - 1);
+
+        if (pathLen != -1)
+        {
+            pathToBootstrapper[pathLen] = '\0';
+            runtimeDirectory.assign(pathToBootstrapper);
+
+            size_t lastSlash = runtimeDirectory.rfind('/');
+            runtimeDirectory.erase(lastSlash);
```
mmdriley: What if no slash is found?

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+extern "C" HRESULT CallApplicationMain(PCALL_APPLICATION_MAIN_DATA data)
+{
+    HRESULT hr = S_OK;
+    size_t cchTrustedPlatformAssemblies = 0;
+    std::string runtimeDirectory;
+
+    if (data->runtimeDirectory != NULL)
+    {
+        runtimeDirectory = std::string(data->runtimeDirectory);
```
mmdriley: just assignment is fine here

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+    // Try native images first
+    if (!GetTrustedPlatformAssembliesList(coreClrDirectory.c_str(), true, trustedPlatformAssemblies))
+    {
+        if (!GetTrustedPlatformAssembliesList(coreClrDirectory.c_str(), false, trustedPlatformAssemblies))
+        {
+            fprintf(stderr, "Failed to find files in the coreclr directory\n");
+            hr = E_FAIL;
+            return hr;
+            // goto Finished;
```
mmdriley: ?

src/dnx.coreclr.unix/stdafx.h
```diff
@@ -0,0 +1,13 @@
+#include <dlfcn.h>
+#include <limits.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <string>
+#include <unistd.h>
+
+typedef char* LPWSTR;
```
mmdriley: seems wrong?

src/dnx.coreclr.unix/tpa.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include "tpa.h"
+
+BOOL CreateTpaBase(LPWSTR** ppNames, size_t* pcNames, bool bNative)
+{
+    const size_t count = 35;
+    LPWSTR* pArray = new LPWSTR[count];
+
+    if (bNative)
+    {
+        pArray[0] = strdup("Internal.Runtime.Uri.ni.dll");
```
mmdriley: this may be a job for vector<string> and std::initializer_list.
e.g.
std::vector<string> names = {
  "a!",
  "b!",
  // ...
};

src/dnx.coreclr.unix/dnx.coreclr.cpp
ellismg: The code here makes more sense if you look at how dnx.coreclr works in windows.  This call normally has the side effect of probing on disk to make sure everything on the TPA list is present.  The logic matches the existing windows probing behavior for the second level loader.

src/dnx.coreclr.unix/tpa.cpp
ellismg: I'll defer this feedback.  I want to unify this file with the existing generated version (but I will need to do a LPWSTR -> LPTSTR conversion first). Because of that, I don't want to make major changes.

src/dnx.coreclr.unix/tpa.cpp
mmdriley: Each of these values was allocated with strdup(), which is spec'd as allocating memory that must be deleted with free().
http://pubs.opengroup.org/onlinepubs/009695399/functions/strdup.html

Probably best we just get rid of the naked pointers altogether and use std::strings.
Best thing here is probably to use real strings and avoid the

src/dnx.coreclr.unix/dnx.coreclr.cpp
mmdriley: More curious about the 'goto' to a nonexistent label.

src/dnx.coreclr.unix/stdafx.h
ellismg: I agree but this is a point in time issue.  This is needed because I wanted to share as much of tpa.[h|cpp] between windows and *nix.  I will fix this later by moving to TCHAR like elsewhere.

src/dnx.coreclr.unix/tpa.h
```diff
@@ -0,0 +1,4 @@
+#pragma once
+
+BOOL CreateTpaBase(LPWSTR** ppNames, size_t* pcNames, bool bNative);
```
mmdriley: this header file uses a lot of types it doesn't include.

src/dnx.coreclr.unix/dnx.coreclr.cpp
ellismg: Ahh, yes.  I will remove this line.  It's vestigial code from a previous attempt at structuring this file.  Thanks!

src/dnx/dnx.cpp
```diff
@@ -1,97 +1,107 @@
...
-    for (dirLength--; dirLength >= 0 && szPath[dirLength] != '\\'; dirLength--);
-    szPath[dirLength + 1] = L'\0';
-}
-
-int LastIndexOfChar(LPCWSTR const pszStr, WCHAR c)
+bool LastIndexOfChar(LPCTSTR const pszStr, TCHAR c, size_t* pIndex)
 {
-    int nIndex = wcsnlen_s(pszStr, MAX_PATH) - 1;
-    do
+    size_t nIndex = _tcsnlen(pszStr, MAX_PATH) - 1;
```
mmdriley: The use of MAX_PATH seems really odd here. There's nothing in the function name to suggest this is dealing with paths.

src/dnx.coreclr.unix/tpa.h
ellismg: The existing version of the file already does the same thing.  They assume that windows.h and friends will already be included.

src/dnx/dnx.cpp
```diff
@@ -1,97 +1,107 @@
...
-    return nIndex;
+    *pIndex = nIndex;
+    return pszStr[nIndex] == c;
 }
 
-bool StringsEqual(LPCWSTR const pszStrA, LPCWSTR const pszStrB)
+bool StringsEqual(LPCTSTR const pszStrA, LPCTSTR const pszStrB)
 {
-    return ::_wcsnicmp(pszStrA, pszStrB, MAX_PATH) == 0;
+    return ::_tcsnicmp(pszStrA, pszStrB, MAX_PATH) == 0;
```
mmdriley: Likewise on the use of MAX_PATH. Is this supposed to be "PathsEqual"?

src/dnx.coreclr.unix/tpa.cpp
ellismg: I'll talk with the larger team about this.  I don't want to start making a bunch of changes to this file, since ideally we can unify it with the existing version.

davidfowl: agreed

mmdriley: erm, agreed with which part? the current code is incorrect, so it should not be checked in.

src/dnx.coreclr.unix/dnx.coreclr.cpp
erkinalp: Yes, `typedef int32_t DWORD;` will be more appropriate

ellismg: DWORD is unsigned.  Did you mean uint32_t?

src/Microsoft.Framework.Runtime.Loader/LoadContext.cs
davidfowl: Yep, we'll cross that bridge when we get to it. We're aware of the limitation.

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+    for (size_t i = 0; i < cTpaAssemblyNames; i++)
+    {
+        trustedPlatformAssemblies.append(szDirectory);
+        trustedPlatformAssemblies.append("/");
+        trustedPlatformAssemblies.append(ppszTpaAssemblyNames[i]);
+        trustedPlatformAssemblies.append(":");
+    }
+
+    if (ppszTpaAssemblyNames != NULL)
```
BrennanConroy: Seems odd to be doing a null check after the array has been used

src/dnx.coreclr.unix/dnx.coreclr.cpp
BrennanConroy: I think you meant
runtimeDirectory = coreClrEnvVar;

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+    std::string coreClrDllPath(coreClrDirectory);
+    coreClrDllPath.append("/");
+    coreClrDllPath.append("libcoreclr.so");
+
+    // TODO: The PAL needs an path which it can use as the app that is hosting the PAL.  Can this be some dummy thing?
+    // It looked like in some places in the CoreCLR it was hardcoded to just "CoreCLR".
+
+    char pathToBootstrapper[PATH_MAX];
+    ssize_t ret = readlink("/proc/self/exe", pathToBootstrapper, PATH_MAX - 1);
```
BrennanConroy: I liked the variable name pathLen you used earlier for similar code.

src/dnx.coreclr.unix/tpa.cpp
troydai: Should have used `free`. I'll go update the build scripts to generate correct result.

ellismg: @troydai I want to edit these scripts anyway, so I can take care of it.  Do you also know why we do strdup at all?  No one ever seems to modify the strings.  Can we just use literal strings and then get out of the free game?

troydai: @ellismg I don't think anything blocking us from doing this. When I wrote these codes only my cat (she stay by my side at the time) and me knows why I use `_strdup`. Now only my cat knows.   JK. I guess one of the reasons was trying to reuse the data between crossgened file names and original names. But end with hard coded all the codes because it helps to allow simplier build scripts.

src/dnx.coreclr.unix/dnx.coreclr.cpp
ellismg: Fixed.

src/dnx.coreclr.unix/dnx.coreclr.cpp
ellismg: typedef unsigned int DWORD -> typedef uint32_t DWORD

src/dnx.coreclr.unix/dnx.coreclr.cpp
ellismg: Fixed.

src/dnx.coreclr.unix/dnx.coreclr.cpp
ellismg: Fixed.

src/dnx.coreclr.unix/dnx.coreclr.cpp
ellismg: Fixed.

src/dnx.coreclr.unix/dnx.coreclr.cpp
ellismg: Fixed all the places we did an explicit test against nullptr.

src/dnx.coreclr.unix/dnx.coreclr.cpp
ellismg: Fixed.

src/dnx.coreclr.unix/dnx.coreclr.cpp
ellismg: Fixed (in the case rfind returns std::string::npos we just don't continue).

src/dnx.coreclr.unix/dnx.coreclr.cpp
ellismg: Fixed.

src/dnx.coreclr.unix/dnx.coreclr.cpp
ellismg: Fixed.

src/dnx.coreclr.unix/dnx.coreclr.cpp
ellismg: I think the (unstated) invariant here is that if ppszTpaAssemblyNames is null, then cTpaAssemblyNames is zero.  This same weird pattern is already in dnx.coreclr.  I will add an assert.

src/dnx.coreclr.unix/dnx.coreclr.cpp
ellismg: Fixed.

src/dnx.coreclr.unix/stdafx.h
ellismg: Fixed now instead, since I ended up touching tpa.cpp anyway.

src/dnx.coreclr.unix/tpa.cpp
ellismg: Fixed this all up and updated the generators.

src/dnx/dnx.cpp
ellismg: Fixed.

src/dnx/dnx.cpp
ellismg: Fixed.

ellismg:
Thank you @BrennanConroy and @mmdriley.  I have pushed new history which addresses the feedback.  When it was appropriate, I just rewrote the commits that had problems instead of introducing new fixup commits.  Please take another look.

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -0,0 +1,229 @@
...
+{
+    size_t cTpaAssemblyNames = 0;
+    LPCTSTR* ppszTpaAssemblyNames = nullptr;
+
+    CreateTpaBase(&ppszTpaAssemblyNames, &cTpaAssemblyNames, isNative);
+
+    assert(ppszTpaAssemblyNames != nullptr || cTpaAssemblyNames == 0);
+
+    //TODO: The Windows version of this actaully ensures the files are present.  We just fail for native and assume MSIL is present
+    if (isNative)
```
BrennanConroy: nit: Move if check to above CreateTpaBase

ellismg: I'm going to push back against this.  We need to remove the if (isNative) check in favor of doing the same thing for both native and the msil cases, which is looking on the file system to ensure all of the files from the TPA list are present.  For now, we just use the isNative check as a heuristic instead of actually probing the disk.  So long term the "should we return true or false from this method" decision is going to be made starting around L48.  If we move the code now, we'll just have to delete it later and it gets things further away from what we want, IMHO.

makefile.shade
```diff
@@ -909,13 +911,8 @@ functions @{
         content.Add("    return true;");
         content.Add("}");
         content.Add(string.Empty);
-        content.Add("BOOL FreeTpaBase(const LPWSTR* values, const size_t count)");
+        content.Add("BOOL FreeTpaBase(const LPCTSTR* values, const size_t count)");
```
mmdriley: This no longer needs a "count" parameter.

ellismg: Yes.  Fixed.

ellismg:
I don't believe there are any open issues here.  Anything left for me to do before we can merge this?

BrennanConroy:
I think it's good.

victorhurdugaci:
I was able to run it on Ubuntu :shipit:

ellismg:
Rebased on dev and pushed, 24dd1f8d

Issue #1405, opened 2015-03-13
Enable DTH plugins to recover.


NTaylorMullen:
Right now if a user attempts to register a DTH plugin and it fails no more work occurs. We need to notify the caller that the plugin couldn't be registered and add the potential plugin to a recovery state mechanic where it could potentially come online.

NTaylorMullen:
a8ad36225706eb0972f5695a3ad8fe0fef1c176f

PR #1413, opened 2015-03-16
Add DTH Plugin Registration recovery.

NTaylorMullen:
- Also modified all plugin activities to no longer throw but to send down error messages
- Separated message handling and message processing.
- Added faulted registration recovery bits.

#1405

### DIFF
--- a/src/Microsoft.Framework.DesignTimeHost/ApplicationContext.cs
+++ b/src/Microsoft.Framework.DesignTimeHost/ApplicationContext.cs
@@ -31,6 +31,8 @@ namespace Microsoft.Framework.DesignTimeHost
 
         private readonly Trigger<string> _appPath = new Trigger<string>();
         private readonly Trigger<string> _configuration = new Trigger<string>();
+        private readonly Trigger<Void> _pluginRegistration = new Trigger<Void>();
+        private readonly Trigger<Void> _pluginWorkNeeded = new Trigger<Void>();
         private readonly Trigger<Void> _filesChanged = new Trigger<Void>();
         private readonly Trigger<Void> _rebuild = new Trigger<Void>();
         private readonly Trigger<Void> _restoreComplete = new Trigger<Void>();
@@ -169,6 +171,8 @@ namespace Microsoft.Framework.DesignTimeHost
                     SendOutgoingMessages();
                 }
 
+                PerformPluginWork();
+
                 lock (_inbox)
                 {
                     // If there's no more messages queued then bail out.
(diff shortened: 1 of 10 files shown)


msotclas:
Hi __@NTaylorMullen__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <p>
        It looks like you're a Microsoft contributor (Taylor Mullen). If you're full-time, we DON'T require a Contribution License Agreement. If you are a vendor, please DO sign the electronic Contribution License Agreement. It will take 2 minutes and there's no faxing! https://cla2.msopentech.com.
    </p>

TTYL, MSOTBOT;

src/Microsoft.Framework.DesignTimeHost/ApplicationContext.cs
```diff
@@ -490,6 +497,34 @@ private bool PerformCompilation()
...
         }
 
+        private bool PerformPluginWork()
+        {
+            if (_pluginRegistration.WasAssigned ||
+                _pluginWorkNeeded.WasAssigned ||
+                _filesChanged.WasAssigned ||
+                _rebuild.WasAssigned ||
+                _restoreComplete.WasAssigned ||
+                _sourceTextChanged.WasAssigned)
```
davidfowl: Do you need to actually do something for all of these types of messages? That doesn't seem right. You only need to do work if there is a plugin registered.

NTaylorMullen: I can expose a `WorkNeeded` property on the `PluginHandler` to short-circuit all of this. I just figured the extra checking here was pretty quick and the `Process` method basically no-ops if there's nothing to do.

davidfowl: I'm not worried about the verbosity of the if statement. I'm not sure this is correct. No work should happen if a plugin is never registered.

NTaylorMullen:
Updated.

src/Microsoft.Framework.DesignTimeHost/ApplicationContext.cs
```diff
@@ -490,6 +501,37 @@ private bool PerformCompilation()
...
+        private bool ShouldPerformPluginWork()
+        {
+            return 
+                _pluginRegistration.WasAssigned ||
+                _pluginWorkNeeded.WasAssigned ||
+                (_pluginHandler.FaultedRegistrations &&
+                (_filesChanged.WasAssigned ||
+                _rebuild.WasAssigned ||
+                _restoreComplete.WasAssigned ||
+                _sourceTextChanged.WasAssigned));
```
davidfowl: Get rid of this one.

NTaylorMullen:
Updated.

NTaylorMullen:
Updated. Did not make `IAssemblyLoadContext` dynamic per `IPlugin`.

NTaylorMullen:
Good?

NTaylorMullen:
Getting this in tonight unless there's an objection :smile:

Issue #1386, opened 2015-03-12
Add lock file version check


davidfowl:
We have a version property in the lock file so that clients can fail and regen the lockfile when the format changes (or fail with an error message).

The reading should take in a required version.

/cc @anurse

PR #1419, opened 2015-03-16
Invalidate lockfile that doesn't meet required version

ChengTian:
parent https://github.com/aspnet/dnx/issues/1386

@davidfowl , once this PR is checked in, we can start to practice lockfile versioning and fix ApplicationHost2 build failure on *nix.


ChengTian:
Got personal sign-off from @davidfowl

Issue #1422, opened 2015-03-17
Kpm functional tests should show info of the runtime in use when they fail


ChengTian:
Thanks @davidfowl for pointing out the pain point.

PR #1423, opened 2015-03-17
Make tests show info of runtime in use when they are failed

ChengTian:
parent https://github.com/aspnet/dnx/issues/1422

@davidfowl 
The failed test case will show something like
```c#
NameSpace.Class.TestCaseMethod(flavor: "coreclr", os: "win", architecture: "x64") [FAIL]
```


davidfowl:
Much better! :shipit:

Issue #1418, opened 2015-03-16
--debug passes argument on, resulting in unrecognized option


BrennanConroy:
After attaching debugger get the below error message.

Waiting for the debugger to attach...
Debugger attached.
Specify --help for a list of available options and commands.
System.Exception: TODO: Error: unrecognized option '--debug' 
at Microsoft.Framework.Runtime.Common.CommandLine.CommandLineApplication.HandleUnexpectedArg(CommandLineApplication command, String[] args, Int32 index, String argTypeName)
   at Microsoft.Framework.Runtime.Common.CommandLine.CommandLineApplication.Execute(String[] args)
   at Microsoft.Framework.PackageManager.Program.Main(String[] args)

BrennanConroy:
I made a fix for this just now

victorhurdugaci:
@anurse said he will address this as part of the work he's doing to dnx

anurse:
I'm not actually working on the dnx native code right now, so feel free to push that change @BrennanConroy

PR #1426, opened 2015-03-17
Fix --debug flag

BrennanConroy:
#1418

### DIFF
--- a/src/dnx.host/RuntimeBootstrapper.cs
+++ b/src/dnx.host/RuntimeBootstrapper.cs
@@ -65,7 +65,6 @@ namespace dnx.host
             // We just need to capture them and clean them up.
             var optionAppbase = app.Option("--appbase <PATH>", "Application base directory path",
                 CommandOptionType.SingleValue);
-            var optionDebug = app.Option("--debug", "Wait for debugger to attach", CommandOptionType.NoValue);
             var optionLib = app.Option("--lib <LIB_PATHS>", "Paths used for library look-up",
                 CommandOptionType.MultipleValue);
             var optionDebug = app.Option("--debug", "Waits for the debugger to attach before beginning execution.",
--- a/src/dnx.host/RuntimeBootstrapper.cs
+++ b/src/dnx.host/RuntimeBootstrapper.cs
@@ -65,6 +65,7 @@ namespace dnx.host
             // We just need to capture them and clean them up.
             var optionAppbase = app.Option("--appbase <PATH>", "Application base directory path",
                 CommandOptionType.SingleValue);
+            var optionDebug = app.Option("--debug", "Wait for debugger to attach", CommandOptionType.NoValue);
             var optionLib = app.Option("--lib <LIB_PATHS>", "Paths used for library look-up",
                 CommandOptionType.MultipleValue);
             var optionDebug = app.Option("--debug", "Waits for the debugger to attach before beginning execution.",
--- a/src/dnx.mono.managed/EntryPoint.cs
+++ b/src/dnx.mono.managed/EntryPoint.cs
@@ -16,9 +16,29 @@ public class EntryPoint
     public static int Main(string[] arguments)
     {
         // Check for the debug flag before doing anything else
-        bool hasDebugWaitFlag = arguments.Any(arg => string.Equals(arg, "--debug", StringComparison.OrdinalIgnoreCase));
+        bool hasDebugWaitFlag = false;
+        for (var i = 0; i < arguments.Length; ++i)
+        {
+            //anything without - or -- is appbase or non-dnx command
+            if (arguments[i][0] != '-')
+            {
+                break;
+            }
+            if(arguments[i] == "--appbase")
+            {
+                //skip path argument
+                ++i;
+                continue;
+            }
+            if(arguments[i] == "--debug")
+            {
+                hasDebugWaitFlag = true;
+                break;
+            }
+        }
+
         if (hasDebugWaitFlag)
-        { 
+        {
             if (!Debugger.IsAttached)
             {
                 Process currentProc = Process.GetCurrentProcess();
@@ -50,7 +70,7 @@ public class EntryPoint
 
         return RuntimeBootstrapper.Execute(arguments);
     }
-    
+
     private static string[] ExpandCommandLineArguments(string[] arguments)
     {
         // If '--appbase' is already given and it has a value, don't need to expand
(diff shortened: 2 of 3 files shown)


BrennanConroy:
@davidfowl @victorhurdugaci

src/dnx/dnx.cpp
```diff
@@ -391,6 +391,14 @@ int wmain(int argc, wchar_t* argv[])
                 }
 
                 ::wprintf_s(L"Debugger attached.\r\n");
+
+                //remove --debug flag from arguments
+                if (argc > i)
```
BrennanConroy: Should be argc > i + 1
Also, need to move the argv[i] = L"\0"; to outside of the if check

src/dnx/dnx.cpp
```diff
@@ -391,6 +391,14 @@ int wmain(int argc, wchar_t* argv[])
...
+                {
+                    memcpy(argv + i, argv + i + 1, (argc - i) * sizeof(wchar_t));
+                    argv[i] = L"\0";
```
BrennanConroy: argv[argc - 1] = L"\0";

ellismg:
FYI: I will need to rebase my dnx changes after this goes in.

victorhurdugaci:
Looks good :shipit: (after you rebase)

davidfowl:
Don't like this fix. Handle it like appbase. Consume the argument in managed code.

anurse:
That makes the experience inconsistent when using batch/shell scripts like "kpm". If I run: `kpm --debug restore`, and the AppHost is the one responsible for removing the arg, it doesn't work:

1. DNX.exe waits for the debugger
2. AppHost gets "dnx --appbase ... Microsoft.Framework.ApplicationHost Microsoft.Framework.PackageManager --debug restore"
3. AppHost doesn't see the "--debug" arg so it flows through to KPM, which has to know about it.

IF we're going to support --debug anywhere (which the current code does), then the native code has to strip it. The alternative would be to only support `--debug` prior to the actual command (which means `kpm --debug restore` would NOT trigger the debug behavior unless kpm ALSO has a debug switch of it's own)

davidfowl:
It doesn't because app host isn't removing the flag. Look at how appbase is handled and do the same

anurse:
We can handle it the same way, but we have to change the native code that reads it to only accept it as an argument TO apphost. If it ends up somewhere AFTER the app command name, apphost ignores it and passes it along (yes, EVEN when you register the option in AppHost's ParseArgs). 

Right now, app host options can only appear BEFORE the name of the DLL to run because CommandLineOptions stops parsing Options after the first Argument appears.

davidfowl:
Ok, to be clear. Undo how the debug flag works today and treat this flag light --appbase. Look at the existing code and do whatever appbase does. Today --appbase is a flag that is parsed in native code and consumed by the first set of managed code before the app runs.

BrennanConroy:
Updated

anurse:
To truely be like `--appbase`, we also need to change the way the `--debug` flag is detected and acted on in native code. Right now, if `--debug` appears ANYWHERE in the command line, we wait for the debugger, but we'll only strip it out if it appears BEFORE the project.json path (because AppHost strips it out).

We should change the native code to **only** support the `--debug` switch **before** the project.json path. Individual K apps are responsible for setting up their own debug hook if they want it (i.e. `kpm --debug` would be handled in `kpm`)

BrennanConroy:
@anurse @davidfowl

anurse:
Looks good now :). :shipit:

src/dnx/dnx.cpp
```diff
@@ -375,8 +375,17 @@ int CallApplicationProcessMain(int argc, wchar_t* argv[])
 int wmain(int argc, wchar_t* argv[])
 {
     // Check for the debug flag before doing anything else
-    for (int i = 0; i < argc; i++)
+    for (int i = 1; i < argc; i++)
     {
+        //anything without - or -- is appbase or non-dnx command
+        if (::_wcsnicmp(argv[i], L"-", 1) != 0)
```
moozzyk: why not just `argv[I][0] == L'-'` ?

src/dnx/dnx.cpp
```diff
@@ -375,8 +375,17 @@ int CallApplicationProcessMain(int argc, wchar_t* argv[])
...
+        {
+            break;
+        }
+        if (StringsEqual(argv[i], L"--appbase"))
```
moozzyk: These two could be nested - if it does not start wit - it won't start with --

moozzyk: Forget it

src/dnx.mono.managed/EntryPoint.cs
```diff
@@ -16,9 +16,27 @@ public class EntryPoint
...
-        bool hasDebugWaitFlag = arguments.Any(arg => string.Equals(arg, "--debug", StringComparison.OrdinalIgnoreCase));
+        bool hasDebugWaitFlag = false;
+        foreach (var arg in arguments)
+        {
+            //anything without - or -- is appbase or non-dnx command
+            if (arg.ElementAt(0) != "-")
+            {
+                break;
+            }
+            if(arg == "--appbase")
```
victorhurdugaci: not needed

src/dnx.mono.managed/EntryPoint.cs
```diff
@@ -16,9 +16,27 @@ public class EntryPoint
     public static int Main(string[] arguments)
     {
         // Check for the debug flag before doing anything else
-        bool hasDebugWaitFlag = arguments.Any(arg => string.Equals(arg, "--debug", StringComparison.OrdinalIgnoreCase));
+        bool hasDebugWaitFlag = false;
+        foreach (var arg in arguments)
+        {
+            //anything without - or -- is appbase or non-dnx command
+            if (arg.ElementAt(0) != "-")
```
victorhurdugaci: ugh... not `ElementAt`. just use the indexer `[0]`

src/dnx.mono.managed/EntryPoint.cs
victorhurdugaci: Wait a second... why do you stop after `appbase`?

ellismg:
Please rebase on top of 24dd1f8d and fix up the merge conflicts.  You'll need to adopt the _T() macros in dnx.cpp and I think I moved some code you touched into the PAL.  Please let me know if you need help fixing things up.

BrennanConroy:
@ellismg rebased, could you take a quick look to see if I missed anything?

src/dnx.host/RuntimeBootstrapper.cs
```diff
@@ -65,6 +65,7 @@ public static Task<int> ExecuteAsync(string[] args)
             // We just need to capture them and clean them up.
             var optionAppbase = app.Option("--appbase <PATH>", "Application base directory path",
                 CommandOptionType.SingleValue);
+            var optionDebug = app.Option("--debug", "Wait for debugger to attach", CommandOptionType.NoValue);
```
BrennanConroy: Looks like someone else added this, will remove

ellismg:
@BrennanConroy LGTM

src/dnx.mono.managed/EntryPoint.cs
```diff
@@ -16,9 +16,29 @@ public class EntryPoint
...
-        bool hasDebugWaitFlag = arguments.Any(arg => string.Equals(arg, "--debug", StringComparison.OrdinalIgnoreCase));
+        bool hasDebugWaitFlag = false;
+        for (var i = 0; i < arguments.Length; ++i)
+        {
+            //anything without - or -- is appbase or non-dnx command
+            if (arguments[i][0] != '-')
+            {
+                break;
+            }
+            if(arguments[i] == "--appbase")
```
davidfowl: spacing!

davidfowl:
Did you add this to the *nix PAL as well?

BrennanConroy:
Change is only in dnx.cpp. No extra work needed.

davidfowl:
:shipit:

Issue #1371, opened 2015-03-11
Lockfile validation should handle parsing erros properly


ChengTian:
Reported by @davidfowl 

If the lockfile is not a valid JSON, don't throw exception. We should treat it as a out-of-date lockfile and give error messages reminding users to do `kpm restore`

PR #1429, opened 2015-03-17
Mark lock file as invalid when see parsing errors

ChengTian:
parent https://github.com/aspnet/dnx/issues/1371

### DIFF
--- a/src/Microsoft.Framework.Runtime/DependencyManagement/LockFileFormat.cs
+++ b/src/Microsoft.Framework.Runtime/DependencyManagement/LockFileFormat.cs
@@ -29,18 +29,29 @@ namespace Microsoft.Framework.Runtime.DependencyManagement
         {
             using (var textReader = new StreamReader(stream))
             {
-                using (var jsonReader = new JsonTextReader(textReader))
+                try
                 {
-                    while (jsonReader.TokenType != JsonToken.StartObject)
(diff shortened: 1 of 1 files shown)


src/Microsoft.Framework.Runtime/DependencyManagement/LockFileFormat.cs
```diff
@@ -71,7 +82,7 @@ private LockFile ReadLockFile(JObject cursor)
         {
             var lockFile = new LockFile();
             lockFile.Islocked = ReadBool(cursor, "locked", defaultValue: false);
-            lockFile.Version = ReadInt(cursor, "version", defaultValue: int.MaxValue);
+            lockFile.Version = ReadInt(cursor, "version", defaultValue: int.MinValue);
```
ChengTian: Version int.MinValue is always out-of-date. We use this way to mark a lockfile as invalid.

davidfowl: Why is the parameter defaultValue instead of minimumValue?

ChengTian: `defaultValue` means this value should be used when there is no `version` property. Here we want to use `int.MinValue` when `version` property doesn't appear in lockfile.

davidfowl:
:shipit:

PR #1432, opened 2015-03-18
Added fast path for known framework names

davidfowl:
- Cache desktop framework population to avoid disk IO
at startup.
- This should show a bigger difference when the dnx is
ngened.
- In the future we can look at deduping similar
assembly names across frameworks.

/cc @anurse


anurse:
Can't do a line level comment for this apparently:
`FrameworkDefinitions.TryPopulateFrameworkFastPath` -> `FrameworkDefinitions.TryPopulateKnownFramework`

anurse:
Also, should we be doing the Path.Combine during this populate or just storing the relative path and Path.Combining when we need them? That's a lot of string manipulation for a bunch of assemblies that aren't often loaded...

davidfowl:
Yea, Path.Combine does a lot of crap. We can probably to a concat since we know it's windows.

anurse:
:shipit: with comments. Can't we just defer the Path.Combine? We only need it at final load time don't we?

davidfowl:
Yep I was thinking about that too but it required more changes than I felt like doing at the moment.

PR #1433, opened 2015-03-18
Change the breadcrumbs folder

victorhurdugaci:
@ChengTian @glennc


glennc:
Yes, that's the path we agreed on in email. :shipit:

PR #1441, opened 2015-03-18
Breadcrumbs on background thread

victorhurdugaci:
@ChengTian @glennc @davidfowl

### DIFF
--- a/src/Microsoft.Framework.Runtime/DependencyManagement/NuGetDependencyResolver.cs
+++ b/src/Microsoft.Framework.Runtime/DependencyManagement/NuGetDependencyResolver.cs
@@ -200,9 +200,6 @@ namespace Microsoft.Framework.Runtime
             var cacheResolvers = GetCacheResolvers();
             var defaultResolver = new DefaultPackagePathResolver(_repository.RepositoryRoot);
 
-            Servicing.Breadcrumbs breadcrumbs = new Servicing.Breadcrumbs();
-            breadcrumbs.CreateRuntimeBreadcrumb();
-
             foreach (var dependency in packages)
             {
                 var packageInfo = _repository.FindPackagesById(dependency.Identity.Name)
@@ -225,9 +222,9 @@ namespace Microsoft.Framework.Runtime
 
                 _packageDescriptions[packageInfo.Id] = packageDescription;
 
-                if (Servicing.Breadcrumbs.IsPackageServiceable(packageDescription.Package))
+                if (Servicing.Breadcrumbs.Instance.IsPackageServiceable(packageDescription.Package))
                 {
-                    breadcrumbs.CreateBreadcrumb(packageInfo.Id, packageInfo.Version);
+                    Servicing.Breadcrumbs.Instance.AddBreadcrumb(packageInfo.Id, packageInfo.Version);
                 }
 
                 var group = packageInfo.LockFileLibrary.FrameworkGroups.FirstOrDefault(g => g.TargetFramework == targetFramework);
--- a/src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
+++ b/src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
@@ -2,15 +2,25 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.IO;
-using System.Reflection;
+using System.Threading;
 using NuGet;
 
 namespace Microsoft.Framework.Runtime.Servicing
 {
     public class Breadcrumbs
     {
+        // It is recommended that you disable breadcrumbs only for perf runs
+        private const string NoBreadcrumbsEnvironmentVariableName = "DNX_NO_BREADCRUMBS";
+
+        private static readonly string _logType = typeof(Breadcrumbs).Name;
+
+        private readonly bool _isEnabled;
         private readonly string _breadcrumbsFolder;
+        private readonly List<BreadcrumbInfo> _breadcrumbsToWrite = new List<BreadcrumbInfo>();
+
+        public static Breadcrumbs Instance { get; private set; } = new Breadcrumbs();
 
         public Breadcrumbs()
             : this(ResolveBreadcrumbsFolder())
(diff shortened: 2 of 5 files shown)


src/Microsoft.Framework.Runtime/DependencyManagement/NuGetDependencyResolver.cs
```diff
@@ -8,6 +8,7 @@
 using System.Runtime.Versioning;
 using Microsoft.Framework.Runtime.DependencyManagement;
 using NuGet;
+using System.Threading;
```
davidfowl: This is in the wrong place

victorhurdugaci:
Forgot to mention, we save ~100ms at startup with the background thread

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
```diff
@@ -109,7 +133,43 @@ private void CreateBreadcrumbFile(string fileName)
...
         {
-            Logger.TraceInformation("Failed to write servicing breadcrumb for {0} because an exception was thrown: {1}", fileName, exception);
+            Logger.TraceInformation(
+                "[{0}] Failed to write servicing breadcrumb for {1} because an exception was thrown: {2}",
+                _logType,
+                fileName,
+                exception);
+        }
+
+        private static BreadcrumbInfo GetRuntimeBreadcrumb()
```
davidfowl: Can we do this somewhere else?

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
```diff
@@ -5,12 +5,18 @@
...
 
 namespace Microsoft.Framework.Runtime.Servicing
 {
     public class Breadcrumbs
     {
+        private static readonly string _logType = typeof(Breadcrumbs).GetType().Name;
         private readonly string _breadcrumbsFolder;
+        private readonly Queue<BreadcrumbInfo> _breadcrumbsQueue = new Queue<BreadcrumbInfo>();
+
+        public static Breadcrumbs Current { get; private set; } = new Breadcrumbs();
```
davidfowl: Current makes it sounds like there's more than one. Get rid of this altogether.

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
```diff
@@ -5,12 +5,18 @@
...
+using System.Collections.Generic;
+using System.Threading;
 
 namespace Microsoft.Framework.Runtime.Servicing
 {
     public class Breadcrumbs
     {
+        private static readonly string _logType = typeof(Breadcrumbs).GetType().Name;
         private readonly string _breadcrumbsFolder;
+        private readonly Queue<BreadcrumbInfo> _breadcrumbsQueue = new Queue<BreadcrumbInfo>();
```
davidfowl: Just use a list.

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
```diff
@@ -37,36 +46,48 @@ public static bool IsPackageServiceable(PackageInfo package)
             return false;
         }
 
-        public void CreateBreadcrumb(string packageId, SemanticVersion packageVersion)
+        public void EnqueueBreadcrumb(string packageId, SemanticVersion packageVersion)
```
davidfowl: AddBreadcrumb?

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
```diff
@@ -37,36 +46,48 @@ public static bool IsPackageServiceable(PackageInfo package)
...
+        }
 
-            if (!string.IsNullOrWhiteSpace(version))
+        /// <summary>
+        /// Writes all the enqueued breadcrumbs to disk. 
+        /// </summary>
+        /// <returns></returns>
+        public void ProcessBreadcrumbsQueue()
+        {
+            while (_breadcrumbsQueue.Count > 0)
```
davidfowl: Just loop over the list and mark each one. The clear it.

src/Microsoft.Framework.Runtime/DependencyManagement/NuGetDependencyResolver.cs
```diff
@@ -272,6 +272,16 @@ public void Initialize(IEnumerable<LibraryDescription> packages, FrameworkName t
 
                 dependency.LoadableAssemblies = assemblies;
             }
+
+            // Process the breadcrumbs on a different thread
+            // but don't wait for it
+            var breadcrumbsThread = new Thread(() =>
```
davidfowl: Use the thread pool instead of making a dedicated thread.

victorhurdugaci:
@davidfowl updated the PR. What do you think?

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
```diff
@@ -109,7 +139,17 @@ private void CreateBreadcrumbFile(string fileName)
 
         private static void LogBreadcrumbsCreationFailure(string fileName, Exception exception)
         {
-            Logger.TraceInformation("Failed to write servicing breadcrumb for {0} because an exception was thrown: {1}", fileName, exception);
+            Logger.TraceInformation(
```
davidfowl: TraceError

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
```diff
@@ -37,36 +45,55 @@ public static bool IsPackageServiceable(PackageInfo package)
...
+        }
 
-            // Create both files for now until we get clear instructions about the format of the name
-            CreateBreadcrumbFile(packageId);
-            CreateBreadcrumbFile(packageId + "." + packageVersion);
+        public void WriteAllBreadcrumbs(bool background = false)
+        {
+            if (background)
+            {
+                ThreadPool.QueueUserWorkItem(_ =>
```
davidfowl: If we actually want the process to wait until breadcrumbs are written then we need to use a normal thread :cry:

davidfowl: Add an environment variable that makes this NOOP. I want to disable this when I'm running performance tests.

davidfowl: Leave the ThreadPool.QueueUserWorkItem it's fine.

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
```diff
@@ -2,15 +2,25 @@
...
-using System.Reflection;
+using System.Threading;
 using NuGet;
 
 namespace Microsoft.Framework.Runtime.Servicing
 {
     public class Breadcrumbs
     {
+        // It is recommended that you disable breadcrumbs only for perf runs
+        private const string NoBreadcrumbsEnvironmentVariableName = "DNX_NO_BREADCRUMBS";
```
victorhurdugaci: @troydai you might want to set this env var for perf runs

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
victorhurdugaci: Added the NOOP variable

davidfowl:
What happens when there is no servicing folder? None of this code runs right? It shoud all noop in that case. 

Did you test this out on mono?

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
troydai: Thanks for reminding.

victorhurdugaci:
@davidfowl No service folder = noop. Mono = tested already (no folder there)

davidfowl:
Where exactly does it noop?

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
```diff
@@ -27,46 +46,83 @@ public Breadcrumbs(string breadcrumbsFolder)
...
+        /// <summary>
+        /// Writes a breadcrumb on the disk 
+        /// </summary>
+        /// <param name="packageId">The ID of the package</param>
+        /// <param name="packageVersion">The version of the the package</param>
+        private void CreateBreadcrumb(string packageId, SemanticVersion packageVersion)
+        {
+            if (string.IsNullOrWhiteSpace(_breadcrumbsFolder))
+            {
+                return;
```
victorhurdugaci: @davidfowl "Where exactly does it noop?" Here. I could change it to use the `_isEnabled` property instead an really do nothing.

davidfowl: That was one of those leading questions. Use that flag to represent both cases.

victorhurdugaci: @davidfowl updated

victorhurdugaci:
Any more comments/concerns on this?

Issue #1439, opened 2015-03-18
Add RefreshDependencies message to DTH


davidfowl:
To separate rebuilding from refreshing dependencies.

PR #1442, opened 2015-03-18
Made some changes to DTH

davidfowl:
- Added RefreshDependencies message
- Fixed issue in DesignTimeAssemblyLoadContextFactory where it
wasn't picking up changes because of the trigger name changes
- Renamed _restoreComplete to _refreshDependencies

#1439


NTaylorMullen:
:shipit:

src/Microsoft.Framework.DesignTimeHost/DesignTimeAssemblyLoadContextFactory.cs
```diff
@@ -63,7 +63,7 @@ private IAssemblyLoadContextFactory GetRuntimeFactory()
                 }
 
                 // Add a cache dependency on restore complete to reevaluate dependencies
```
NTaylorMullen: update comment.

Issue #1366, opened 2015-03-10
Fix sporadic exception thrown during "kpm restore"


ChengTian:
Directory.Move() sporadically throws exceptions.

ChengTian:
Found a problem after pushing to CI.

The fix #1324 doesn't work because when `kpm restore` to a remote folder, `Directory.Move()` the temp folder from current machine to the remote machine throws exception:
```
System.IO.IOException: Source and destination path must have identical roots. Move will not work acr
oss volumes.
```

davidfowl:
Nice

ChengTian:
This can be frequently reproduced with a VM with following configuration:
OS: Windows 7 Enterprise (EN-US)
Hardware: CPU=2, RAM=3584, Arch=AMD64

muratg:
Looks like we're still looking for the right fix. Moving to Beta 5.

davidfowl:
Nooooooo

victorhurdugaci:
We should fix this as soon as possible. The cross plat CI machines are affected by this all the time. The runs are very unstable.

muratg:
adding @pranavkm in case he has some ideas..

pranavkm:
I'll spend some time looking at this - if we can't find a reasonable solution for this we can consider reverting the parallel restore behavior until we have something more reliable.

davidfowl:
Whats the state of this issue?

pranavkm:
I'll close this one and use https://github.com/aspnet/dnx/issues/1190 to track the issue.

PR #1443, opened 2015-03-19
Move Copy file from Bundle to FileOperationUtils

BrennanConroy:
Use Copy instead of Directory.Move since it sometimes throws
#1366 
@davidfowl @ChengTian


BrennanConroy:
Not a big fan of adding catch and retry. But I tested it and it works, the recursive copy didn't work.

ChengTian:
I just tested this branch, although it mitigates the problem by making the failure less frequent, the bug is not completely fixed. Here is the sporadic error I got:
untitled

Issue #1428, opened 2015-03-17
Rename kpm and subcommands


ChengTian:
`kpm` -> `dnu`
`kpm bundle` -> `dnu publish`
Cut `kpm add` (need to notify Tooling team, I think they are using it)

ChengTian:
Just got suggestion from @davidfowl , we will keep `kpm.cmd` and add `dnu.cmd` to have a gradual adoption of `dnu.cmd`. This can make the transition easier.

ChengTian:
The first stage is done. We are supporting both `kpm` and `dnu` now. This should allow a non-breaking transition.

However, we should force adoption of `dnu` and remove support of `kpm` at some point later.

@muratg , moving to next milestone.

ChengTian:
@muratg , @davidfowl decides that we do this now. Moving back to beta4 and working on it.

xtianus79:
is there any official documentation to this?  where is this coming from?

ChengTian:
Hi @xtianus79 , you should be able to find official announcements for breaking changes in our https://github.com/aspnet/Announcements repo.
For this specific topic:
https://github.com/aspnet/Announcements/issues/4

PR #1450, opened 2015-03-19
Add dnu scripts

ChengTian:
parent https://github.com/aspnet/dnx/issues/1428

@davidfowl , adding `dnu` without removing `kpm`


davidfowl:
:shipit:

PR #1451, opened 2015-03-19
Add "kpm publish" command

ChengTian:
parent https://github.com/aspnet/dnx/issues/1428

@davidfowl , adding `kpm publish` without removing `kpm bundle`

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Program.cs
+++ b/src/Microsoft.Framework.PackageManager/Program.cs
@@ -134,6 +134,58 @@ namespace Microsoft.Framework.PackageManager
                 });
             });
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.Framework.PackageManager/Program.cs
```diff
@@ -134,6 +134,58 @@ public int Main(string[] args)
                 });
             });
 
+            app.Command("publish", c =>
+            {
+                c.Description = "Bundle application for deployment";
```
davidfowl: Publish

davidfowl:
:shipit:

Issue #1388, opened 2015-03-12
'kpm restore' fails when api.nuget.org (a nuget v3 feed) is enabled


deepakaravindr:
'kpm restore' fails when api.nuget.org (a nuget v3 feed) is enabled

Either
1) We need 'kpm restore' to understand and consume nuget v3 feeds OR
2) skip nuget v3 feeds and not fail restore

PR #1455, opened 2015-03-20
"kpm restore" ignores NuGet v3 feeds

ChengTian:
parent https://github.com/aspnet/dnx/issues/1388

@davidfowl @deepakaravindr @feiling


anurse:
Bit of a hack, but it should work for now. :shipit:

PR #1468, opened 2015-03-24
Build errors during no-source bundle should give non-zero exit code

ChengTian:
Replacing https://github.com/aspnet/dnx/pull/1031

@davidfowl @muratg


muratg:
:shipit:

PR #1471, opened 2015-03-24
Rename kpm to dnu

ChengTian:
https://github.com/aspnet/Universe/issues/186

@davidfowl @muratg


src/Microsoft.Framework.Runtime.Common/Impl/Constants.cs
```diff
@@ -12,7 +12,7 @@ internal static class Constants
         public const string WebConfigRuntimeVersion = RuntimeNamePrefix + "version";
         public const string WebConfigRuntimeFlavor = RuntimeNamePrefix + "clr";
         public const string WebConfigRuntimeAppBase = RuntimeNamePrefix + "app-base";
-        public const string WebConfigKpmPackagePath = "kpm-package-path";
+        public const string WebConfigPackagePath = "package-path";
```
davidfowl: Did you update helios too?

davidfowl:
:shipit:

src/Microsoft.Framework.Runtime.Common/Impl/Constants.cs
ChengTian: No, Helios is not reading this key

Issue #1476, opened 2015-03-24
Regression with latest dnx


davidfowl:
File new project in visual studio shows this error in DTH:

```
Information: [ConnectionContext]: Creating new application context for 1
Information: [ApplicationContext]: Set DTH protocol version to 1.
Information: [ApplicationContext]: Received Initialize
Information: [ApplicationContext]: Set DTH protocol version to 1.
Error: [ApplicationContext]: Error occured: System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Framework.DesignTimeHost.ApplicationContext.<>c__DisplayClass56_0.<GetApplicationHostContext>b__0(CacheContext ctx)
   at Microsoft.Framework.Runtime.CacheExtensions.<>c__DisplayClass0_0`1.<Get>b__0(CacheContext ctx)
   at Microsoft.Framework.Runtime.Cache.CreateEntry(Object k, Func`2 acquire)
   at Microsoft.Framework.Runtime.Cache.<>c__DisplayClass5_0.<AddEntry>b__0()
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.LazyInitValue()
   at System.Lazy`1.get_Value()
   at Microsoft.Framework.Runtime.Cache.Get(Object key, Func`2 factory)
   at Microsoft.Framework.Runtime.CacheExtensions.Get[T](ICache cache, Object key, Func`2 factory)
   at Microsoft.Framework.DesignTimeHost.ApplicationContext.GetApplicationHostContext(Project project, String configuration, FrameworkName frameworkName, Boolean useRuntimeLoadContextFactory)
   at Microsoft.Framework.DesignTimeHost.ApplicationContext.<>c__DisplayClass58_0.<ResolveProjectDepencies>b__0(CacheContext ctx)
   at Microsoft.Framework.Runtime.CacheExtensions.<>c__DisplayClass0_0`1.<Get>b__0(CacheContext ctx)
   at Microsoft.Framework.Runtime.Cache.CreateEntry(Object k, Func`2 acquire)
   at Microsoft.Framework.Runtime.Cache.<>c__DisplayClass5_0.<AddEntry>b__0()
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.LazyInitValue()
   at System.Lazy`1.get_Value()
   at Microsoft.Framework.Runtime.Cache.Get(Object key, Func`2 factory)
[...]

PR #1478, opened 2015-03-24
Fix a bug in DTH when there is no lockfile

ChengTian:
parent https://github.com/aspnet/dnx/issues/1476

@davidfowl @muratg , verified with both VS and command-line


ChengTian:
It's a trivial one, merging now to get it into beta4 release branch

Issue #1472, opened 2015-03-24
Remove "kpm-package-path" from web.config


ChengTian:
Looks like nobody is reading it except for `Stress`:
https://github.com/search?q=kpm-package-path+user%3Aaspnet&ref=simplesearch&type=Code&utf8=%E2%9C%93

@troydai

PR #1480, opened 2015-03-24
Remove "package-path" key from web.config

ChengTian:
parent https://github.com/aspnet/dnx/issues/1472

@davidfowl @muratg


ChengTian:
Trivial one, already did local `build verify`

muratg:
:shipit:

PR #1491, opened 2015-03-25
Fix dnx trace message

ellismg:
My previous changes caused garbage to be printed to the console when DNX_TRACE=1 was set on Windows.  This change fixes that.

I also updated the TPA list since when I built new stuff was pended.

### DIFF
--- a/src/dnx.coreclr.unix/tpa.cpp
+++ b/src/dnx.coreclr.unix/tpa.cpp
@@ -6,7 +6,7 @@
 
 BOOL CreateTpaBase(LPCTSTR** ppNames, size_t* pcNames, bool bNative)
 {
-    const size_t count = 34;
+    const size_t count = 35;
     LPCTSTR* pArray = new LPCTSTR[count];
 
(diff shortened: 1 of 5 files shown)


ellismg:
@davidfowl PTAL, this addresses the issue you told me about.

davidfowl:
@anurse Can you make sure you didn't introduce this new dependency into the TPA list: - System.Diagnostics.Tracing

davidfowl:
:shipit:

ellismg:
Per conversation with @davidfowl, I will just merge the actual dnx change and hold of on the TPA updates

anurse:
System.Diagnostics.Tracing is ETW stuff, I don't see any reference to it in Logging, so I'm not sure how it would have arrived from that. I'll do some spelunking.

anurse:
@davidfowl I don't see System.Diagnostics.Tracing anywhere, not sure how it would end up in the TPA in this PR. It's not appearing in the TPA in my local dev build.

PR #1497, opened 2015-03-25
Update the bundle manager to use file globbing correctly

troydai:
This is intended to address the issue that long paths will break the the `dnu bundle/publish` scenario. 

This issue is caused by incorrect usage of file globbing. Instead of make bundle excluding pattern played as part of excluding patterns in matcher, the original code uses it as include pattern to search files and then explicitly excludes them from content files set. The logic derives from the original implementation, a precedent to the introduction of file globbing.

This change will use the bundle excluding patterns in matcher as part of the excluding pattern group.


troydai:
/cc @DamianEdwards @davidfowl @ChengTian @muratg 

This is intend for beta4. I preliminary tested on @DamianEdwards 's case. More tests need to done before it is merged.

A few more tests will be added soon to guard the long path scenario.

src/Microsoft.Framework.PackageManager/Bundle/BundleOperations.cs
```diff
@@ -17,15 +18,40 @@ public void Delete(string folderPath)
             FileOperationUtils.DeleteFolder(folderPath);
         }
 
-        public void Copy(string sourcePath, string targetPath)
+        public void Copy(IEnumerable<string> sourceFiles, string sourceDirectory, string targetDirectory)
         {
```
moozzyk: This is a public method on a public class - doesn't it need some argument Null checking?

ChengTian:
Just a reminder: beta4 changes should be targeting `release` branch now

troydai:
@ChengTian I'll soon rebase this on release.

src/Microsoft.Framework.PackageManager/Bundle/BundleOperations.cs
troydai: Will check.

src/Microsoft.Framework.PackageManager/Bundle/BundleOperations.cs
```diff
@@ -17,15 +18,40 @@ public void Delete(string folderPath)
...
-                targetPath, 
-                shouldInclude: _ => true);
-        }
+            sourceDirectory = PathUtility.EnsureTrailingSlash(sourceDirectory);
+            targetDirectory = PathUtility.EnsureTrailingSlash(targetDirectory);
+
+            foreach (var sourceFilePath in sourceFiles)
+            {
+                var fileName = Path.GetFileName(sourceFilePath);
+                Debug.Assert(fileName != null, "fileName != null");
```
ChengTian: In addition to null value check, I guess you also want to check the case where `sourceFilePath` is a path ending with path separator (i.e. a path to dir). In this case, `fileName` is `string.Empty`.

So I suggest use `string.IsNullOrEmpty(fileName)`

src/Microsoft.Framework.Runtime.Hosting/ProjectFilesCollection.cs
```diff
@@ -88,9 +82,20 @@ public IEnumerable<string> SharedFiles
             get { return _sharedPatternsGroup.SearchFiles(_projectDirectory).Distinct(); }
         }
 
-        public IEnumerable<string> ContentFiles
+        public IEnumerable<string> GetFilesForBunding(bool includeSource, IEnumerable<string> additionalExcludePatterns)
```
ChengTian: Bundling?

src/Microsoft.Framework.PackageManager/Bundle/BundleOperations.cs
troydai: I don't want to change the existing behavior to complicate the case. The source files collection return from `Matcher` is always a absolute file paths collection.

troydai:
https://github.com/aspnet/dnx/issues/1500

davidfowl:
We need to have some kind of test here

troydai:
Working on tests now

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuBundleTests.cs
```diff
@@ -453,11 +453,13 @@ public void WildcardMatchingFacts(string flavor, string os, string architecture)
...
         '.': ['project.json', 'project.lock.json'],
         'MixFolder1': {
+          '.': ['uselessfile1.txt', 'uselessfile2'],
           'UsefulSub': ['useful.txt', 'useful']
         },
         'MixFolder2': {
+          '.': ['uselessfile1.txt', 'uselessfile2'],
           'UsefulSub': ['useful.txt', 'useful']
-        }
+        },
```
Eilon: Invalid JSON

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuBundleTests.cs
```diff
@@ -471,8 +473,6 @@ public void WildcardMatchingFacts(string flavor, string os, string architecture)
     ""UselessFolder1\\**"",
     ""UselessFolder2/**/*"",
     ""UselessFolder3\\**/*.*"",
```
Eilon: Invalid JSON

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuBundleTests.cs
```diff
@@ -497,8 +497,6 @@ public void WildcardMatchingFacts(string flavor, string os, string architecture)
...
```
Eilon: Invalid JSON (BTW these are all trailing commas)

troydai:
:watch: @davidfowl @DamianEdwards

troydai:
rebased to `release` branch. build verified.

troydai:
Guys, Ping!

troydai:
@DamianEdwards is testing this change. So far so good.

Getting more people to review the code.
/cc @muratg @Praburaj @Eilon

src/Microsoft.Framework.PackageManager/Bundle/BundleOperations.cs
muratg: In that case, IsNullOrEmpty wouldn't change the behavior, right? I think Wei's suggestion makes sense.  In case Matcher changes in the future, that would keep this code path correct.  Alternatively, you can consider adding a test case to cover this.

src/Microsoft.Framework.PackageManager/Bundle/BundleProject.cs
```diff
@@ -241,15 +241,13 @@ private static void UpdateJson(string jsonFile, Action<JObject> modifier)
 
         private void CopyProject(BundleRoot root, Runtime.Project project, string targetPath, bool includeSource)
         {
-            // A set of excluded files/directories used as a filter when doing copy
-            var excludeSet = new HashSet<string>(project.Files.BundleExcludeFiles, StringComparer.OrdinalIgnoreCase);
-            var contentFiles = new HashSet<string>(project.Files.ContentFiles, StringComparer.OrdinalIgnoreCase);
+            var addtionalExcluding = new List<string>();
```
muratg: Typo nit: "add```it```ional"

muratg:
Looks good to me (apart from the comments above)

:shipit:

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuBundleTests.cs
```diff
@@ -1256,5 +1254,111 @@ public void BundleWithNoSourceOptionUpdatesLockFileOnClr(string flavor, string o
...
+                        BasicLockFile)
+                    .WithFileContents(Path.Combine("approot", "global.json"), @"{
+  ""packages"": ""packages""
+}");
+                Assert.True(expectedOutputDir.MatchDirectoryOnDisk(testEnv.BundleOutputDirPath,
+                    compareFileContents: true));
+            }
+        }
+
+        private string BuildLongPath(string baseDir)
```
davidfowl: Is this going to create a path that I can't erase?

Eilon: BTW in Windows in Explorer you can *delete* long paths, but only *if* you can navigate to the parent folder (so the parent folder is not itself a long path), and then Explorer will warn you that the files won't go to the Recycle Bin.

Eilon: And of course you can always navigate to the folder via UNC, i.e. `\\localhost\c$\some\long\path\folder` and do whatever operations you want (UNC paths don't have the same short MAX_PATH limit).

dougbu: or you can use `rimraf` (as part of your `node` installation)

troydai:
Rebased. Will be merge into release in seconds.

Issue #1492, opened 2015-03-25
Running dnu commands on mono null refs


davidfowl:
```
Davids-MacBook-Pro:helloworld dfowler$ dnu commands
System.NullReferenceException: Object reference not set to an instance of an object
  at Microsoft.Framework.Runtime.Common.CommandLine.CommandLineApplication.ShowHelp (System.String commandName) [0x00000] in <filename unknown>:0 
  at Microsoft.Framework.PackageManager.Program+<>c__DisplayClass3_13.<Main>b__26 () [0x00000] in <filename unknown>:0 
  at Microsoft.Framework.Runtime.Common.CommandLine.CommandLineApplication.Execute (System.String[] args) [0x00000] in <filename unknown>:0 
  at Microsoft.Framework.PackageManager.Program.Main (System.String[] args) [0x00000] in <filename unknown>:0 
  at (wrapper managed-to-native) System.Reflection.MonoMethod:InternalInvoke (System.Reflection.MonoMethod,object,object[],System.Exception&)
  at System.Reflection.MonoMethod.Invoke (System.Object obj, BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00000] in <filename unknown>:0 
Davids-MacBook-Pro:helloworld dfowler$ 
```

/cc @victorhurdugaci

victorhurdugaci:
It happens both on mono and windows and it caused by [this fix](https://github.com/aspnet/dnx/commit/5d8ab2b1eb5c0a702a73a27034b0f19cda8a96ca).

Particular, [this line](https://github.com/aspnet/dnx/commit/5d8ab2b1eb5c0a702a73a27034b0f19cda8a96ca) because `VersionGetter` is null.

@BrennanConroy can you please take a look?

davidfowl:
@victorhurdugaci I can't repro on windows

BrennanConroy:
I can, using dnx v11360

victorhurdugaci:
![untitled](https://cloud.githubusercontent.com/assets/5921642/6832417/8bc46698-d2e3-11e4-8916-29d0cc9adae9.png)

davidfowl:
Ah hah, I was using 51

davidfowl:
So these kinds of regressions are bad. We need to beef up our testing. Manual and automated.

victorhurdugaci:
I agree, we should have more tests for scenarios that are not hit everyday

[...]

PR #1498, opened 2015-03-25
Checking if version exists before using it

BrennanConroy:
@davidfowl @victorhurdugaci @Eilon @danroth27 
#1492

### DIFF
--- a/src/Microsoft.Framework.CommandLineUtils/CommandLine/CommandLineApplication.cs
+++ b/src/Microsoft.Framework.CommandLineUtils/CommandLine/CommandLineApplication.cs
@@ -392,8 +392,15 @@ namespace Microsoft.Framework.Runtime.Common.CommandLine
             }
             headerBuilder.AppendLine();
 
-            var VersionAndName = string.Format("{0} v{1}{2}{2}", FullName, VersionGetter(), Environment.NewLine);
-            Console.Write("{0}{1}{2}{3}{4}", VersionAndName, headerBuilder, argumentsBuilder, optionsBuilder, commandsBuilder);
+            var versionAndName = new StringBuilder(FullName);
+            if (VersionGetter != null)
+            {
+                versionAndName.Append(string.Format(" v{0}", VersionGetter()));
+                versionAndName.AppendLine();
+            }
+            versionAndName.AppendLine();
+
+            Console.Write("{0}{1}{2}{3}{4}", versionAndName, headerBuilder, argumentsBuilder, optionsBuilder, commandsBuilder);
         }
 
         public void ShowVersion()


davidfowl:
No test?

BrennanConroy:
Wrote on the issue that I will write tests for the dnu/dnx commands after this fix.

Sent from my HTC

----- Reply message -----

davidfowl:
Why? A simple unit test will suffice.

BrennanConroy:
Because I'm going to write a bunch of them and wanted this fix out asap. I can write 1 for now.

Sent from my HTC

----- Reply message -----

Eilon:
Why exactly is the version getter null?

src/Microsoft.Framework.CommandLineUtils/CommandLine/CommandLineApplication.cs
```diff
@@ -392,7 +392,13 @@ public void ShowHelp(string commandName = null)
             }
             headerBuilder.AppendLine();
 
-            var VersionAndName = string.Format("{0} v{1}{2}{2}", FullName, VersionGetter(), Environment.NewLine);
+            var VersionAndName = new StringBuilder(FullName);
+            if (VersionGetter != null)
+            {
+                VersionAndName.Append(string.Format(" v{0}{1}", VersionGetter(), Environment.NewLine));
```
moozzyk: `AppendLine` ?

src/Microsoft.Framework.CommandLineUtils/CommandLine/CommandLineApplication.cs
```diff
@@ -392,7 +392,13 @@ public void ShowHelp(string commandName = null)
...
+            }
+            VersionAndName.Append(Environment.NewLine);
```
moozzyk: same as above...

src/Microsoft.Framework.CommandLineUtils/CommandLine/CommandLineApplication.cs
```diff
@@ -392,7 +392,13 @@ public void ShowHelp(string commandName = null)
             }
             headerBuilder.AppendLine();
 
-            var VersionAndName = string.Format("{0} v{1}{2}{2}", FullName, VersionGetter(), Environment.NewLine);
+            var VersionAndName = new StringBuilder(FullName);
```
davidfowl: uppercase? Fix this since you're changing it

src/Microsoft.Framework.CommandLineUtils/CommandLine/CommandLineApplication.cs
davidfowl: +1

src/Microsoft.Framework.CommandLineUtils/CommandLine/CommandLineApplication.cs
davidfowl: +1

BrennanConroy:
It is only null for Microsoft.Framework.Project, we could add it, similar to Microsoft.Framework.ApplicationHost's

Eilon:
:shipit: for beta4, and then add tests to beta5 I guess.

Issue #1488, opened 2015-03-24
Appbase passed to dnx cannot have trailing slash if there is no global.json


ChengTian:
Repro steps:
(1) git clone `Home` repo
(2) cd `Home/samples/ConsoleApp`
(3) `dnu restore`
(4) `dnx . run`, works well
(5) `dnx ./ run`, fails with the following exception:
```
System.InvalidOperationException: Unable to resolve project 'ConsoleApp' from C:\AspNet\Home\samples\ConsoleApp\
   at Microsoft.Framework.Runtime.ApplicationHostContext..ctor(IServiceProvider serviceProvider, String projectDirectory
, String packagesDirectory, String configuration, FrameworkName targetFramework, ICache cache, ICacheContextAccessor cac
heContextAccessor, INamedCacheDependencyProvider namedCacheDependencyProvider, IAssemblyLoadContextFactory loadContextFa
ctory, Boolean skipLockFileValidation) in C:\AspNet\dnx\src\Microsoft.Framework.Runtime\ApplicationHostContext.cs:line 5
7
   at Microsoft.Framework.Runtime.DefaultHost.Initialize(DefaultHostOptions options, IServiceProvider hostServices) in C
:\AspNet\dnx\src\Microsoft.Framework.Runtime\DefaultHost.cs:line 116
   at Microsoft.Framework.Runtime.DefaultHost..ctor(DefaultHostOptions options, IServiceProvider hostServices) in C:\Asp
Net\dnx\src\Microsoft.Framework.Runtime\DefaultHost.cs:line 37
   at Microsoft.Framework.ApplicationHost.Program.Main(String[] args) in C:\AspNet\dnx\src\Microsoft.Framework.Applicati
onHost\Program.cs:line 42
```
(6) Add a `global.json` to root of `Home` repo with the following contents:
```json
{
  "projects": ["samples"]
}
```
Then `dnx ./ run` works.

@davidfowl , looks like a project dependency resolver bug? Need to investigate why there is difference between using `.` and `./` as appbase.

davidfowl:
Is this a regression?

ChengTian:
Yes, I remember I was able to run `klr ./ run` in `KRE/samples/HelloWorld`

PR #1505, opened 2015-03-26
Fix regression in ProjectResolver

ChengTian:
https://github.com/aspnet/dnx/issues/1488

@davidfowl @muratg

### DIFF
--- a/src/Microsoft.Framework.Runtime/ProjectResolver.cs
+++ b/src/Microsoft.Framework.Runtime/ProjectResolver.cs
@@ -49,7 +49,7 @@ namespace Microsoft.Framework.Runtime
 
         private void Initialize(string projectPath, string rootPath)
         {
-            _searchPaths.Add(Path.GetDirectoryName(projectPath));
+            _searchPaths.Add(new DirectoryInfo(projectPath).Parent.FullName);
 
             GlobalSettings global;
 
--- a/test/Bootstrapper.FunctionalTests/BootstrapperTests.cs
+++ b/test/Bootstrapper.FunctionalTests/BootstrapperTests.cs
@@ -74,7 +74,7 @@ namespace Bootstrapper.FunctionalTests
 
         [Theory]
         [MemberData("RuntimeComponents")]
-        public void BootstrapperInvokesApplicationHostWithInferredAppBase(string flavor, string os, string architecture)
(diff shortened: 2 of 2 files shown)


src/Microsoft.Framework.Runtime/ProjectResolver.cs
```diff
@@ -49,7 +49,7 @@ public bool TryResolveProject(string name, out Project project)
 
         private void Initialize(string projectPath, string rootPath)
         {
-            _searchPaths.Add(Path.GetDirectoryName(projectPath));
+            _searchPaths.Add(new DirectoryInfo(projectPath).Parent.FullName);
```
davidfowl: Path.GetFullPath?

ChengTian: I am using `FullName`

davidfowl: I meant instead of this entire line

ChengTian:
I should add a test case covering this case. Adding test now.

src/Microsoft.Framework.Runtime/ProjectResolver.cs
anurse: `Path.GetDirectoryName(Path.GetFullPath(projectPath))` should do it

davidfowl: ^

ChengTian: `Path.GetFullPath()` doesn't help here.

To make it clear, `projectPath` is a directory, that may have a trailing slash. We want to get the parent dir of `projectPath`.

The current approach is the most readable one. If you think newing an object here may increase program startup time, I can do a naïve trailing slash removal here.

ChengTian: Path.GetDirectoryName("this/is/a/path") returns "this/is/a"
Path.GetDirectoryName("this/is/a/path/") returns "this/is/a/path"
Path.GetFullPath() doesn't remove trailing slash for you

ChengTian:
Added test

davidfowl:
:shipit:

src/Microsoft.Framework.Runtime/ProjectResolver.cs
anurse: Ah, good catch. I stand (or rather sit) corrected :)

PR #1509, opened 2015-03-27
Add a --parallel flag to restore command

pranavkm:
Based on http://stackoverflow.com/a/3408642/235591.

Running `dnu restore` in a loop with this change seems to suggest the issue is no longer exhibited. We could start by taking this change and looking at something more substantial if we continue seeing the issue in the wild.

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Program.cs
+++ b/src/Microsoft.Framework.PackageManager/Program.cs
@@ -61,6 +61,9 @@ namespace Microsoft.Framework.PackageManager
                 var optUnlock = c.Option("--unlock",
                     "Creates dependencies file with locked property set to false. Overwrites file if it exists.",
                     CommandOptionType.NoValue);
+                var optParallel = c.Option("--parallel",
+                    "Restores in parallel when more than one project.json is discovered.",
+                    CommandOptionType.NoValue);
                 c.HelpOption("-?|-h|--help");
 
                 c.OnExecute(async () =>
@@ -71,6 +74,7 @@ namespace Microsoft.Framework.PackageManager
                     command.FeedOptions = feedOptions;
                     command.Lock = optLock.HasValue();
                     command.Unlock = optUnlock.HasValue();
+                    command.Parallel = optParallel.HasValue();
 
                     if (feedOptions.ProxyOptions.HasValue())
                     {
--- a/src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
@@ -13,7 +13,6 @@ using System.Security.Cryptography;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Framework.PackageManager.Bundle;
-using Microsoft.Framework.PackageManager.Restore.NuGet;
 using Microsoft.Framework.PackageManager.Utils;
 using Microsoft.Framework.Runtime;
 using Microsoft.Framework.Runtime.DependencyManagement;
@@ -43,6 +42,11 @@ namespace Microsoft.Framework.PackageManager
         public bool Lock { get; set; }
         public bool Unlock { get; set; }
         public string PackageFolder { get; set; }
+
+        /// <summary>
+        /// Gets or sets a flag that determines if restore is performed on multiple project.json files in parallel.
+        /// </summary>
+        public bool Parallel { get; set; }
 
         public ScriptExecutor ScriptExecutor { get; private set; }
 
@@ -106,7 +110,7 @@ namespace Microsoft.Framework.PackageManager
                     }
                 };
 
-                if (PlatformHelper.IsMono)
+                if (!Parallel || PlatformHelper.IsMono)
                 {
                     // Restoring in parallel on Mono throws native exception
                     foreach (var projectJsonFile in projectJsonFiles)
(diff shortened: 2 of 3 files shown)


pranavkm:
cc @muratg

BrennanConroy:
@ChengTian 
I will run this a few times on my box that was getting this issue.

BrennanConroy:
This didn't fix the issue, got the exact same error as before but instead of from Directory.Move it says from DirectoryInfo.MoveTo

pranavkm:
:frowning: yeah, guess we need to use plan B. I'll send an update.

pranavkm:
cc @davidfowl based on discussion.

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
```diff
@@ -105,8 +109,7 @@ public RestoreCommand(IApplicationEnvironment env)
                         Interlocked.Increment(ref successCount);
                     }
                 };
-
-                if (PlatformHelper.IsMono)
+                if (!RestoreInParallel || PlatformHelper.IsMono)
```
BrennanConroy: nit: space above if

BrennanConroy:
Does this actually fix it? Or is this just "security through obscurity"?

davidfowl:
This should be in the release branch and no it doesn't fix anything. It's the workaround for beta4. We'll have more time to investigate for beta5

src/Microsoft.Framework.PackageManager/Program.cs
```diff
@@ -61,6 +61,9 @@ public int Main(string[] args)
                 var optUnlock = c.Option("--unlock",
                     "Creates dependencies file with locked property set to false. Overwrites file if it exists.",
                     CommandOptionType.NoValue);
+                var restoreInParallel = c.Option("--parallel",
```
ChengTian: Should follow naming convention and make it `optParallel`

src/Microsoft.Framework.PackageManager/Program.cs
```diff
@@ -71,6 +74,7 @@ public int Main(string[] args)
                     command.FeedOptions = feedOptions;
                     command.Lock = optLock.HasValue();
                     command.Unlock = optUnlock.HasValue();
+                    command.RestoreInParallel = restoreInParallel.HasValue();
```
ChengTian: `RestoreCommand.RestoreInParallel`? Should we get rid of unnecessary redundancy and make it `RestoreCommand.Parallel`?

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
dougbu: is `PlatformHelper.IsMono` special case still needed?  `RestoreInParallel` defaults to `false`

BrennanConroy: On mono if you do --parallel it will not work, unless we just remove the option on mono

src/Microsoft.Framework.PackageManager/Utils/NuGetPackageUtils.cs
```diff
@@ -75,6 +88,11 @@ internal static class NuGetPackageUtils
                     stream.Seek(0, SeekOrigin.Begin);
                     var nupkgSHA = Convert.ToBase64String(sha512.ComputeHash(stream));
                     File.WriteAllText(hashPath, nupkgSHA);
+
+                    if (performingParallelInstalls)
+                    {
+                        extractDirectory.MoveTo(targetPath);
```
dougbu: `extractDirectory` is the `DirectoryInfo` for `targetPath`.  as-is this shouldn't do anything.

src/Microsoft.Framework.PackageManager/Utils/NuGetPackageUtils.cs
```diff
@@ -34,20 +37,30 @@ internal static class NuGetPackageUtils
...
+                    {
+                        // Extracting to the {id}/{version} has an issue with concurrent installs - when a package has been partially
+                        // extracted, the Restore Operation can inadvertly conclude the package is available locally and proceed to read
+                        // partially written package contents. To avoid this we'll extract the package to a sibling directory and Move it 
+                        // to the target path.
+                        extractPath = Path.Combine(Path.GetDirectoryName(targetPath), Path.GetRandomFileName());
+                        targetNupkg = Path.Combine(extractPath, Path.GetFileName(targetNupkg));
+                    }
+
+                    var extractDirectory = Directory.CreateDirectory(targetPath);
```
dougbu: suspect this and a couple of other `targetPath` references should be using `extractPath`.  right now the .nupkg is streamed into a file in the `extractPath` directory.  but files are extracted directly into the `targetPath` directory and the .nupkg is left around in `extractPath`.

src/Microsoft.Framework.PackageManager/Program.cs
pranavkm: Hungarian notation? :disappointed:

src/Microsoft.Framework.PackageManager/Program.cs
pranavkm: Sure

src/Microsoft.Framework.PackageManager/Program.cs
davidfowl: Look at all of the other variables. Follow the pattern bro :small_blue_diamond:

src/Microsoft.Framework.PackageManager/Utils/NuGetPackageUtils.cs
pranavkm: Oops, yeah. Missed that bit.

src/Microsoft.Framework.PackageManager/Program.cs
pranavkm: Oh sure I'm changing it. Doesn't mean it's not :disappointed:

pranavkm:
Updated.

ChengTian:
:shipit: after you test this change on MusicStore or similar project

Issue #1470, opened 2015-03-24
Fix casing of package IDs in lockfile


ChengTian:
https://github.com/aspnet/dnx/issues/581

Current `dnu restore` behavior is:
When restoring a package with wrong id casing, for example restoring package `XYZ.1.0.0` by having `"xyz": "1.0.0"` in `project.json`, it shows `Unable to resolve xyz.1.0.0, do you mean XYZ.1.0.0?` and doesn't install the package.
It prevents users from using some NuGet packages (see cases below).

New behavior should be:
Instead of exiting with error message above, we install the package to `.dnx/xyz/1.0.0` and make sure the casing of package name in lockfile is correct (`xyz`). This should fix the cases below.

**Basically, we always agree on the casing used by dependency graph.**

davidfowl:
```JSON
{
    "version": "1.0.0-*",
    "dependencies": {
        "NuGet.Core": "2.8.3"
    },
    "frameworks": {
        "dnx451": { },
        "dnxcore50": { }
    }
}
```

Results in

```
Information: [LoaderContainer]: Load name=Microsoft.Framework.PackageManager
Information: [PathBasedAssemblyLoader]: Loaded name=Microsoft.Framework.PackageManager in 61ms
Restoring packages for C:\Users\davifowl\documents\visual studio 14\Projects\WebApplication25\src\WebApplication1\project.json
  CACHE https://nuget.org/api/v2/FindPackagesById()?Id='NuGet.Core'
  CACHE https://www.myget.org/F/aspnetvolatile/api/v2/FindPackagesById()?Id='NuGet.Core'
  CACHE https://www.nuget.org/api/v2/package/Nuget.Core/2.8.3
  CACHE https://nuget.org/api/v2/FindPackagesById()?Id='Microsoft.Web.Xdt'
  CACHE https://www.myget.org/F/aspnetvolatile/api/v2/FindPackagesById()?Id='Microsoft.Web.Xdt'
Unable to locate NuGet.Core >= 2.8.3. Do you mean Nuget.Core?
Writing lock file C:\Users\davifowl\documents\visual studio 14\Projects\WebApplication25\src\WebApplication1\project.lock.json
Restore complete, 241ms elapsed
Errors in C:\Users\davifowl\documents\visual studio 14\Projects\WebApplication25\src\WebApplication1\project.json
    Unable to locate NuGet.Core >= 2.8.3. Do you mean Nuget.Core?
Loaded Module: dnx.clr.dll
[...]

PR #1511, opened 2015-03-27
Fix package name casing in lock file and installation process

ChengTian:
parent https://github.com/aspnet/dnx/issues/1470

We normalize all other package name casing (in lockfile & on disk) to the casing used by dependency graph (runtime)

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
@@ -389,19 +389,10 @@ namespace Microsoft.Framework.PackageManager
                     return;
                 }
 
-                // "dnu restore" is case-sensitive
                 if (!string.Equals(node.Item.Match.Library.Name, node.LibraryRange.Name, StringComparison.Ordinal))
                 {
-                    if (missingItems.Add(node.LibraryRange))
-                    {
-                        var errorMessage = string.Format("Unable to locate {0} {1}. Do you mean {2}?",
-                            node.LibraryRange.Name.Red().Bold(), node.LibraryRange.VersionRange, node.Item.Match.Library.Name.Bold());
-                        ErrorMessages.GetOrAdd(projectJsonPath, _ => new List<string>()).Add(errorMessage);
-                        Reports.Error.WriteLine(errorMessage);
-                        success = false;
-                    }
-
-                    return;
+                    // Fix casing of the library name to be installed
+                    node.Item.Match.Library.Name = node.LibraryRange.Name;
                 }
 
                 var isRemote = remoteProviders.Contains(node.Item.Match.Provider);
@@ -702,7 +693,8 @@ namespace Microsoft.Framework.PackageManager
                         package,
                         sha512,
                         frameworks,
-                        new DefaultPackagePathResolver(repository.RepositoryRoot));
+                        new DefaultPackagePathResolver(repository.RepositoryRoot),
+                        correctedPackageName: library.Name);
 
                     lockFile.Libraries.Add(lockFileLib);
                 }
--- a/src/Microsoft.Framework.PackageManager/Utils/LockFileUtils.cs
+++ b/src/Microsoft.Framework.PackageManager/Utils/LockFileUtils.cs
@@ -18,10 +18,15 @@ namespace Microsoft.Framework.PackageManager.Utils
             IPackage package,
             SHA512 sha512,
             IEnumerable<FrameworkName> frameworks,
-            IPackagePathResolver resolver)
+            IPackagePathResolver resolver,
+            string correctedPackageName = null)
         {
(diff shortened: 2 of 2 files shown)


ChengTian:
@davidfowl , I didn't get rid of the nuspec casing fix here: https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Framework.PackageManager/Utils/NuGetPackageUtils.cs#L54
because this fix is still necessary. We want to make sure `DefaultPackagePathResolver.GetManifestFilePath(string packageId, SemanticVersion version)` gives correct path on a case-sensitive filesystem.

davidfowl:
Nice! That simple huh? Do all of the test cases just work?

ChengTian:
Yes. Clean you packages folder, `dnu restore` and then `dnx . run` the following test app:
```c#
// Program.cs
using System;
namespace ConsoleApp2
{
    public class Program
    {
        public void Main(string[] args)
        {
            Console.WriteLine(typeof(NuGet.Constants).FullName);
            Console.WriteLine(typeof(NuGet.Indexing.DocumentKey).FullName);
            Console.WriteLine(typeof(Hangfire.SqlServer.SqlServerStorage).FullName);
            Console.WriteLine(typeof(Newtonsoft.Json.JsonReader).FullName);
            Console.ReadLine();
        }
    }
}
```
```json
{
    "version": "1.0.0-*",
    "dependencies": {
        "NuGet.Indexing": "3.0.75-r-master",
        "HangFire": "1.3.0",
        "newtonsoft.json": "6.0.8"
    },
    "frameworks": {
        "dnx451": { }
    }
}
```
This should cover all your test cases

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
```diff
@@ -389,19 +389,10 @@ public RestoreCommand(IApplicationEnvironment env)
...
-                        var errorMessage = string.Format("Unable to locate {0} {1}. Do you mean {2}?",
-                            node.LibraryRange.Name.Red().Bold(), node.LibraryRange.VersionRange, node.Item.Match.Library.Name.Bold());
-                        ErrorMessages.GetOrAdd(projectJsonPath, _ => new List<string>()).Add(errorMessage);
-                        Reports.Error.WriteLine(errorMessage);
-                        success = false;
-                    }
-
-                    return;
+                    // Fix casing of the library name to be installed
+                    node.Item.Match.Library.Name = node.LibraryRange.Name;
```
ChengTian: @davidfowl 
`node.Item.Match.Library.Name` is the package ID used by installation and lockfile. We correct its value to ID used in **runtime dependency graph**. So everything is agreeing on the ID used by runtime dependency graph.

davidfowl:
Why do I need to clean the packages folder?

ChengTian:
On case-insensitive filesystem of Windows/OSX, if you already have `NuGet.Core` installed, there is no way to install `Nuget.Core`. So even we fix the casing of package ID to install, it will fail to install the package with correct casing.

davidfowl:
Hmm so you still have to do nuspec rewriting for the package restore case? Or does this handle both cases?

ChengTian:
@davidfowl , my bad, wrong instructions. You don't need to clean your packages folder, no matter your filesystem is case-sensitive or case-insensitive. I just verified.

ChengTian:
We always do nuspec rewriting to make its casing same as the one used by runtime dependency graph. So `DefaultPackagePathResolver.GetManifestFilePath(string packageId, SemanticVersion version)` can always find the nuspec, no matter the filesystem is case-sensitive or not.

ChengTian:
I found a bug. Try this:
1. Clean your packages folder
2. `dnu restore` in the sample app above (then `dnx . run` should work)
3. Modify `project.json`: `NuGet.Indexing -> nuget.indexing`
4. `dnu restore` and then `dnx . run` fails

I am fixing this bug.

src/Microsoft.Framework.PackageManager/Utils/LockFileUtils.cs
```diff
@@ -18,10 +18,11 @@ internal static class LockFileUtils
             IPackage package,
             SHA512 sha512,
             IEnumerable<FrameworkName> frameworks,
-            IPackagePathResolver resolver)
+            IPackagePathResolver resolver,
+            string correctedPackageName = null)
         {
             var lockFileLib = new LockFileLibrary();
-            lockFileLib.Name = package.Id;
+            lockFileLib.Name = correctedPackageName ?? package.Id;
```
ChengTian: @davidfowl , I fixed the bug with this minimal change.
`package.Id` is read from nuspec of an installed package and its casing might be wrong. Here we use the package name from runtime dependency graph as corrected package name.

davidfowl:
:shipit:

Issue #1504, opened 2015-03-26
Add AssemblyFileVersion attribute to all assemblies


joeloff:
Assemblies do not currently contain an AssemblyFileVersion attribute. This attribute is used by Windows Installer to determine whether or not a file can be upgraded.

We should add something like  DNX_ASSEMBLY_FILE_VERSION

Any installer that consumes these will be broken when doing upgrades from one build to the next as files could be skipped.

pranavkm:
There's also the `AssemblyInformationalVersion` which might need something similar.

davidfowl:
We already infer that from the version in project.json

pranavkm:
Doesn't help if I want to use build time information to set it (for instance it would be nice to set the commit shas in the binaries we produce)

victorhurdugaci:
Here is a proposal that's not for the two PRs that I sent because they go in the release branch. It is for the dev branch.

I don't think that the Roslyn compiler should have this logic embedded in it.The file version and the commit hash should be compiler preprocess steps added by the app. But, instead of adding the preprocess step to each project, we could either add it to `global.json` (does that work today?) or have an environment variable that specifies extra folders to look for compiler process files. So, KoreBuild would included these preprocess steps and apps consume them.

What do you think?

PR #1529, opened 2015-03-28
Victorhu/fileversion

victorhurdugaci:
Inject the file version attribute in the generated assembly, only if the attribute is not already present.

Goes together with https://github.com/aspnet/Universe/pull/192 and fixes https://github.com/aspnet/dnx/issues/1504

Please review: @davidfowl, @anurse, @pranavkm, @ChengTian 
Please note this is for beta4/release branch

### DIFF
--- a/src/Microsoft.Framework.Runtime.Roslyn/RoslynCompiler.cs
+++ b/src/Microsoft.Framework.Runtime.Roslyn/RoslynCompiler.cs
@@ -213,15 +213,55 @@ namespace Microsoft.Framework.Runtime.Roslyn
         private static CSharpCompilation ApplyVersionInfo(CSharpCompilation compilation, Project project,
             CSharpParseOptions parseOptions)
         {
-            var emptyVersion = new Version(0, 0, 0, 0);
+            const string assemblyFileVersionName = "System.Reflection.AssemblyFileVersionAttribute";
+            const string assemblyVersionName = "System.Reflection.AssemblyVersion";
+            const string assemblyInformationalVersion = "System.Reflection.AssemblyInformationalVersion";
+            
+            var assemblyAttributes = compilation.Assembly.GetAttributes();
 
-            // If the assembly version is empty then set the version
-            if (compilation.Assembly.Identity.Version == emptyVersion)
+            var foundAssemblyFileVersion = false;
+            var foundAssemblyVersion = false;
+            var foundAssemblyInformationalVersion = false;
+
+            foreach (var assembly in assemblyAttributes)
             {
-                return compilation.AddSyntaxTrees(new[]
+                string attributeName = assembly.AttributeClass.ToString();
+
+                if (string.Equals(attributeName, assemblyFileVersionName, StringComparison.Ordinal))
                 {
-                    CSharpSyntaxTree.ParseText("[assembly: System.Reflection.AssemblyVersion(\"" + project.Version.Version + "\")]", parseOptions),
-                    CSharpSyntaxTree.ParseText("[assembly: System.Reflection.AssemblyInformationalVersion(\"" + project.Version + "\")]", parseOptions)
+                    foundAssemblyFileVersion = true;
+                }
+                else if (string.Equals(attributeName, assemblyVersionName, StringComparison.Ordinal))
+                {
+                    foundAssemblyVersion = true;
+                }
+                else if (string.Equals(attributeName, assemblyInformationalVersion, StringComparison.Ordinal))
+                {
+                    foundAssemblyInformationalVersion = true;
+                }
+            }
+
+            var versionAttributes = new StringBuilder();
+            if (!foundAssemblyFileVersion)
+            {
+                versionAttributes.AppendLine($"[assembly:{assemblyFileVersionName}(\"{project.AssemblyFileVersion}\")]");
+            }
+
+            if (!foundAssemblyVersion)
+            {
+                versionAttributes.AppendLine($"[assembly:{assemblyVersionName}(\"{project.Version.Version}\")]");
+            }
+
+            if (!foundAssemblyInformationalVersion)
+            {
+                versionAttributes.AppendLine($"[assembly:{assemblyInformationalVersion}(\"{project.Version}\")]");
+            }
+
+            if (versionAttributes.Length != 0)
+            {
+                compilation = compilation.AddSyntaxTrees(new[]
+                {
+                    CSharpSyntaxTree.ParseText(versionAttributes.ToString(), parseOptions)
                 });
             }
 
--- a/src/Microsoft.Framework.Runtime/Project.cs
+++ b/src/Microsoft.Framework.Runtime/Project.cs
@@ -62,6 +62,8 @@ namespace Microsoft.Framework.Runtime
 
         public SemanticVersion Version { get; private set; }
 
+        public Version AssemblyFileVersion { get; private set; }
+
         public IList<LibraryDependency> Dependencies { get; private set; }
 
         public LanguageServices LanguageServices { get; private set; }
@@ -188,6 +190,23 @@ namespace Microsoft.Framework.Runtime
                 }
             }
 
+            var fileVersion = Environment.GetEnvironmentVariable("DNX_ASSEMBLY_FILE_VERSION");
+            if (string.IsNullOrWhiteSpace(fileVersion))
+            {
+                project.AssemblyFileVersion = project.Version.Version;
+            }
+            else
+            {
+                try
+                {
+                    project.AssemblyFileVersion = SpecifyFileVersionSnapshot(version?.Value<string>(), fileVersion);
+                }
+                catch (FormatException ex)
+                {
+                    throw new FormatException("The assembly file version is invalid: " + fileVersion, ex);
+                }
+            }
+
             project.Description = rawProject.GetValue<string>("description");
             project.Summary = rawProject.GetValue<string>("summary");
             project.Copyright = rawProject.GetValue<string>("copyright");
@@ -282,6 +301,20 @@ namespace Microsoft.Framework.Runtime
             }
 
             return new SemanticVersion(version);
+        }
+
+        private static Version SpecifyFileVersionSnapshot(string version, string fileSnapshotValue)
+        {
+            if (string.IsNullOrEmpty(version))
+            {
+                version = "1.0.0." + fileSnapshotValue;
+            }
+            else if (version.EndsWith("-*"))
+            {
+                version = version.Substring(0, version.Length - 2) + "." + fileSnapshotValue;
+            }
+
+            return new Version(version);
         }
 
         private static void PopulateDependencies(


src/Microsoft.Framework.Runtime.Roslyn/RoslynCompiler.cs
```diff
@@ -214,17 +214,23 @@ private CompilationModules GetCompileModules(ILibraryKey target)
             CSharpParseOptions parseOptions)
         {
             var emptyVersion = new Version(0, 0, 0, 0);
+            var assemblyAttributes = compilation.Assembly.GetAttributes();
 
-            // If the assembly version is empty then set the version
-            if (compilation.Assembly.Identity.Version == emptyVersion)
+            var addAttributeIfNotExists = new Action<string, string>((attributeName, attributeValue) =>
```
pranavkm: Why not make this a regular method?

src/Microsoft.Framework.Runtime/Project.cs
```diff
@@ -188,6 +190,13 @@ internal static Project GetProjectFromStream(Stream stream, string projectName,
                 }
             }
 
+            var fileVersion = Environment.GetEnvironmentVariable("DNX_ASSEMBLY_FILE_VERSION");
```
pranavkm: No love for `DNX_ASSEMBLY_INFORMATIONAL_VERSION`?

src/Microsoft.Framework.Runtime.Roslyn/RoslynCompiler.cs
victorhurdugaci: Less arguments :)

src/Microsoft.Framework.Runtime/Project.cs
```diff
@@ -62,6 +62,8 @@ public string ProjectDirectory
 
         public SemanticVersion Version { get; private set; }
 
+        public string AssemblyFileVersion { get; private set; }
```
pranavkm: Should this be of type `Version`?

src/Microsoft.Framework.Runtime.Roslyn/RoslynCompiler.cs
```diff
@@ -214,17 +214,23 @@ private CompilationModules GetCompileModules(ILibraryKey target)
             CSharpParseOptions parseOptions)
         {
             var emptyVersion = new Version(0, 0, 0, 0);
+            var assemblyAttributes = compilation.Assembly.GetAttributes();
 
-            // If the assembly version is empty then set the version
-            if (compilation.Assembly.Identity.Version == emptyVersion)
```
pranavkm: Did this check get removed?

victorhurdugaci: Yes. Because we already check the attributes

src/Microsoft.Framework.Runtime/Project.cs
victorhurdugaci: Sure, why not

src/Microsoft.Framework.Runtime/Project.cs
victorhurdugaci: Nope, no love for it :) That value is automatically computed

victorhurdugaci: That value is taken from the `DNX_BUILD_VERSION` variable

victorhurdugaci:
@pranavkm updated the PR. please take a look

src/Microsoft.Framework.Runtime/Project.cs
```diff
@@ -188,6 +190,21 @@ internal static Project GetProjectFromStream(Stream stream, string projectName,
...
+                fileVersion = project.Version.ToString();
+            }
+
+            try
+            {
+                project.AssemblyFileVersion = new Version(fileVersion);
+            }
+            catch(FormatException ex)
+            {
+                throw new FormatException("The assembly file version is invalid: " + fileVersion, ex);
```
davidfowl: Nah. Just use Version.Parse and let that throw.

src/Microsoft.Framework.Runtime/Project.cs
```diff
@@ -188,6 +190,21 @@ internal static Project GetProjectFromStream(Stream stream, string projectName,
                 }
             }
 
+            var fileVersion = Environment.GetEnvironmentVariable("DNX_ASSEMBLY_FILE_VERSION");
+            if (string.IsNullOrWhiteSpace(fileVersion))
+            {
+                fileVersion = project.Version.ToString();
```
davidfowl: Why to ToString and new Verison right underneath? If you're defaulting to the project.Version then just assign it directly.

src/Microsoft.Framework.Runtime.Roslyn/RoslynCompiler.cs
```diff
@@ -213,18 +213,59 @@ private CompilationModules GetCompileModules(ILibraryKey target)
         private static CSharpCompilation ApplyVersionInfo(CSharpCompilation compilation, Project project,
             CSharpParseOptions parseOptions)
         {
+            const string assemblyFileVersionName = "System.Reflection.AssemblyFileVersionAttribute";
+            const string assemblyVersionName = "System.Reflection.AssemblyVersion";
+            const string assemblyInformationalVersion = "System.Reflection.AssemblyInformationalVersion";
+
             var emptyVersion = new Version(0, 0, 0, 0);
```
davidfowl: You should be able to remove this

src/Microsoft.Framework.Runtime.Roslyn/RoslynCompiler.cs
```diff
@@ -213,18 +213,59 @@ private CompilationModules GetCompileModules(ILibraryKey target)
...
+            var assemblyAttributes = compilation.Assembly.GetAttributes();
+
+            bool foundAssemblyFileVersion = false;
+            bool foundAssemblyVersion = false;
+            bool foundAssemblyInformationalVersion = false;
 
-            // If the assembly version is empty then set the version
-            if (compilation.Assembly.Identity.Version == emptyVersion)
+            foreach(var assembly in assemblyAttributes)
```
davidfowl: spacing

src/Microsoft.Framework.Runtime.Roslyn/RoslynCompiler.cs
```diff
@@ -213,18 +213,59 @@ private CompilationModules GetCompileModules(ILibraryKey target)
...
             {
-                return compilation.AddSyntaxTrees(new[]
+                string attributeName = assembly.AttributeClass.ToString();
+
+                if (attributeName == assemblyFileVersionName)
                 {
-                    CSharpSyntaxTree.ParseText("[assembly: System.Reflection.AssemblyVersion(\"" + project.Version.Version + "\")]", parseOptions),
-                    CSharpSyntaxTree.ParseText("[assembly: System.Reflection.AssemblyInformationalVersion(\"" + project.Version + "\")]", parseOptions)
+                    foundAssemblyFileVersion = true;
```
davidfowl: Why not just add the syntax trees inline instead of having a bool? You can't have more than one of those attributes anyways.

src/Microsoft.Framework.Runtime.Roslyn/RoslynCompiler.cs
Eilon: Also, a more typical empty version is to call the parameterless ctor (so that it has no build/revision in it, as opposed to explicitly 0).

src/Microsoft.Framework.Runtime.Roslyn/RoslynCompiler.cs
```diff
@@ -213,18 +213,59 @@ private CompilationModules GetCompileModules(ILibraryKey target)
...
             CSharpParseOptions parseOptions)
         {
+            const string assemblyFileVersionName = "System.Reflection.AssemblyFileVersionAttribute";
+            const string assemblyVersionName = "System.Reflection.AssemblyVersion";
+            const string assemblyInformationalVersion = "System.Reflection.AssemblyInformationalVersion";
+
             var emptyVersion = new Version(0, 0, 0, 0);
+            var assemblyAttributes = compilation.Assembly.GetAttributes();
+
+            bool foundAssemblyFileVersion = false;
```
pranavkm: `var`

src/Microsoft.Framework.Runtime/Project.cs
pranavkm: Seems like a simple enough change - but we could do it as a separate PR.

src/Microsoft.Framework.Runtime.Roslyn/RoslynCompiler.cs
```diff
@@ -213,18 +213,59 @@ private CompilationModules GetCompileModules(ILibraryKey target)
...
+                {
+                    foundAssemblyInformationalVersion = true;
+                }
+            }
+
+            if (!foundAssemblyFileVersion)
+            {
+                compilation = compilation.AddSyntaxTrees(new[]
+                {
+                    CSharpSyntaxTree.ParseText("[assembly: " +  assemblyFileVersionName + "(\"" + project.AssemblyFileVersion + "\")]", parseOptions)
```
pranavkm: Embrace the C# 6.0 Luke.
`CSharpSyntaxTree.ParseText($"[assembly: {assemblyFileVersionName}(\"${project.AssemblyFileVersion}\")]", parseOptions)`

src/Microsoft.Framework.Runtime.Roslyn/RoslynCompiler.cs
```diff
@@ -213,18 +213,59 @@ private CompilationModules GetCompileModules(ILibraryKey target)
...
+            bool foundAssemblyInformationalVersion = false;
 
-            // If the assembly version is empty then set the version
-            if (compilation.Assembly.Identity.Version == emptyVersion)
+            foreach(var assembly in assemblyAttributes)
             {
-                return compilation.AddSyntaxTrees(new[]
+                string attributeName = assembly.AttributeClass.ToString();
+
+                if (attributeName == assemblyFileVersionName)
```
pranavkm: Maybe do `string.Equals(attributeName, assemblyFileVersionName, StringComparison.Ordinal)` here. Also is the value guaranteed to be an fully qualified name all the time? Even if the actual code didn't fully qualify it?

```C#
using System.Reflection;
[assembly: AssemblyVersion("1.0.0")]
```

victorhurdugaci: > Maybe do  ... 

Ok

> Also is the value guaranteed to be an fully qualified name all the time? Even if the actual code didn't fully qualify it?

Good question and the answer is yes, it is always the full qualified name.

src/Microsoft.Framework.Runtime/Project.cs
victorhurdugaci: We can change that in the `dev` branch

src/Microsoft.Framework.Runtime/Project.cs
victorhurdugaci: Nah. The error message is really bad. It just says that the string has an invalid format but doesn't tell you what the value is.

davidfowl: Then at least fix the formatting in the file :smile:

victorhurdugaci:
Updated

src/Microsoft.Framework.Runtime/Project.cs
```diff
@@ -188,6 +190,23 @@ internal static Project GetProjectFromStream(Stream stream, string projectName,
...
+            var fileVersion = Environment.GetEnvironmentVariable("DNX_ASSEMBLY_FILE_VERSION");
+            if (string.IsNullOrWhiteSpace(fileVersion))
+            {
+                project.AssemblyFileVersion = project.Version.Version;
+            }
+            else
+            {
+                try
+                {
+                    project.AssemblyFileVersion = new Version(fileVersion);
```
pranavkm: Maybe `TryParse` instead of try-catching?

pranavkm:
:shipit: once somebody from DNX signs off.

src/Microsoft.Framework.Runtime/Project.cs
```diff
@@ -188,6 +190,23 @@ internal static Project GetProjectFromStream(Stream stream, string projectName,
...
+                }
+                catch (FormatException ex)
+                {
+                    throw new FormatException("The assembly file version is invalid: " + fileVersion, ex);
```
muratg: Consider using String.Format instead of concat.

muratg:
/cc: @joeloff 

:shipit:

src/Microsoft.Framework.Runtime/Project.cs
moozzyk: Doesn't the exception message have to be localizable?

victorhurdugaci: No strings in the runtime are localizable. I think there is an issue about that somewhere...

victorhurdugaci:
Updated the PR with one change: the prefix for the version is picked up from the `project.json` file.

PR #1537, opened 2015-03-29
Revamp how diagnostics are handled to work more consistently

davidfowl:
- Added GetDependencyDiagnostics to DependencyWalker in order to
get warnings and errors about the resolved dependency graph.
- Removed ProjectWarningsMessage and used DiagnosticsMessage instead.
- Surface dependency diagnostics at build time when using dnu build
- Surface dependency diagnostics in the DTH via a new message type and
as part of AllDiagnostics.
- Added Line number information to the LibraryRange (when it makes sense)
- Split DependencyDiagnostics and CompilationDiagnostics in DTH


src/Microsoft.Framework.Runtime/DependencyManagement/DependencyWalker.cs
```diff
@@ -57,6 +58,45 @@ public void Walk(string name, SemanticVersion version, FrameworkName targetFrame
...
+                    {
+                        StartLine = library.LibraryRange.Line,
+                        StartColumn = library.LibraryRange.Column
+                    });
+                }
+                else
+                {
+                    // If we ended up with a version that isn't the minimum and we're not floating
+                    // add a warning. e.g. "A": "1.0" resulting in "A": "4.0"
+                    if (library.LibraryRange.VersionRange != null &&
```
davidfowl: This logic can be removed. It was an example of other diagnostics we can add.

PR #1566, opened 2015-04-01
Steps towards build verify working on *nix

davidfowl:
- Fixed the makefile so that it works on mono
- Fixed a bunch of tests (paths etc)
- Replaced System.IO.Compression extract calls with unzip because
of flakyness on mono
- Still need to fix globbing tests


test/Microsoft.Framework.FunctionalTestUtils/TestUtils.cs
```diff
@@ -65,21 +78,59 @@ public static DisposableDir GetRuntimeHomeDir(string flavor, string os, string a
...
+
+            if (!PlatformHelper.IsMono)
+            {
+                System.IO.Compression.ZipFile.ExtractToDirectory(runtimeNupkg, runtimeRoot);
+            }
+            else
+            {
+                Directory.CreateDirectory(runtimeRoot);
+                Exec("unzip", runtimeNupkg + " -d " + runtimeRoot);
+                // We need to mark these files because unzip doesn't preserve the exec bit
```
Tragetaschen: Is the exec bit in there in the first place?

davidfowl: I marked them in this change

PR #1568, opened 2015-04-01
Use absolute path for lock file

BrennanConroy:
File.Exists was using a relative path

@victorhurdugaci

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Install/UninstallCommand.cs
+++ b/src/Microsoft.Framework.PackageManager/Install/UninstallCommand.cs
@@ -12,12 +12,10 @@ namespace Microsoft.Framework.PackageManager
     internal class UninstallCommand
     {
         private readonly IAppCommandsRepository _commandsRepo;
-        private readonly IApplicationEnvironment _environment;
         private readonly Reports _reports;
 
-        public UninstallCommand(IApplicationEnvironment applicationEnvironment, IAppCommandsRepository commandsRepo, Reports reports)
+        public UninstallCommand(IAppCommandsRepository commandsRepo, Reports reports)
         {
-            _environment = applicationEnvironment;
             _commandsRepo = commandsRepo;
             _reports = reports;
         }
@@ -78,6 +76,7 @@ namespace Microsoft.Framework.PackageManager
             foreach(var applicationPackage in applicationPackagesStillUsed)
             {
                 var appLockFileFullPath = Path.Combine(
+                    _commandsRepo.PackagesRoot.Root,
                     _commandsRepo.PathResolver.GetPackageDirectory(applicationPackage.Id, applicationPackage.Version),
                     InstallBuilder.CommandsFolderName,
                     LockFileFormat.LockFileName);
--- a/test/Microsoft.Framework.PackageManager.FunctionalTests/DnuCommandsTests.cs
+++ b/test/Microsoft.Framework.PackageManager.FunctionalTests/DnuCommandsTests.cs
@@ -1,0 +1,112 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.Framework.CommonTestUtils;
+using Microsoft.Framework.Runtime.DependencyManagement;
+using Xunit;
+
+namespace Microsoft.Framework.PackageManager
+{
+    public class DnuCommandsTests
+    {
+        public static IEnumerable<object[]> RuntimeComponents
+        {
+            get { return TestUtils.GetRuntimeComponentsCombinations(); }
+        }
+
+        [Theory]
+        [MemberData(nameof(RuntimeComponents))]
+        public void DnuCommands_Uninstall_PreservesPackagesUsedByOtherInstalledApps(string flavor, string os, string architecture)
+        {
+            var runtimeHomeDir = TestUtils.GetRuntimeHomeDir(flavor, os, architecture);
+            using (var testEnv = new DnuTestEnvironment(runtimeHomeDir))
+            {
+                var directory = Path.Combine(testEnv.RootDir, Runtime.Constants.DefaultLocalRuntimeHomeDir, "bin");
+                InstallFakeApp(directory, "pack1", "0.0.0");
+                InstallFakeApp(directory, "pack2", "0.0.0");
+                WriteLockFile($"{directory}/packages/pack2/0.0.0/app", "pack2", "0.0.0");
+
+                var environment = new Dictionary<string, string> { { "USERPROFILE", testEnv.RootDir } };
+                DnuTestUtils.ExecDnu(runtimeHomeDir, "commands", "uninstall pack1", environment);
+
+                // Pack2 is in use by the pack2 app so should not be removed
+                Assert.True(Directory.Exists($"{directory}/packages/pack2"));
+                // Pack1 only used by pack1 app so should be removed
+                Assert.False(Directory.Exists($"{directory}/packages/pack1"));
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(RuntimeComponents))]
+        public void DnuCommands_Uninstall_RemovesUnusedPackageNotUsedByApp(string flavor, string os, string architecture)
+        {
+            var runtimeHomeDir = TestUtils.GetRuntimeHomeDir(flavor, os, architecture);
+            using (var testEnv = new DnuTestEnvironment(runtimeHomeDir))
+            {
+                var directory = Path.Combine(testEnv.RootDir, Runtime.Constants.DefaultLocalRuntimeHomeDir, "bin");
+                InstallFakeApp(directory, "pack1", "0.0.0");
+                Directory.CreateDirectory($"{directory}/packages/pack2/0.0.0/");
+                WriteLockFile($"{directory}/packages/pack1/0.0.0/app", "pack1", "0.0.0");
+
+                var environment = new Dictionary<string, string> { { "USERPROFILE", testEnv.RootDir } };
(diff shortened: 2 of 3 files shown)


src/Microsoft.Framework.PackageManager/Install/UninstallCommand.cs
```diff
@@ -78,6 +78,7 @@ private void PurgeOrphanPackages()
             foreach(var applicationPackage in applicationPackagesStillUsed)
             {
                 var appLockFileFullPath = Path.Combine(
+                    packagesRepo.RepositoryRoot.Root,
```
victorhurdugaci: `_commandsRepo.PackagesRoot` I think has the same value and might be a better choice

davidfowl:
No test?

BrennanConroy:
I'd like to add a test, but either need to make UninstallCommand a public class or make internals visible to the test project

davidfowl:
Without a test I have no idea what this actually fixes

BrennanConroy:
So uninstall would try to look at the lock file of the app command it was uninstalling to look at dependencies to figure out which packages to remove etc. but it was using a relative path to look for the lock file and when it doesn't find it, it just removes all the packages for all commands, so any other commands you have will not work anymore.

Now, should I make the UninstallCommand class public or add the InternalsVisibleTo attribute?

davidfowl:
InternalsVisibleTo

BrennanConroy:
Added test, and yes the test fails if I remove the fix and passes if I put the fix back.

BrennanConroy:
@victorhurdugaci @davidfowl 
Ping

src/Microsoft.Framework.PackageManager/Properties/AssemblyInfo.cs
```diff
@@ -4,5 +4,6 @@
 using System.Reflection;
 using System.Runtime.CompilerServices;
 
+[assembly: InternalsVisibleTo("Microsoft.Framework.PackageManager.FunctionalTests")]
```
victorhurdugaci: Nope. We don't want this. Internal visible to is only for unit tests

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuCommandsTests.cs
```diff
@@ -0,0 +1,80 @@
...
+            File.WriteAllText($"{rootDir}/packages/{name}/{version}/{name}.{version}.nupkg.sha512", "TestSha");
+        }
+
+        private void WriteLockFile(string dir, string libName, string version)
+        {
+            var lockFile = new LockFile
+            {
+                Islocked = false,
+                Libraries = new List<LockFileLibrary>
+                                {
```
victorhurdugaci: Indentation

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuCommandsTests.cs
```diff
@@ -0,0 +1,80 @@
...
+                    Error = new NullReport(),
+                    Information = new NullReport(),
+                    Quiet = new NullReport(),
+                    Verbose = new NullReport()
+                };
+                var command = new UninstallCommand(repo, reports);
+
+                command.Execute("pack1");
+
+                //Pack2 is inuse by the pack2 app so should not be removed
```
victorhurdugaci: Nit: "inuse" -> "in use"

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuCommandsTests.cs
```diff
@@ -0,0 +1,80 @@
...
+                                    }
+                                }
+            };
+            var lockFormat = new LockFileFormat();
+            lockFormat.Write($"{dir}/project.lock.json", lockFile);
+        }
+
+        [Theory]
+        [MemberData("RuntimeComponents")]
+        public void DnuCommands_Uninstall_PreservesPackagesUsedByOtherInstalledApps(string flavor, string os, string architecture)
```
victorhurdugaci: Add 2 more test that does:
First:
1. Add a package that is not used by anyone
2. Have a single application installed (that doesn't use that packge)
3. Uninstall the app
4. Make sure there is no package left

Second:
Exactly like the previous test but pass `--no-purge` and make sure that the orphan package is not deleted

src/Microsoft.Framework.PackageManager/Properties/AssemblyInfo.cs
BrennanConroy: I could change this to unit tests, since it is a weird mix between unit and functional, but we need internals visible otherwise the only way to test this is to compromise the users current installs

victorhurdugaci: It is a tricky situation because unit tests shouldn't touch the disk...

BrennanConroy:
@victorhurdugaci 
Updated
The travis failure is because of a change in Universe that was fixed in: https://github.com/aspnet/Universe/commit/789544a1e80c7ef0e011b1bf9b8a3ec9fd501939

BrennanConroy:
@troydai

src/Microsoft.Framework.PackageManager/Install/AppCommandsFolderRepository.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace Microsoft.Framework.PackageManager
 {
-    internal class AppCommandsFolderRepository : IAppCommandsRepository
+    public class AppCommandsFolderRepository : IAppCommandsRepository
```
troydai: Why does it have to be _public_? If for test, how about use `InternalsVisibleTo` attribute?
https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Framework.Runtime/Properties/AssemblyInfo.cs#L7

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuCommandsTests.cs
```diff
@@ -0,0 +1,145 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
```
troydai: Update the license

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuCommandsTests.cs
```diff
@@ -0,0 +1,145 @@
...
+namespace Microsoft.Framework.PackageManager
+{
+    public class DnuCommandsTests
+    {
+        public static IEnumerable<object[]> RuntimeComponents
+        {
+            get
+            {
+                return TestUtils.GetRuntimeComponentsCombinations();
+            }
```
troydai: Why not put them on one line to save more lines of code to show up in on your screen?

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuCommandsTests.cs
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        private void InstallFakeApp(string rootDir, string name, string version)
```
troydai: Put the _private method_ at the end?

troydai:
:watch:

src/Microsoft.Framework.PackageManager/Install/AppCommandsFolderRepository.cs
BrennanConroy: Because we don't use InternalsVisibleTo on FunctionalTests

troydai: Then can you mock one?

BrennanConroy: No? I want the functionality to stay the same

BrennanConroy:
Ping @troydai @victorhurdugaci

troydai:
Still concern about filpping the `internal` to `public` just for the sake of tests.

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuCommandsTests.cs
```diff
@@ -0,0 +1,125 @@
...
+        {
+            var runtimeHomeDir = TestUtils.GetRuntimeHomeDir(flavor, os, architecture);
+            using (var testEnv = new DnuTestEnvironment(runtimeHomeDir))
+            {
+                var directory = Path.Combine(testEnv.RootDir, Runtime.Constants.DefaultLocalRuntimeHomeDir, "bin");
+                InstallFakeApp(directory, "pack1", "0.0.0");
+                InstallFakeApp(directory, "pack2", "0.0.0");
+                WriteLockFile($"{directory}/packages/pack2/0.0.0/app", "pack2", "0.0.0");
+
+                //var repo = new AppCommandsFolderRepository(testEnv.RootDir);
```
troydai: Remove the commented code.

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuCommandsTests.cs
```diff
@@ -0,0 +1,125 @@
...
+            {
+                var directory = Path.Combine(testEnv.RootDir, Runtime.Constants.DefaultLocalRuntimeHomeDir, "bin");
+                InstallFakeApp(directory, "pack1", "0.0.0");
+                Directory.CreateDirectory($"{directory}/packages/pack2/0.0.0/");
+                WriteLockFile($"{directory}/packages/pack1/0.0.0/app", "pack1", "0.0.0");
+
+                var environment = new Dictionary<string, string> { { "USERPROFILE", testEnv.RootDir } };
+                DnuTestUtils.ExecDnu(runtimeHomeDir, "commands", "uninstall pack1", environment);
+
+                //Pack1 only used by pack1 app so should be removed
```
troydai: Add space after `//`. Same for the four occurrences below.

troydai:
:shipit: after fix the comments.

PR #1574, opened 2015-04-02
Quick strong-naming in Roslyn

anurse:
1. Only works when running `dnu build/pack` on desktop (you **can** build **for** CoreCLR, just not **on** CoreCLR)
1. No project.json support
1. Enabled via the presence of the `DNX_BUILD_KEY_FILE` environment variable
1. Set `DNX_BUILD_DELAY_SIGN` to `true` to do delay signing instead of full signing

/cc @emgarten @davidfowl

### DIFF
--- a/src/Microsoft.Framework.Runtime.Roslyn/RoslynCompiler.cs
+++ b/src/Microsoft.Framework.Runtime.Roslyn/RoslynCompiler.cs
@@ -122,6 +122,9 @@ namespace Microsoft.Framework.Runtime.Roslyn
                 references,
                 compilationSettings.CompilationOptions);
 
+            // Apply strong-name settings
+            compilation = ApplyStrongNameSettings(compilation);
+
             compilation = ApplyVersionInfo(compilation, project, parseOptions);
 
             var compilationContext = new CompilationContext(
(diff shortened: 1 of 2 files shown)


davidfowl:
:shipit: assuming you tested it on NuGet.Packaging?

anurse:
aye, and it strong named all the things

anurse:
I'll merge it after the fancy new PR CI builds pass

anurse:
Not sure why the travis build failed... seems to be related to Mono build issues in globbing tests. AppVeyor seems to be stuck in a long queue so if my local build verify works I'm gonna yolo push it and babysit the internal CI :)

src/Microsoft.Framework.Runtime.Roslyn/RoslynCompiler.cs
```diff
@@ -186,6 +189,29 @@ public class RoslynCompiler
...
+                    string.Equals(delaySignString, "1", StringComparison.OrdinalIgnoreCase));
+
+                var strongNameProvider = new DesktopStrongNameProvider();
+                var newOptions = compilation.Options
+                    .WithStrongNameProvider(strongNameProvider)
+                    .WithCryptoKeyFile(keyFile)
+                    .WithDelaySign(!string.IsNullOrEmpty(delaySignString) && string.Equals("true", delaySignString, StringComparison.OrdinalIgnoreCase));
+                return compilation.WithOptions(newOptions);
+            }
+#endif
```
Eilon: `#elseif throw <something>` or add a diagnostic or something? Or maybe put the `#ifdef` inside the main `if()`? I could imagine debugging this would be painful otherwise...

nil4: A few lines above the `delaySign` variable is computed, allowing either `true` or `1` as values to enable delay signing. Why not reuse that variable in the call to `WithDelaySign` instead of recomputing from `delaySignString` and only checking for a `true` value? I assume this is an oversight, but it's inconsistent and unnecessary.

anurse: It's not just inconsistent, it's incorrect, since the delaySign value computed above is completely unused! Going to have to see why the compiler warning didn't fire for that. Anyway, cleaning this and @Eilon 's comment in an upcoming PR. This was a quick hack-job to unblock other work :).

Eilon: @anurse sounds good to me.

PR #1582, opened 2015-04-03
Update PatternsCollectionHelper and corresponding tests

troydai:
1. Return normalized patterns even for default values;
2. Update tests


davidfowl:
:shipit:

PR #1596, opened 2015-04-06
Warn user when attempting to strong-name on CoreCLR

anurse:
We just silently fail to strong-name when the environment variable is present on CoreCLR. We can't strong-name on CoreCLR (yet) but we should warn the user of this.

capture

### DIFF
--- a/src/Microsoft.Framework.Runtime.Roslyn/RoslynDiagnostics.cs
+++ b/src/Microsoft.Framework.Runtime.Roslyn/RoslynDiagnostics.cs
@@ -1,0 +1,15 @@
+﻿using Microsoft.CodeAnalysis;
+
+namespace Microsoft.Framework.Runtime.Roslyn
+{
+    internal class RoslynDiagnostics
+    {
+        internal static readonly DiagnosticDescriptor StrongNamingNotSupported = new DiagnosticDescriptor(
+            id: "DNX1001",
+            title: "Strong name generation is not supported on this platform",
+            messageFormat: "Strong name generation is not supported on CoreCLR. Skipping strong name generation.",
+            category: "StrongNaming",
+            defaultSeverity: DiagnosticSeverity.Warning,
+            isEnabledByDefault: true);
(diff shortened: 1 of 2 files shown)


anurse:
Not sure why travis failed... it seems very unrelated. Trying again

src/Microsoft.Framework.Runtime.Roslyn/RoslynDiagnostics.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Microsoft.CodeAnalysis;
+
+namespace Microsoft.Framework.Runtime.Roslyn
+{
+    internal class RoslynDiagnostics
+    {
+        internal static readonly DiagnosticDescriptor StrongNamingNotSupported = new DiagnosticDescriptor(
```
davidfowl: It actually is supported

anurse: I'll check in to that... last I checked (this morning) we had to implement `StrongNameProvider` and something wasn't working, but maybe I missed something.

anurse: Yeah, looks like https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/Portable/StrongNameProvider.cs still has some `internal abstract` properties :(.

Eilon:
While you guys debate whether it's supported or not, the code itself looks good to me :smile:

src/Microsoft.Framework.Runtime.Roslyn/RoslynDiagnostics.cs
anurse: The are indeed still `internal abstract` in Roslyn. I'm leave this as-is until we get an implementable strong name system. Not using a `StrongNameProvider` yields "CS7027: Error signing ... Assembly signing not supported."

PR #1599, opened 2015-04-06
* Use the presence of hash file to denote completion of package install

pranavkm:
operation removing the need for Directory.Move.

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
@@ -13,16 +13,18 @@ using System.Security.Cryptography;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Framework.PackageManager.Publish;
+using Microsoft.Framework.PackageManager.Restore.RuntimeModel;
 using Microsoft.Framework.PackageManager.Utils;
 using Microsoft.Framework.Runtime;
 using Microsoft.Framework.Runtime.DependencyManagement;
 using NuGet;
-using Microsoft.Framework.PackageManager.Restore.RuntimeModel;
 
 namespace Microsoft.Framework.PackageManager
 {
     public class RestoreCommand
     {
+        private static readonly int MaxDegreesOfConcurrency = Environment.ProcessorCount;
+
         public RestoreCommand(IApplicationEnvironment env)
         {
             ApplicationEnvironment = env;
@@ -110,7 +112,7 @@ namespace Microsoft.Framework.PackageManager
                     }
                 };
 
-                if (!Parallel || PlatformHelper.IsMono)
+                if (!RestoringInParallel())
                 {
                     // Restoring in parallel on Mono throws native exception
                     foreach (var projectJsonFile in projectJsonFiles)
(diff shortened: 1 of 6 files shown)


pranavkm:
cc @davidfowl @ChengTian 

@BrennanConroy, I tried this on your VM and it does look like it's resolved correctly this time around.

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
```diff
@@ -414,7 +416,7 @@ public RestoreCommand(IApplicationEnvironment env)
                 }
             });
 
-            await InstallPackages(installItems, packagesDirectory, packageFilter: (library, nupkgSHA) => true);
```
pranavkm: Unless I'm missing something super obvious, `packageFilter` never did anything particularly useful.

ChengTian:
I can help to verify this change on my super slow Win7 VM. Will post feedback here after that.

ChengTian:
Being blocked by an issue on Win7: https://github.com/aspnet/dnx/issues/1616
I cannot do the verification temporarily.

ChengTian:
Wrote a script and ran 1000+ times of `dnu restore --parallel` for `MusicStore` on a very slow Win7 VM. Couldn't reproduce the bug. Looks like it does fix the bug.

pranavkm:
Great. Could you review the PR so we can check it in and start using it as part of the CI build script?

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
ChengTian: Yeah, I also removed this part during `dnu refactoring`

davidfowl:
@pranavkm Why is the build is failing?

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
```diff
@@ -635,30 +637,33 @@ internal bool IsBestVersion(string name, SemanticVersion version)
...
+            using (var memoryStream = new MemoryStream())
+            {
+                var match = installItem.Match;
+                await match.Provider.CopyToAsync(installItem.Match, memoryStream);
 
-                    Reports.Information.WriteLine("Installing {0} {1}", library.Name.Bold(), library.Version);
-                    memStream.Seek(0, SeekOrigin.Begin);
-                    await NuGetPackageUtils.InstallFromStream(memStream, library, packagesDirectory, sha512);
-                }
+                memoryStream.Position = 0;
```
ChengTian: Why do we replace `memStream.Seek(0, SeekOrigin.Begin);` with this?

pranavkm:
Travis seems to hung due to package restore - `No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.` AppVeyor doesn't have any build logs. Let me try rebasing and pushing to this branch - so we'll get a new build.

davidfowl:
It's because you added --quiet. You might need to turn it back on.

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
pranavkm: I can revert this - but they're functionally identical.

ChengTian: Yes, please revert it. I guess it's better to use the same approach consistently.

pranavkm:
Updated.

ChengTian:
:shipit:

PR #1607, opened 2015-04-07
Add serviceable attribute to projects.

NTaylorMullen:
I don't believe `ApplicationHost`/`ApplicationHost2` are needed but I added them for now since they were in the `src` folder


anurse:
AppHost(2) are part of DNX and are not serviceable via package servicing, so they don't need the serviceable attribute. I don't see why it would be a problem for them to have it though, so meh.

NTaylorMullen:
Updated.

anurse:
:shipit:

PR #1613, opened 2015-04-07
Add submodule NuGet.Configuration

ChengTian:
Pull NuGet.Configuration in for `dnu restore` refactoring

@davidfowl @anurse


ChengTian:
Just realized that dnu doesn't need submodule. We can use them as NuGet dependences directly.

Issue #1547, opened 2015-03-30
Make Unit Tests work on CoreCLR


BrennanConroy:
Should be easy for most tests. i.e. CommandLineUtils.Tests just needed to target dnxcore50.

PR #1614, opened 2015-04-07
Enable Unit tests on CoreCLR

BrennanConroy:
#1547 
DesignTimeHost tests aren't on CoreCLR since Assembly class is non-inheritable on CoreCLR and it is used heavily.


test/Microsoft.Framework.Runtime.Tests/PathHelpers.cs
```diff
@@ -25,13 +25,17 @@ public static string GetRootedPath(params string[] paths)
                 return root;
             }
 
-            return Path.Combine(root, paths.Aggregate(Path.Combine));
+            return Path.Combine(root, paths.Aggregate((path1, path2) => Path.Combine(path1, path2)));
         }
 
         private static bool IsWindows()
```
victorhurdugaci: We should get rid of this method and use `PlatformHelper` instead

test/Microsoft.Framework.PackageManager.FunctionalTests/project.json
```diff
@@ -3,6 +3,7 @@
         "Microsoft.Framework.FunctionalTestUtils": "1.0.0-*",
         "Microsoft.Framework.Runtime.Common": "1.0.0-*",
         "Microsoft.Framework.PackageManager": "1.0.0-*",
+        "Microsoft.Framework.PackageManager.Tests": "1.0.0-*",
```
victorhurdugaci: Why is functional tests depending on the (unit) tests project?

test/Bootstrapper.FunctionalTests/project.json
```diff
@@ -1,7 +1,7 @@
 {
     "dependencies": {
         "Microsoft.Framework.FunctionalTestUtils": "1.0.0-*",
-        "Microsoft.Framework.PackageManager.FunctionalTests": "1.0.0-*",
+        "Microsoft.Framework.PackageManager.Tests": "1.0.0-*",
```
victorhurdugaci: Same question as below. Why this dependency?

test/Microsoft.Framework.PackageManager.FunctionalTests/project.json
BrennanConroy: We used to do it the other way; Unit tests depended on functional tests. They share some Utilities. But now that the unit tests have dnxcore50 as a target framework and the functional tests "don't", moving the utilities to the unit tests makes it all work.

victorhurdugaci: Ugh... That's ugly. Those utilities should be in a test common project

BrennanConroy: Probably

davidfowl: Not probably :smile:

BrennanConroy: That could be taken either way D:

davidfowl:
> DesignTimeHost tests aren't on CoreCLR since Assembly class is non-inheritable on CoreCLR and it is used heavily.

What?

BrennanConroy:
Look at the TestAssembly class and try to make it compile for both CoreCLR and Desktop.

test/Microsoft.Framework.PackageManager.FunctionalTests/project.json
BrennanConroy: So do you want a "test common project" or not?

davidfowl:
Make sure travis keeps working when you make this change

BrennanConroy:
@davidfowl 
Travis passed!

test/Microsoft.Framework.Runtime.Tests/PathHelpers.cs
```diff
@@ -25,13 +25,7 @@ public static string GetRootedPath(params string[] paths)
...
             }
 
-            return Path.Combine(root, paths.Aggregate(Path.Combine));
-        }
-
-        private static bool IsWindows()
-        {
-            var p = (int)Environment.OSVersion.Platform;
-            return (p != 4) && (p != 6) && (p != 128);
+            return Path.Combine(root, paths.Aggregate((path1, path2) => Path.Combine(path1, path2)));
```
davidfowl: So you made my beautiful 1 liner more verbose why? :smile:

BrennanConroy: Because CoreCLR only has Path.Combine(string []) and it needed Path.Combine(string, string)

davidfowl: @ellismg Can we fix this :smile:

test/Microsoft.Framework.PackageManager.FunctionalTests/project.json
davidfowl: Yes, fix it please, or move the logic elsewhere.

victorhurdugaci: Yes, please

BrennanConroy:
Made a common test project

davidfowl:
:shipit:

PR #1621, opened 2015-04-08
Build script do over

davidfowl:
- Made it work for linux (osx support coming soon)
- Always build mono (on all platforms)
- Build platform specific artifacts when on that platform
- Don't build the runtime with the new runtime
- Make it possible to run unit tests on all CLRs and oses
- Make functional tests reuse unzipped folders (for performance)

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -345,7 +345,7 @@ var RUNTIME_TARGETS='${new List<RuntimeTarget>()}'
         // Travis ci has a timelimit for no activity and the CoreCLR pacakge is
         // about 51MB. We're going to print to the console while this is happening 
         // so the build doesn't timeout
-        if (Travis)
+        if (IsTravis)
         {
             new System.Threading.Thread(() => 
             {
--- a/makefile.shade
+++ b/makefile.shade
@@ -836,7 +836,7 @@ functions @{
         get
         {
             var travis = Environment.GetEnvironmentVariable("TRAVIS");
-            return string.Equals(travis, "TRUE");
+            return string.Equals(travis, "true");
         }
     }
 
--- a/makefile.shade
+++ b/makefile.shade
@@ -24,6 +24,9 @@ var SAMPLES_DIR = '${Path.Combine(ROOT, "samples")}'
 var FULL_VERSION = '${PRODUCT_VERSION + "-" + E("DNX_BUILD_VERSION")}'
 var CORECLR_TARGET_PATH = '${Path.Combine(BUILD_DIR2, "CoreCLR")}'
 var RUNTIME_NAME_PREFIX = 'dnx-'
+
+- // Runtime names
+
 var RUNTIME_MONO_NAME = '${RUNTIME_NAME_PREFIX + "mono"}'
 var RUNTIME_CLR_WIN_x86_NAME = '${RUNTIME_NAME_PREFIX + "clr-win-x86"}'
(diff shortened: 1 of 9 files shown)


makefile.shade
```diff
@@ -688,128 +774,97 @@ var RC_FILES = '${FindAllFiles("Resource.rc", BOOTSTRAPPER_EXE_NAME, BOOTSTRAPPE
...
-        var process = Process.Start(procStartInfo);
-        var content = process.StandardOutput.ReadToEnd();
-        process.WaitForExit();
-
-        File.WriteAllText(outputFile, content);
-
-        return process.ExitCode;
+#ensure-clang
+    @{
+        // TODO: Check for clang 3.5 on the path
```
moozzyk: todo

makefile.shade
```diff
@@ -101,124 +121,53 @@ var RC_FILES = '${FindAllFiles("Resource.rc", BOOTSTRAPPER_EXE_NAME, BOOTSTRAPPE
...
+-// Specific to building the dnx
+#build-common target='build-compile'
+#package-common target='package-runtime'
 
-#update-rc-files if='!IsMono'
-    if condition='IsTeamCity'
-        for each='var rcfile in RC_FILES'
-            @{
-                var digitVer = PRODUCT_VERSION.Replace(".", ",") + ",0";
+-// Define platform specific builds as a n00p if the current platform can't produce a build
```
ellismg: Did you mean "no-op" ;-)

makefile.shade
```diff
@@ -688,128 +774,97 @@ var RC_FILES = '${FindAllFiles("Resource.rc", BOOTSTRAPPER_EXE_NAME, BOOTSTRAPPE
...
-      return FindAllFiles("project.json", folders);
+    public string Uname()
+    {    
+        string uname;
+        ExecuteAndRedirectOutput("uname", "", out uname);
+        return string.IsNullOrEmpty(uname) ? null : uname.Trim();
     }
 
-    string[] FindAllFiles(string fileName, params string[] folders)
+    public void Clang(string args)
```
ellismg: In the future I think we should just search for clang/clang++-3.5 on the PATH and use what we find instead of scoping this to travis.  The -3.5 suffix is a Ubuntu-ism and an artifact of us bringing things up on Ubuntu 14.04.

If you file an issue against me I can fix this.

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -142,7 +142,7 @@ extern "C" HRESULT CallApplicationMain(PCALL_APPLICATION_MAIN_DATA data)
     if (!coreClr)
     {
         char* error = dlerror();
-        fprintf(stderr, "failed to locate coreclr.dll with error %s\n", error);
+        fprintf(stderr, "failed to locate coreclr.so with error %s\n", error);
```
ellismg: Unrelated to your change, but we should not hardcode .so since it will be a .dylib on Darwin.  This is also hardcoded in `LoadCoreClr`

ellismg:
LGTM, but you probably want another set of eyes since my sake-foo is not great.  Thanks for the rewrite.

src/dnx.coreclr.unix/dnx.coreclr.cpp
davidfowl: We'll ifdef then

makefile.shade
davidfowl: sounds good

makefile.shade
davidfowl: lolz

PR #1627, opened 2015-04-09
Build DNX for Darwin

ellismg:
- Clean up .so/.dylib differences in dnx/dnx.coreclr
- Refactor DNX PAL and implement it on Darwin
- Update makefile.shade to probe for clang on $PATH
- Add darwin builds to makefile.shade

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -58,6 +58,7 @@ var BOOTSTRAPPER_HOST_NAME = 'dnx.host'
 
 var PROGRAM_FILES_X86 = '${Environment.GetFolderPath(Environment.SpecialFolder.ProgramFilesX86)}'
 var MSBUILD = '${Path.Combine(PROGRAM_FILES_X86, "MSBuild", "14.0", "Bin", "MSBuild.exe")}'
+var CLANG = '${SearchForClang()}'
 
 var MANAGED_PROJECTS = '${FindAllProjects(
     BOOTSTRAPPER_HOST_NAME,
@@ -299,7 +300,7 @@ var RUNTIME_TARGETS='${new List<RuntimeTarget>()}'
 
         Directory.CreateDirectory(soOutputDir);
 
-        Clang(string.Format("-fPIC -shared {0} -g -o {1} --std=c++11 -ldl", string.Join(" ", sourceFiles), soOutputPath));
+        Exec(CLANG, string.Format("-fPIC -shared {0} -g -o {1} -DPLATFORM_LINUX --std=c++11 -ldl", string.Join(" ", sourceFiles), soOutputPath));
     }
 
     copy sourceDir='${soOutputDir}' include='*.so' outputDir='${Path.Combine(BUILD_DIR2, BOOTSTRAPPER_CORECLR_NAME + ".unix")}' overwrite='${true}'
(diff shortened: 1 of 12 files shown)


makefile.shade
```diff
@@ -866,16 +907,28 @@ functions @{
...
-    public void Clang(string args)
+    public string SearchForClang()
     {
-        if (IsTravis)
+        var CXX = Environment.GetEnvironmentVariable("CXX");
+
+        if (!string.IsNullOrEmpty(CXX))
         {
-            Exec("clang++", args);
+            return null;
```
BrennanConroy: return CXX;?

ellismg: Duh.  Yes.

makefile.shade
```diff
@@ -786,11 +825,13 @@ var RUNTIME_TARGETS='${new List<RuntimeTarget>()}'
...
     }
   
-
-
 #ensure-clang
     @{
-        // TODO: Check for clang 3.5 on the path
+        if (CLANG == null || !File.Exists(CLANG))
+        {
+            Log.Warn("clang++ not found. Please ensure you have clang++ or clang++3-5 on the path or set CXX to a path to clang++.");
```
BrennanConroy: clang++-3.5

ellismg:
I did a build/dev-install on both OSX and Linux and could run stuff.  I was using the volatile feed to pick up CoreCLR.

makefile.shade
ellismg: Fixed, thanks!

src/dnx/pal.linux.cpp
```diff
@@ -5,111 +5,32 @@
...
     assert(ret != -1);
 
-    szPath[ret] = '\0';
+    // readlink does not null terminate the path
+    szPath[ret] = _T('\0');
 
     size_t lastSlash = 0;
 
-    for (size_t i = 0; szPath[i] != '\0'; i++)
+    for (size_t i = 0; szPath[i] != _T('\0'); i++)
```
BrennanConroy: ```i <= ret``` saves an array index

ellismg: Will fix.

BrennanConroy: Actually just < I believe

ellismg: Fixed

ellismg:
@davidfowl PTAL

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -90,11 +96,19 @@ void* LoadCoreClr(std::string& runtimeDirectory)
     if (!ret)
     {
         // Try to load coreclr from application path.
-        char pathToBootstrapper[PATH_MAX];
-        ssize_t pathLen = readlink("/proc/self/exe", pathToBootstrapper, PATH_MAX - 1);
+
+#ifdef PLATFORM_DARWIN
```
davidfowl: Can we do this once and pass it around so we can centralize the ifdefs

ellismg: Can do.

ellismg: Fixed, and pushed new history.

src/dnx/pal.h
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-void GetNativeBootstrapperDirectory(LPTSTR szPath);
+LPTSTR GetNativeBootstrapperDirectory();
```
moozzyk: Why this can't just return std::string? It would save all the hassle with cleaning up etc. I have started working on cleaning LPCTSTR/LPTSTR/LPSTR mess by introducing cross platform types definitions (WIP: https://github.com/aspnet/dnx/commit/cfe5955281c87825877842f8bc21c068a2203671#diff-f83d3802c96eed0aa99ee2b81e85e9ffR9). Using stl would also make it much easier to do string operations.

src/dnx/pal.linux.cpp
```diff
@@ -5,111 +5,32 @@
...
     assert(ret != -1);
 
-    szPath[ret] = '\0';
+    // readlink does not null terminate the path
+    szPath[ret] = _T('\0');
 
     size_t lastSlash = 0;
 
-    for (size_t i = 0; szPath[i] != '\0'; i++)
+    for (size_t i = 0; i < ret && szPath[i] != _T('\0'); i++)
```
moozzyk: why are we looking for the last slash from the first character and not from the last?

src/dnx/pal.win32.cpp
```diff
@@ -3,11 +3,13 @@
 
 #include "stdafx.h"
 
-void GetNativeBootstrapperDirectory(LPTSTR szPath)
+LPTSTR GetNativeBootstrapperDirectory()
 {
+    LPTSTR szPath = (LPTSTR)calloc(MAX_PATH, sizeof(TCHAR));
```
moozzyk: why `calloc` and not not `new`?

src/dnx/pal.h
ellismg: When I did the work initially, I was under the impression there was a desire to not use the stl string types in the first level boot strapper.  If it's okay to use them, I'm happy to do so.

ellismg: Actually, I would prefer to check this in as is if the cross platform string types haven't landed yet.

src/dnx/pal.linux.cpp
ellismg: Will fix.

src/dnx/pal.win32.cpp
ellismg: Will fix.

src/dnx/pal.h
moozzyk: That's fine. I will look at it after I refactor invoking `CallApplicationMain` because I will have the types then,

davidfowl:
Tested on osx, and it works. :shipit: when other review comments are done

ellismg:
Rebased, squashed and pushed as 2c6deeca.

PR #1628, opened 2015-04-09
Add Platform Specific Assemblies to cross plat DNX

ellismg:
We will include assemblies with platform specific code (for Linux and
Darwin) in DNX until CoreFX starts shipping nupkg's with all platform
spcific versions and then DNX stack starts to consume them.

This adds the current set of CoreFX assemblies from the CoreCLR package
that have platform specific versions.

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -483,9 +483,26 @@ var RUNTIME_TARGETS='${new List<RuntimeTarget>()}'
 
                 if (target.OS != "win")
                 {
-                    // Platform specific binaries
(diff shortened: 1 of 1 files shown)


ellismg:
@davidfowl Here are the platform specific assemblies as of today.  More will be coming soon (Networking and some Crypto stuff).

davidfowl:
:shipit:

ellismg:
rebased and pushed to dev as 73e1f99.

PR #1639, opened 2015-04-13
Removed stuff from .gitignore

davidfowl:


davidfowl:
We should clean it up in universe

Issue #1637, opened 2015-04-12
Regression in dnu list


davidfowl:
It prints garbage now :smile: 

```
framework/Microsoft.CSharp 4.0.0.0
[...]

PR #1643, opened 2015-04-13
Fix dnu list regression

troydai:
https://github.com/aspnet/dnx/issues/1637

### DIFF
--- a/src/Microsoft.Framework.PackageManager/List/LibraryDependencyFlatRenderer.cs
+++ b/src/Microsoft.Framework.PackageManager/List/LibraryDependencyFlatRenderer.cs
@@ -89,7 +89,7 @@ namespace Microsoft.Framework.PackageManager.List
 
                 if (!_hideDependent)
                 {
-                    var dependents = string.Join(", ", dependenciesMap[description].Select(dep => dep.ToString()).OrderBy(name => name));
+                    var dependents = string.Join(", ", dependenciesMap[description].Select(dep => dep.Identity.ToString()).OrderBy(name => name));
                     results.Add(string.Format("    -> {0}", dependents));
                 }
             }


src/Microsoft.Framework.PackageManager/List/LibraryDependencyFlatRenderer.cs
```diff
@@ -89,7 +89,7 @@ public IEnumerable<string> GetRenderContent(IGraphNode<LibraryDescription> root)
 
                 if (!_hideDependent)
                 {
-                    var dependents = string.Join(", ", dependenciesMap[description].Select(dep => dep.ToString()).OrderBy(name => name));
+                    var dependents = string.Join(", ", dependenciesMap[description].Select(dep => dep.Identity.ToString()).OrderBy(name => name));
```
davidfowl: Identity can obviously never be null right?

troydai: Not here. If you track down the source of the LibraryDescriptions' source, their Identities are already checked before the tree is created. *If* the identity is null, the failure will happen in the very beginning of dnu list.

davidfowl: Can you write a test for this?

troydai: Yup. But lets finish the analyzing first. The original source of all the LibraryDescriptions is ApplicaitonHostContext.DependencyWalker https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Framework.PackageManager/List/DependencyListOperation.cs#L31. Any chance it returns any LibraryDescription with null Identity?

pranavkm:
:shipit: !

PR #1646, opened 2015-04-13
Fix path for --out in dnu pack

BrennanConroy:
Running ```dnu pack --out <somefoldername>``` would not include the full path in the output for nupkg location

### DIFF
--- a/test/Microsoft.Framework.PackageManager.FunctionalTests/DnuPackTests.cs
+++ b/test/Microsoft.Framework.PackageManager.FunctionalTests/DnuPackTests.cs
@@ -1,0 +1,160 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.Framework.CommonTestUtils;
+using Xunit;
+
+namespace Microsoft.Framework.PackageManager
+{
+    public class DnuPackTests
+    {
+        public static IEnumerable<object[]> RuntimeComponents
+        {
+            get
+            {
+                return TestUtils.GetRuntimeComponentsCombinations();
+            }
+        }
+
+        [Theory]
+        [MemberData("RuntimeComponents")]
+        public void DnuPack_NoArgs(string flavor, string os, string architecture)
+        {
+            string expectedDNX =
+            @"Building {0} for DNX,Version=v4.5.1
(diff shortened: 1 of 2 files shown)


davidfowl:
No tests?

BrennanConroy:
Haha, good question. I actually found this when a test I wrote failed because it didn't have the full path I was expecting. I will make a PR soon with a bunch of tests.

davidfowl:
Add it to this PR

BrennanConroy:
Added tests

BrennanConroy:
@davidfowl 
Ping

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuPackTests.cs
```diff
@@ -0,0 +1,157 @@
...
+            string stdOut;
+            string stdError;
+            var runtimeHomeDir = TestUtils.GetRuntimeHomeDir(flavor, os, architecture);
+            int exitCode;
+
+            using (var testEnv = new DnuTestEnvironment(runtimeHomeDir))
+            {
+                File.WriteAllText($"{testEnv.RootDir}/project.json",
+                @"{
+                    ""frameworks"": {
```
victorhurdugaci: Use the `Project` object and generate the json from it instead of hardcoding json here

BrennanConroy: What Project object? All the other tests hardcode the json.

victorhurdugaci: https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Framework.Runtime/Project.cs

If everyone does it, you don't have to follow them :) I really don't like these hardcoded strings.

moozzyk: Is it supposed to be a unit test (in which case I would say using the string is preferable) or a functional/end-2-end test (in which case I could see using Project)?

BrennanConroy:
@victorhurdugaci @davidfowl 
The travis failure is because of console colors, which will be fixed by #1860

victorhurdugaci:
Looks good :shipit:

davidfowl:
:shipit:

davidfowl:
Did travis fail for this change?

BrennanConroy:
Yea because of the console color showing up in stdout on crossplat.

Sent from my HTC

----- Reply message -----

Issue #1634, opened 2015-04-11
Random issue with resource location on Ubuntu + mono (3.12.1) when running dnu restore


davidfowl:
Super weird, and it only happened once

```
fied culture or the neutral culture.  Make sure "Microsoft.Framework.Runtime.NuGetResources.resources" was correctly embedded or linked into assembly "Microsoft.Framework.Runtime" at compile time, or that all the satellite assemblies required are loadable and fully signed.
  at System.Resources.ResourceManager.AssemblyResourceMissing (System.String fileName) [0x00000] in <filename unknown>:0
  at System.Resources.ResourceManager.InternalGetResourceSet (System.Globalization.CultureInfo culture, Boolean createIfNotExists, Boolean tryParents) [0x00000] in <filename unknown>:0
  at System.Resources.ResourceManager.InternalGetResourceSet (System.Globalization.CultureInfo culture, Boolean createIfNotExists, Boolean tryParents) [0x00000] in <filename unknown>:0
  at System.Resources.ResourceManager.InternalGetResourceSet (System.Globalization.CultureInfo culture, Boolean createIfNotExists, Boolean tryParents) [0x00000] in <filename unknown>:0
  at System.Resources.ResourceManager.GetString (System.String name, System.Globalization.CultureInfo culture) [0x00000] in <filename unknown>:0
  at NuGet.Resources.NuGetResources.get_Debug_AddedFile () [0x00000] in <filename unknown>:0
  at NuGet.PhysicalFileSystem.WriteAddedFileAndDirectory (System.String path) [0x00000] in <filename unknown>:0
  at NuGet.PhysicalFileSystem.AddFileCore (System.String path, System.Action`1 writeToStream) [0x00000] in <filename unknown>:0
  at NuGet.PhysicalFileSystem.AddFile (System.String path, System.Action`1 writeToStream) [0x00000] in <filename unknown>:0
  at NuGet.XmlUtility.CreateDocument (System.Xml.Linq.XName rootName, IFileSystem fileSystem, System.String path) [0x00000] in <filename unknown>:0
  at NuGet.XmlUtility.GetOrCreateDocument (System.Xml.Linq.XName rootName, IFileSystem fileSystem, System.String path) [0x00000] in <filename unknown>:0
  at NuGet.Settings+<>c__DisplayClass8_0.<.ctor>b__0 () [0x00000] in <filename unknown>:0
[...]

PR #1648, opened 2015-04-14
Missed a few resource names

victorhurdugaci:
Fixes https://github.com/aspnet/dnx/issues/1634

cc @davidfowl


davidfowl:
:shipit:

Issue #1636, opened 2015-04-12
Typing dnu commands install prints error message twice


davidfowl:
```
Davids-MacBook-Pro:aspnet-buzz dfowler$ dnu commands install
The name of the package to be installed was not specified.
The name of the package to be installed was not specified.
```

/cc @victorhurdugaci

victorhurdugaci:
cc @troydai @BrennanConroy

PR #1649, opened 2015-04-14
Do not print error twice from install global

victorhurdugaci:
Fixes https://github.com/aspnet/dnx/issues/1636

cc @davidfowl


victorhurdugaci:
cc @troydai @BrennanConroy

BrennanConroy:
Looks good to me

troydai:
:shipit:

Issue #310, opened 2014-06-11
Surface runtime environment information 


davidfowl:
Things like, Debug/Release, platform, OS?

ellismg:
.NETCore will bring by Environment.OSVersion, fwiw.  It will let you get the PlatformID of the OS.

davidfowl:
Yay! We'll expose that information via this interface and more :smile:

PR #1650, opened 2015-04-14
Add IRuntimeEnvironment

victorhurdugaci:
Fixes https://github.com/aspnet/dnx/issues/310

I also moved the `RuntimeFramework` from `IApplicationEnvironment` to `IRuntimeEnvironment`. However, I cannot remove the old one because xunit is using it. Therefore, I'll keep both for now, send a PR to xunit and then remove the old one. If I remove the old one, our build will fail.

cc @davidfowl

### DIFF
--- a/src/Microsoft.Framework.Runtime.Interfaces/IRuntimeEnvironment.cs
+++ b/src/Microsoft.Framework.Runtime.Interfaces/IRuntimeEnvironment.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Runtime.Versioning;
+
+namespace Microsoft.Framework.Runtime
+{
+    public interface IRuntimeEnvironment
+    {
+        string OperatingSystem { get; }
+
+        string OperatingSystemVersion { get; }
+
+        string RuntimeType { get; }
+
+        string RuntimeArchitecture { get; }
+
+        string RuntimeVersion { get; }
+    }
+}
--- a/src/dnx.host/Bootstrapper.cs
+++ b/src/dnx.host/Bootstrapper.cs
@@ -73,6 +73,7 @@ namespace dnx.host
                 serviceProvider.Add(typeof(IAssemblyLoaderContainer), container);
                 serviceProvider.Add(typeof(IAssemblyLoadContextAccessor), accessor);
                 serviceProvider.Add(typeof(IApplicationEnvironment), applicationEnvironment);
+                serviceProvider.Add(typeof(IRuntimeEnvironment), new RuntimeEnvironment());
 
                 CallContextServiceLocator.Locator.ServiceProvider = serviceProvider;
 
--- a/src/dnx.host/NativeMethods.cs
+++ b/src/dnx.host/NativeMethods.cs
@@ -1,0 +1,68 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Framework.Runtime
+{
+    internal static class NativeMethods
+    {
+        public unsafe static string Uname()
+        {
+            var buffer = stackalloc byte[8192];
+            try
+            {
+                if (uname((IntPtr)buffer) == 0)
+                {
+                    return Marshal.PtrToStringAnsi((IntPtr)buffer);
+                }
+            }
+            catch
+            {
+            }
+
+            return null;
+        }
+
+        // Linux and Mac import
+        [DllImport("libc")]
+        private static extern int uname(IntPtr buf);
+
+#if DNXCORE50
+        public static Version OSVersion
+        {
+            get
+            {
+                uint dwVersion = GetVersion();
+
+                int major = (int)(dwVersion & 0xFF);
+                int minor = (int)((dwVersion >> 8) & 0xFF);
+
+                return new Version(major, minor);
+            }
+        }
+
+        private static uint GetVersion()
+        {
+            try
+            {
+                return GetVersion_ApiSet();
+            }
+            catch (DllNotFoundException)
+            {
+                return GetVersion_Kernel32();
+            }
+        }
+
+        // The API set is required by OneCore based systems
+        // and it is available only on Win 8 and newer
+        [DllImport("api-ms-win-core-sysinfo-l1-2-1", EntryPoint = "GetVersion")]
+        private static extern uint GetVersion_ApiSet();
+
+        // For Win 7 and Win 2008 compatibility
+        [DllImport("kernel32.dll", EntryPoint = "GetVersion")]
+        private static extern uint GetVersion_Kernel32();
+#endif
+    }
+}
--- a/src/dnx.host/RuntimeEnvironment.cs
+++ b/src/dnx.host/RuntimeEnvironment.cs
@@ -1,0 +1,67 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Reflection;
+using System.Runtime.Versioning;
+
+namespace Microsoft.Framework.Runtime
+{
+    public class RuntimeEnvironment : IRuntimeEnvironment
+    {
+        private string _runtimeVersion;
+
+        public RuntimeEnvironment()
+        {
+#if DNXCORE50
+            RuntimeType = "CoreCLR";
+            RuntimeArchitecture = IntPtr.Size == 8 ? "x64" : "x86";
+#else
+            RuntimeType = Type.GetType("Mono.Runtime") == null ? "CLR" : "Mono";
+            RuntimeArchitecture = Environment.Is64BitProcess ? "x64" : "x86";
+#endif
+
+            string uname = NativeMethods.Uname();
+            if (!string.IsNullOrEmpty(uname))
+            {
+                OperatingSystem = uname;
+                OperatingSystemVersion = null;
+            }
+            else
+            {
+                OperatingSystem = "Windows";
+#if DNXCORE50
+                OperatingSystemVersion = NativeMethods.OSVersion.ToString();
+#else
+                OperatingSystemVersion = Environment.OSVersion.Version.ToString();
+#endif
+            }
+        }
+
+        public string OperatingSystem { get; private set; }
+
+        public string OperatingSystemVersion { get; private set; }
+
+        public string RuntimeType { get; private set; }
+
+        public string RuntimeArchitecture { get; private set; }
+
+        public string RuntimeVersion
+        {
+            get
+            {
+                if (_runtimeVersion == null)
+                {
+                    _runtimeVersion = typeof(RuntimeEnvironment)
+                        .GetTypeInfo()
+                        .Assembly
+                        .GetCustomAttribute<AssemblyInformationalVersionAttribute>()
+                        .InformationalVersion;
+                }
+
+                return _runtimeVersion;
+
+            }
+        }
+    }
+}
--- a/DNX.sln
+++ b/DNX.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.22806.0
+VisualStudioVersion = 14.0.22809.0
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "scripts", "scripts", "{189961D1-DFF4-406A-9D42-39B61221A73A}"
 	ProjectSection(SolutionItems) = preProject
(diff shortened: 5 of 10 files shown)


victorhurdugaci:
cc @troydai @ChengTian

src/Microsoft.Framework.Runtime.Common/Impl/WindowsApis.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace Microsoft.Framework.Runtime
+{
+    internal static class WindowsApis
+    {
+        public static Version OSVersion
+        {
+            get
+            {
+                uint dwVersion = GetVersion();
```
moozzyk: hungarian?

src/Microsoft.Framework.Runtime.Interfaces/IApplicationEnvironment.cs
```diff
@@ -11,6 +12,8 @@ public interface IApplicationEnvironment
         string Version { get; }
         string ApplicationBasePath { get; }
         string Configuration { get; }
+
+        [Obsolete("This method will be removed in a future version. Use IRuntimeEnvironment.Framework instead")]
```
moozzyk: What are the rules about deprecating APIs that have not officially shipped (i.e. why not just remove this property)?

victorhurdugaci: From the PR description:

> also moved the  RuntimeFramework  from  IApplicationEnvironment  to  IRuntimeEnvironment . However, I cannot remove the old one because xunit is using it. Therefore, I'll keep both for now, send a PR to xunit and then remove the old one. If I remove the old one, our build will fail.

moozzyk: Thanks. Jumped to the code without reading the description.

moozzyk:
Looks good to me but I don't know the code too much so have someone else review it.

src/Microsoft.Framework.Runtime.Interfaces/IRuntimeEnvironment.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Microsoft.Framework.Runtime
+{
+    public interface IRuntimeEnvironment
+    {
+        OperatingSystem OperatingSystem { get; }
+        
+        Version OperatingSystemVersion { get; }
+
+        FrameworkName Framework { get; }
```
davidfowl: This isnt right. It should be flavor and it should say what CLR you're running on not what framework.

src/dnx.host/ApplicationEnvironment.cs
```diff
@@ -37,12 +39,6 @@ public string ApplicationBasePath
             private set;
         }
 
-        public FrameworkName RuntimeFramework
```
davidfowl: Bring this back.

src/Microsoft.Framework.Runtime.Common/Impl/RuntimeEnvironment.cs
```diff
@@ -0,0 +1,68 @@
...
+#endif
+
+        private static readonly OperatingSystem _operatingSystem;
+
+        static RuntimeEnvironment()
+        {
+#if DNXCORE50
+            // TODO: Weak assumption, will not work well on non Windows
+            // waiting for the CLR team to provide the right API
+            _operatingSystem = OperatingSystem.Windows;
```
davidfowl: Nope, this doesn't work. Copy the code from kestrel

davidfowl:
Leave IApplicationEnvironment as is and change IRuntimeEnvironment to have Flavor?

src/Microsoft.Framework.Runtime/ApplicationEnvironment.cs
```diff
@@ -44,10 +44,12 @@ public string Version
...
         public string Configuration { get; private set; }
 
-
         public FrameworkName RuntimeFramework
         {
-            get { return _targetFramework; }
+            get
+            {
+                return _runtimeEnv.Framework;
+            }
```
troydai: Would be nice to keep these four lines in one line so as to compact the source files.

src/Microsoft.Framework.Runtime.Common/Impl/RuntimeEnvironment.cs
```diff
@@ -0,0 +1,68 @@
...
+            Version osVersion = Environment.OSVersion.Version;
+            switch (Environment.OSVersion.Platform)
+            {
+                case PlatformID.Win32NT:
+                    _operatingSystem = OperatingSystem.Windows;
+                    break;
+                case PlatformID.Unix:
+                    _operatingSystem = OperatingSystem.Unix;
+                    break;
+                case PlatformID.MacOSX:
```
Tragetaschen: This doesn't work on mono: https://github.com/mono/mono/blob/master/mcs/class/corlib/System/Environment.cs#L238

davidfowl: Yep. Like I said, @victorhurdugaci we need to use uname instead of relying on Environment.OSVersion where possible.

davidfowl:
After you fix everything to not change IApplicationEnvironment, add some tests with the conditionals to test that this logic works on all oses and platforms.

src/Microsoft.Framework.Runtime.Common/Impl/WindowsApis.cs
davidfowl: cool like in the 90s

src/Microsoft.Framework.Runtime.Interfaces/IApplicationEnvironment.cs
davidfowl: That's ok, we should revert this anyways.

src/Microsoft.Framework.Runtime/ApplicationHostContext.cs
```diff
@@ -108,6 +108,7 @@ public class ApplicationHostContext
 
             // Default services
             _serviceProvider.Add(typeof(IApplicationEnvironment), new ApplicationEnvironment(Project, targetFramework, configuration));
+            _serviceProvider.Add(typeof(IRuntimeEnvironment), new RuntimeEnvironment());
```
davidfowl: Not sure you need both. It should just flow from the host.

src/Microsoft.Framework.Runtime.Interfaces/IRuntimeEnvironment.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Microsoft.Framework.Runtime
+{
+    public interface IRuntimeEnvironment
+    {
+        OperatingSystem OperatingSystem { get; }
+
+        Version OperatingSystemVersion { get; }
+
+        string FrameworkFlavor { get; }
```
davidfowl: RuntimeFlavor not framework.

src/Microsoft.Framework.Runtime.Interfaces/IRuntimeEnvironment.cs
```diff
@@ -0,0 +1,19 @@
...
+
+        string FrameworkArchitecture { get; }
```
davidfowl: RuntimeArchitecture

victorhurdugaci: `Architecture`? I don't want to repeat `Runtime` from the class name

davidfowl: Yep sure.

src/Microsoft.Framework.Runtime.Interfaces/IRuntimeEnvironment.cs
davidfowl: Maybe explicit and call it ClrFlavor? ClrType ?

src/Microsoft.Framework.Runtime.Interfaces/IRuntimeEnvironment.cs
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Runtime.Versioning;
+
+namespace Microsoft.Framework.Runtime
+{
+    public interface IRuntimeEnvironment
+    {
+        OperatingSystem OperatingSystem { get; }
```
davidfowl: Lets make this a string

src/Microsoft.Framework.Runtime.Common/Impl/RuntimeEnvironment.cs
```diff
@@ -0,0 +1,48 @@
...
+            FrameworkArchitecture = IntPtr.Size == 8 ? "x64" : "x86";
+#else
+            FrameworkFlavor = Type.GetType("Mono.Runtime") == null ? "CLR" : "Mono";
+            FrameworkArchitecture = Environment.Is64BitProcess ? "x64" : "x86";
+#endif
+
+            string uname = ExternalApis.Uname();
+            if (!string.IsNullOrEmpty(uname))
+            {
+                OperatingSystem = uname.Equals("Darwin", StringComparison.Ordinal) ?
```
davidfowl: string.Equals

src/Microsoft.Framework.Runtime.Common/Impl/ExternalApis.cs
```diff
@@ -0,0 +1,80 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Framework.Runtime
+{
+    internal static class ExternalApis
```
davidfowl: Call this NativeMethods

src/dnx.host/RuntimeEnvironment.cs
```diff
@@ -0,0 +1,46 @@
...
+#else
+            FrameworkFlavor = Type.GetType("Mono.Runtime") == null ? "CLR" : "Mono";
+            FrameworkArchitecture = Environment.Is64BitProcess ? "x64" : "x86";
+#endif
+
+            string uname = NativeMethods.Uname();
+            if (!string.IsNullOrEmpty(uname))
+            {
+                OperatingSystem = uname;
+                OperatingSystemVersion = string.Empty;
```
davidfowl: Leave it null

src/Microsoft.Framework.Runtime.Interfaces/IRuntimeEnvironment.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public interface IRuntimeEnvironment
+    {
+        string OperatingSystem { get; }
+
+        string OperatingSystemVersion { get; }
+
+        string FrameworkFlavor { get; }
+
+        string FrameworkArchitecture { get; }
```
davidfowl: Architecture

src/Microsoft.Framework.Runtime.Interfaces/IRuntimeEnvironment.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Microsoft.Framework.Runtime
+{
+    public interface IRuntimeEnvironment
+    {
+        string OperatingSystem { get; }
+
+        string OperatingSystemVersion { get; }
+
+        string FrameworkFlavor { get; }
```
davidfowl: RuntimeType

src/Microsoft.Framework.Runtime.Interfaces/IRuntimeEnvironment.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Runtime.Versioning;
+
+namespace Microsoft.Framework.Runtime
+{
+    public interface IRuntimeEnvironment
```
davidfowl: Add one more thing to this. RuntimeVersion

src/dnx.host/RuntimeEnvironment.cs
```diff
@@ -0,0 +1,46 @@
...
+
+namespace Microsoft.Framework.Runtime
+{
+    public class RuntimeEnvironment : IRuntimeEnvironment
+    {
+        public RuntimeEnvironment()
+        {
+#if DNXCORE50
+            FrameworkFlavor = "CoreCLR";
+            FrameworkArchitecture = IntPtr.Size == 8 ? "x64" : "x86";
```
Tragetaschen: Isn't [Is64BitProcess](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Environment.cs#L1320) available yet?

davidfowl:
Build broke

src/dnx.host/NativeMethods.cs
```diff
@@ -0,0 +1,80 @@
...
+            catch (DllNotFoundException)
+            {
+                return null;
+            }
+        }
+
+
+        // Linux and Mac import
+        [DllImport("libc")]
+        private static extern int uname(IntPtr buf);
```
Tragetaschen: [In my big Kestrel PR](https://github.com/Tragetaschen/KestrelHttpServer/blob/pinvoke-cleanup/src/Microsoft.AspNet.Server.Kestrel/Networking/Libuv.cs#L26), I'm using a variant that omits unsafe entirely.
[CoreFX](https://github.com/dotnet/corefx/blob/master/src/Common/src/Interop/Interop.PlatformDetection.cs#L58) uses `stackalloc`

src/Microsoft.Framework.Runtime.Interfaces/IRuntimeEnvironment.cs
victorhurdugaci: I'll go with RuntimeArchitecture because I don't want anyone to confuse it with OS architecture

victorhurdugaci:
Updated with runtime version

src/dnx.host/Bootstrapper.cs
```diff
@@ -67,12 +67,19 @@ public Task<int> RunAsync(List<string> args)
                                                                         configuration,
                                                                         assembly);
 
+                var versionAttribute = typeof(Bootstrapper)
```
davidfowl: Do this lazily, we should never need to do reflection if nobody asks for it.

src/dnx.host/NativeMethods.cs
```diff
@@ -0,0 +1,68 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Framework.Runtime
+{
+    internal static class NativeMethods
+    {
+        public unsafe static string Uname()
```
davidfowl: does this still need to be unsafe?

victorhurdugaci: Yes, stackalloc and fixed size buffers require an unsafe context

davidfowl:
:shipit: Once the CI passes. I still want to add tests for these things.

PR #1652, opened 2015-04-14
Refactoring DNX - part 1

moozzyk:
@davidfowl @ellismg Hopefully I did not get carried away too much (because there is more work to do)

build verify passed on win, linux and mac


src/dnx/dnx.cpp
```diff
@@ -226,150 +227,78 @@ bool ExpandCommandLineArguments(int nArgc, LPTSTR* ppszArgv, int& nExpandedArgc,
...
-        m_hHostModule = nullptr;
-    }
+    data.applicationBase = appBaseBuffer;
 
-    if (bExpanded)
+    try
     {
-        FreeExpandedCommandLineArguments(nExpandedArgc, ppszExpandedArgv);
-    }
+        const dnx::char_t* hostModuleName =
```
anurse: This seems like maybe it belongs as a constant in a header file somewhere.

src/dnx/pal.unix.cpp
```diff
@@ -47,33 +52,44 @@ BOOL GetFullPath(LPCTSTR szPath, LPTSTR szNormalizedPath)
     return TRUE;
 }
 
-HMODULE LoadNativeHost(LPCTSTR szHostModuleName)
+int CallApplicationMain(const char* moduleName, const char* functionName, CALL_APPLICATION_MAIN_DATA* data, TraceWriter traceWriter)
 {
-    LPTSTR localPath = GetNativeBootstrapperDirectory();
+    auto localPath =  GetNativeBootstrapperDirectory().append("/").append(moduleName);
```
anurse: Extra space before `GetNativeBootstrapperDirectory`

src/dnx/pal.unix.cpp
```diff
@@ -47,33 +52,44 @@ BOOL GetFullPath(LPCTSTR szPath, LPTSTR szNormalizedPath)
...
+        if (!pfnCallApplicationMain)
+        {
+            std::ostringstream oss;
+            oss << "Failed to find export '" << functionName << "' in " << moduleName;
+            throw std::runtime_error(oss.str());
+        }
+
+        traceWriter.Write(dnx::xstring_t(_X("Found export: ")).append(moduleName), true);
+
+        auto result  = pfnCallApplicationMain(data);
```
anurse: Extra space after `result`

src/dnx/dnx.cpp
```diff
@@ -226,150 +227,78 @@ bool ExpandCommandLineArguments(int nArgc, LPTSTR* ppszArgv, int& nExpandedArgc,
...
-    // Set the DNX_CONOSLE_HOST flag which will print exceptions
-    // to stderr instead of throwing
-    SetConsoleHost();
-
-    CALL_APPLICATION_MAIN_DATA data = { 0 };
-    int nExpandedArgc = -1;
-    LPTSTR* ppszExpandedArgv = nullptr;
-    bool bExpanded = ExpandCommandLineArguments(argc - 1, &(argv[1]), nExpandedArgc, ppszExpandedArgv);
-    if (bExpanded)
+    for (int i = 0; i < argc; ++i)
```
BrennanConroy: If you do ```i < argc - 1``` here you don't need it in the if below and keep the same behavior.

src/dnx/dnx.cpp
moozzyk: I did think about moving this elsewhere but I am on the edge here because it is being used only in this one place and currently we don't have a header for constants. I decided to leave it as is for now.

src/dnx/pal.unix.cpp
moozzyk: Could not install my awesome [trailing space flagger](https://visualstudiogallery.msdn.microsoft.com/bcb3ef8e-a3a4-401e-9e63-fdf00b45268f) on emacs. Will fix.

src/dnx/dnx.cpp
```diff
@@ -226,150 +227,78 @@ bool ExpandCommandLineArguments(int nArgc, LPTSTR* ppszArgv, int& nExpandedArgc,
...
-        exitCode = data.exitcode;
-    }
-    else
-    {
-        fSuccess = false;
-        exitCode = hr;
-    }
+    CALL_APPLICATION_MAIN_DATA data = { 0 };
+    data.argc = argc;
+    data.argv = const_cast<const dnx::char_t**>(argv);
```
BrennanConroy: Why const? Should just be able to reinterpret_cast<dnx::char_t **>(argv)

src/dnx/pal.unix.cpp
moozzyk: Will fix.

src/dnx/dnx.cpp
BrennanConroy: This can also start at ```i = 1``` I believe

src/dnx/dnx.cpp
moozzyk: To remove or add const you use `const_cast` and not `reinterpret_cast` (I **did** to have to look it up)

src/dnx/dnx.cpp
moozzyk: Will check - this is more or less what we had and I just moved the block.

moozzyk: Will fix.

src/dnx/dnx.cpp
anurse: Fair enough. I noticed it was only used here.

anurse:
Looks good but I'll let other more experience C++ folks give a squirrel :)

src/dnx/dnx.cpp
BrennanConroy: Ya, I just noticed that data.argv is const, all good here :)

src/dnx/pal.unix.cpp
```diff
@@ -47,33 +52,44 @@ BOOL GetFullPath(LPCTSTR szPath, LPTSTR szNormalizedPath)
...
+    try
     {
-        return dlclose(hModule);
+        host = dlopen(localPath.c_str(), RTLD_NOW | RTLD_GLOBAL);
+        if (!host)
+        {
+            throw std::runtime_error(std::string("Failed to load: ").append(moduleName));
+        }
+
+        traceWriter.Write(std::string("Loaded module: ").append(moduleName), true);
```
BrennanConroy: You use std::string() and dnx::xstring_t() randomly throughout this file.

moozzyk: I remember fixing this - must have overwritten the file when moving changes between boxes. Should be `std::string` everywhere since this is a platform specific file and we know what types we should be using.

src/dnx/pal.unix.cpp
```diff
@@ -47,33 +52,44 @@ BOOL GetFullPath(LPCTSTR szPath, LPTSTR szNormalizedPath)
...
+
+        auto pfnCallApplicationMain = (FnCallApplicationMain)dlsym(host, functionName);
+        if (!pfnCallApplicationMain)
+        {
+            std::ostringstream oss;
+            oss << "Failed to find export '" << functionName << "' in " << moduleName;
+            throw std::runtime_error(oss.str());
+        }
+
+        traceWriter.Write(dnx::xstring_t(_X("Found export: ")).append(moduleName), true);
```
moozzyk: should be `functionName` instead of `moduleName`

src/dnx/pal.win32.cpp
```diff
@@ -76,17 +81,41 @@ BOOL GetFullPath(LPCTSTR szPath, LPTSTR pszNormalizedPath)
...
-BOOL FreeNativeHost(HMODULE hHost)
-{
-    return FreeLibrary(hHost);
-}
+        traceWriter.Write(std::wstring(L"Loaded module: ").append(moduleName), true);
 
-FARPROC GetEntryPointFromHost(HMODULE hHost, LPCSTR lpProcName)
-{
-    return GetProcAddress(hHost, lpProcName);
+        auto pfnCallApplicationMain = (FnCallApplicationMain)GetProcAddress(hostModule, functionName);
```
moozzyk: `(FnCallApplicationMain)GetProcAddress` - should be `reinterpret_cast`

davidfowl: zomg

moozzyk: while it does not make much difference here "the thinking goes that casting is inherently an ugly operation and it requires ugly syntax" :smile:

src/dnx/dnx.cpp
```diff
@@ -122,20 +123,25 @@ void FreeExpandedCommandLineArguments(int nArgc, LPTSTR* ppszArgv)
...
-bool ExpandCommandLineArguments(int nArgc, LPTSTR* ppszArgv, int& nExpandedArgc, LPTSTR*& ppszExpandedArgv)
+dnx::char_t* GetAppBaseParameterValue(int argc, dnx::char_t* argv[])
 {
-    if (nArgc == 0)
+    for (auto i = 0; i < argc - 1; ++i)
     {
-        return false;
+        if (StringsEqual(argv[i], _X("--appbase")))
+        {
+            return argv[i + 1];
```
troydai: Do you need to check if index is out of range?

troydai: Ah, my bad just saw the `i < argc - 1`. good.

moozzyk:
can't use stl in dnx :facepunch: - learnt that hard way

BrennanConroy:
Oh no D: why not?

moozzyk:
FYI. Resurrecting this PR. I rebased on my latest changes and now it runs on both Win7 and Nano just by copy paste.

davidfowl:
@moozzyk Seems like appveyor is broken. Lets make sure this change doesn't break that in general.

src/dnx/TraceWriter.h
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+#pragma once
+
+#include "xplat.h"
+
+class TraceWriter
```
davidfowl: Inconsistent casing?

src/dnx/dnx.cpp
```diff
@@ -386,22 +307,34 @@ extern "C" int __stdcall DnxMain(int argc, wchar_t* argv[])
...
+    LPTSTR* ppszExpandedArgv = nullptr;
+    auto expanded = ExpandCommandLineArguments(argc - 1, &(argv[1]), nExpandedArgc, ppszExpandedArgv);
+
+    auto traceWriter = TraceWriter{ IsTracingEnabled() };
+    if (!expanded)
+    {
+        return CallApplicationProcessMain(argc - 1, &argv[1], traceWriter);
+    }
+
+    auto exitCode = CallApplicationProcessMain(nExpandedArgc, ppszExpandedArgv, traceWriter);
```
davidfowl: How does the exception escape to the console?

moozzyk: There is no exception here - `CallApplicationProcessMain` does not throw but it catches any exception and turns it to `1` as the `exitCode`.

src/dnx/TraceWriter.h
moozzyk: Hard to tell. I am a big fan of `snake_case` in C/C++ but these files are using hungarian + camelCasing + Windows. I don't think spending time on just changing casing is really worth it, so I wanted to make it kind of similar to what we have. Note that this is the only class in these files so far so as far as class names go it can't be inconsistent.

src/dnx/dnx.cpp
moozzyk: (the exception message is being written to the console in the exception handler in the `CallApplicationProcessMain`)

Issue #1665, opened 2015-04-16
What does this DNU exception mean?


GuardRex:
I ran into this exception trying to jockey my runtimes to only have only the CoreCLR x64 runtime installed on my Win7 box. Calling `dnu restore` triggers this.
```
Unable to find an entry point named 'GetVersion' in DLL 'api-ms-win-core-sysinfo-l1-2-1'.
at Microsoft.Framework.Runtime.NativeMethods.GetVersion_ApiSet()
at Microsoft.Framework.Runtime.NativeMethods.GetVersion()
at Microsoft.Framework.Runtime.NativeMethods.get_OSVersion()
at Microsoft.Framework.Runtime.RuntimeEnvironment...ctor()
at dnx.host.Bootstrapper.RunAsync(List`1 args)
at dnx.host.RuntimeBootstrapper.ExecuteAsync(String[] args)
at dnx.host.RuntimeBootstrapper.Execute(String[] args)
```

davidfowl:
Means @victorhurdugaci caused a regression

victorhurdugaci:
Ugh... that's my change from yesterday. I'll take a look

victorhurdugaci:
Sent a PR with a fix

PR #1666, opened 2015-04-16
Fix IRuntimeEnvironment exception

victorhurdugaci:
Fixes https://github.com/aspnet/dnx/issues/1665

cc @davidfowl

### DIFF
--- a/src/dnx.host/NativeMethods.cs
+++ b/src/dnx.host/NativeMethods.cs
@@ -49,10 +49,18 @@ namespace Microsoft.Framework.Runtime
             {
                 return GetVersion_ApiSet();
             }
-            catch (DllNotFoundException)
+            catch
             {
-                return GetVersion_Kernel32();
+                try
+                {
+                    return GetVersion_Kernel32();
+                }
+                catch
+                {
+                    return 0;
+                }
             }
+
         }
 
         // The API set is required by OneCore based systems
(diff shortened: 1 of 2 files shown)


davidfowl:
The try catch is fine but why doesn't it work on win7?

victorhurdugaci:
Because I was catching the wrong exception. On Win7 the api set DLL is there but it is missing that API. So it was throwing an entry point not found instead of DLL not found

davidfowl:
Can we also change this code to be lazy? It would be great if we could avoid all first chance exceptions on the startup path.

GuardRex:
Queued at AppVeyor for over an hour. Is that normal?

... and after the AppVeyor build, will `dnvm update-self` get me back to work?

victorhurdugaci:
@GuardRex yes, it can take a while. Now it is done
@davidfowl sure, let me update the code

victorhurdugaci:
@GuardRex no, AppVeyor does not produce any bits. You will have to wait for this PR to be merged and then run "dnvm upgrade". Or, you can get my code and build dnx locally

GuardRex:
ok, thanks

GuardRex:
@victorhurdugaci My attempts to `dnvm install 1.0.0-beta5-11561 -r coreclr -arch x64` (and `-11560`) failed to get me a `dnu restore` that works. What's the ETA on a merge like this (I see a bunch of PR's have been around for days)? If it's going to be a while, then I will need to build locally ... or do you know what I'm doing wrong in trying to get an older, working runtime installed?

victorhurdugaci:
I'll merge this tomorrow after I update the PR. 

Meanwhile, try an even older build like 11540 or even 11530 because we produce more than one build per day. I can't check right now which is last build before my previous change.

GuardRex:
ah ... that explains it then ... I simply didn't go back far enough. Thanks.

Issue #1663, opened 2015-04-16
Update dnx.coreclr.dll on *NIX to handle coreclr PAL changes


ellismg:
Recent builds of CoreCLR on *NIX now depends on a shared library (libcoreclrpal.[so|dylib]).  We need to update DNX to load the PAL first, otherwise loading coreclr itself will fail.

davidfowl:
Why doesn't core clr load them?

ellismg:
You can't even load coreclr.so, since the imports can't be resolved, so it can't run code and do it on its behalf.

PR #1668, opened 2015-04-17
Cross Plat DNX Changes

ellismg:
This has two sets of changes
- Update the list of platform specific binaries to include with dnx today.  We have made more things platform specific as we implement them.
- React to a CoreCLR change requiring hosts to load the PAL before loading the CLR itself.

The second issue is prevents me from updating the coreclr we package up for DNX, so I would like to get this in sooner rather than later.

I am still validating that a clean build of DNX with these changes + a CoreCLR package with the new runtime installs and runs hello world, but I figured I would put up the PR now while I do the validation, I have done some by hand stuff and things look good.

### DIFF
--- a/src/dnx.coreclr.unix/dnx.coreclr.cpp
+++ b/src/dnx.coreclr.unix/dnx.coreclr.cpp
@@ -38,6 +38,9 @@ typedef HRESULT (*ExecuteAssemblyFunction)(
 const HRESULT S_OK = 0;
 const HRESULT E_FAIL = -1;
 
+namespace
+{
+
 #ifdef PLATFORM_DARWIN
 const char* LIBCORECLR_NAME = "libcoreclr.dylib";
 const char* LIBCORECLRPAL_NAME = "libcoreclrpal.dylib";
@@ -98,6 +101,23 @@ bool GetTrustedPlatformAssembliesList(const std::string& tpaDirectory, bool isNa
 void* pLibCoreClr = nullptr;
 void* pLibCoreClrPal = nullptr;
 
+bool LoadCoreClrAtPath(const std::string loadPath, void** ppLibCoreClr, void** ppLibCoreClrPal)
+{
+    std::string coreClrDllPath = loadPath;
+    std::string coreClrPalPath = loadPath;
+
+    coreClrDllPath.append("/");
+    coreClrDllPath.append(LIBCORECLR_NAME);
+
+    coreClrPalPath.append("/");
+    coreClrPalPath.append(LIBCORECLRPAL_NAME);
+
+    *ppLibCoreClrPal = dlopen(coreClrPalPath.c_str(), RTLD_NOW | RTLD_GLOBAL);
+    *ppLibCoreClr = dlopen(coreClrDllPath.c_str(), RTLD_NOW | RTLD_GLOBAL);
+
+    return *ppLibCoreClrPal != nullptr && *ppLibCoreClr != nullptr;
+}
+
 // libcoreclr has a dependency on libcoreclrpal, which is commonly not on LD_LIBRARY_PATH, so for every 
 // location we try to load libcoreclr from, we first try to load libcoreclrpal so when we load coreclr
 // itself the linker is happy.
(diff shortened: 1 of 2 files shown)


ellismg:
@davidfowl PTAL

ellismg:
I will add fixes metadata to 1f5d0ef to refer to #1663

ellismg:
I completed my validation.

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -124,13 +148,30 @@ void* LoadCoreClr(std::string& runtimeDirectory)
         runtimeDirectory.erase(lastSlash);
 
         std::string coreClrDllPath = runtimeDirectory;
```
moozzyk: nit: lines 150-160 should be a function - right now we basically copied the code from the above

ellismg: OK

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -124,13 +148,30 @@ void* LoadCoreClr(std::string& runtimeDirectory)
         runtimeDirectory.erase(lastSlash);
```
moozzyk: I don't know what happens in the code after this function exits but since `runtimeDirectory` is passed by reference this line changes the `runtimeDirectory` that is used by upstream callers. Is it intentional?

ellismg: Yes.  The caller of this function needs to know the directory the runtime was loaded from so it can use it later.

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -105,13 +116,26 @@ void* LoadCoreClr(std::string& runtimeDirectory)
         runtimeDirectory = coreClrEnvVar;
 
         std::string coreClrDllPath = runtimeDirectory;
```
moozzyk: (This is a general note - I wonder if we need a general try/catch(const &std:exception) at the entry point somewhere - these could potentially throw `std::bad_alloc`  which will lead to a crash if not handled somehow)

ellismg: Is there a reasonable policy to be applied that isn't just crashing?

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -93,8 +95,17 @@ bool GetTrustedPlatformAssembliesList(const std::string& tpaDirectory, bool isNa
     return true;
 }
 
-// TODO: Figure out if runtimeDirectory should have a trailing "/".  Need to look at what happens on Windows.
-void* LoadCoreClr(std::string& runtimeDirectory)
+void* pLibCoreClr = nullptr;
```
moozzyk: consider putting all but `CallApplicationMain` in the unnamed namespace to hide it from everyone else (won't be visible outside this translation unit) just for safety

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -157,9 +198,9 @@ extern "C" HRESULT CallApplicationMain(PCALL_APPLICATION_MAIN_DATA data)
     }
 
     std::string coreClrDirectory;
-    void* coreClr = LoadCoreClr(coreClrDirectory);
+    LoadCoreClr(coreClrDirectory);
 
-    if (!coreClr)
+    if (!pLibCoreClr)
     {
         char* error = dlerror();
```
moozzyk: shouldn't you unload pLibCoreClrPal here?

ellismg: Yes.  I will restructure this (as well as the return if the tpa can't be built).

src/dnx.coreclr.unix/dnx.coreclr.cpp
moozzyk: I don't know - but if writing the error to the console succeeded then it could be helpful.

src/dnx.coreclr.unix/dnx.coreclr.cpp
moozzyk: :ok_hand:

src/dnx.coreclr.unix/dnx.coreclr.cpp
ellismg: Good idea.

moozzyk:
@davidfowl - this does not affect me yet - I haven't left dnx yet - fighting with nano and SignalR C++ Client
@ellismg other than freeing `pLibCoreClrPal` it looks good to me. The other comments are mostly about cleanliness not functionality per se.

ellismg:
Thanks @moozzyk.  Quick question, do you want me to indent the code in the default namespace? I am not sure what your guidelines are here.

ellismg:
(In f4ae2ef I responded to your feedback, except for the try/catch thing and did not indent the code in the namespace)

moozzyk:
try/catch is totally unrelated to this - it was more kind of a mental note for myself. 
I don't think it's important to indent the code in the unnamed namespce - I don't think we have any guidelines about namespaces because so far there are no namespaces in the dev branch so I would just leave it as is. 
:shipit:

ellismg:
Rebased and pushed.  Thanks!

PR #1691, opened 2015-04-22
Don't use Uri class for GetRelativePath

BrennanConroy:


test/Microsoft.Framework.Runtime.Tests/PathUtilityFacts.cs
```diff
@@ -39,5 +39,21 @@ public void IsChildOfDirectoryWorksOnBaseDirWithoutTrailingPathSeparator()
             Assert.True(PathUtility.IsChildOfDirectory(baseDir, childPath));
             Assert.False(PathUtility.IsChildOfDirectory(baseDir, nonChildPath));
         }
+
+        [Fact]
+        public void GetRelativePathReturnsCorrectRelativePaths()
```
davidfowl: Use theories instead and add some tests that handle windows style paths.

BrennanConroy:
Fixed, also the CI's passed just before I changed it

src/Microsoft.Framework.Runtime/NuGet/Utility/PathUtility.cs
```diff
@@ -72,7 +72,58 @@ public static string GetRelativePath(string path1, string path2)
...
+            var path2Segments = path2.Split(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
+            // if path1 does not end with / it is assumed the end is not a directory
+            // we will assume that is isn't a directory by ignoring the last split
+            var len1 = path1Segments.Length - 1;
+            var len2 = path2Segments.Length;
+
+            int min = Math.Min(len1, len2);
+            while (min > index)
+            {
+                if (string.Compare(path1Segments[index], path2Segments[index], StringComparison.OrdinalIgnoreCase) != 0)
```
moozzyk: `StringComparison.OrdinalIgnoreCase` - paths are case sensitive on non-windows

moozzyk:
Honestly, I think we should not have this code. If the Uri class is broken then it may be good as a stopgap but ultimately this should be fixed in the Uri class... Maybe find someone who own Uri and have them confirm this is a bug and maybe they will give you a workaround?

BrennanConroy:
@davidfowl 
Should we get this in?
I filed an issue on CoreFX
https://github.com/dotnet/corefx/issues/1745

davidfowl:
@BrennanConroy no. I'm on the same page as @moozzyk and I'm also worried about regressions.

BrennanConroy:
@davidfowl @moozzyk 
Please review before Monday

test/Microsoft.Dnx.Runtime.Tests/PathUtilityFacts.cs
```diff
@@ -40,5 +41,54 @@ public void IsChildOfDirectoryWorksOnBaseDirWithoutTrailingPathSeparator()
             Assert.True(PathUtility.IsChildOfDirectory(baseDir, childPath));
             Assert.False(PathUtility.IsChildOfDirectory(baseDir, nonChildPath));
         }
+
+        public static IEnumerable<object[]> FolderPaths
+        {
+            get {
+                var list =  new List<object[]>
```
davidfowl: You  need some tests with the slashes in the opposite direction.

davidfowl: Also add file paths.

BrennanConroy:
Updated

src/Microsoft.Dnx.Runtime/NuGet/Utility/PathUtility.cs
```diff
@@ -72,7 +73,72 @@ public static string GetRelativePath(string path1, string path2)
         /// </summary>
         public static string GetRelativePath(string path1, string path2, char separator)
         {
-            return UriUtility.GetPath(GetRelativeUri(path1, path2), separator);
+            if (path1 == null)
```
moozzyk: does this function work if one of the paths (or both) is `string.Empty`?

test/Microsoft.Dnx.Runtime.Tests/PathUtilityFacts.cs
```diff
@@ -40,5 +41,56 @@ public void IsChildOfDirectoryWorksOnBaseDirWithoutTrailingPathSeparator()
             Assert.True(PathUtility.IsChildOfDirectory(baseDir, childPath));
             Assert.False(PathUtility.IsChildOfDirectory(baseDir, nonChildPath));
         }
+
+        public static IEnumerable<object[]> FolderPaths
+        {
+            get {
```
moozzyk: nit: put `{` on the next line

src/Microsoft.Dnx.Runtime/NuGet/Utility/PathUtility.cs
moozzyk: also all the tests are using absolute paths - what if a relative path is passed?

test/Microsoft.Dnx.Runtime.Tests/PathUtilityFacts.cs
davidfowl: Add some tests that try to resolve 2 absolute paths

src/Microsoft.Dnx.Runtime/NuGet/Utility/PathUtility.cs
BrennanConroy: The function needs absolute paths, what does it even mean to get a relative path from two relative paths? Its not possible without assuming they have the same base path.
If string.Empty is passed it should probably error too?

test/Microsoft.Dnx.Runtime.Tests/PathUtilityFacts.cs
BrennanConroy: Isn't that what all the tests are doing?

src/Microsoft.Dnx.Runtime/NuGet/Utility/PathUtility.cs
moozzyk: This was my point. You should throw in either case.

src/Microsoft.Dnx.Runtime/NuGet/Utility/PathUtility.cs
```diff
@@ -145,6 +211,10 @@ public static string GetPathWithBackSlashes(string path)
 
         public static string GetPathWithDirectorySeparator(string path)
         {
+            if (!RuntimeEnvironmentHelper.IsWindows)
```
davidfowl: This is incorrect. This helpers is used in other places to normalize slashes in paths.

src/Microsoft.Dnx.Runtime/NuGet/Utility/PathUtility.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.IO;
+using Microsoft.Dnx.Runtime;
```
davidfowl: Remove this

src/Microsoft.Dnx.Runtime/NuGet/Utility/PathUtility.cs
```diff
@@ -72,7 +72,72 @@ public static string GetRelativePath(string path1, string path2)
...
+                    break;
+                }
+                ++index;
+            }
+
+            var path = "";
+
+            // check if path2 ends with a non-directory separator and if path1 has the same non-directory at the end
+            if (len1 + 1 == len2 && !string.IsNullOrEmpty(path1Segments[index]) &&
+                string.Compare(path1Segments[index], path2Segments[index], compare) == 0)
```
davidfowl: string.Equals?

src/Microsoft.Dnx.Runtime/NuGet/Utility/PathUtility.cs
```diff
@@ -72,7 +72,72 @@ public static string GetRelativePath(string path1, string path2)
...
+            var path2Segments = path2.Split(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
+            // if path1 does not end with / it is assumed the end is not a directory
+            // we will assume that is isn't a directory by ignoring the last split
+            var len1 = path1Segments.Length - 1;
+            var len2 = path2Segments.Length;
+
+            var min = Math.Min(len1, len2);
+            while (min > index)
+            {
+                if (string.Compare(path1Segments[index], path2Segments[index], compare) != 0)
```
davidfowl: !string.Equals?

davidfowl:
:shipit:

src/Microsoft.Dnx.Runtime/NuGet/Utility/PathUtility.cs
```diff
@@ -72,7 +72,75 @@ public static string GetRelativePath(string path1, string path2)
         /// </summary>
         public static string GetRelativePath(string path1, string path2, char separator)
         {
-            return UriUtility.GetPath(GetRelativeUri(path1, path2), separator);
+            if (string.IsNullOrEmpty(path1))
+            {
+                throw new ArgumentException(nameof(path1));
```
moozzyk: The string you passed is the error message.... The second parameter is the argument name.

BrennanConroy:
046a0595d32357b7ef5c5c58fe877b2225cad8fb

PR #1692, opened 2015-04-22
Use IRuntimeEnvironment for PlatformHelper

BrennanConroy:
@victorhurdugaci @davidfowl


src/Microsoft.Framework.Runtime.Common/Impl/PlatformHelper.cs
```diff
@@ -2,18 +2,23 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.Framework.Runtime.Infrastructure;
 
 namespace Microsoft.Framework.Runtime
 {
     internal static class PlatformHelper
```
davidfowl: Noooo, get rid of this and stop using CallContextServiceLocator everywhere :smile:

BrennanConroy: What am I supposed to use instead D:<

src/Microsoft.Framework.Runtime.Common/Impl/PlatformHelper.cs
```diff
@@ -2,18 +2,23 @@
...
+                    .Locator
+                    .ServiceProvider
+                    .GetService(typeof(IRuntimeEnvironment))).OperatingSystem);
 
         public static bool IsMono
         {
             get
             {
-                return _isMono.Value;
+                return _osName.Value == "Darwin";
```
moozzyk: forgive my ignorance but how this is supposed to work when we have CoreClr running on non-windows OSes?

davidfowl: This check is wrong :smile:

src/Microsoft.Framework.Runtime.Common/Impl/PlatformHelper.cs
davidfowl: Use `IRuntimeEnvironment` in place of PlatformHelper

src/Microsoft.Framework.Runtime.Common/Impl/PlatformHelper.cs
BrennanConroy: Oh wow, I wasn't thinking D:

src/Microsoft.Framework.Runtime.Common/Impl/PlatformHelper.cs
BrennanConroy: IServiceProvider unfortunately isn't easily accessible to a lot of the locations that use PlatformHelper. For the places it is, something similar to ```if((((IRuntimeEnvironment)_services.GetService(typeof(IRuntimeEnvironment))).RuntimeType == "Mono"))``` will need to be used. And the places it isn't available we will need to use CallContextServiceLocator.

BrennanConroy:
@davidfowl @ChengTian 
Passing services everywhere

src/Microsoft.Framework.Runtime.Hosting/Dependencies/FrameworkReferenceResolver.cs
```diff
@@ -162,10 +164,10 @@ public static string GetReferenceAssembliesPath()
                     "Reference Assemblies", "Microsoft", "Framework");
         }
 
-        
         private FrameworkInformation GetFrameworkInformation(NuGetFramework targetFramework)
         {
-            string referenceAssembliesPath = GetReferenceAssembliesPath();
+            var isMono = ((IRuntimeEnvironment)_services.GetService(typeof(IRuntimeEnvironment))).RuntimeType == "Mono";
```
ChengTian: I see this several times. Can we encapsulate it with something like
```c#
SomeUtils.IsMono(IServiceProvider)
```

ChengTian: And `IsWindows` as well

src/Microsoft.Framework.PackageManager/Scripts/ScriptExecutor.cs
```diff
@@ -15,7 +15,7 @@ public class ScriptExecutor
     {
         private static readonly string ErrorMessageTemplate = "The '{0}' script failed with status code {1}.";
 
-        public bool Execute(Runtime.Project project, string scriptName, Func<string, string> getVariable)
+        public bool Execute(Runtime.Project project, string scriptName, Func<string, string> getVariable, bool isMono)
         {
```
ChengTian: Create a constructor taking `isMono` as a parameter.

src/Microsoft.Framework.PackageManager/Publish/PublishRuntime.cs
```diff
@@ -11,11 +11,13 @@ public class PublishRuntime
     {
         private readonly FrameworkName _frameworkName;
         private readonly string _runtimePath;
+        private readonly bool _isMono;
 
-        public PublishRuntime(PublishRoot root, FrameworkName frameworkName, string runtimePath)
+        public PublishRuntime(PublishRoot root, FrameworkName frameworkName, string runtimePath, bool isMono)
```
ChengTian: I guess it's better to have `PublishRoot.IsMono` instead of passing it in this way.

ChengTian:
Need rebase

test/Microsoft.Framework.ApplicationHost.FunctionalTests/AppHostTests.cs
```diff
@@ -135,7 +135,7 @@ public void AppHostShowsErrorWhenCurrentTargetFrameworkWasNotFoundInProjectJson(
             var runtimeTargetFramework = new FrameworkName(runtimeTargetFrameworkString);
             var runtimeTargetFrameworkShortName = VersionUtility.GetShortFrameworkName(runtimeTargetFramework);
             var runtimeType = flavor == "coreclr" ? "CoreCLR" : "CLR";
-            runtimeType = PlatformHelper.IsMono ? "Mono" : runtimeType;
+            runtimeType = TestUtils.CurrentRuntimeEnvironment.RuntimeType == "Mono" ? "Mono" : runtimeType;
```
BrennanConroy: Missed one

test/Microsoft.Framework.CommonTestUtils/TestUtils.cs
```diff
@@ -157,8 +157,8 @@ public static DisposableDir GetRuntimeHomeDir(string flavor, string os, string a
             runtimeHomePath = CreateTempDir();
             runtimeName = Path.GetFileNameWithoutExtension(runtimeNupkg);
             var runtimeRoot = Path.Combine(runtimeHomePath, "runtimes", runtimeName);
-
-            if (!PlatformHelper.IsMono)
+
```
BrennanConroy: Remove space

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuWrapTests.cs
```diff
@@ -276,7 +280,7 @@ public void DnuWrapCreatesGlobalJsonIfSolutionRootDoesNotHaveOne(string flavor,
         {
             var runtimeHomeDir = TestUtils.GetRuntimeHomeDir(flavor, os, architecture);
 
-            if (PlatformHelper.IsMono)
+            if (TestUtils.CurrentRuntimeEnvironment.RuntimeType == "Mono")
```
BrennanConroy: Missed another

test/Microsoft.Framework.PackageManager.FunctionalTests/Microsoft.Framework.PackageManager.FunctionalTests.xproj
```diff
@@ -13,5 +13,8 @@
   <PropertyGroup>
     <SchemaVersion>2.0</SchemaVersion>
   </PropertyGroup>
+  <ItemGroup>
```
BrennanConroy: Not sure, will remove

test/Microsoft.Framework.Runtime.Roslyn.Tests/ProjectExtensionsFacts.cs
```diff
@@ -176,7 +177,7 @@ public static CompilationSettings GetCompilationSettings(this Project project, s
         {
             var framework = FrameworkNameHelper.ParseFrameworkName(frameworkName);
             return project.GetCompilerOptions(framework, "Debug")
-                          .ToCompilationSettings(framework);
+                          .ToCompilationSettings(framework, CallContextServiceLocator.Locator.ServiceProvider);
```
BrennanConroy: Use common test utils

src/Microsoft.Framework.PackageManager/Install/AppCommandsFolderRepository.cs
```diff
@@ -66,7 +68,7 @@ public NuGet.PackageInfo FindCommandOwner(string command)
         public void Remove(string commandName)
         {
             string commandFileName = commandName +
-                (PlatformHelper.IsWindows ? ".cmd" : string.Empty);
+                (_isWindows ? ".cmd" : string.Empty);
```
moozzyk: nit: use var (same below)

src/Microsoft.Framework.PackageManager/Install/AppCommandsFolderRepository.cs
```diff
@@ -76,7 +78,7 @@ public void Load()
         {
             _commands = new Dictionary<string, NuGet.PackageInfo>();
 
-            string pathFilter = PlatformHelper.IsWindows ? "*.cmd" : "*.*";
+            string pathFilter = _isWindows ? "*.cmd" : "*.*";
```
ChengTian: Let's do some cleanup by the way. Use var here.

src/Microsoft.Framework.PackageManager/Program.cs
```diff
@@ -412,7 +412,7 @@ public int Main(string[] args)
                     }
 
                     var command = new DependencyListCommand(options, _environment.RuntimeFramework);
-                    return command.Execute();
+                    return command.Execute(_hostServices);
```
ChengTian: I suggest we follow one consistent pattern in this file. That is, for class `XYZCommand`, we pass environment info/services it needs to its ctor and `XYZCommand.Execute()` should take no parameter.

src/Microsoft.Framework.PackageManager/Program.cs
```diff
@@ -518,7 +519,7 @@ public int Main(string[] args)
                     command.InPlace = optInPlace.HasValue();
                     command.Framework = optFramework.Value();
 
-                    var success = command.ExecuteCommand();
+                    var success = command.ExecuteCommand(RuntimeEnvironmentHelper.IsMono(_runtimeEnv));
```
ChengTian: Same here. Runtime info should be passed to ctor and ExecuteCommand() should take no parameter.

Please help do some cleanup BTW, let's rename this `WrapCommand.ExecuteCommand()` to `WrapCommand.Execute()` to remove redundancy in the name.

src/Microsoft.Framework.PackageManager/Utils/VersionUtils.cs
```diff
@@ -25,7 +26,7 @@ private static string GetActiveRuntimeName()
 
             if (!string.IsNullOrEmpty(pathVariable))
             {
-                string dnuExecutable = PlatformHelper.IsWindows ? "dnu.cmd" : "dnu";
+                string dnuExecutable = RuntimeEnvironmentHelper.IsWindows(CallContextServiceLocator.Locator.ServiceProvider) ? "dnu.cmd" : "dnu";
```
ChengTian: If we can do this, why do we want to pass `IServiceProvider/IRuntimeEnvironment` around? If we have a global instance of service provider, can we just have a property `RuntimeEnvironmentHelper.IsWindows`, in which the boolean is lazily evaluated with this global service provider?

BrennanConroy:
After talking with David, we would like to do something similar to what Platformhelper did, so my proposed solution is:
```
internal static class RuntimeEnvironmentHelper
{
    private static Lazy<bool> _isMono = new Lazy<bool>(() => ((IRuntimeEnvironment)Services.GetService(typeof(IRuntimeEnvironment))).RuntimeType == "Mono");
    private static Lazy<bool> _isWindows = new Lazy<bool>(() => ((IRuntimeEnvironment)Services.GetService(typeof(IRuntimeEnvironment))).RuntimeType == "Mono");

    public static IServiceProvider Services { private get; set; }

    public static bool IsWindows
    {
        get
        {
            if(Services == null)
            {
                throw new NullReferenceException("Services hasn't been assigned yet");
            }

            return _isWindows.Value;
        }
    }

    public static bool IsMono
    {
        get
        {
            if (Services == null)
            {
                throw new NullReferenceException("Services hasn't been assigned yet");
            }

            return _isMono.Value;
        }
    }
}
```

We will need to initialize the Services at the beginning, and then we can use the Helper just like we did PlatformHelper before, which will keep the code a lot cleaner.

ChengTian:
:+1:

ChengTian:
How do you initialize Services? Make `RuntimeEnvironmentHelper.Services` lazily initialized from `CallContextServiceLocator.Locator.ServiceProvider`?

BrennanConroy:
Oh ya, that's a good idea

src/Microsoft.Framework.Runtime.Sources/Impl/RuntimeEnvironmentHelper.cs
```diff
@@ -0,0 +1,31 @@
...
+using System;
+
+namespace Microsoft.Framework.Runtime
+{
+    internal static class RuntimeEnvironmentHelper
+    {
+        private static Lazy<bool> _isMono = new Lazy<bool>(() => ((IRuntimeEnvironment)Services.Value.GetService(typeof(IRuntimeEnvironment))).RuntimeType == "Mono");
+        private static Lazy<bool> _isWindows = new Lazy<bool>(() => ((IRuntimeEnvironment)Services.Value.GetService(typeof(IRuntimeEnvironment))).OperatingSystem == "Windows");
+
+        public static Lazy<IServiceProvider> Services = new Lazy<IServiceProvider>(() => Infrastructure.CallContextServiceLocator.Locator.ServiceProvider);
```
BrennanConroy: Change to private

src/Microsoft.Framework.PackageManager/Program.cs
```diff
@@ -529,7 +527,7 @@ public int Main(string[] args)
 
         private Reports CreateReports(bool verbose, bool quiet)
         {
-            var useConsoleColor = _runtimeEnv.OperatingSystem == "Windows";
+            var useConsoleColor = false;
```
ChengTian: Why? Shouldn't it be
```c#
var useConsoleColor = RuntimeEnvironmentHelper.IsWindows
```

BrennanConroy: Oh ya, well the AnsiConsole will be changed in another PR before this goes out, so I didn't change this.

src/Microsoft.Framework.Runtime.Roslyn/RoslynProjectReference.cs
```diff
@@ -270,7 +270,7 @@ private static bool SupportsPdbGeneration()
         {
             try
             {
-                if (PlatformHelper.IsMono)
```
BrennanConroy: Should this be !IsWindows?

victorhurdugaci: Sure, unless the CLR team has a solution to generate PDBs on non windows

moozzyk: nit: I don't know if it makes a lot of difference but why this check is inside the try/catch block?

src/Microsoft.Framework.PackageManager/Commands/WrapCommand.cs
```diff
@@ -30,7 +30,7 @@ public class WrapCommand
         public string Framework { get; set; }
         public Reports Reports { get; set; }
 
-        public bool ExecuteCommand()
+        public bool Execute()
```
troydai: The change of the method name seems not to be related to this PR.

src/Microsoft.Framework.Runtime.Sources/Impl/RuntimeEnvironmentHelper.cs
```diff
@@ -0,0 +1,31 @@
...
+        private static Lazy<bool> _isWindows = new Lazy<bool>(() => ((IRuntimeEnvironment)Services.Value.GetService(typeof(IRuntimeEnvironment))).OperatingSystem == "Windows");
+
+        private static Lazy<IServiceProvider> Services = new Lazy<IServiceProvider>(() => Infrastructure.CallContextServiceLocator.Locator.ServiceProvider);
+
+        public static bool IsWindows
+        {
+            get
+            {
+                return _isWindows.Value;
+            }
```
troydai: Put these in one line to save space.

src/Microsoft.Framework.Runtime.Sources/Impl/RuntimeEnvironmentHelper.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Framework.Runtime
+{
+    internal static class RuntimeEnvironmentHelper
+    {
+        private static Lazy<bool> _isMono = new Lazy<bool>(() => ((IRuntimeEnvironment)Services.Value.GetService(typeof(IRuntimeEnvironment))).RuntimeType == "Mono");
```
troydai: This line is too long. Break it.

src/Microsoft.Framework.Runtime.Sources/Impl/RuntimeEnvironmentHelper.cs
```diff
@@ -0,0 +1,31 @@
...
+        private static Lazy<bool> _isWindows = new Lazy<bool>(() => ((IRuntimeEnvironment)Services.Value.GetService(typeof(IRuntimeEnvironment))).OperatingSystem == "Windows");
```
troydai: Break the line.

src/Microsoft.Framework.Runtime.Sources/Impl/RuntimeEnvironmentHelper.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        private static Lazy<IServiceProvider> Services = new Lazy<IServiceProvider>(() => Infrastructure.CallContextServiceLocator.Locator.ServiceProvider);
```
troydai: Break the line.

troydai:
:watch:

src/Microsoft.Framework.PackageManager/Commands/WrapCommand.cs
ChengTian: We can keep this. In previous approach, @BrennanConroy 's changes conflict with mine so I asked him to make this change. All other `XYZCommand.ExecuteCommand()` has been renamed to `XYZCommand.Execute()` by me. This is the only one left.

test/Microsoft.Framework.PackageManager.FunctionalTests/Microsoft.Framework.PackageManager.FunctionalTests.xproj
```diff
@@ -13,8 +13,5 @@
   <PropertyGroup>
     <SchemaVersion>2.0</SchemaVersion>
   </PropertyGroup>
-  <ItemGroup>
-    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
-  </ItemGroup>
```
dougbu: Don't remove this. It's a test service reference and helps VS load the solution faster.

src/Microsoft.Framework.PackageManager/Utils/FileOperationUtils.cs
```diff
@@ -38,12 +37,6 @@ private static void DeleteRecursive(string deletePath)
 
         public static bool MarkExecutable(string file)
         {
-            if (PlatformHelper.IsWindows)
```
dougbu: Why attempt to execute `chmod` on Windows?

src/Microsoft.Framework.PackageManager/Commands/WrapCommand.cs
```diff
@@ -47,7 +47,7 @@ public bool ExecuteCommand()
             var extension = Path.GetExtension(InputFilePath);
             if (string.Equals(extension, ".csproj", StringComparison.OrdinalIgnoreCase))
             {
-                if (PlatformHelper.IsMono)
+                if (!RuntimeEnvironmentHelper.IsWindows)
```
dougbu: Curious: What are there rules by which you've chosen `IsMono` -> `IsMono` in some places and `IsMono` -> `!IsWindows` in others?

BrennanConroy: IsMono was used incorrectly in most places because we were only doing stuff on Windows and on Mono, now that we are getting cross-plat CoreCLR support in we need to fix all of these incorrect IsMono's

ChengTian: Yeah, previously without the utils added by this PR, we have no way to tell the OS we are running on. So we assumed `IsMono == non-Windows`. Now we are fixing this.

src/Microsoft.Framework.PackageManager/Utils/FileOperationUtils.cs
BrennanConroy: I removed this a while ago during an earlier iteration because I noticed everytime we call this function we already did the check for Windows. Would it be better to keep the check in here and remove it from the places we call it?

ChengTian: Sounds good. This is a shared util method and putting OS check in it makes code DRYer.

dougbu: Definitely safer and results in less code duplication. I'd also have to change #1931 if this went in first :smiling_imp:

src/Microsoft.Framework.PackageManager/Commands/WrapCommand.cs
dougbu: thx

src/Microsoft.Framework.PackageManager/Utils/FileOperationUtils.cs
moozzyk: Add an assert.

ChengTian:
Please address the last few comments and :shipit:

BrennanConroy:
We can't merge this until we use a version of mono that doesn't null-ref

src/Microsoft.Framework.Runtime.Sources/Impl/RuntimeEnvironmentHelper.cs
ChengTian: `_services`?

src/dnx.host/Bootstrapper.cs
```diff
@@ -76,6 +76,7 @@ public Task<int> RunAsync(List<string> args, IRuntimeEnvironment env)
                 serviceProvider.Add(typeof(IRuntimeEnvironment), env);
 
                 CallContextServiceLocator.Locator.ServiceProvider = serviceProvider;
+                var mono = RuntimeEnvironmentHelper.IsMono;
```
ChengTian: Why are we adding this? Seems we don't use this newly added variable

BrennanConroy: I should add a comment. This is my proposed temporary fix for mono null-ref. If we grab the value here it wont fail later.

davidfowl: Add the comment and add an initialize method that sets it up once instead of all the lazy crazyness

src/Microsoft.Framework.PackageManager/Commands/WrapCommand.cs
davidfowl: I'm not sure this one is corect

src/Microsoft.Framework.Runtime/FrameworkReferenceResolver.cs
```diff
@@ -185,7 +185,8 @@ private static FrameworkInformation GetFrameworkInformation(FrameworkName target
 
             // Skip this on mono since it has a slightly different set of reference assemblies at a different
             // location
-            if (!PlatformHelper.IsMono && FrameworkDefinitions.TryPopulateFrameworkFastPath(targetFramework.Identifier, targetFramework.Version, referenceAssembliesPath, out frameworkInfo))
+            var isMono = RuntimeEnvironmentHelper.IsMono;
```
davidfowl: Why did you declare a variable?

src/Microsoft.Framework.PackageManager/Commands/WrapCommand.cs
BrennanConroy: Ok, who is the owner of Wrap?

BrennanConroy: Wei said this is correct because wrap calls MSBuild

davidfowl: MSBuild exists on windows and since it shells out it will work with CoreCLR on windows.

BrennanConroy: Ok? How is this wrong though

BrennanConroy:
Looks like Mono has fixed the Execution Context issue in the released bits, will run some more to be sure.

src/Microsoft.Framework.PackageManager/Publish/PublishRuntime.cs
```diff
@@ -43,7 +43,7 @@ public bool Emit(PublishRoot root)
 
             new PublishOperations().Copy(_runtimePath, TargetPath);
 
-            if (PlatformHelper.IsMono)
+            if (!RuntimeEnvironmentHelper.IsWindows)
```
ChengTian: Do we want something like `RuntimeEnvironmentHelper.IsUnixLike` (which is `IsLinux || IsDarwin`)?

test/Microsoft.Framework.CommonTestUtils/TestUtils.cs
```diff
@@ -403,7 +403,7 @@ private static bool IsWindows()
             var p = (int)Environment.OSVersion.Platform;
             return (p != 4) && (p != 6) && (p != 128);
 #else
-            return PlatformHelper.IsWindows;
+            return RuntimeEnvironmentHelper.IsWindows;
```
davidfowl: Use the runtime helper for everything, not just for DNXCORE50

src/Microsoft.Framework.Runtime/DependencyManagement/GacDependencyResolver.cs
```diff
@@ -17,7 +17,7 @@ public class GacDependencyResolver : IDependencyProvider, ILibraryExportProvider
 
         public IEnumerable<string> GetAttemptedPaths(FrameworkName targetFramework)
         {
-            if (PlatformHelper.IsMono)
+            if (!RuntimeEnvironmentHelper.IsWindows)
```
davidfowl: Add the mono check as well (windows on mono should fail)

src/Microsoft.Framework.Runtime.Sources/Impl/RuntimeEnvironmentHelper.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    internal static class RuntimeEnvironmentHelper
+    {
+        private static Lazy<bool> _isMono = new Lazy<bool>(() =>
+            ((IRuntimeEnvironment)_services.Value.GetService(typeof(IRuntimeEnvironment))).RuntimeType == "Mono");
+
+        private static Lazy<bool> _isWindows = new Lazy<bool>(() =>
+            ((IRuntimeEnvironment)_services.Value.GetService(typeof(IRuntimeEnvironment))).OperatingSystem == "Windows");
+
+        private static Lazy<IServiceProvider> _services = new Lazy<IServiceProvider>(() =>
```
davidfowl: Make this `Lazy<IRuntimeEnvironment>`

src/Microsoft.Framework.PackageManager/Publish/PublishRuntime.cs
davidfowl: That's interesting. We should think about that.

davidfowl:
:shipit: After those few comments. Make sure that older versions of mono don't start null reffing all over the place because of a broken call context. This change needs to work without that fix.

muratg:
When can we get this in?  @ChengTian's fix is waiting on this one.

BrennanConroy:
@davidfowl 
Any more comments? Or ship?

davidfowl:
@BrennanConroy I assume you know why appveyor is failing and that this change didn't actually regress anything.

If that's the case then :shipit:

BrennanConroy:
Ya, appveyor failed because of the System.Reflection.Primitives not being found.

kichalla:
Several other repos have their own copy of `PlatformHelper`. So is this the new recommended way of checking for platform? @BrennanConroy 

https://github.com/search?q=user%3Aaspnet+platformhelper&ref=searchresults&type=Code&utf8=%E2%9C%93

cc: @bricelam

victorhurdugaci:
@kichalla no. you should use `IApplicationEnvironment`

PR #1694, opened 2015-04-23
Make sure we use newly built runtime to generate lock file for tests

ChengTian:
@davidfowl 

Tested locally with experimental lock file format change (bumping version up by 1)

### DIFF
--- a/test/Bootstrapper.FunctionalTests/BootstrapperTests.cs
+++ b/test/Bootstrapper.FunctionalTests/BootstrapperTests.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.IO;
 using Microsoft.Framework.CommonTestUtils;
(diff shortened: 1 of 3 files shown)


ChengTian:
Each sample-app-based test should have a copy of the sample app because, when parallel testing is turned on, we don't want multiple `dnu restore` writing to one lock file.

davidfowl:
When the ci machine is happy

ChengTian:
@davidfowl , Happy CI X 3

test/Bootstrapper.FunctionalTests/BootstrapperTests.cs
```diff
@@ -151,19 +157,25 @@ public void BootstrapperInvokesAssemblyWithInferredAppBaseAndLibPathOnClr(string
...
                 var exitCode = DnuTestUtils.ExecDnu(
                     runtimeHomeDir,
                     subcommand: "build",
-                    arguments: string.Format("{0} --configuration=Release --out {1}", sampleAppRoot, tempDir.DirPath));
+                    arguments: string.Format("{0} --configuration=Release --out {1}", sampleAppRoot, tempDir),
+                    stdOut: out stdOut,
+                    stdErr: out stdErr);
 
-                Assert.Equal(0, exitCode);
+                if (exitCode != 0)
```
victorhurdugaci: Do you want to continue running if this failed?

ChengTian: Yes, the failure will be caught by later assertion. We are not testing `dnu restore` here so we don't assert on the exit code of `dnu restore`

PR #1700, opened 2015-04-24
Include native resource files for CoreCLR Linux

ellismg:
We should include the compiled message files as part of the package so
we can put them in DNX.  These are used when the runtime throws an
exception internally and are useful for debugging (e.g. it makes
FileNotFound exceptions actually have the text of the file that was not
found).

They are not produced for OSX yet (see dotnet/coreclr#678) so we can
only include the Linux versions.


ellismg:
@Eilon @davidfowl

BrennanConroy:
Yes please!

Eilon:
:shipit:

PR #1702, opened 2015-04-24
DRY

moozzyk:
Debug|Win32:   DRY
Release|Win32: DRY
Debug|x64:     DRY
Release|x64:   DRY

Reorganizing dnx.vcxproj file to avoid duplication. Seperating common compile and link settings so that we can later use them in other native projects in the dnx repo and therefore make sure that we build all native projects with the same settings. These changes should also make it possible to run API Scan on native binaries.


BrennanConroy:
Looks good

muratg:
Looks good to me too, but AppVeyor seems to be failing? It seems to be passing with other branches, so I wonder if these changes is the cause of the failure.

moozzyk:
I saw that but it seems that AppVeyor apparently hung on cloning the repo. The last line in the log is:
`git clone -q --depth=1 --branch=pawelka/vcxproj git://github.com/aspnet/dnx.git C:\projects\dnx`
Will try to trigger the build again.

PR #1711, opened 2015-04-25
Add a few more platform specific DLLs

ellismg:
The Linux and Darwin implementation of these DLLs differ from Windows.
Also add a missing native component that was being excluded because of
its name, which is used by parts of the crypto stack.


davidfowl:
:shipit:

Issue #1708, opened 2015-04-24
Don't use console color with CoreCLR on Linux/Darwin


ChengTian:
Because CoreCLR temporarily doesn't support console color on those platforms.

We should bring tests back on those platforms when this issue is done: https://github.com/aspnet/dnx/commit/7b297209f2031a0343a346719345e7db1eac0bfd

ellismg:
There's no way to get the current foreground or background color of the terminal in *Nix.  We are hesitant to try to lie because given the default settings for different terminals, there's no good initial value we can use.

Changing the color and resetting both work, it's just getting the current color that throws PlatformNotSupported exception.  I do not expect this will change.

PR #1724, opened 2015-04-28
Remove color parsing from Coreclr on Linux and Mac

BrennanConroy:
#1708

### DIFF
--- a/src/Microsoft.Framework.CommandLineUtils/CommandLine/AnsiConsole.cs
+++ b/src/Microsoft.Framework.CommandLineUtils/CommandLine/AnsiConsole.cs
@@ -9,22 +9,40 @@ namespace Microsoft.Framework.Runtime.Common.CommandLine
 {
     internal class AnsiConsole
     {
-        private AnsiConsole(TextWriter writer)
+        private AnsiConsole(TextWriter writer, IRuntimeEnvironment runtimeEnv)
         {
             Writer = writer;
-            OriginalForegroundColor = Console.ForegroundColor;
+
+            var useColors = runtimeEnv.OperatingSystem == "Windows" || runtimeEnv.RuntimeType == "Mono";
+
+            if (useColors)
+            {
+                WriteLine = WriteLineParse;
+                OriginalForegroundColor = Console.ForegroundColor;
+            }
+            else
+            {
+                WriteLine = WriteLineNoParse;
+            }
         }
 
+        public Action<string> WriteLine { get; private set; }
         private int _boldRecursion;
 
-        public static AnsiConsole Output = new AnsiConsole(Console.Out);
+        public static AnsiConsole GetOutput(IRuntimeEnvironment runtimeEnv)
+        {
+            return new AnsiConsole(Console.Out, runtimeEnv);
+        }
 
-        public static AnsiConsole Error = new AnsiConsole(Console.Error);
+        public static AnsiConsole GetError(IRuntimeEnvironment runtimeEnv)
+        {
+            return new AnsiConsole(Console.Error, runtimeEnv);
+        }
 
         public TextWriter Writer { get; }
 
         public ConsoleColor OriginalForegroundColor { get; }
-        
+
         private void SetColor(ConsoleColor color)
         {
             Console.ForegroundColor = (ConsoleColor)(((int)Console.ForegroundColor & 0x08) | ((int)color & 0x07));
@@ -41,7 +59,12 @@ namespace Microsoft.Framework.Runtime.Common.CommandLine
             Console.ForegroundColor = (ConsoleColor)((int)Console.ForegroundColor ^ 0x08);
         }
 
-        public void WriteLine(string message)
+        private void WriteLineNoParse(string message)
+        {
+            Writer.WriteLine(message);
+        }
+
+        private void WriteLineParse(string message)
         {
             var sb = new StringBuilder();
             var escapeScan = 0;
--- a/src/Microsoft.Framework.CommandLineUtils/CommandLine/AnsiConsole.cs
+++ b/src/Microsoft.Framework.CommandLineUtils/CommandLine/AnsiConsole.cs
@@ -9,34 +9,28 @@ namespace Microsoft.Framework.Runtime.Common.CommandLine
 {
     internal class AnsiConsole
     {
-        private AnsiConsole(TextWriter writer, IRuntimeEnvironment runtimeEnv)
+        private AnsiConsole(TextWriter writer, bool useConsoleColor)
         {
             Writer = writer;
 
-            var useColors = runtimeEnv.OperatingSystem == "Windows" || runtimeEnv.RuntimeType == "Mono";
-
-            if (useColors)
+            _useConsoleColor = useConsoleColor;
+            if (_useConsoleColor)
             {
-                WriteLine = WriteLineParse;
                 OriginalForegroundColor = Console.ForegroundColor;
             }
-            else
-            {
-                WriteLine = WriteLineNoParse;
-            }
         }
 
-        public Action<string> WriteLine { get; private set; }
         private int _boldRecursion;
+        private bool _useConsoleColor;
 
-        public static AnsiConsole GetOutput(IRuntimeEnvironment runtimeEnv)
+        public static AnsiConsole GetOutput(bool useConsoleColor)
         {
-            return new AnsiConsole(Console.Out, runtimeEnv);
+            return new AnsiConsole(Console.Out, useConsoleColor);
         }
 
-        public static AnsiConsole GetError(IRuntimeEnvironment runtimeEnv)
+        public static AnsiConsole GetError(bool useConsoleColor)
         {
-            return new AnsiConsole(Console.Error, runtimeEnv);
+            return new AnsiConsole(Console.Error, useConsoleColor);
         }
 
         public TextWriter Writer { get; }
@@ -59,13 +53,14 @@ namespace Microsoft.Framework.Runtime.Common.CommandLine
             Console.ForegroundColor = (ConsoleColor)((int)Console.ForegroundColor ^ 0x08);
         }
 
-        private void WriteLineNoParse(string message)
+        public void WriteLine(string message)
         {
-            Writer.WriteLine(message);
-        }
+            if (!_useConsoleColor)
+            {
+                Writer.WriteLine(message);
+                return;
+            }
 
-        private void WriteLineParse(string message)
-        {
             var sb = new StringBuilder();
             var escapeScan = 0;
             for (; ;)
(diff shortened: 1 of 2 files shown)


src/Microsoft.Framework.CommandLineUtils/CommandLine/AnsiConsole.cs
```diff
@@ -9,22 +9,52 @@ namespace Microsoft.Framework.Runtime.Common.CommandLine
...
-        private AnsiConsole(TextWriter writer)
+        private AnsiConsole(TextWriter writer, IServiceProvider services)
         {
             Writer = writer;
-            OriginalForegroundColor = Console.ForegroundColor;
+
+            var runtimeEnv = (IRuntimeEnvironment)services.GetService(typeof(IRuntimeEnvironment));
+            var shouldParse = runtimeEnv.OperatingSystem == "Windows" || runtimeEnv.RuntimeType == "Mono";
+
+            if (shouldParse)
```
victorhurdugaci: I don't think `parse` is the right term here. Maybe `useColors` or something on those lines would be more appropriate

src/Microsoft.Framework.CommandLineUtils/CommandLine/AnsiConsole.cs
```diff
@@ -9,22 +9,52 @@ namespace Microsoft.Framework.Runtime.Common.CommandLine
...
+            else
+            {
+                WriteLine = new WriteDelegate(WriteLineNoParse);
+            }
         }
 
         private int _boldRecursion;
+        public delegate void WriteDelegate(string message);
+        public WriteDelegate WriteLine;
+        static private AnsiConsole Output = null;
```
victorhurdugaci: visibility modifier first: `private static` not `static private`

src/Microsoft.Framework.CommandLineUtils/CommandLine/AnsiConsole.cs
```diff
@@ -9,22 +9,52 @@ namespace Microsoft.Framework.Runtime.Common.CommandLine
 {
     internal class AnsiConsole
     {
-        private AnsiConsole(TextWriter writer)
+        private AnsiConsole(TextWriter writer, IServiceProvider services)
```
victorhurdugaci: Can you inject directly the `IRuntimeEnvironment`?

src/Microsoft.Framework.CommandLineUtils/CommandLine/AnsiConsole.cs
```diff
@@ -9,22 +9,52 @@ namespace Microsoft.Framework.Runtime.Common.CommandLine
...
+                OriginalForegroundColor = Console.ForegroundColor;
+            }
+            else
+            {
+                WriteLine = new WriteDelegate(WriteLineNoParse);
+            }
         }
 
         private int _boldRecursion;
+        public delegate void WriteDelegate(string message);
```
victorhurdugaci: Use an `Action<string>` instead?

src/Microsoft.Framework.CommandLineUtils/CommandLine/AnsiConsole.cs
BrennanConroy: How? We only create it in the static methods below

victorhurdugaci:
Looks good overall, except the places where I commented. :shipit:

src/Microsoft.Framework.CommandLineUtils/CommandLine/AnsiConsole.cs
victorhurdugaci: Inject it in program.cs and pass it down the stream?

BrennanConroy: Oh, ya, we can do IRuntimeEnvironment instead of IServiceProvider

src/Microsoft.Framework.CommandLineUtils/CommandLine/AnsiConsole.cs
Eilon: And don't initialize fields to their default values.

BrennanConroy:
@victorhurdugaci 
Made changes, take a quick look?

src/Microsoft.Framework.CommandLineUtils/CommandLine/AnsiConsole.cs
```diff
@@ -9,22 +9,50 @@ namespace Microsoft.Framework.Runtime.Common.CommandLine
...
+            }
+            else
+            {
+                WriteLine = WriteLineNoParse;
+            }
         }
 
         private int _boldRecursion;
+        public Action<string> WriteLine;
+        private static AnsiConsole Output;
```
victorhurdugaci: Naming convention for fields `_camelCaseName'

src/Microsoft.Framework.CommandLineUtils/CommandLine/AnsiConsole.cs
```diff
@@ -9,22 +9,50 @@ namespace Microsoft.Framework.Runtime.Common.CommandLine
...
+                OriginalForegroundColor = Console.ForegroundColor;
+            }
+            else
+            {
+                WriteLine = WriteLineNoParse;
+            }
         }
 
         private int _boldRecursion;
+        public Action<string> WriteLine;
```
victorhurdugaci: Convestion: fields are private. Use a property if you need them outside of this class

src/Microsoft.Framework.CommandLineUtils/CommandLine/AnsiConsole.cs
```diff
@@ -9,22 +9,50 @@ namespace Microsoft.Framework.Runtime.Common.CommandLine
...
+                WriteLine = WriteLineParse;
+                OriginalForegroundColor = Console.ForegroundColor;
+            }
+            else
+            {
+                WriteLine = WriteLineNoParse;
+            }
         }
 
         private int _boldRecursion;
```
victorhurdugaci: Not enforced but it is pretty common to structure a class like this:

- Fields
- Constructors
- Properties
- Methods

victorhurdugaci:
looks good. :shipit:

src/Microsoft.Framework.CommandLineUtils/CommandLine/AnsiConsole.cs
```diff
@@ -9,22 +9,50 @@ namespace Microsoft.Framework.Runtime.Common.CommandLine
...
+            }
+            else
+            {
+                WriteLine = WriteLineNoParse;
+            }
         }
 
+        public Action<string> WriteLine { get; private set; }
         private int _boldRecursion;
+        private static AnsiConsole _output;
```
moozzyk: Designwise, I don't think we should cache `_output` and `_error` here. These are static field and the value depends on `runtimeEnv`. It will probably work fine now if `runtimeEnv` does not change but if for whatever reason `GetOutput()` was called multiple times with different `runtimeEnv` value the returned value for the second call would be wrong because it would use runtimeEnv from the first call. I think the `GetOutput` and `GetError` should just create and return the value and the code that invokes these methods should take care of caching the returned instance if it makes sense.

src/Microsoft.Framework.PackageManager/Program.cs
```diff
@@ -520,12 +522,12 @@ public int Main(string[] args)
...
         private Reports CreateReports(bool verbose, bool quiet)
         {
-            IReport output = new Report(AnsiConsole.Output);
+            IReport output = new Report(AnsiConsole.GetOutput(_runtimeEnv));
             var reports = new Reports()
             {
                 Information = output,
                 Verbose = verbose ? output : new NullReport(),
-                Error = new Report(AnsiConsole.Output),
+                Error = new Report(AnsiConsole.GetError(_runtimeEnv)),
```
moozzyk: nit: move `Quiet` from the line below to the initializer and then you would be able to return `reports` directly (RVO :smile:)

BrennanConroy:
@davidfowl 
Is this more like what you want?

src/Microsoft.Framework.CommandLineUtils/CommandLine/AnsiConsole.cs
```diff
@@ -43,6 +55,12 @@ private void SetBold(bool bold)
 
         public void WriteLine(string message)
         {
+            if(!_useColors)
```
troydai: Format: space after `if`

troydai:
:shipit: after format the code.

moozzyk:
Agree with @davidfowl  that `useAnsiCodes` sounds weird but otherwise :sheep::it:

BrennanConroy:
He suggested ```useAnsiCodes``` >.>

davidfowl:
Why is the build failing?

src/Microsoft.Framework.CommandLineUtils/CommandLine/AnsiConsole.cs
```diff
@@ -9,22 +9,34 @@ namespace Microsoft.Framework.Runtime.Common.CommandLine
 {
     internal class AnsiConsole
     {
-        private AnsiConsole(TextWriter writer)
+        private AnsiConsole(TextWriter writer, bool useAnsiCodes)
```
glennc: This name doesn't feel right. You are using ansiCodes either way, just processing them differently. It is more like ansiCodePassthrough, but given that nobody knows what that means and this is internal it doesn't worry me too much.

src/Microsoft.Framework.CommandLineUtils/CommandLine/AnsiConsole.cs
```diff
@@ -9,22 +9,34 @@ namespace Microsoft.Framework.Runtime.Common.CommandLine
...
+            {
+                OriginalForegroundColor = Console.ForegroundColor;
+            }
         }
 
         private int _boldRecursion;
+        private bool _useAnsiCodes;
 
-        public static AnsiConsole Output = new AnsiConsole(Console.Out);
+        public static AnsiConsole GetOutput(bool useAnsiCodes)
```
glennc: This changes the behaviour of this does it not? Previously everyone who called this would've gotten the same instance. Now they will all get a different one. Not sure it matters, but it is a thing. @davidfowl?

src/Microsoft.Framework.PackageManager/Program.cs
```diff
@@ -520,12 +522,14 @@ public int Main(string[] args)
 
         private Reports CreateReports(bool verbose, bool quiet)
         {
-            IReport output = new Report(AnsiConsole.Output);
+            var useAnsiCodes = _runtimeEnv.OperatingSystem == "Windows";
```
davidfowl: Ok we need to rename the bool. useConsoleColor

davidfowl:
:shipit:

PR #1725, opened 2015-04-28
"build dev-install" overwrites existing symbol links

ChengTian:
This makes it convenient to switch between different local dnx repos

@davidfowl @victorhurdugaci

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -1138,11 +1138,11 @@ functions @{
         var sourceRuntime= Path.GetFileName(sourceDir);
         var destinationRuntime= sourceRuntime+ ".1.0.0-dev";
         var destinationDir = Path.Combine(FindRuntimePackagesFolder(), destinationRuntime);
-    
+
+        // Overwrite existing symbol link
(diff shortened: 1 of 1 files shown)


makefile.shade
```diff
@@ -1138,11 +1138,18 @@ functions @{
...
         {
-            // Nothing to do, just return
-            return;
+            if (IsMono)
+            {
+                Exec("rm", "-rf " + destinationDir);
+            }
+            else
+            {
+                Exec("cmd", "/C rmdir " + destinationDir);
```
victorhurdugaci: 1. Why don't you call `rmdir` directly?
2. Why not use `Directory.Delete` ?
3. Will `rm -rf` remove the contents of the symlink or just the symlink itself?

davidfowl: Use 2.

ChengTian:
Thanks for the feedback. Please check the updated patch.

victorhurdugaci:
Looks good. :shipit:

PR #1727, opened 2015-04-28
Really treating warnings as errors

moozzyk:


victorhurdugaci:
:shipit:

build/Dnx.Native.Settings
```diff
@@ -24,7 +24,7 @@
       <PrecompiledHeader>Use</PrecompiledHeader>
       <WarningLevel>Level4</WarningLevel>
       <PreprocessorDefinitions>_CONSOLE;%(PreprocessorDefinitions);$(Defines)</PreprocessorDefinitions>
-      <TreatWarningAsError Condition="'$(TreatWarningsAsErrors)' != ''">true</TreatWarningAsError>
+      <TreatWarningAsError Condition="'$(TreatWarningsAsErrors)' == ''">true</TreatWarningAsError>
```
Eilon: Kind of confused... isn't this now backwards?

1. Why would we enable TWAE only if the setting is *empty*?
2. Why not just force it to always be on? (That is, don't have a setting for it.)

moozzyk: It is empty by default. This allows you to override if you need for any reason (e.g. when hacking on it on your own box)

Eilon: Makes sense, thanks!

PR #1728, opened 2015-04-28
Undo removing EmitReferenceAssembly

troydai:
/cc @davidfowl @BillHiebert


src/Microsoft.Framework.Runtime.Interfaces/Compilation/IMetadataProjectReference.cs
```diff
@@ -16,6 +17,8 @@ public interface IMetadataProjectReference : IMetadataReference
 
         Assembly Load(IAssemblyLoadContext loadContext);
 
+	void EmitReferenceAssembly(Stream stream);
```
troydai: I'll replace the tab.

davidfowl:
Can you revert the change instead of doing this?

troydai:
Updated.

Issue #1654, opened 2015-04-15
wrap command should create global.json if it doesn't exist


glennc:
If running wrap and there is no global.json then we could create one. Possibly predicated on the existence of a .sln file.

Specific scenario is that i create a csproj, then try and add a web project to the solution and reference it. In this scenario there is no global.json and wrap fails.

PR #1738, opened 2015-04-29
dnu wrap also looks for sln when resolving root dir

ChengTian:
parent https://github.com/aspnet/dnx/issues/1654

Resolving root with sln file only happens in dnu wrap. It deals with the case where you first create a solution with a csproj and add a xproj to the solution later. In this case VS doesn't generate a global.json file and `dnu wrap` cannot locate correct root dir in order to create `wrap` folder.

@davidfowl @glennc

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Commands/WrapCommand.cs
+++ b/src/Microsoft.Framework.PackageManager/Commands/WrapCommand.cs
@@ -104,7 +104,7 @@ namespace Microsoft.Framework.PackageManager
                 return false;
             }
 
-            var rootDir = ProjectResolver.ResolveRootDirectory(Directory.GetCurrentDirectory());
+            var rootDir = ResolveRootDirectory(Directory.GetCurrentDirectory());
             var wrapRoot = Path.Combine(rootDir, WrapRootName);
             EmitAssemblyWrapper(wrapRoot, frameworkName, InputFilePath);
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Framework.PackageManager/Commands/WrapCommand.cs
```diff
@@ -692,7 +692,7 @@ private static string GetNuGetPackagePath(string packagesDir, string id, string
 
         private static string ResolvePackagesDirectory(string projectDir)
         {
-            var rootDir = ProjectResolver.ResolveRootDirectory(projectDir);
```
troydai: If `ProjectResolver.ResolveRootDirectory` is not used anywhere else, it can be deleted. (I don't know if it is used anywhere else.)

src/Microsoft.Framework.PackageManager/Commands/WrapCommand.cs
```diff
@@ -731,6 +731,33 @@ private static void SetPropertyValueIfEmpty(JObject obj, string name, string val
             }
         }
 
+        private static string ResolveRootDirectory(string projectPath)
```
troydai: This looks like common helper method. So should this actually be placed in `ProjectResolver`?

troydai:
Besides the questions, looks good to me. My suggestions should not block.

src/Microsoft.Framework.PackageManager/Commands/WrapCommand.cs
davidfowl: No this logic needs to be special for wrap.

src/Microsoft.Framework.PackageManager/Commands/WrapCommand.cs
ChengTian: It is used to locate the root of current solution (the ancestor folder containing `global.json`). This method is used widely in our runtime and functional tests.

muratg:
Looks good to me.  :shipit:

PR #1754, opened 2015-04-30
Fix "dnu wrap" help information

ChengTian:
`dnu wrap` can wrap dlls. Fix the help information to make users aware of this.

@davidfowl @muratg

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Program.cs
+++ b/src/Microsoft.Framework.PackageManager/Program.cs
@@ -483,9 +483,9 @@ namespace Microsoft.Framework.PackageManager
 
             app.Command("wrap", c =>
             {
-                c.Description = "Wrap a csproj into a project.json, which can be referenced by project.json files";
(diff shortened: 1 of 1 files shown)


src/Microsoft.Framework.PackageManager/Program.cs
```diff
@@ -481,9 +481,9 @@ public int Main(string[] args)
 
             app.Command("wrap", c =>
             {
-                c.Description = "Wrap a csproj into a project.json, which can be referenced by project.json files";
+                c.Description = "Wrap a csproj/dll into a project.json, which can be referenced by project.json files";
```
muratg: Does "assembly" make more sense than dll?

ChengTian: Yeah, "assembly" sounds more formal. Fixed

troydai:
:shipit:

muratg:
:shipit:

PR #1759, opened 2015-05-01
Restore performance improvements

davidfowl:
- Cache matches across contexts
- Reuse previous libraries for old lock files if sha didn't change
- Use lock file resolver as local input to speed up
already installed local package walk
- Don't use JToken.FromObject it's slow
- Don't wait for all tasks to complete if there's an
exact match that finished earlier.

/cc @anurse @PradeepKadubandi @pranavkm

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
@@ -228,11 +228,20 @@ namespace Microsoft.Framework.PackageManager
             var localProviders = new List<IWalkProvider>();
             var remoteProviders = new List<IWalkProvider>();
             var contexts = new List<RestoreContext>();
+            var cache = new Dictionary<LibraryRange, Task<WalkProviderMatch>>();
 
             projectProviders.Add(
                 new LocalWalkProvider(
                     new ProjectReferenceDependencyProvider(
                         projectResolver)));
+
+            if (lockFile != null)
+            {
+                var lockFileProvider = new NuGetDependencyResolver(new PackageRepository(packagesDirectory));
+                lockFileProvider.ApplyLockFile(lockFile);
+                localProviders.Add(
+                    new LocalWalkProvider(lockFileProvider));
+            }
 
             localProviders.Add(
                 new LocalWalkProvider(
@@ -257,6 +266,7 @@ namespace Microsoft.Framework.PackageManager
                     ProjectLibraryProviders = projectProviders,
                     LocalLibraryProviders = localProviders,
                     RemoteLibraryProviders = remoteProviders,
+                    MatchCache = cache
                 };
 
                 contexts.Add(context);
@@ -290,6 +300,7 @@ namespace Microsoft.Framework.PackageManager
                         ProjectLibraryProviders = projectProviders,
                         LocalLibraryProviders = localProviders,
                         RemoteLibraryProviders = remoteProviders,
+                        MatchCache = cache
                     };
                     contexts.Add(context);
                 }
@@ -302,6 +313,7 @@ namespace Microsoft.Framework.PackageManager
                         ProjectLibraryProviders = projectProviders,
                         LocalLibraryProviders = localProviders,
                         RemoteLibraryProviders = remoteProviders,
+                        MatchCache = cache
                     });
                 }
 
@@ -361,7 +373,8 @@ namespace Microsoft.Framework.PackageManager
                                 LocalLibraryProviders = pair.context.LocalLibraryProviders,
                                 RemoteLibraryProviders = pair.context.RemoteLibraryProviders,
                                 RuntimeName = runtimeName,
-                                RuntimeSpecs = runtimeSpecs
+                                RuntimeSpecs = runtimeSpecs,
+                                MatchCache = cache
                             };
                             var projectLibrary = new LibraryRange
                             {
@@ -449,7 +462,9 @@ namespace Microsoft.Framework.PackageManager
                 Reports.Information.WriteLine(string.Format("Writing lock file {0}", projectLockFilePath.White().Bold()));
 
                 var repository = new PackageRepository(packagesDirectory);
-                WriteLockFile(projectLockFilePath,
+
+                WriteLockFile(lockFile,
+                              projectLockFilePath,
                               project,
                               graphItems,
                               repository,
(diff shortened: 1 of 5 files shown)


src/Microsoft.Framework.PackageManager/Restore/RestoreOperations.cs
```diff
@@ -314,7 +324,26 @@ public Task<GraphItem> FindLibraryCached(RestoreContext context, LibraryRange li
             }
 
             WalkProviderMatch bestMatch = null;
-            var matches = await Task.WhenAll(tasks);
+            var matches = new List<WalkProviderMatch>();
+
+            // Short circuit if we find an exact match
+            while (tasks.Any())
+            {
+                var task = await Task.WhenAny(tasks);
```
pranavkm: Ahh this was the change you were speaking of. Could you also send a PR to NuGet with this change?

src/Microsoft.Framework.PackageManager/Utils/LockFileUtils.cs
```diff
@@ -24,34 +22,38 @@ public static LockFileLibrary CreateLockFileLibrary(IPackagePathResolver resolve
             // it has the correct casing that runtime needs during dependency resolution.
             lockFileLib.Name = correctedPackageName ?? package.Id;
             lockFileLib.Version = package.Version;
+            lockFileLib.Sha512 = File.ReadAllText(resolver.GetHashPath(package.Id, package.Version));
 
-            using (var nupkgStream = package.GetStream())
+            // If the shas are equal then do nothing
+            if (previousLibrary?.Sha512 == lockFileLib.Sha512)
```
pranavkm: `string.Equal(.., Ordinal)`

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
```diff
@@ -754,6 +747,34 @@ private Task<LockFile> ReadLockFile(string projectLockFilePath)
...
+                if (packageInfo == null)
+                {
+                    continue;
+                }
+
+                var package = packageInfo.Package;
+
+                LockFileLibrary previousLibrary = null;
+                previousLibraries?.TryGetValue(Tuple.Create(library.Name, library.Version), out previousLibrary);
+
```
pranavkm: Could you add an assert here saying `previousLibrary` is not expected to be null? Or Does `CreateLockFileLibrary` handle nulls?

src/Microsoft.Framework.PackageManager/Utils/LockFileUtils.cs
```diff
@@ -63,9 +65,7 @@ public static LockFileTargetLibrary CreateLockFileTargetLibrary(IPackage package
             // it has the correct casing that runtime needs during dependency resolution.
             lockFileLib.Name = correctedPackageName ?? package.Id;
             lockFileLib.Version = package.Version;
-
-            var files = package.GetFiles().Select(p => p.Path.Replace(Path.DirectorySeparatorChar, '/')).ToList();
-
+            var files = library.Files.Select(p => p.Replace(Path.DirectorySeparatorChar, '/'));
```
pranavkm: Could we do this up front (in line 35)? `Path.Combine` should work with well with forward slashes.

src/Microsoft.Framework.PackageManager/Utils/LockFileUtils.cs
```diff
@@ -443,6 +443,8 @@ internal static bool TargetFrameworkName_IsCriteriaSatisfied(object criteria, ob
 
         public class PatternDefinitions
         {
+            public static PatternDefinitions DotNetPatterns = new PatternDefinitions();
```
pranavkm: `readonly`

pranavkm:
:shipit:

src/Microsoft.Framework.PackageManager/Utils/LockFileUtils.cs
```diff
@@ -24,34 +22,38 @@ public static LockFileLibrary CreateLockFileLibrary(IPackagePathResolver resolve
...
+            else
             {
-                if (!string.Equals(Path.GetExtension(filePath), ".dll"))
+                lockFileLib.Files = package.GetFiles().Select(p => p.Path).ToList();
+                var installPath = resolver.GetInstallPath(package.Id, package.Version);
+                foreach (var filePath in lockFileLib.Files)
                 {
-                    continue;
-                }
+                    if (!string.Equals(Path.GetExtension(filePath), ".dll"))
```
moozzyk: `OrdinalIgnoreCase` ?

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
davidfowl: Yes it handles nulls

src/Microsoft.Framework.PackageManager/Utils/LockFileUtils.cs
davidfowl: Path.Combine does alot of work. I'll do this in another PR, when the Files node gets read from the lock file, it does some path resolution and that might have a ripple effect on other parts of the code. This was tested and works.

I'll take a look at the other part later.

src/Microsoft.Framework.PackageManager/Utils/LockFileUtils.cs
davidfowl: .DLL :smile:

moozzyk: (System32) :boom:
-a---l        3/14/2015   1:54 AM           7168 KBDTAILE.DLL
-a---l        3/14/2015   1:53 AM           6656 KBDTAJIK.DLL
-a---l        3/14/2015   1:53 AM           7168 KBDTAT.DLL
-a---l        3/14/2015   1:53 AM           7168 KBDTH0.DLL
-a---l        3/14/2015   1:53 AM           7168 KBDTH1.DLL
-a---l        3/14/2015   1:54 AM           7168 KBDTH2.DLL
-a---l        3/14/2015   1:54 AM           7168 KBDTH3.DLL
-a---l        3/14/2015   1:53 AM           7168 KBDTIFI.DLL

davidfowl: Thats fine, we absolutely don't care about those. Find me a nuget package that has that :stuck_out_tongue_closed_eyes:

moozzyk: Wait a few minutes... need to create one ;) - 

it's just a nit - I don't really care too much (for the same reason you just said).

muratg: Just curious, do you not to do OrdinalIgnoreCase because performance?

src/Microsoft.Framework.PackageManager/Utils/LockFileUtils.cs
pranavkm: I wasn't suggesting using `Path.Combine`. It was more along the lines of - if we store these files with transformed paths to start with, `Path.Combine` would continue working without issues and we'll avoid iterating over this list twice.

PR #1767, opened 2015-05-02
Update the makefile.shade for file header

troydai:
To auto-generate tpa.cpp with correct file headers

@Eilon @NTaylorMullen

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -1309,7 +1309,7 @@ functions @{
 
         var content = new List<string>();
 
-        content.Add(@"// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.");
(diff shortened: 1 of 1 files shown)


NTaylorMullen:
:shipit:

Eilon:
Make sure you run this too and update the generated CPP files as part of this commit.

davidfowl:
I just ran into this. :shipit:

Eilon:
You ran into it? It's not merged yet is it?

davidfowl:
I ran a build and it set the header back to the wrong one.

Eilon:
Ohhh I see. So we already updated the headers, right. Well then merge @troydai 's PR then :smile: I forgot the headers already updated.

Issue #780, opened 2014-10-19
Throw exception if running KRE doesn't have a matching target framework in project.json


davidfowl:
KRE-CoreCLR-* should require aspnetcore50
KRE-CLR/KRE-Mono-* should require aspnet50

PR #1773, opened 2015-05-04
Give more frienly error when the app doesn't support current runtime

ChengTian:
parent https://github.com/aspnet/dnx/issues/780

@davidfowl

### DIFF
--- a/src/Microsoft.Framework.Runtime/DefaultHost.cs
+++ b/src/Microsoft.Framework.Runtime/DefaultHost.cs
@@ -10,10 +10,12 @@ using System.Reflection;
 using System.Runtime.Versioning;
 using Microsoft.Framework.Runtime.Caching;
 using Microsoft.Framework.Runtime.Common.DependencyInjection;
+using Microsoft.Framework.Runtime.Common.Impl;
 using Microsoft.Framework.Runtime.Compilation;
 using Microsoft.Framework.Runtime.FileSystem;
 using Microsoft.Framework.Runtime.Infrastructure;
 using Microsoft.Framework.Runtime.Loader;
+using NuGet;
 
 namespace Microsoft.Framework.Runtime
 {
@@ -58,11 +60,34 @@ namespace Microsoft.Framework.Runtime
 
             Initialize();
 
+            var unresolvedLibs = _applicationHostContext.DependencyWalker.Libraries.Where(l => !l.Resolved);
+
             // If there's any unresolved dependencies then complain
-            if (_applicationHostContext.DependencyWalker.Libraries.Any(l => !l.Resolved))
+            if (unresolvedLibs.Any())
             {
-                var exceptionMsg = _applicationHostContext.DependencyWalker.GetMissingDependenciesWarning(
-                    _targetFramework);
+                string exceptionMsg;
+
+                // If the main project cannot be resolved, it means the app doesn't support current target framework
+                // (i.e. project.json doesn't contain a framework that is compatible with target framework of current runtime)
+                if (unresolvedLibs.Any(l => string.Equals(l.Identity.Name, Project.Name)))
+                {
+                    var runtimeEnv = ServiceProvider.GetService(typeof(IRuntimeEnvironment)) as IRuntimeEnvironment;
+                    var shortName = VersionUtility.GetShortFrameworkName(_targetFramework);
+                    exceptionMsg = $@"The current runtime target framework is not compatible with '{Project.Name}'.
+
+Current runtime Target Framework: '{_targetFramework} ({shortName})'
+  Type: {runtimeEnv.RuntimeType}
+  Architecture: {runtimeEnv.RuntimeArchitecture}
+  Version: {runtimeEnv.RuntimeVersion}
+
+Please make sure the runtime matches a framework specified in {Project.ProjectFileName}";
+                }
+                else
+                {
+                    exceptionMsg = _applicationHostContext.DependencyWalker.GetMissingDependenciesWarning(
+                        _targetFramework);
+                }
+
                 throw new InvalidOperationException(exceptionMsg);
             }
 
--- a/test/Microsoft.Framework.ApplicationHost.FunctionalTests/AppHostTests.cs
+++ b/test/Microsoft.Framework.ApplicationHost.FunctionalTests/AppHostTests.cs
@@ -1,11 +1,15 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Runtime.Versioning;
 using Bootstrapper.FunctionalTests;
 using Microsoft.Framework.CommonTestUtils;
 using Microsoft.Framework.Runtime;
+using Microsoft.Framework.Runtime.Common.Impl;
+using NuGet;
 using Xunit;
 
 namespace Microsoft.Framework.ApplicationHost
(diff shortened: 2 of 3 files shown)


test/Microsoft.Framework.ApplicationHost.FunctionalTests/AppHostTests.cs
```diff
@@ -117,5 +117,36 @@ public void AppHostShowsErrorWhenGivenSubcommandWasNotFoundInProjectJson(string
...
+}".Replace("FRAMEWORK_NAME", flavor == "coreclr" ? "dnx451" : "dnxcore50");
+
+            using (runtimeHomeDir)
+            using (var projectPath = new DisposableDir())
+            {
+                var projectName = new DirectoryInfo(projectPath).Name;
+                var projectJsonPath = Path.Combine(projectPath, Project.ProjectFileName);
+                File.WriteAllText(projectJsonPath, projectJsonContents);
+
+                string stdOut, stdErr;
```
moozzyk: `out var` ?

ChengTian: This feature was removed from C# 6

moozzyk: Thanks! Did not know about that.

src/Microsoft.Framework.Runtime/DefaultHost.cs
```diff
@@ -58,6 +58,16 @@ public Assembly GetEntryPoint(string applicationName)
 
             Initialize();
 
+            // If the main project cannot be resolved, it means the app doesn't support current target framework
+            // (i.e. project.json doesn't contain a framework that is compatible to target framework of current runtime)
+            var mainProject = _applicationHostContext.DependencyWalker.Libraries
```
davidfowl: Why do this loop twice? Why not do it in the condition below.

src/Microsoft.Framework.Runtime/DefaultHost.cs
```diff
@@ -58,6 +58,16 @@ public Assembly GetEntryPoint(string applicationName)
...
+                .Single(l => string.Equals(Project.Name, l.Identity.Name));
+            if (!mainProject.Resolved)
+            {
+                throw new InvalidOperationException(
+                    $"'{Project.Name}' doesn't support current runtime target framework '{_targetFramework}'");
+            }
+
             // If there's any unresolved dependencies then complain
             if (_applicationHostContext.DependencyWalker.Libraries.Any(l => !l.Resolved))
             {
```
davidfowl: Put the code in here. Change this to:

```
var unresolved = _applicationHostContext.DependencyWalker.Libraries.Where(l => !l.Resolved);
if (unresolved.Any())
{
      if (unresolved.Any(l => string.Equals(l.Identity.Name, Project.Name)))
      {
           ... show new error ...
      }
      else
      {
            ... show error ...
      }
}
```

src/Microsoft.Framework.Runtime/DefaultHost.cs
```diff
@@ -58,6 +58,16 @@ public Assembly GetEntryPoint(string applicationName)
...
             Initialize();
 
+            // If the main project cannot be resolved, it means the app doesn't support current target framework
+            // (i.e. project.json doesn't contain a framework that is compatible to target framework of current runtime)
+            var mainProject = _applicationHostContext.DependencyWalker.Libraries
+                .Single(l => string.Equals(Project.Name, l.Identity.Name));
+            if (!mainProject.Resolved)
+            {
+                throw new InvalidOperationException(
+                    $"'{Project.Name}' doesn't support current runtime target framework '{_targetFramework}'");
```
davidfowl: This error also isn't good enough. We need to tell people what runnable target frameworks are in their project.json and how they can potentially fix it. We can also use the IRuntimeEnvironment to show what DNX was invoked and tell the user which one should be invoked.

Something like:

```
'Project' doesn't support the current runtime target framework 'DNX Version=4.5.1'. 
```

/cc @Eilon @tdykstra here for help on the error message.

src/Microsoft.Framework.Runtime/DefaultHost.cs
```diff
@@ -58,11 +59,31 @@ public Assembly GetEntryPoint(string applicationName)
...
-                    _targetFramework);
+                string exceptionMsg;
+
+                // If the main project cannot be resolved, it means the app doesn't support current target framework
+                // (i.e. project.json doesn't contain a framework that is compatible to target framework of current runtime)
+                if (unresolvedLibs.Any(l => string.Equals(l.Identity.Name, Project.Name)))
+                {
+                    exceptionMsg = string.Format(@"'{0}' doesn't support current runtime target framework '{1}'.
+
+Runtime Target Framework: '{2}'
```
davidfowl: Use string interpolation.

src/Microsoft.Framework.Runtime/DefaultHost.cs
```diff
@@ -58,11 +60,34 @@ public Assembly GetEntryPoint(string applicationName)
...
+
+                // If the main project cannot be resolved, it means the app doesn't support current target framework
+                // (i.e. project.json doesn't contain a framework that is compatible to target framework of current runtime)
+                if (unresolvedLibs.Any(l => string.Equals(l.Identity.Name, Project.Name)))
+                {
+                    var runtimeEnv = ServiceProvider.GetService(typeof(IRuntimeEnvironment)) as IRuntimeEnvironment;
+                    var alternativeType = VersionUtility.IsDesktop(_targetFramework) ? "CoreCLR" : "CLR";
+                    var shortName = VersionUtility.GetShortFrameworkName(_targetFramework);
+                    exceptionMsg = string.Join(Environment.NewLine, new[]{
+                        $"'{Project.Name}' doesn't support current runtime target framework." + Environment.NewLine,
```
Eilon: Instead of the extra newlines to make it a double newline, can you add an empty string on the next line? That way the source code formatting will be closer to the runtime output.

Eilon: Change string to `'{Project.Name}' is not supported by the current runtime target framework.`

src/Microsoft.Framework.Runtime/DefaultHost.cs
```diff
@@ -58,11 +60,34 @@ public Assembly GetEntryPoint(string applicationName)
...
+                        $"Runtime Target Framework: '{_targetFramework} ({shortName})'",
+                        $"Runtime Type: {runtimeEnv.RuntimeType}",
+                        $"Runtime Architecture: {runtimeEnv.RuntimeArchitecture}",
+                        $"Runtime Version: {runtimeEnv.RuntimeVersion}" + Environment.NewLine,
+                        $"Please make sure the runtime matches a framework specified in {Project.ProjectFileName}" });
```
Eilon: Should we mention using `dnvm` or something? How do we expect the user to know how to fix this?

src/Microsoft.Framework.Runtime/DefaultHost.cs
```diff
@@ -58,11 +60,34 @@ public Assembly GetEntryPoint(string applicationName)
...
+                // If the main project cannot be resolved, it means the app doesn't support current target framework
+                // (i.e. project.json doesn't contain a framework that is compatible to target framework of current runtime)
+                if (unresolvedLibs.Any(l => string.Equals(l.Identity.Name, Project.Name)))
+                {
+                    var runtimeEnv = ServiceProvider.GetService(typeof(IRuntimeEnvironment)) as IRuntimeEnvironment;
+                    var alternativeType = VersionUtility.IsDesktop(_targetFramework) ? "CoreCLR" : "CLR";
+                    var shortName = VersionUtility.GetShortFrameworkName(_targetFramework);
+                    exceptionMsg = string.Join(Environment.NewLine, new[]{
+                        $"'{Project.Name}' doesn't support current runtime target framework." + Environment.NewLine,
+                        $"Runtime Target Framework: '{_targetFramework} ({shortName})'",
```
Eilon: Suggested output reformatting:

```
The current runtime target framework is not compatible with '{projectName}'.

Current runtime target framework:
    Type: {type}
    Architecture: {arch}
    Version: {version}

Please use the `dnvm` command to select a runtime target framework compatible with '{projectName}'.
```

(Note that I removed the filename from the end of the error and used the project name instead... do you think having the project file + path is helpful?)

src/Microsoft.Framework.Runtime/DefaultHost.cs
ChengTian: I discussed this with @davidfowl yesterday. There are two possible solutions when a user runs into this:
1. Switch to another flavor of runtime
2. Add proper framework into `frameworks` section of `project.json`

We don't which one the best suggestion because both of them are possible. So we came up with a summary that covers both case:
`Please make sure the runtime matches a framework specified in project.json`

Eilon: OK how about:

> Please use the `dnvm` command to select a runtime target framework compatible with '{projectName}' or change the framework specified in '{projectFileName}'.

ChengTian: @davidfowl , seems it's better to give concrete and concise suggestions instead of a short summary without any instruction. I'll go with @Eilon 's suggestion and this should save us from answering numerous questions

davidfowl: Except in helios when you see this message it wont make any sense.

ChengTian: I guess it still makes sense when my project (no matter it is web app or console app) has project.json
```
{
  "frameworks": {
    "dnx451": { }
  }
}
```
but run the project with dnx-coreclr

Eilon: @davidfowl is there any concrete suggestion we can put in there for Helios and similar scenarios?

davidfowl: I'm ok coupling dnvm but not web.config with respect to error messages. Helios does have its own bootstrapper logic so maybe we can put that message there if things fail.

Eilon: OK let's stick with this for now then.

ChengTian:
Thanks for the feedbacks. Please check the updated commit.

davidfowl:
:shipit:

davidfowl:
@ChengTian I fixed travis so see if the tests pass there now

ChengTian:
@davidfowl , fixed tests on Mono in the last commit. Looks like Appveyor build was cancelled for this PR because you pushed to dev. I already `build verify` on Windows locally.

Please check the latest commit. If you find it Ok, I'll merge this.

davidfowl:
:shipit:

Issue #1733, opened 2015-04-28
Sharing violation on breadcrumbs files


moozzyk:
I think we should catch the exception and retry n-times before failing. 

Stack trace:
```
info    : [HelloWorld:IISExpress:CoreClr:x86] IncrementMessages called
error   : [Ntlm:IISExpress:CoreClr:x86] Unhandled Exception: System.IO.IOException: The process cannot access the file 'C:\ProgramData\Microsoft DNX\BreadcrumbS
tore\Microsoft.Framework.ConfigurationModel.Json.1.0.0-beta5-11252' because it is being used by another process.
error   : [Ntlm:IISExpress:CoreClr:x86]    at System.IO.Win32FileStream.Init(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, F
ileOptions options, SECURITY_ATTRIBUTES secAttrs)
error   : [Ntlm:IISExpress:CoreClr:x86]    at System.IO.Win32FileSystem.Open(String fullPath, FileMode mode, FileAccess access, FileShare share, Int32 bufferSiz
e, FileOptions options, FileStream parent)
error   : [Ntlm:IISExpress:CoreClr:x86]    at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOp
tions options)
error   : [Ntlm:IISExpress:CoreClr:x86]    at Microsoft.Framework.Runtime.Servicing.Breadcrumbs.CreateBreadcrumbFile(String fileName)
error   : [Ntlm:IISExpress:CoreClr:x86]    at Microsoft.Framework.Runtime.Servicing.Breadcrumbs.WriteAllBreadcrumbsInternal()
error   : [Ntlm:IISExpress:CoreClr:x86]    at Microsoft.Framework.Runtime.Servicing.Breadcrumbs.<WriteAllBreadcrumbs>b__13_0(Object _)
error   : [Ntlm:IISExpress:CoreClr:x86]    at System.Threading.QueueUserWorkItemCallback.WaitCallback_Context(Object state)
error   : [Ntlm:IISExpress:CoreClr:x86]    at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
error   : [Ntlm:IISExpress:CoreClr:x86]    at System.Threading.QueueUserWorkItemCallback.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem()
[...]

Issue #1771, opened 2015-05-03
Writing breadcrumbs appears to have threading bug


Eilon:
I'm seeing a lot of errors for several different repos on the internal CI machine with call stacks in the breadcrumb code:

```
[12:15:50] :		 [xunit-test] System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
[12:15:50] :		 [xunit-test]    at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
[12:15:50] :		 [xunit-test]    at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
[12:15:50] :		 [xunit-test]    at System.Collections.Generic.List`1.Enumerator.MoveNext()
[12:15:50] :		 [xunit-test]    at Microsoft.Framework.Runtime.Servicing.Breadcrumbs.WriteAllBreadcrumbsInternal()
[12:15:50] :		 [xunit-test]    at Microsoft.Framework.Runtime.Servicing.Breadcrumbs.<WriteAllBreadcrumbs>b__13_0(Object _)
[12:15:50] :		 [xunit-test]    at System.Threading.QueueUserWorkItemCallback.WaitCallback_Context(Object state)
[12:15:50] :		 [xunit-test]    at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
[12:15:50] :		 [xunit-test]    at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
[12:15:50] :		 [xunit-test]    at System.Threading.QueueUserWorkItemCallback.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem()
[12:15:50] :		 [xunit-test]    at System.Threading.ThreadPoolWorkQueue.Dispatch()
[12:15:50] :		 [xunit-test]    at System.Threading._ThreadPoolWaitCallback.PerformWaitCallback()
```

Looking at the code that manages the breadcrumbs and the data types (e.g. https://github.com/aspnet/dnx/blob/e937e25ab3453ea86bda13c1297c41011dfda9de/src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs#L21), it appears that the code *needs* be thread safe, but isn't.

@victorhurdugaci can you take a look?

[...]

PR #1775, opened 2015-05-04
Sync the breadcrumbs writer 

victorhurdugaci:
Fixes: https://github.com/aspnet/dnx/issues/1733 and https://github.com/aspnet/dnx/issues/1771

The root cause is this change: https://github.com/aspnet/dnx/blob/be5bb7b77550bc2297b1827687206341c3fb0ec2/src/Microsoft.Framework.Runtime/ExportProviders/ProjectLibraryExportProvider.cs#L102 . We write the breadcrumbs during the dependency graph walking. That code was called only once before and it happened before breadcrumb writing.

Now that is it called multiple times, it can happen to run at the same time as the breadcrumb writing that that's running on a different thread. 

This change will allow `Add` to be called at anytime and only throw if you try to add a new breadcrumb (one that didn't exists when `write` was called). 

Also, the change includes a lock for multi-process breadcrumb writing. It locks the breadcrumb file and doesn't allow other processes to access it while writing.

Please review: @davidfowl @pranavkm

### DIFF
--- a/src/Microsoft.Framework.Runtime/Internal/ConcurrencyUtilities.cs
+++ b/src/Microsoft.Framework.Runtime/Internal/ConcurrencyUtilities.cs
@@ -6,9 +6,9 @@ using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
 
-namespace Microsoft.Framework.PackageManager
+namespace Microsoft.Framework.Runtime.Internal
 {
-    internal static class ConcurrencyUtilities
+    public static class ConcurrencyUtilities
     {
         internal static string FilePathToLockName(string filePath)
         {
@@ -19,7 +19,7 @@ namespace Microsoft.Framework.PackageManager
             return filePath.Replace(Path.DirectorySeparatorChar, '_');
         }
 
-        internal async static Task<T> ExecuteWithFileLocked<T>(string filePath, Func<bool, Task<T>> action)
+        public async static Task<T> ExecuteWithFileLocked<T>(string filePath, Func<bool, Task<T>> action)
         {
             var createdNew = false;
             var fileLock = new Semaphore(initialCount: 0, maximumCount: 1, name: FilePathToLockName(filePath),
--- a/src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
+++ b/src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
@@ -5,6 +5,8 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Framework.Runtime.Internal;
 using NuGet;
 
 namespace Microsoft.Framework.Runtime.Servicing
@@ -18,7 +20,10 @@ namespace Microsoft.Framework.Runtime.Servicing
 
         private readonly bool _isEnabled;
         private readonly string _breadcrumbsFolder;
-        private readonly List<BreadcrumbInfo> _breadcrumbsToWrite = new List<BreadcrumbInfo>();
+        private readonly List<string> _breadcrumbsToWrite = new List<string>();
+        private readonly object _addLock = new object();
+
+        private bool _writeWasCalled;
 
         public static Breadcrumbs Instance { get; private set; } = new Breadcrumbs();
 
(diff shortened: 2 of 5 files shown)


Eilon:
Two things we discussed that require follow-up:
1. Why is the dependency tree being walked twice? Is that a bug? Or on purpose? Are the results the same? Cache?
2. Is it allowed/supported for user code to load NuGet packages while the app is running? If so, we should theoretically breadcrumb those, even though it might be technically challenging.

victorhurdugaci:
@davidfowl the two questions above are for you. @Eilon and I discussed offline

davidfowl:
Hah! My change caused those problem :blush:

Eilon:
@davidfowl you owe @victorhurdugaci $10 plus a dozen donuts.

victorhurdugaci:
@davidfowl we can make than just a dozen donuts if you answer the Eilon's two questions :smile:

davidfowl:
I'm wondering if we can turn of breadcrumbing for compilation purposes or move where it is done so it doesn't happen more than once?

victorhurdugaci:
We can do that. It would be the same ignoring any breadcrumb add once writing has started. That leads to @Eilon 's second question: is it possible to load NuGet packages dynamically while the app is running?

davidfowl:
Ok here's what we need to do:
- Make the servicing classes services that are passed into the NuGetDependencyResolver instead of being a singleton (a singleton may be passed in but don't assume it is one).
- Make it so that a bool can be passed in that will disable ball bread crumbing

victorhurdugaci:
> •Make it so that a bool can be passed in that will disable ball bread crumbing 

Are you okay with simply ignoring adds after writing has started?

davidfowl:
You can absolutely call Assembly.Load or write your own loader that does something similar to what we do but that's not our problem

davidfowl:
Yes we actually have this in the wrong layer. Everyone uses this api to reason about the dependency graph. We really shouldn't be bread crumbing anything that wouldn't get loaded. As much, it should only ever be true when the ApplicationHost runs

victorhurdugaci:
Updated. No more adds after write

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
```diff
@@ -72,12 +77,20 @@ public void AddBreadcrumb(string packageId, SemanticVersion packageVersion)
             {
                 return;
             }
-
-            _breadcrumbsToWrite.Add(new BreadcrumbInfo()
+
```
moozzyk: nit: remove whitespaces

davidfowl:
Whats the deal with this?

src/Microsoft.Framework.Runtime/Internal/ConcurrencyUtilities.cs
```diff
@@ -6,9 +6,9 @@
 using System.Threading;
 using System.Threading.Tasks;
 
-namespace Microsoft.Framework.PackageManager
+namespace Microsoft.Framework.Runtime.Internal
 {
-    internal static class ConcurrencyUtilities
+    public static class ConcurrencyUtilities
```
davidfowl: Why is this public?

src/Microsoft.Framework.Runtime/Internal/ConcurrencyUtilities.cs
```diff
@@ -19,7 +19,7 @@ internal static string FilePathToLockName(string filePath)
             return filePath.Replace(Path.DirectorySeparatorChar, '_');
         }
 
-        internal async static Task<T> ExecuteWithFileLocked<T>(string filePath, Func<bool, Task<T>> action)
+        public async static Task<T> ExecuteWithFileLocked<T>(string filePath, Func<bool, Task<T>> action)
```
pranavkm: Could we have a synchronous overload for this which takes an `Action`? We can have it call this method and perform `GetAwaiter().GetResult()` in that method. @ChengTian is using this in another PR to perform synchronous operations and it would be cleaner to commonize the pattern.

src/Microsoft.Framework.Runtime/Internal/ConcurrencyUtilities.cs
victorhurdugaci: Because it is used by the package manager too. Please note that I moved the file in a different project

davidfowl:
:shipit:

src/Microsoft.Framework.Runtime/Internal/ConcurrencyUtilities.cs
victorhurdugaci: Yes, I'm going to do that in a PR that follows. I need to push this because it is blocking the CI

Issue #30, opened 2014-03-04
Internalize JSON.NET dependency


davidfowl:
Today we're using JSON.NET as a package and that's no good for apps and frameworks that use this framework. This is part of the work #29

ChengTian:
We should close the following issues after this is done:
https://github.com/aspnet/dnx/issues/1788
https://github.com/aspnet/dnx/issues/1816

muratg:
https://github.com/aspnet/dnx/issues/1816#issuecomment-100063615

muratg:
https://github.com/aspnet/Announcements/issues/24

troydai:
Fixed in f0a852af06282494168100a83b3b79e0cc155c15

PR #1778, opened 2015-05-04
Decouple Microsoft.Framework.Runtime from Newtonsoft.Json

troydai:
Issue: https://github.com/aspnet/dnx/issues/30

1. `JavaScriptSerializer` is trimmed down and refactoried into `JsonDeserializer`. It is entirely internal to `Microsoft.Framework.Runtime.`
2. Logic of writing _lock file_ is moved to `Microsoft.Framework.PackageManager`, where `Newtonsoft.Json` is referenced.
3. All exceptions are updated. 

I assume it will take a while to finish the code review. In the meantime, I'll test this change with latest VS.


Eilon:
Should we ignore this and use #1779 instead? Or?

troydai:
Please use #1779 

Thanks.

PR #1779, opened 2015-05-05
[Prototype] Use JsonDeserializer in Microsoft.Framework.Runtime

troydai:
Address issue: #30

`JsonDeserializer` is entirely internal to `Microsoft.Framework.Runtime`.

Logic of writing lock file is moved to `Microsoft.Framework.PackageManager`, where `Newtonsoft.Json` will be referenced.

All exceptions are updated.

Unit tests are added to covert `JsonDeserializer`


src/Microsoft.Framework.Runtime.Hosting/Json/JsonObject.cs
```diff
@@ -0,0 +1,108 @@
...
+        public JsonString ValueAsString(string key)
+        {
+            return ValueAs(key, value => value as JsonString);
+        }
+
+        public bool ValueAsBoolean(string key, bool defaultValue = false)
+        {
+            return ValueAs(key, value =>
+            {
+                if (value != null && value is JsonBoolean)
```
Maxwe11: condition could be simplified:
a) `as` cast and check on null
b) remove null check because if `value` is null `is` will return false

src/Microsoft.Framework.Runtime.Hosting/Json/JsonObject.cs
```diff
@@ -0,0 +1,108 @@
...
+
+                return defaultValue;
+            });
+        }
+
+        public bool? ValueAsNullableBoolean(string key)
+        {
+            return ValueAs<bool?>(key, value =>
+            {
+                if (value != null && value is JsonBoolean)
```
Maxwe11: ditto

src/Microsoft.Framework.Runtime.Hosting/Json/JsonContent.cs
```diff
@@ -0,0 +1,199 @@
...
+        {
+            _content = content;
+        }
+
+        public int TotalLines { get { return _content.Count; } }
+
+        /// <summary>
+        /// Current line number in zero-based index.
+        /// </summary>
+        public int CurrentLine { get; private set; } = 0;
```
Eilon: 0 is the default value.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonContent.cs
```diff
@@ -0,0 +1,199 @@
...
+        /// </summary>
+        /// <returns>Returns false if the cursor reach the end of the content.</returns>
+        public bool MoveToNextNonEmptyChar()
+        {
+            while (TotalLines > CurrentLine)
+            {
+                while (_content[CurrentLine].Length > CurrentColumn + 1)
+                {
+                    char c = _content[CurrentLine][++CurrentColumn];
+                    if (!char.IsWhiteSpace(c))
```
Eilon: Be careful with these `char` helper methods - they rarely do what you want. Please check the JSON / JavaScript specs to see exactly what *they* define as whitespace and use that. It's *possible* that they define it in the same way that Unicode does (which is what `char.IsWhiteSpace` uses), but that's generally unlikely.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
```diff
@@ -0,0 +1,493 @@
...
+            }
+
+            _input = JsonContent.CreateFromStream(stream);
+
+            return Deserialize();
+        }
+
+        private JsonValue Deserialize()
+        {
+            JsonValue result = DeserializeInternal(0);
```
Eilon: `var` and named param.

src/Microsoft.Framework.PackageManager/Restore/LockFileWriter.cs
```diff
@@ -0,0 +1,235 @@
...
+                WriteArray(frameworkInfo.Dependencies, WriteString));
+        }
+
+        private JToken WriteFrameworkAssemblies(IList<FrameworkAssemblyReference> frameworkAssemblies)
+        {
+            var groups = frameworkAssemblies.SelectMany(x =>
+            {
+                if (x.SupportedFrameworks.Any())
+                {
+                    return x.SupportedFrameworks.Select(xx => new { x.AssemblyName, FrameworkName = xx });
```
victorhurdugaci: x? xx?

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
```diff
@@ -0,0 +1,493 @@
...
+            {
+                return DeserializeString();
+            }
+
+            return DeserializePrimitiveObject();
+        }
+
+        private JsonDeserializerException CreateExceptionFromContent(string message)
+        {
+            message = message ?? "Failed to deserialize. ";
```
Eilon: Is trailing space deliberate?

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
```diff
@@ -0,0 +1,493 @@
...
+                _input.CurrentColumn);
+        }
+
+        private JsonArray DeserializeList(int depth)
+        {
+            var list = new List<JsonValue>();
+
+            if (!_input.MoveNext())
+            {
+                throw CreateExceptionFromContent("Parsing reach the end of the content before it can finish");
```
Eilon: Need a more grammatically correct message here.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
```diff
@@ -0,0 +1,493 @@
...
+            return new JsonArray(list.ToArray(), position);
+        }
+
+        private JsonObject DeserializeDictionary(int depth)
+        {
+            IDictionary<string, JsonValue> dictionary = null;
+
+            if (!_input.MoveNext())
+            {
+                throw CreateExceptionFromContent("Parsing reach the end of the content before it can finish");
```
Eilon: This one too.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonContent.cs
```diff
@@ -0,0 +1,199 @@
...
+            if (CurrentColumn + 1 < _content[CurrentLine].Length)
+            {
+                CurrentColumn += 1;
+                return true;
+            }
+            else
+            {
+                // find the first non empty line after current line
+                var targetLine = CurrentLine;
+                while (++targetLine < TotalLines && _content[targetLine].Length == 0)
```
victorhurdugaci: I think it would be easier to understand if you move the increment inside the while block

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
```diff
@@ -0,0 +1,493 @@
...
+
+            if (token.String.Equals("false"))
+            {
+                return new JsonBoolean(false, token.Position);
+            }
+
+            // Is it a floating point value
+            bool hasDecimalPoint = token.String.IndexOf('.') >= 0;
+
+            // DevDiv 56892: don't try to parse to Int32/64/Decimal if it has an exponent sign
```
Eilon: Remove the old bug number

Eilon:
BTW several other style issues (multiple types per file, `var`, etc.) but I won't comment on them further because this is a prototype.

victorhurdugaci:
So this json deserializer doesn't support comments? That's a breaking change

src/Microsoft.Framework.Runtime.Hosting/Json/JsonContent.cs
troydai: got it.

src/Microsoft.Framework.PackageManager/Restore/LockFileWriter.cs
troydai: I copied the same code over there. It's not in the scope of this PR to bring sense to them :) Let's rename next time

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
troydai: Nop. will clear.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonContent.cs
troydai: Fair, will update.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
troydai: Got it.

troydai:
Two features are required not to break current scenario:

1. Comments support;
2. Redundant trailing _comma_.

troydai:
@davidfowl @Eilon @victorhurdugaci 
 There are quite a few implementations I want to improve, but not until we have a evaluation on the prototype and decide whether we want to move forward with additional features.

Eilon:
@victorhurdugaci BTW it's not a "breaking change" in the traditional change because this is all new... but it would be breaking from previous previews, of course. But I agree with @troydai that in the meantime we'll do things one at a time in terms of honing the prototype into production quality.

victorhurdugaci:
Then we need to update some project.json files like this one: https://github.com/aspnet/MusicStore/blob/dev/src/MusicStore/project.json

Eilon:
@troydai  BTW how close is this code to the stuff I sent you? Is the basic parsing logic the same, except that you added the line number support etc.? Are there any other significant differences? I ask because the old parsing code was fairly robust; it was just missing some desirable features (line numbers, comment support, etc.).

troydai:
I make sure the the original parsing logic is kept the same, what I added is the line information support. But adding line information support introduces extract logic to `MoveNext` and `MovePrev`, which open to potential bugs. Besides that the deserializing logic itself is intact.

Eilon:
@victorhurdugaci I think before we take this change we'd have to support comments (assuming we want to continue to support comments, which I hope we do).

Eilon:
@troydai great!

davidfowl:
Feels like the parser needs to be trimmed down for things we don't need to care about:

- Utf16StringValidator.cs (do we need it)
- Json* types (why do we need them?), `Dictionary<string, object>`, `string`, `int`, primitive types should be enough right?
- Depth checking (why bother)? Do we need this safety?

troydai:
Types are for line information. By combining the line formation and data in one class, it helps to consume the line information easily.

src/Microsoft.Framework.PackageManager/Restore/LockFileWriter.cs
```diff
@@ -0,0 +1,235 @@
...
+
+        private JToken WriteFrameworkAssemblyReference(FrameworkAssemblyReference item)
+        {
+            return new JValue(item.AssemblyName);
+        }
+
+        private JToken WritePackageReferenceSet(PackageReferenceSet item)
+        {
+            var json = new JObject();
+            json["targetFramework"] = item.TargetFramework?.ToString();
```
BrennanConroy: Is possibly assigning null here okay?

src/Microsoft.Framework.Runtime.Hosting/Json/JsonContent.cs
```diff
@@ -0,0 +1,199 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
```
BrennanConroy: Some files have the old copyright, fix them all!

src/Microsoft.Framework.Runtime.Hosting/Json/JsonContent.cs
```diff
@@ -0,0 +1,199 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace Microsoft.Framework.Runtime.Json
+{
+    /// <summary>
+    /// JsonBuffer represents a piece of loaded json content.
```
BrennanConroy: JsonContent*

src/Microsoft.Framework.Runtime.Hosting/Json/JsonContent.cs
```diff
@@ -0,0 +1,199 @@
...
+                var targetLine = CurrentLine;
+
+                // find the first non empty line before current line
+                while (--targetLine >= 0 && _content[targetLine].Length == 0)
+                {
+                }
+
+                if (targetLine < 0)
+                {
+                    // Even the first line is empty, move it to the initial position.
```
BrennanConroy: "If the first line..."

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializerResource.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        internal static string JSON_InvalidRecursionLimit
+        {
+            get { return "RecursionLimit must be a positive integer."; }
+        }
+
+        internal static string JSON_MaxJsonLengthExceeded
+        {
+            get { return "Error during serialization or deserialization using the JSON JavaScriptSerializer. The length of the string exceeds the value set on the maxJsonLength property."; }
```
BrennanConroy: Do we want to say JavaScript here?

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializerResource.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        internal static string JSON_UnterminatedString
+        {
+            get { return "Unterminated string passed in."; }
+        }
+
+        internal static string JSON_MaxJsonDeserializerMembers
+        {
+            get { return "The maximum number of items has already been deserialized into a single dictionary by the JavaScriptSerializer.The value is '{0}'."; }
```
BrennanConroy: Space after Serializer, also JavaScript again

src/Microsoft.Framework.Runtime.Hosting/Json/Utf16StringValidator.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            // slow case: surrogates exist, so we need to validate them
+            char[] chars = input.ToCharArray();
+            for (int i = idxOfFirstSurrogate; i < chars.Length; i++)
+            {
+                char thisChar = chars[i];
+
+                // If this character is a low surrogate, then it was not preceded by
+                // a high surrogate, so we'll replace it.
+                if (Char.IsLowSurrogate(thisChar))
```
BrennanConroy: lowercase char

src/Microsoft.Framework.Runtime.Hosting/PatternsCollectionHelper.cs
```diff
@@ -5,53 +5,44 @@
...
-                if (token.Type == JTokenType.String)
+                string valueInString = rawProject.ValueAsString(propertyName);
+                if (valueInString != null)
                 {
-                    return CreateCollection(projectDirectory, token.Value<string>());
+                    return CreateCollection(projectDirectory, valueInString);
                 }
 
-                if (token.Type == JTokenType.Array)
+                // TODO: simplify after introduce JsonArray
```
BrennanConroy: We have JsonArray

src/Microsoft.Framework.Runtime/DependencyManagement/LockFileFormat.cs
```diff
@@ -109,37 +85,20 @@ private LockFileLibrary ReadLibrary(string property, JToken json)
...
-        private JProperty WriteLibrary(LockFileLibrary library)
+        private LockFileTarget ReadTarget(string property, JsonValue json)
         {
-            var json = new JObject();
-            if (library.IsServiceable)
+            var jobject = json as JsonObject;
+            if (jobject == null)
             {
-                WriteBool(json, "serviceable", library.IsServiceable);
+                throw FileFormatException.Create("The value type is not object.", json, _currentLockFilePath);
```
BrennanConroy: an object

src/Microsoft.Framework.Runtime/DependencyManagement/LockFileFormat.cs
```diff
@@ -148,13 +107,19 @@ private LockFileTarget ReadTarget(string property, JToken json)
...
             return target;
         }
 
-        private LockFileTargetLibrary ReadTargetLibrary(string property, JToken json)
+        private LockFileTargetLibrary ReadTargetLibrary(string property, JsonValue json)
         {
+            var jobject = json as JsonObject;
+            if (jobject == null)
+            {
+                throw FileFormatException.Create("The value type is not object.", json, _currentLockFilePath);
```
BrennanConroy: same

src/Microsoft.Framework.Runtime/Project.cs
```diff
@@ -318,86 +342,79 @@ private static SemanticVersion SpecifySnapshot(string version, string snapshotVa
         private static void PopulateDependencies(
             string projectPath,
             IList<LibraryDependency> results,
-            JObject settings,
+            JsonObject settings,
             string propertyName,
             bool isGacOrFrameworkReference)
         {
-            var dependencies = settings[propertyName] as JObject;
+            var dependencies = settings.ValueAsJsonObject(propertyName);  //settings[propertyName] as JObject;
```
BrennanConroy: Remove comment

src/Microsoft.Framework.Runtime/Project.cs
```diff
@@ -594,13 +620,13 @@ private bool BuildTargetFrameworkNode(KeyValuePair<string, JToken> targetFramewo
                 Dependencies = new List<LibraryDependency>()
             };
 
-            var properties = targetFramework.Value.Value<JObject>();
+            //var properties = d targetFramework.Value.Value<JObject>();
```
BrennanConroy: Remove

troydai:
Thanks for all the comments. I'll begin a major refactorying to the deserializer now.

troydai:
Iteration 2!

/cc @davidfowl @victorhurdugaci @BrennanConroy @Eilon

src/Microsoft.Framework.PackageManager/Restore/LockFileWriter.cs
troydai: Fixing `LockFileWriter` is not in the scope of this PR. This is merely copied from `LockFileFormat` to separate read and write.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
```diff
@@ -0,0 +1,301 @@
...
+
+        private int _line = 1;
+        private int _column = 0;
+
+        public JsonBuffer(TextReader reader)
+        {
+            _reader = reader;
+        }
+
+        public JsonToken Read()
```
victorhurdugaci: Better name: `GetNextToken`

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
```diff
@@ -0,0 +1,301 @@
...
+                        buf.Add((char)JsonConstants.Quotation);
+                    }
+                    else if (next == JsonConstants.ReverseSolidus)
+                    {
+                        buf.Add((char)JsonConstants.ReverseSolidus);
+                    }
+                    else if (next == JsonConstants.Solidus)
+                    {
+                        buf.Add((char)JsonConstants.Solidus);
+                    }
```
victorhurdugaci: For all these if blocks:

```
if (next == ... ) || (next == ... ) || (next == ...) 
{ 
    buf.Add((char)next); 
}
```

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
```diff
@@ -0,0 +1,301 @@
...
+                    {
+                        // '\r' carriage return
+                        buf.Add((char)JsonConstants.CR);
+                    }
+                    else if (next == 0x74)
+                    {
+                        // '\t' tab
+                        buf.Add((char)JsonConstants.HT);
+                    }
+                    else if (next == 0x75)
```
victorhurdugaci: what characters is `0x75` ? use chars instead `if (next == 'X')`

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
```diff
@@ -0,0 +1,301 @@
...
+            {
+                var next = ReadNextChar();
+
+                if (next == -1 || next == JsonConstants.LF)
+                {
+                    throw new JsonDeserializerException(JsonDeserializerResource.JSON_OpenString, _line, _column);
+                }
+                else if (escaped)
+                {
+                    if (next == JsonConstants.Quotation)
```
victorhurdugaci: Just compare to a char, no need to these constants

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
```diff
@@ -0,0 +1,301 @@
...
+            {
+                ReadLiteral(JsonConstants.ValueFalse);
+                result.Type = JsonTokenType.False;
+            }
+            else if (first == (int)'n')
+            {
+                ReadLiteral(JsonConstants.ValueNull);
+                result.Type = JsonTokenType.Null;
+            }
+            else if (char.IsDigit((char)first) || first == (int)'-')
```
victorhurdugaci: `.` can be the first character. `.42` is a valid number

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
```diff
@@ -0,0 +1,301 @@
...
+
+                if (first == -1)
+                {
+                    result.Type = JsonTokenType.EOL;
+                    return result;
+                }
+                else if (!IsWhitespace(first))
+                {
+                    break;
+                }
```
victorhurdugaci: Don't you have to increase the column here once you skip a space?

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
```diff
@@ -0,0 +1,301 @@
...
+                        _line, _column);
+                }
+                else
+                {
+                    ReadNextChar();
+                }
+            }
+
+            var tail = _reader.Peek();
+            if (tail != JsonConstants.RightCurlyBracket &&
```
victorhurdugaci: This is not something that the scanner should check

victorhurdugaci: Also, if you decide to take such an approach, at least make sure that you do the same for numbers and strings

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
```diff
@@ -0,0 +1,301 @@
...
+            {
+                throw new JsonDeserializerException(
+                    JsonDeserializerResource.Format_IllegalTrailingCharacterAfterLiteral(tail, literal),
+                    _line, _column);
+            }
+        }
+
+        private string ReadString()
+        {
+            var buf = new List<char>();
```
victorhurdugaci: `StringBuilder`

test/Microsoft.Framework.Runtime.Tests/TestSample/project.lock.sample
```diff
@@ -0,0 +1,2180 @@
+{
```
ChengTian: Can we generate this dynamically by calling `dnu restore` in tests? Then we don't need to update this later when we change lock file format.

You can generate a lock file for `HelloWorld` sample project by calling this: https://github.com/aspnet/dnx/blob/dev/test/Bootstrapper.FunctionalTests/BootstrapperTestUtils.cs#L35

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
victorhurdugaci: Suggestion: instead of allocating a buffer for each token, have a class level token that has the `current token`

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
```diff
@@ -0,0 +1,301 @@
...
+                            }
+                            else
+                            {
+                                unicodes[i] = (char)next;
+                            }
+                        }
+
+                        try
+                        {
+                            var unicodeValue = int.Parse(new string(unicodes), NumberStyles.HexNumber, CultureInfo.InvariantCulture);
```
victorhurdugaci: Catch parse failure and throw the exception below?

src/Microsoft.Framework.Runtime.Hosting/Json/JsonConstants.cs
```diff
@@ -0,0 +1,34 @@
...
+        public const int HT = 0x09;
+        public const int FF = 0x0C;
+        public const int BACKSPACE = 0x08;
+
+        public const int LeftSquareBracket  = '['; // 0x5B
+        public const int RightSquareBracket = ']'; // 0x5D
+        public const int LeftCurlyBracket   = '{'; // 0x7B
+        public const int RightCurlyBracket  = '}'; // 0x7D
+
+        public const int Colon = ':';           // 0x3A
```
victorhurdugaci: Just use chars  directly in code

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
troydai: No, the column is increased when a char is read (line 30). It doesn't matter if it s a space or not.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
```diff
@@ -0,0 +1,301 @@
...
+                            buf.Add((char)unicodeValue);
+                        }
+                        catch (FormatException ex)
+                        {
+                            throw new JsonDeserializerException("Invalid Unicode format [" + new string(unicodes) + "]", ex, unicodeLine, unicodeColumn);
```
victorhurdugaci: resource?

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
```diff
@@ -0,0 +1,167 @@
...
+
+        private JsonValue Deserialize()
+        {
+            var result = DeserializeInternal(_buffer.Read());
+
+            // There are still unprocessed char. The parsing is not finished. Error happened.
+            var nextToken = _buffer.Read();
+            if (nextToken.Type != JsonTokenType.EOL)
+            {
+                throw new JsonDeserializerException("Failed to continue deserializing. Additional token " + nextToken.Value, nextToken);
```
victorhurdugaci: Resource?

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
troydai: I didn't find this to be significant enough to change. Is there a scenario in which it is a bad idea to guarantee the return literal is correct?

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
```diff
@@ -0,0 +1,167 @@
...
+                throw new ArgumentNullException(nameof(reader));
+            }
+
+            _buffer = new JsonBuffer(reader);
+            return Deserialize();
+        }
+
+        private JsonValue Deserialize()
+        {
+            var result = DeserializeInternal(_buffer.Read());
```
victorhurdugaci: I think your parser accepts `1` as a valid json. Not sure if that's in the spec

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
```diff
@@ -0,0 +1,167 @@
...
+            {
+                return DeserializeObject(next);
+            }
+
+            if (next.Type == JsonTokenType.String)
+            {
+                return new JsonString(next.Value, next.GetPosition());
+            }
+
+            if (next.Type == JsonTokenType.True || next.Type == JsonTokenType.False)
```
victorhurdugaci: Token type . boolean ?

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
troydai: No. It's not.
http://tools.ietf.org/html/rfc7159#section-6

```
      number = [ minus ] int [ frac ] [ exp ]

      decimal-point = %x2E       ; .

      digit1-9 = %x31-39         ; 1-9

      e = %x65 / %x45            ; e E

      exp = e [ minus / plus ] 1*DIGIT

      frac = decimal-point 1*DIGIT

      int = zero / ( digit1-9 *DIGIT )

      minus = %x2D               ; -

      plus = %x2B                ; +

      zero = %x30                ; 0
```

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
```diff
@@ -0,0 +1,167 @@
...
+                {
+                    throw new JsonDeserializerException(JsonDeserializerResource.JSON_InvalidArrayEnd, next);
+                }
+                else if (next.Type == JsonTokenType.RightSquareBracket)
+                {
+                    break;
+                }
+                else if (next.Type != JsonTokenType.Comma)
+                {
+                    throw new JsonDeserializerException(JsonDeserializerResource.JSON_InvalidArrayExpectComma, next);
```
victorhurdugaci: Rather than having different message for each expected character, you can have a generic one like `Unexpected character. Expected: {0}`

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
```diff
@@ -0,0 +1,167 @@
...
+                    dictionary[memberName] = DeserializeInternal(_buffer.Read());
+
+                    next = _buffer.Read();
+                    if (next.Type == JsonTokenType.RightCurlyBracket)
+                    {
+                        break;
+                    }
+                    else if (next.Type != JsonTokenType.Comma)
+                    {
+                        throw new JsonDeserializerException(JsonDeserializerResource.JSON_InvalidObject, next);
```
victorhurdugaci: Throw an unexpected character/token exception

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
troydai: Open to rename the method. Need more ideas here.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
troydai: Yes. `1` is a valid json.

http://tools.ietf.org/html/rfc7159#section-2
```
JSON-text = ws value ws

value = false / null / true / object / array / number / string
```

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
troydai: I find this easier to consume and save allocating additional string to store the boolean value.

troydai:
I need more review to the resource / exception messages.

src/Microsoft.Framework.Runtime/project.json
```diff
@@ -23,8 +24,7 @@
         "Microsoft.Framework.Runtime.Compilation.Sources": { "version": "1.0.0-*", "type": "build" },
         "Microsoft.Framework.Runtime.Sources": { "version": "1.0.0-*", "type": "build" },
         "Microsoft.CodeAnalysis.CSharp": { "version": "1.0.0-*", "type": "build" },
-        "Microsoft.Framework.Runtime.Roslyn.Abstractions": { "version": "1.0.0-*", "type": "build" },
-        "Newtonsoft.Json": "6.0.6"
```
troydai: No more.

muratg: :+1:

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
```diff
@@ -0,0 +1,301 @@
...
+using System.IO;
+
+namespace Microsoft.Framework.Runtime.Json
+{
+    internal class JsonBuffer
+    {
+        private readonly TextReader _reader;
+
+        private int _line = 1;
+        private int _column = 0;
```
Eilon: Don't initialize to default value.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
Eilon: `char.IsDigit` does not do what you think it does. It checks for any Unicode char, which is like hundreds of different chars. Just check for the valid ASCII range per the JSON spec.

Eilon: (As a bonus, it'll be much faster to do the ASCII check.)

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
```diff
@@ -0,0 +1,301 @@
...
+                {
+                    // This is the end of file
+                    return -1;
+                }
+                else if (value == JsonConstants.LF)
+                {
+                    // This is a new line. Let the next loop read the first charactor of the following line.
+                    // Set position ahead of next line
+                    _column = 0;
+                    _line += 1;
```
Eilon: `_line++;` should be fine here :smile:

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
```diff
@@ -0,0 +1,301 @@
...
+        }
+
+        private string ReadNumber(int firstRead)
+        {
+            var buf = new List<char> { (char)firstRead };
+            while (true)
+            {
+                var next = _reader.Peek();
+
+                if (char.IsDigit((char)next) ||
```
Eilon: See earlier comment. Don't use `char.IsDigit`.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
```diff
@@ -0,0 +1,301 @@
...
+                {
+                    buf.Add((char)ReadNextChar());
+                }
+                else
+                {
+                    break;
+                }
+            }
+
+            return new string(buf.ToArray());
```
Eilon: I think I saw a comment from @victorhurdugaci or something to just use StringBuilder directly. With the code here there are *three* copies of the data: 1. `List<char>`, 2. `char[]`, 3. `string`.

I recommend either using StringBuilder (which will have less copying/allocations), or find a way to use the `List<char>` or whatever directly (if possible).

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
```diff
@@ -0,0 +1,167 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace Microsoft.Framework.Runtime.Json
+{
+    internal class JsonDeserializer
```
Eilon: Can this class just be a `static` class? It has no meaningful state and it doesn't seem like an instance of this type can be meaningfully reused. And, all current usage of this type is single use anyway. Recommend making it static.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
Eilon: `An unexpected token '{0}' was found at position {position}.`

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializerException.cs
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        public JsonDeserializerException(string message, JsonToken nextToken)
+            : base(message)
+        {
+            Line = nextToken.Line;
+            Column = nextToken.Column;
+        }
+
+        public int Line { get; private set; }
```
Eilon: C# 6 - I think you can remove the private setters.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializerResource.cs
```diff
@@ -0,0 +1,88 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Framework.Runtime.Json
+{
+    internal class JsonDeserializerResource
+    {
+        internal static string Format_IllegalTrailingCharacterAfterLiteral(int value, string literal)
+        {
+            return string.Format("Illegal character {0} ({1:X4}) trails the literal name {2}", (char)value, value, literal);
```
Eilon: `trails` --> `after`

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializerResource.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        internal static string Format_UnrecognizedLiteral(string literal)
+        {
+            return string.Format("Unrecognized json literal. {0} is expected.", literal);
```
Eilon: `json` --> `JSON` and please check this in all resources / exceptions / doc comments in the PR

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializerResource.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        internal static string JSON_DepthLimitExceeded
+        {
+            get { return "RecursionLimit exceeded."; }
+        }
+
+        internal static string JSON_ExpectedOpenBrace
+        {
+            get { return "Invalid object passed in, '{' expected."; }
```
Eilon: The format of all these strings should be `Invalid {whatever}. Expected '{0}'.`

src/Microsoft.Framework.Runtime.Hosting/Json/JsonPrimitive.cs
```diff
@@ -0,0 +1,87 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Globalization;
+
+namespace Microsoft.Framework.Runtime.Json
+{
+    internal class JsonPrimitive : JsonValue
```
Eilon: Split out into multiple files, of course.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonPrimitive.cs
```diff
@@ -0,0 +1,87 @@
...
+            {
+                Value = true;
+            }
+            else if (token.Type == JsonTokenType.False)
+            {
+                Value = false;
+            }
+            else
+            {
+                throw new ArgumentException("Token value should be either True or False.", nameof(token));
```
Eilon: `should` --> `must`

src/Microsoft.Framework.Runtime/Project.cs
```diff
@@ -324,86 +350,79 @@ private static SemanticVersion SpecifySnapshot(string version, string snapshotVa
...
             if (dependencies != null)
             {
-                foreach (var dependency in dependencies.Properties())
+                foreach (var dependencyKey in dependencies.Keys)
                 {
-                    if (string.IsNullOrEmpty(dependency.Name))
+                    if (string.IsNullOrEmpty(dependencyKey))
                     {
-
+                        // TODO: add line information
```
Eilon: Fix or log a bug?

src/Microsoft.Framework.Runtime.Hosting/Json/JsonPrimitive.cs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        public double Double
+        {
+            get { return _double; }
+        }
+
+        public int Int
+        {
+            get { return (int)_double; }
```
Eilon: This does not seem safe. This will simply ignore conversion errors, such as when we read the lock file version. I think to be safe this type needs to know whether a value is an integer or not, e.g. by having a `bool IsValidInt { get; }` property or something. Then if the value isn't an int and an int was requested, we can throw.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
troydai: Yup. I was thinking it should be a static during refactorying.

Will do.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
troydai: Will use StringBuilder

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
troydai: Will update.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
troydai: Will add to resource.

src/Microsoft.Framework.Runtime/Project.cs
troydai: Fixed. Just forget to remove the comment.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonPrimitive.cs
troydai: Will do.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonPrimitive.cs
troydai: Will do.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializerResource.cs
troydai: Got it.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializerResource.cs
troydai: Will do.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializerResource.cs
troydai: Got it.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializerException.cs
troydai: yes.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
victorhurdugaci: ++_line :unamused:

src/Microsoft.Framework.Runtime.Hosting/Json/JsonPosition.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            var other = obj as JsonPosition;
+            return other != null &&
+                   other.Line == Line &&
+                   other.Column == Column;
+        }
+
+        public override int GetHashCode()
+        {
+            return base.GetHashCode();
```
victorhurdugaci: `return Line^Column`

src/Microsoft.Framework.Runtime.Hosting/Json/JsonToken.cs
```diff
@@ -0,0 +1,34 @@
...
+        public int Line;
+        public int Column;
+
+        public JsonPosition GetPosition()
+        {
+            return new JsonPosition(Line, Column);
+        }
+    }
+
+    public enum JsonTokenType
```
victorhurdugaci: split in 2 files?

src/Microsoft.Framework.Runtime.Hosting/Json/JsonValue.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Framework.Runtime.Json
+{
+    internal class JsonValue
+    {
+        public JsonValue(JsonPosition position)
```
victorhurdugaci: I think `JsonToken` and `JsonValue` are swapped because `JsonToken` has a value while `JsonValue` does not

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
troydai: Will do.

test/Microsoft.Framework.Runtime.Tests/JsonDeserializerFacts.cs
```diff
@@ -0,0 +1,204 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.IO;
+using Microsoft.Framework.Runtime.Json;
+using Xunit;
+
+namespace Microsoft.Framework.Runtime.Tests
+{
+    public class JsonDeserializerFacts
+    {
```
victorhurdugaci: I suggest a few more error case tests:

1. Unbalanced `}`
2. Unbalanced `]`
3. Wrong nesting like: `{ [ } ]`

src/Microsoft.Framework.Runtime.Hosting/Json/JsonToken.cs
troydai: Will do.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonValue.cs
troydai: They serves different purpose. 
`JsonToken` is the value returned by `JsonBuffer` it's the data interface between `JsonBuffer` and `JsonDeserializer`. It's a value type.

`JsonValue` is the base class of return values of `JsonDeserializer`, it is carries position information to form a `FileFormatException`. As base class it doesn't carry the value not it knows the type of the value.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonPosition.cs
troydai: not sure if its necessary since it is not used in any hashset or dicitionary, but yeah, why not.

test/Microsoft.Framework.Runtime.Tests/TestSample/project.lock.sample
troydai: The unit test shouldn't reference to functional test util. I prefer a hardcopy of lock file in the git so that we leave trails whenever the baseline is updated. 

The functional test will surely cover the deserializing a real lock file once the runtime is updated.

victorhurdugaci: Will this lock file be changed if I do dnu restore in the root of the solution folder?

troydai: It won't. The name is not `project.lock.json`. It's a simple text file serve as a test resource.

victorhurdugaci: Ah, I missed the file name :smile:

troydai:
:watch: 
/c c @Eilon @victorhurdugaci @BrennanConroy @muratg @davidfowl

test/Microsoft.Framework.Runtime.Tests/JsonDeserializerFacts.cs
troydai: Added.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
```diff
@@ -0,0 +1,312 @@
...
+                    buf.Append((char)next);
+                }
+            }
+
+            return buf.ToString();
+        }
+
+        private static bool IsWhitespace(int value)
+        {
+            return value == JsonConstants.SP || value == JsonConstants.HorizontalTab || value == JsonConstants.CarriageReturn;
```
davidfowl: SP?

src/Microsoft.Framework.Runtime.Hosting/Json/JsonConstants.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Framework.Runtime.Json
+{
+    internal class JsonConstants
+    {
+        public const int CarriageReturn = 0x0D;
```
davidfowl: What's up with the inconsistent casing?

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
```diff
@@ -0,0 +1,312 @@
...
+            {
+                ReadLiteral(JsonConstants.ValueFalse);
+                _token.Type = JsonTokenType.False;
+            }
+            else if (first == 'n')
+            {
+                ReadLiteral(JsonConstants.ValueNull);
+                _token.Type = JsonTokenType.Null;
+            }
+            else if ((first >= '0' && first <= '9') || first == '-')
```
ChengTian: The first condition should be `(first > '0' && first <= '9')`
```json
{
  "num": 01
}
```
is an invalid JSON

ChengTian: Wait, we should also consider
```json
{
  "num": 0
}
```
Need more sophisticated logic here

troydai: You're right.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonConstants.cs
troydai: Fixed.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
troydai: It's complicate here. `0`, `0.1` are legal but `01` is illegal. The responsibility of parsing number is delegated to `JsonNumber`. Should we want to throw on the case of leading zero, we can throw there.

But I do wondering if it worth to detect this error.

ChengTian: Yeah, let's throw inside JsonNumber. It's not complicated if we do it there.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializerResource.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        internal static string JSON_OpenString
+        {
+            get { return Format_InvalidSyntaxExpectation("JSON string", '\"'); }
+        }
+
+        internal static string JSON_InvalidEnd
+        {
+            get { return "Invalid JSON end unexpectedly."; }
```
Eilon: `Invalid JSON. Unexpected end of file.`

src/Microsoft.Framework.Runtime.Hosting/Json/JsonNumber.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        public double Double
+        {
+            get { return _double; }
+        }
+
+        public int Int
+        {
+            get { return (int)_double; }
```
Eilon: Not sure I saw a response to my earlier question about this. This cast doesn't seem safe to me...

src/Microsoft.Framework.Runtime.Hosting/Json/JsonObject.cs
```diff
@@ -0,0 +1,108 @@
...
+            JsonValue result;
+            if (!_data.TryGetValue(key, out result))
+            {
+                result = null;
+            }
+
+            return result;
+        }
+
+        public T ValueAs<T>(string key, Func<JsonValue, T> cast)
```
Eilon: This method doesn't seem terribly useful. Why can't the caller just grab the `JsonValue` and do whatever it wants? A method that takes in a delegate and just calls that delegate seems very roundabout.

src/Microsoft.Framework.Runtime.Hosting/project.json
```diff
@@ -23,8 +23,17 @@
         "dnxcore50": {
             "dependencies": {
                 "System.Xml.XDocument": "4.0.10-beta-*",
-                "System.Runtime.Loader": "4.0.0-beta-*"
+                "System.Runtime.Loader": "4.0.0-beta-*",
+                "System.Reflection": "4.0.10-beta-*",
```
Eilon: Sort 'em.

src/Microsoft.Framework.Runtime/GlobalSettings.cs
```diff
@@ -37,27 +37,32 @@ public static bool TryGetGlobalSettings(string path, out GlobalSettings globalSe
...
             {
-                settings = JObject.Parse(json);
+                using (var fs = File.OpenRead(globalJsonPath))
+                {
+                    var reader = new StreamReader(fs);
+                    var jobject = JsonDeserializer.Deserialize(reader) as JsonObject;
+
+                    if (jobject == null)
+                    {
+                        throw new InvalidOperationException("The json file can't be deserialized to a Json object.");
```
Eilon: `JSON` & `JSON`

Eilon: (And in other files too.)

src/Microsoft.Framework.Runtime/Project.cs
```diff
@@ -324,86 +349,78 @@ private static SemanticVersion SpecifySnapshot(string version, string snapshotVa
...
                     }
+                    else if (dependencyValue is JsonString)
+                    {
+                        // "dependencies" : { "Name" : "1.0" }
+                        dependencyVersionAsString = (JsonString)dependencyValue;
+                    }
+                    else
+                    {
+                        throw FileFormatException.Create(
+                            "Unrecoganizable format of dependency version of " + dependencyKey,
```
Eilon: Unrecognizable word `Unrecoganizable` :trollface:

muratg: Unidentifiable? Though I couldn't recognize the rest of the error either.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonNumber.cs
muratg: Ya, consider using `Convert.ToInt32(_double)` instead. Range of `double` is much larger than `int` and casting would not be throwing an exception, while the ToInt32 would.

Eilon: What I really want is an ability for callers of this to make smart decisions about whether the number is an integer or not, and for the *caller* to throw the exception because that'll have the most context.

src/Microsoft.Framework.Runtime/Project.cs
Eilon: Ok just to be clear: I was joking about the spelling :smile: But the message overall could probably be improved as well (the phrasing is awkward and unclear).

src/Microsoft.Framework.Runtime.Hosting/Json/JsonNumber.cs
victorhurdugaci: AFAIK, JSON (and Javascript) don't make the distinction between different types of numbers. Basically, if it looks like a number and parses like number, it must be a number (of some kind). Can we do the same and let the caller decide how to parse it further?

davidfowl: Yes plz.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonArray.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            if (array == null)
+            {
+                throw new ArgumentNullException(nameof(array));
+            }
+
+            _array = array;
+        }
+
+        public int Count
```
victorhurdugaci: `Length` (to be consistent with array)

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
```diff
@@ -0,0 +1,312 @@
...
+                {
+                    break;
+                }
+            }
+
+            _token.Value = null;
+            _token.Line = _line;
+            _token.Column = _column;
+
+            if (first == '{')
```
victorhurdugaci: Here's an idea that I just had: create a dictionary that maps characters to tokens. If the dictionary contains that key, you take the value. It would reduce all this code to 3 lines :)

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
```diff
@@ -0,0 +1,312 @@
...
+                    ReadNextChar();
+                }
+            }
+
+            var tail = _reader.Peek();
+            if (tail != '}' &&
+                tail != ']' &&
+                tail != ',' &&
+                tail != -1 &&
+                tail != JsonConstants.CarriageReturn &&
```
victorhurdugaci: Can we use characters instead of these constants? It is incosistent to have chars in some places and constants in others

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
```diff
@@ -0,0 +1,179 @@
...
+            {
+                throw new JsonDeserializerException(
+                    JsonDeserializerResource.Format_UnfinishedJSON(nextToken.Value),
+                    nextToken);
+            }
+
+            return result;
+        }
+
+        private static JsonValue DeserializeInternal(JsonToken next, JsonBuffer buffer)
```
victorhurdugaci: Instead of passing the buffer to every method, create a class field and make this class non static

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
davidfowl: Lets avoid dictionaries for simple things like this

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
```diff
@@ -0,0 +1,179 @@
...
+                {
+                    throw new JsonDeserializerException(
+                        JsonDeserializerResource.Format_InvalidSyntaxExpectation("JSON object", '}'),
+                        next);
+                }
+
+                if (next.Type == JsonTokenType.Colon)
+                {
+                    throw new JsonDeserializerException(
+                        JsonDeserializerResource.Format_InvalidSyntaxNotExpected("JSON object", ':'),
```
victorhurdugaci: This exception is not consistent with the rest. Everywhere else you tell what is expected, which is the right way to do it, but here you tell that something is not expected and not offer a solution (what's expected). why?

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializerException.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Framework.Runtime.Json
+{
+    internal class JsonDeserializerException : Exception
```
victorhurdugaci: Are you sure this is supposed to be internal? How is someone going to catch this?

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializerResource.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+            return string.Format("Invalid {0} syntax. Expected '{1}' or '{2}'.", syntaxName, expectation1, expectation2);
+        }
+
+        internal static string Format_InvalidUnicode(string unicode)
+        {
+            return string.Format("Invalid Unicode [{0}]", unicode);
+        }
+
+        internal static string Format_UnfinishedJSON(string nextTokenValue)
```
victorhurdugaci: What is the diff between this and `JSON_InvalidEnd` ?

src/Microsoft.Framework.Runtime.Hosting/Json/JsonToken.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Framework.Runtime.Json
+{
+    internal struct JsonToken
```
victorhurdugaci: Why struct?

src/Microsoft.Framework.Runtime.Hosting/Json/JsonTokenType.cs
```diff
@@ -0,0 +1,21 @@
...
+        RightCurlyBracket,  // ]
+        RightSquareBracket, // }
+        Colon,              // :
+        Comma,              // ,
+        Null,
+        True,
+        False,
+        Number,
+        String,
+        EOL
```
victorhurdugaci: EOF?

test/Microsoft.Framework.Runtime.Tests/JsonBufferFacts.cs
```diff
@@ -0,0 +1,151 @@
...
+        [Fact]
+        public void ReadTokenThroughLockFileSample()
+        {
+            using (var stream = File.OpenRead(".\\TestSample\\project.lock.sample"))
+            {
+                var reader = new StreamReader(stream);
+                var buffer = new JsonBuffer(reader);
+                Assert.NotNull(buffer);
+
+                while (buffer.Read().Type != JsonTokenType.EOL)
```
victorhurdugaci: Um... this test could pass even when things go wrong. If the buffer returns EOL after the first read, this test passes.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
victorhurdugaci: btw, nit: `IsWhitespace` already checks for `CR`

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
```diff
@@ -0,0 +1,179 @@
...
+
+                    next = buffer.Read();
+                    if (next.Type == JsonTokenType.RightCurlyBracket)
+                    {
+                        break;
+                    }
+                    else if (next.Type != JsonTokenType.Comma)
+                    {
+                        throw new JsonDeserializerException(
+                            JsonDeserializerResource.Format_InvalidSyntaxExpectation("JSON object", ','),
```
victorhurdugaci: Expected `,` or `}`

test/Microsoft.Framework.Runtime.Tests/JsonBufferFacts.cs
troydai: The `Read` will throw. The buffer won't return EOL after first read because the we control the sample. This is a smoke test on a lock file.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonToken.cs
troydai: Value type. Avoid frequent allocation.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializerResource.cs
troydai: Invalid end is the file finishes to early. This is file finishes to late, there are unprocessed content for example: 
```
{}
{}
```

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializerException.cs
troydai: It is captured and turned into `FileFormatException`

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
troydai: This is informational enough. JSON syntax is easy enough to figure out. And the exception message explain exactly what's the issue it runs into.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializer.cs
troydai: I agree with @Eilon . It is simpler to keep the `JsonDeserializer` a static class. Plus `TestReader` is a disposable class. I won't want to keep a handle of it.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
troydai: I don't get your point.
All the comparison are `int` to `int`.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonBuffer.cs
troydai: @victorhurdugaci did you notice i'm avoiding allocating at all cost?

src/Microsoft.Framework.Runtime.Hosting/Json/JsonNumber.cs
troydai: I'm with @victorhurdugaci. JSON's number type is basically double. Should this method be a concern I would rather remove it and expose the raw string.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonObject.cs
troydai: I just find it make client code feels more ... hipster ...

troydai: All right, removing it.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonDeserializerException.cs
troydai: This is not a deserializer for general purpose. Whoever consume it can definitely access the internal exception.

test/Microsoft.Framework.Runtime.Tests/JsonBufferFacts.cs
troydai: Should things really go wrong in the way that EOL is return in all cases all the tests will blow.

src/Microsoft.Framework.Runtime.Hosting/Json/JsonNumber.cs
Eilon: I don't have a particularly strong opinion on how to fix it. The important aspect is to ensure that callers can make smart decisions. Right now that's basically impossible because some properties will simply return bad data and the caller will have no idea.

troydai:
@Eilon @victorhurdugaci :watch:

troydai:
Rebased and ready for reviews.

Eilon:
:shipit: for me - I don't think that me looking at it anymore will yield anything useful :smile:

troydai:
Yeah. I collected one squirrel now.

src/Microsoft.Framework.Runtime.Hosting/FileFormatException.cs
```diff
@@ -2,69 +2,103 @@
...
         public int Column { get; private set; }
 
-        private FileFormatException WithLineInfo(IJsonLineInfo lineInfo)
+        internal static FileFormatException Create(Exception exception, string filePath)
         {
-            if (lineInfo != null)
+            if (exception is JsonDeserializerException)
+            {
+                return new FileFormatException(exception.Message, exception)
+                   .WithFilePath(filePath)
```
victorhurdugaci: I like this fluent api :)

victorhurdugaci:
:ship: it

troydai:
All right, I got a squirrel sitting on a ship now. All what I need is @davidfowl screaming at me one more time....

JamesNK:
If you want ideas for crazy JSON to throw at your reader, there are a ton of examples here: https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json.Tests/JsonTextReaderTest.cs

Lots of stuff I read is not technically JSON however. Multi-line strings, date constructors, octal numbers, etc.

davidfowl:
@JamesNK This parser is really for 2 things:
1. Don't use JSON.NET before the app's dependency graph loads
2. Performance loading 3 known file formats, project.lock.json, project.json, global.json.

We're going to use JSON.NET for everything else.

JamesNK:
I know. Just letting you know that you either need to handle those examples or fail well. They're what I've built up over the years.

davidfowl:
@JamesNK thanks for the input. Hopefully this will be used on those known quantities and we won't have to handle anything like that.

src/Microsoft.Framework.PackageManager/Restore/LockFileWriter.cs
```diff
@@ -0,0 +1,247 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
davidfowl: I think you should move the old lock file reader writer logic as a whole to the PackageManager and create a copy of the writer (unshared) in the runtime itself. The PackageManager needs to read the lock file as well but it can use the normal JSON.NET stuff.

src/Microsoft.Framework.Runtime/DependencyManagement/LockFileFormat.cs
```diff
@@ -17,91 +16,68 @@ public class LockFileFormat
         public const int Version = -9996;
         public const string LockFileName = "project.lock.json";
 
+        private string _currentLockFilePath;
```
davidfowl: This is pretty gross, I'd get rid of it. We don't call this from multiple threads but it looks kinda messy.

src/Microsoft.Framework.Runtime/Project.cs
```diff
@@ -98,9 +95,9 @@ public string ProjectDirectory
 
         public IProjectFilesCollection Files { get; private set; }
 
-        public IDictionary<string, string> Commands { get; } = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+        public IDictionary<string, string> Commands { get; private set; }
```
davidfowl: What's up with this change?

src/Microsoft.Framework.Runtime/project.json
davidfowl: :clap:

test/Microsoft.Framework.ApplicationHost.FunctionalTests/AppHostTests.cs
```diff
@@ -107,7 +107,7 @@ public void AppHostShowsErrorWhenGivenSubcommandWasNotFoundInProjectJson(string
         {
             var runtimeHomeDir = _fixture.GetRuntimeHomeDir(flavor, os, architecture);
             var projectStructure = @"{
-  'project.json': '{ }'
```
davidfowl: single quotes break?

davidfowl:
A few minor comments but :shipit: it looks great!

I have some things to follow up on after this initial checkin to do with performance :smile: !

src/Microsoft.Framework.Runtime/Project.cs
troydai: sure, I'll revert it.

test/Microsoft.Framework.ApplicationHost.FunctionalTests/AppHostTests.cs
troydai: Yes. I remove the support. Single quote is not supported according to JSON spec.

src/Microsoft.Framework.PackageManager/Restore/LockFileWriter.cs
troydai: I'll get a copy of updated lock file.

troydai: All right. I'll do that.

src/Microsoft.Framework.Runtime/DependencyManagement/LockFileFormat.cs
troydai: It is used to throw format exception. Let me see what I can do there.

Issue #1746, opened 2015-04-30
"dnu restore" can be "corrupted" by package server returning errors


bradwilson:
There is apparently some issue with myget.org right now which causes the DNU cache to become corrupted, which breaks "dnu restore". I'm basically constantly wiping out ~\AppData\Local\dnu to try to get rid of the corruption.

This problem seems preventable, because the HTTP response the MyGet is returning is a 503 with HTML inside it, something which should clearly not be saved in the local cache.

```
The ZIP archive C:\Users\Brad\AppData\Local\dnu\cache\b32144cb6963afda2cfb94aac3e707d76eda01b9$https_www.myget.org_F_xunit_\nupkg_xunit.assert.2.1.0-beta2-build2975.dat is corrupt
The ZIP archive C:\Users\Brad\AppData\Local\dnu\cache\b32144cb6963afda2cfb94aac3e707d76eda01b9$https_www.myget.org_F_xunit_\nupkg_xunit.assert.2.1.0-beta2-build2975.dat is corrupt
----------
System.IO.InvalidDataException: End of Central Directory record could not be found.
   at System.IO.Compression.ZipArchive.ReadEndOfCentralDirectory()
   at System.IO.Compression.ZipArchive.Init(Stream stream, ZipArchiveMode mode, Boolean leaveOpen)
   at System.IO.Compression.ZipArchive..ctor(Stream stream, ZipArchiveMode mode, Boolean leaveOpen, Encoding entryNameEncoding)
   at Microsoft.Framework.PackageManager.PackageUtilities.<OpenNuspecStreamFromNupkgAsync>d__1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.Framework.PackageManager.Restore.NuGet.NuGetv2Feed.<OpenNuspecStreamAsync>d__25.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.Framework.PackageManager.RemoteWalkProvider.<GetDependencies>d__7.MoveNext()
[...]

PR #1787, opened 2015-05-05
"dnu restore" invalidates illegal HTTP cache

ChengTian:
parent https://github.com/aspnet/dnx/issues/1746

`dnu restore` can now deal with misbehaving NuGet service servers. It will never cache an error page, even the server mistakenly returns the error page along with 200 status code.

@davidfowl @muratg @anurse

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
@@ -181,26 +181,35 @@ namespace Microsoft.Framework.PackageManager.Restore.NuGet
             return result;
         }
 
+        internal async Task InvalidateCacheFileAsync(string cacheKey)
+        {
+            var cacheFile = GetCacheFilePath(cacheKey);
+
+            // Acquire the lock on a file before we delete it to prevent this process
+            // from deleting a file that just passed existence check in some other processes
+            await ConcurrencyUtilities.ExecuteWithFileLocked(cacheFile, async _ =>
+            {
+                if (File.Exists(cacheFile))
+                {
+                    await Task.Factory.StartNew(() => File.Delete(cacheFile));
+                }
+
+                return 0;
+            });
+        }
+
         private async Task<HttpSourceResult> TryCache(string uri, string cacheKey, TimeSpan cacheAgeLimit)
         {
-            var baseFolderName = RemoveInvalidFileNameChars(ComputeHash(_baseUri));
-            var baseFileName = RemoveInvalidFileNameChars(cacheKey) + ".dat";
-
-#if DNX451
-            var localAppDataFolder = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
-#else
-            var localAppDataFolder = Environment.GetEnvironmentVariable("LocalAppData");
-#endif
-            var cacheFolder = Path.Combine(localAppDataFolder, "dnu", "cache", baseFolderName);
-            var cacheFile = Path.Combine(cacheFolder, baseFileName);
+            var cacheFile = GetCacheFilePath(cacheKey);
+            var cacheFolder = Path.GetDirectoryName(cacheFile);
 
             if (!Directory.Exists(cacheFolder) && !cacheAgeLimit.Equals(TimeSpan.Zero))
             {
                 Directory.CreateDirectory(cacheFolder);
             }
 
-            // Acquire the lock on a file before we open it to prevent this process
-            // from opening a file deleted by the logic in HttpSource.GetAsync() in another process
+            // Acquire the lock on a file before we open it to prevent this process from opening
+            // a file deleted by HttpSource.GetAsync()/HttpSource.InvalidateCacheFile() in another process
             return await ConcurrencyUtilities.ExecuteWithFileLocked(cacheFile, _ =>
             {
                 if (File.Exists(cacheFile))
@@ -274,6 +283,19 @@ namespace Microsoft.Framework.PackageManager.Restore.NuGet
             }
 
             return false;
+        }
+
+        private string GetCacheFilePath(string cacheKey)
+        {
+            var baseFolderName = RemoveInvalidFileNameChars(ComputeHash(_baseUri));
+            var baseFileName = RemoveInvalidFileNameChars(cacheKey) + ".dat";
+
+#if DNX451
+            var localAppDataFolder = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
+#else
+            var localAppDataFolder = Environment.GetEnvironmentVariable("LocalAppData");
+#endif
+            return Path.Combine(localAppDataFolder, "dnu", "cache", baseFolderName, baseFileName);
         }
 
         private static FileStream CreateAsyncFileStream(string path, FileMode mode, FileAccess access, FileShare share)
--- a/src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
@@ -14,10 +14,11 @@ namespace Microsoft.Framework.PackageManager.Restore.NuGet
 {
     public class NuGetv2Feed : IPackageFeed
     {
-        private static readonly XName _xnameEntry = XName.Get("entry", "http://www.w3.org/2005/Atom");
-        private static readonly XName _xnameTitle = XName.Get("title", "http://www.w3.org/2005/Atom");
-        private static readonly XName _xnameContent = XName.Get("content", "http://www.w3.org/2005/Atom");
-        private static readonly XName _xnameLink = XName.Get("link", "http://www.w3.org/2005/Atom");
+        private static readonly string _defaultNamespace = "http://www.w3.org/2005/Atom";
+        private static readonly XName _xnameEntry = XName.Get("entry", _defaultNamespace);
+        private static readonly XName _xnameTitle = XName.Get("title", _defaultNamespace);
+        private static readonly XName _xnameContent = XName.Get("content", _defaultNamespace);
+        private static readonly XName _xnameLink = XName.Get("link", _defaultNamespace);
         private static readonly XName _xnameProperties = XName.Get("properties", "http://schemas.microsoft.com/ado/2007/08/dataservices/metadata");
         private static readonly XName _xnameId = XName.Get("Id", "http://schemas.microsoft.com/ado/2007/08/dataservices");
         private static readonly XName _xnameVersion = XName.Get("Version", "http://schemas.microsoft.com/ado/2007/08/dataservices");
@@ -91,6 +92,7 @@ namespace Microsoft.Framework.PackageManager.Restore.NuGet
                     var uri = _baseUri + "FindPackagesById()?Id='" + id + "'";
                     var results = new List<PackageInfo>();
                     var page = 1;
+                    var cacheKey = string.Format("list_{0}_page{1}", id, page);
                     while (true)
                     {
                         // TODO: Pages for a package Id are cached separately.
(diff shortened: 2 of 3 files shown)


ChengTian:
Tested with a simple fake NuGet service written by me: https://github.com/ChengTian/NuGetServiceForTest/blob/master/src/NuGetServiceForTest/Controllers/PackageController.cs#L59

When `FindPackagesById` service is misbehaving, it looks like
invalidfindpackagesbyidresponse

When package downloading service is misbehaving, it looks like
invalidnupkgreturned

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
```diff
@@ -232,17 +240,49 @@ public PackageInfo BuildModel(string id, XElement element)
             });
         }
 
+        private void EnsureValidFindPackagesResponse(XDocument doc)
+        {
+            if (!string.Equals(doc.Root.GetDefaultNamespace().ToString(), _defaultNamespace,
+                StringComparison.OrdinalIgnoreCase))
```
ChengTian: A error page can be a valid XML file, so we use the default namespace to tell whether it is a valid NuGet service response

src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
```diff
@@ -181,26 +181,35 @@ internal class HttpSource
...
-#endif
-            var cacheFolder = Path.Combine(localAppDataFolder, "dnu", "cache", baseFolderName);
-            var cacheFile = Path.Combine(cacheFolder, baseFileName);
+            // Acquire the lock on a file before we delete it to prevent this process
+            // from deleting a file that just passed existence check in some other processes
+            await ConcurrencyUtilities.ExecuteWithFileLocked(cacheFile, async _ =>
+            {
+                if (File.Exists(cacheFile))
+                {
+                    await Task.Factory.StartNew(() => File.Delete(cacheFile));
```
moozzyk: any specific reason why this call needs to be `async`?

ChengTian: No, just because we need at least one `await` in an async block to get rid of warning. We can also remove the await here and `return Task.FromResult(0)` below. But there is no big difference I guess.

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
```diff
@@ -14,10 +14,11 @@ namespace Microsoft.Framework.PackageManager.Restore.NuGet
 {
     public class NuGetv2Feed : IPackageFeed
     {
-        private static readonly XName _xnameEntry = XName.Get("entry", "http://www.w3.org/2005/Atom");
-        private static readonly XName _xnameTitle = XName.Get("title", "http://www.w3.org/2005/Atom");
-        private static readonly XName _xnameContent = XName.Get("content", "http://www.w3.org/2005/Atom");
-        private static readonly XName _xnameLink = XName.Get("link", "http://www.w3.org/2005/Atom");
+        private static readonly string _defaultNamespace = "http://www.w3.org/2005/Atom";
```
moozzyk: I believe this should be XNamespace

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
```diff
@@ -232,17 +240,49 @@ public PackageInfo BuildModel(string id, XElement element)
             });
         }
 
+        private void EnsureValidFindPackagesResponse(XDocument doc)
+        {
+            if (!string.Equals(doc.Root.GetDefaultNamespace().ToString(), _defaultNamespace,
```
moozzyk: should be `doc.Root.Name.Namespace`. Currently I believe it would return `abc` for the following xml <ns1:element xmlns="abc" xmlns:ns1="xyz" /> even though `element` is in the `xyz` namespace.

moozzyk: Also if _defaultNamespace is defined as `XNamespace` you should be just able to the Equals operator on the `XNamespace` rather than using string comparison.

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
```diff
@@ -232,17 +240,49 @@ public PackageInfo BuildModel(string id, XElement element)
...
+                StringComparison.OrdinalIgnoreCase))
+            {
+                throw new FormatException("The XML is not a valid NuGet v2 service response to FindPackagesById() request");
```
moozzyk: Shouldn't this just be `InvalidOperationException`?

src/Microsoft.Framework.PackageManager/Restore/NuGet/PackageUtilities.cs
```diff
@@ -103,5 +103,33 @@ internal static ZipArchiveEntry GetEntryOrdinalIgnoreCase(this ZipArchive archiv
...
+            {
+                if (!File.Exists(path))
+                {
+                    return Task.FromResult(false);
+                }
+
+                using (var stream = File.OpenRead(path))
+                {
+                    try {
+                        using (var archive = new ZipArchive(stream, ZipArchiveMode.Read, leaveOpen: false))
```
moozzyk: I think you can just do `new ZipArchive(File.OpenRad(path), ZipArchive..., leaveOpen:false)`. Because you are using `leaveOpen: false` the file stream you are passing should be closed for you.

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
ChengTian: Yeah, you can say it is an invalid operation if you send HTTP requests to a server that responds with wrong contents. However, I guess using `FormatException` can give users more intuitive understanding of what is wrong here.

What's the principle you are following when you decide to use `InvalidOperationException`?

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
```diff
@@ -232,17 +240,49 @@ public PackageInfo BuildModel(string id, XElement element)
...
+                        var isValidNupkg = await PackageUtilities.IsValidNupkgAsync(package.Id, data.CacheFileName);
+                        if (!isValidNupkg)
+                        {
+                            await _httpSource.InvalidateCacheFileAsync(cacheKey);
+
+                            _reports.Information.WriteLine(
+                                "The response from {0} is not a valid NuGet package. Invalidating the cached file.",
+                                package.ContentUri.Yellow().Bold());
+
+                            var message = string.Format("{0} is not a valid ZIP archive of NuGet package",
```
moozzyk: nit: use string interpolation

src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
```diff
@@ -276,6 +285,19 @@ private static bool IsFileAlreadyOpen(string filePath)
             return false;
         }
 
+        private string GetCacheFilePath(string cacheKey)
```
moozzyk: I only see that we are writing the cache file and deleting the cache file but we never read it. This tells me that there is a copy of this logic somewhere where we read the cached file. If this is the case the we should just one function that calculate this path.

src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
moozzyk: I would go with returning `Task.FromResult(0)` - it does not schedule your task and will not use a separate thread. A bit expensive to get rid of a warning you can fix at compile time.

src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
ChengTian: The reading logic is in `TryCache()` and it is calling this method, too: https://github.com/aspnet/dnx/blob/wei/invalidate-illegal-http-cache/src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs#L215

There is no duplicate of this logic.

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
moozzyk: This is semantically not correct and can return `false` for a valid NuGet xml file since it relies on Xml being serialized the way you expect and when using namespaces there are at least, In most cases when an element is in a namespace it actually does not have the the default namespace since the namespace the element belongs to is either inherited from it's parent or the namespace-uri is bound to a namespace prefix. Therefore you need to get the namespace from the element name and not the default namespace that is declared along with attributes and other (non-default) namespaces.

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
moozzyk: As per MSDN "The exception that is thrown when the format of an argument is invalid, or when a composite format string is not well formed." The format here is Xml and because the file could be loaded into XDocument the it is a valid Xml document so the format is correct. If we had an XSD (hopefully we don't) for this Xml file we would have thrown an XmlValidationException because the document is not valid according to the schema but this is not format exception. 
`InvalidOperationException` is probably general enough that you can throw it almost everywhere where you can't find a better exception and you don't want to create a specialized exception.

src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
moozzyk: All good. I did not open the collapsed parts of the file and missed it.

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
ChengTian: I see. Then I'll use `InvalidOperationException` instead. Thanks for the detailed explanation!

davidfowl:
I think I would change how this was done. Make the call to httpSource.GetAsync take a function that lets the caller validate the response before caching. That will stop the bad data from being cached all in one place and keep the validation logic in one place. It can also happen before anything hits the disk.

src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
```diff
@@ -276,6 +285,19 @@ private static bool IsFileAlreadyOpen(string filePath)
...
+        {
+            var baseFolderName = RemoveInvalidFileNameChars(ComputeHash(_baseUri));
+            var baseFileName = RemoveInvalidFileNameChars(cacheKey) + ".dat";
+
+#if DNX451
+            var localAppDataFolder = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
+#else
+            var localAppDataFolder = Environment.GetEnvironmentVariable("LocalAppData");
```
BrennanConroy: I don't think this works for Cross-Plat

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
```diff
@@ -98,13 +100,15 @@ public Task<IEnumerable<PackageInfo>> FindPackagesByIdAsync(string id)
                         // However, (1) In most cases the pages grow rather than shrink;
                         // (2) cache for pages is valid for only 30 min.
                         // So we decide to leave current logic and observe.
-                        using (var data = await _httpSource.GetAsync(uri,
-                        string.Format("list_{0}_page{1}", id, page),
-                        retry == 0 ? _cacheAgeLimitList : TimeSpan.Zero))
+                        using (var data = await _httpSource.GetAsync(
+                            uri,
```
BrennanConroy: Formatting looks weird, could just be me though

src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
ChengTian: I am moving existent code to a helper method. Fixing cross-platform issues should be done in a separate thread. I filed an issue for it: https://github.com/aspnet/dnx/issues/1813

ChengTian:
Will start a new PR using the approach suggested by @davidfowl . Thanks for all your feedbacks. Most feedbacks here are still useful in the new implementation.

PR #1799, opened 2015-05-06
Fixing parsing coreclr package version

suhasj:
@davidfowl @BrennanConroy 

Issue: When building dnx, the copy-coreclr is trying to extract the version number of the core clr package. It is expecting it to be in the form `X.0.0-pre-buildNumber` and extracting the `buildNumber` value. The current version is `1.0.0-beta4` so this is failing with a parse exception. 

Fix: Adding the `TryParse` to see if it can be parsed., Rest functionality is unchanged

Also the fix needs to be in release since we are merging that again. I'll port it to dev once approved


BrennanConroy:
I think David put ship it on the old one... :shipit:

suhasj:
Talked to @Eilon and I can merge this to master and then copy the same to dev.

suhasj:
merged to master

Issue #1726, opened 2015-04-28
dnu packages add doesn't overwrite destination if it already exists


Eilon:
**Repro:**
1. Create a NuGet package (doesn't matter what's in it)
2. Run `dnu packages add MyPackage.1.0.0.nupkg c:\pkgs`
3. Modify the original NuGet package in some verifiable way (e.g. change something in the nuspec, add a file, etc.)
4. Run `dnu packages add MyPackage.1.0.0.nupkg c:\pkgs` *again*

**Result:**
Step (4) does nothing, though there's no warning of any sort. That is, the original "exploded" package remains in `c:\pkgs`, untouched.

**Expected:**
Step (4) should overwrite the destination.

----------

This in fact causes issues in our own build system where we publish packages to the volatile feed on the CI. For example with a lot of internal packages that we consume (e.g. Roslyn), their package id's and version numbers don't change on each build, so our build system thinks it's updating Roslyn builds in the volatile feed with `dnu packages add ...` when in fact it's no-op'ing.

Eilon:
@danroth27 @davidfowl

davidfowl:
Ugh, we should fix that.

Eilon:
@muratg can we get someone to take a look at this this week? This is a must-fix to ship even the next preview because we use this command on the CI and it doesn't work as we want it to.

davidfowl:
@ChengTian a simple change would be to compare the sha's. If they differ then overwrite. Also we should print out a message that shows what action was taken (overwrite, vs already exists).

PR #1811, opened 2015-05-07
"dnu packages add" overwrites when SHAs don't match

ChengTian:
parent https://github.com/aspnet/dnx/issues/1726

@davidfowl @Eilon @muratg , this should help fix issues (and may cause issues) on CI.

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Utils/NuGetPackageUtils.cs
+++ b/src/Microsoft.Framework.PackageManager/Utils/NuGetPackageUtils.cs
@@ -27,18 +27,33 @@ namespace Microsoft.Framework.PackageManager
             var targetPath = packagePathResolver.GetInstallPath(library.Name, library.Version);
             var targetNuspec = packagePathResolver.GetManifestFilePath(library.Name, library.Version);
             var targetNupkg = packagePathResolver.GetPackageFilePath(library.Name, library.Version);
-            var hashPath = packagePathResolver.GetHashPath(library.Name, library.Version);
+            var targetHashPath = packagePathResolver.GetHashPath(library.Name, library.Version);
 
             // Acquire the lock on a nukpg before we extract it to prevent the race condition when multiple
             // processes are extracting to the same destination simultaneously
-            await ConcurrencyUtilities.ExecuteWithFileLocked(targetNupkg, async createdNewLock =>
+            await ConcurrencyUtilities.ExecuteWithFileLocked(targetNupkg, async _ =>
             {
-                // If this is the first process trying to install the target nupkg, go ahead
-                // After this process successfully installs the package, all other processes
-                // waiting on this lock don't need to install it again.
-                if (createdNewLock && !File.Exists(targetNupkg))
+                string packageHash;
+                using (var sha512 = SHA512.Create())
                 {
-                    information.WriteLine($"Installing {library.Name.Bold()} {library.Version}");
+                    packageHash = Convert.ToBase64String(sha512.ComputeHash(stream));
+                }
+
+                var actionName = "Installing";
+                var installedPackageHash = string.Empty;
+                if (File.Exists(targetHashPath))
+                {
+                    installedPackageHash = File.ReadAllText(targetHashPath);
+                    actionName = "Overwriting";
+                }
+
+                if (string.Equals(packageHash, installedPackageHash, StringComparison.Ordinal))
+                {
+                    information.WriteLine($"{library.Name}.{library.Version} already exists");
+                }
+                else
+                {
+                    information.WriteLine($"{actionName} {library.Name}.{library.Version}");
 
                     Directory.CreateDirectory(targetPath);
                     using (var nupkgStream = new FileStream(
@@ -49,6 +64,7 @@ namespace Microsoft.Framework.PackageManager
                         bufferSize: 4096,
                         useAsync: true))
                     {
+                        stream.Seek(0, SeekOrigin.Begin);
                         await stream.CopyToAsync(nupkgStream);
                         nupkgStream.Seek(0, SeekOrigin.Begin);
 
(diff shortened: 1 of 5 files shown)


test/Bootstrapper.FunctionalTests/BootstrapperTestUtils.cs
```diff
@@ -31,43 +31,5 @@ public static class BootstrapperTestUtils
             stdErr = stdErrStr;
             return exitCode;
         }
-
-        public static DisposableDir PrepareTemporarySamplesFolder(string runtimeHomeDir)
```
ChengTian: I am moving this method to a more common util class, which is accessible in a bigger scope.

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuPackagesAddTests.cs
```diff
@@ -0,0 +1,132 @@
...
+using Microsoft.Framework.CommonTestUtils;
+using Newtonsoft.Json.Linq;
+using NuGet;
+using Xunit;
+
+namespace Microsoft.Framework.PackageManager
+{
+    public class DnuPackagesAddTests
+    {
+        private static readonly string ProjectName = "HelloWorld";
```
moozzyk: nit: since the string properties are private they could be just `const`

src/Microsoft.Framework.PackageManager/Utils/NuGetPackageUtils.cs
```diff
@@ -27,18 +27,33 @@ internal static class NuGetPackageUtils
...
+
+                var actionName = "Installing";
+                var installedPackageHash = string.Empty;
+                if (File.Exists(targetHashPath))
+                {
+                    installedPackageHash = File.ReadAllText(targetHashPath);
+                    actionName = "Overwriting";
+                }
+
+                if (string.Equals(packageHash, installedPackageHash, StringComparison.Ordinal))
```
moozzyk: I am wondering why we just can't overwrite the package always? For instance we always write the hash file and not only when we saved the package...

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuPackagesAddTests.cs
```diff
@@ -0,0 +1,132 @@
...
+
+                exitCode = DnuPackagesAddOutputPackage(tempSamplesDir, runtimeHomeDir, out stdOut);
+                Assert.Equal(0, exitCode);
+                Assert.Contains($"Overwriting {ProjectName}.{ProjectVersion}", stdOut);
+
+                var xDoc = XDocument.Load(packagePathResolver.GetManifestFilePath(ProjectName, ProjectVersion));
+                var actualDescription = xDoc.Root.Descendants()
+                    .Single(x => string.Equals(x.Name.LocalName, "description")).Value;
+                Assert.Equal("New", actualDescription);
+                Assert.NotEqual(lastInstallTime, new FileInfo(nuspecPath).LastWriteTimeUtc);
```
moozzyk: You may want to also verify that the SHA was updated.

src/Microsoft.Framework.PackageManager/Utils/NuGetPackageUtils.cs
```diff
@@ -27,18 +27,33 @@ internal static class NuGetPackageUtils
...
+                {
+                    information.WriteLine($"{library.Name.Bold()}.{library.Version} already exists");
```
Eilon: How about `...already exists and is identical`?

moozzyk: Actually thought about '...already exists, is identical and won't be overwritten`

Eilon: `...already exists, is identical and won't be overwritten, try changing the package so its hash will be different and then try again`? :trollface:

Eilon: Ok fine seriously: ` ...already exists and won't be overwritten because it is identical`

moozzyk: :+1: 

` ...already exists and won't be overwritten because it is identical and we don't overwrite identical files except for the SHA files` :smiley:

src/Microsoft.Framework.PackageManager/Utils/NuGetPackageUtils.cs
ChengTian: This util is also used by some other commands like `dnu restore`. We don't want to install the same package ten times during `dnu restore`

moozzyk: Sounds good. Thanks for the explanation.

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuPackagesAddTests.cs
```diff
@@ -0,0 +1,143 @@
...
+                Assert.NotEqual(hashBeforeReAdding, File.ReadAllText(hashFilePath));
+                Assert.Equal(newPackageHash, File.ReadAllText(hashFilePath));
```
ChengTian: @moozzyk , added SHA update verification here

ChengTian:
Thanks for all your feedbacks. Please check the updated patch.

moozzyk:
:ship::it:

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuPackagesAddTests.cs
```diff
@@ -0,0 +1,143 @@
...
+        public static IEnumerable<object[]> RuntimeComponents
+        {
+            get
+            {
+                return TestUtils.GetRuntimeComponentsCombinations();
+            }
+        }
+
+        [Theory]
+        [MemberData("RuntimeComponents")]
```
BrennanConroy: nameof()

ChengTian: Widely spreading problem in our tests, let's fix it in a separate PR.

muratg:
Looks good to me.  Pawel already gave the ship it :)

PR #1827, opened 2015-05-09
Add support for satellite resource assemblies

lodejard:
* Several APIs take AssemblyName instead of string
* nupkg dependencies with resource assemblies are loadable
* projects with localized embedded resources will also respond to culture-specific loading
* dnu build command writes out satellite assemblies to disk in culture subfolders


lodejard:
@davidfowl here's the progress so far - you might want to rebase your uber branch onto lodejard/resources to reduce merge conflicts

one thing that's still pending is the support for DTH assembly loading, not sure how to approach that. to test this in VS i've been setting the env var "DNX_COMPILATION_SERVER_PORT" to "n/a" in the amazing new super-fantastic debug launch tab

another thing is the localized assemblies aren't added to the nupkg yet. they do get written to disk, but the pack command just needs to look for those subfolder names to add more dlls to the nupkg

borgdylan:
Would this work even with non-roslyn projects just as the resx compilation does today?

PR #1836, opened 2015-05-11
Work around a Mono issue that makes restore unbearably slow

akoeplinger:
This is due to some form of contention when requests are processed concurrently. Restoring sequentially is *much* faster in this case.

I tested this on a sample project that would take ~70secs to restore before and it now takes 10secs consistently.

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Program.cs
+++ b/src/Microsoft.Framework.PackageManager/Program.cs
@@ -29,6 +29,14 @@ namespace Microsoft.Framework.PackageManager
 #if DNX451
             Thread.GetDomain().SetData(".appDomain", this);
             ServicePointManager.DefaultConnectionLimit = 1024;
+
+            // Work around a Mono issue that makes restore unbearably slow,
+            // due to some form of contention when requests are processed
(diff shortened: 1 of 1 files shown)


Tratcher:
Does the Mono HttpClient use ServicePoint still? How about setting ServicePoint.ConnectionLimit = 1?

davidfowl:
@Tratcher would be interesting to try that as well. Apparently it does respect service point manager but then it got hung up somewhere else.

akoeplinger:
Yes, it does use the limit defined in ServicePointManager (I see multiple requests being made like on .NET).

I just tried by setting ServicePointManager.DefaultConnectionLimit=1 and it had the same result as this PR (contrary to what I suspected). Any preferences?

davidfowl:
@akoeplinger service point manager solution is cleaner.

akoeplinger:
I agree, PR updated.

Tratcher:
Nice. Did you try it with 2, 5, and 10 for reference?  1024 is a bit high even for Windows.

akoeplinger:
Yes, anything >1 causes contention somewhere and massively slows it down, whatever the reason may be.

Issue #1804, opened 2015-05-06
"dnu restore" shouldn't show ArgumentNullException when it is unable to download a package


ChengTian:
If there are some issues with you network or NuGet service, after 3 attempts of package downloading, `dnu restore` will fail with the following message:

![capture](https://cloud.githubusercontent.com/assets/1383883/7504864/476b1150-f402-11e4-9538-12e7734b74f1.PNG)

It gives useful error messages along with an `ArgumentNullException`, which can be confusing.

PR #1845, opened 2015-05-12
"dnu restore" doesn't cache invalid contents

ChengTian:
parent https://github.com/aspnet/dnx/issues/1746

Use the approach suggested by @davidfowl 

Also did some cleanup, replacing `string.Format()` with string interpolation.

Also fixing https://github.com/aspnet/dnx/issues/1804 by adding just one statement.

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
@@ -118,71 +118,87 @@ namespace Microsoft.Framework.PackageManager.Restore.NuGet
             }
 #endif
 
-            var response = await _client.SendAsync(request);
-            response.EnsureSuccessStatusCode();
-
-            var responseStream = await response.Content.ReadAsStreamAsync();
-            ensureValidContents(responseStream);
-
-            var newFile = result.CacheFileName + "-new";
-
-            // Zero value of TTL means we always download the latest package
-            // So we write to a temp file instead of cache
-            if (cacheAgeLimit.Equals(TimeSpan.Zero))
+            // Write the response to a file and use the file stream to avoid memory explosion
+            using (var responseStream = new FileStream(
+                Path.GetTempFileName(),
+                FileMode.Create,
+                FileAccess.ReadWrite,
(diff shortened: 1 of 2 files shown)


src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
```diff
@@ -239,17 +243,59 @@ public PackageInfo BuildModel(string id, XElement element)
                 {
                     if (retry == 2)
                     {
-                        _reports.Error.WriteLine(string.Format("Error: DownloadPackageAsync: {1}\r\n  {0}", ex.Message, package.ContentUri.Red().Bold()));
+                        _reports.Error.WriteLine(
+                            $"Error: DownloadPackageAsync: {package.ContentUri}{Environment.NewLine}  {ex.Message}".Red().Bold());
+                        throw;
```
ChengTian: Adding re-throw here can fix https://github.com/aspnet/dnx/issues/1804

ChengTian:
@davidfowl , using file stream to avoid memory explosion now.

src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
```diff
@@ -117,67 +118,87 @@ internal class HttpSource
...
-            response.EnsureSuccessStatusCode();
-
-            var newFile = result.CacheFileName + "-new";
-
-            // Zero value of TTL means we always download the latest package
-            // So we write to a temp file instead of cache
-            if (cacheAgeLimit.Equals(TimeSpan.Zero))
+            // Write the response to a file and use the file stream to avoid memory explosion
+            using (var responseStream = new FileStream(
+                Path.GetTempFileName(),
```
davidfowl: Do we make sure this file is always deleted?

ChengTian: Yes, see `FileOptions.DeleteOnClose` option below :smile:

davidfowl:
:shipit:

PR #1846, opened 2015-05-12
[Proposal] Check global.json sdk version on startup

davidfowl:
- If there's a specific version in global.json it must match the runtime version
exactly. Show a nice error message if this isn't the case.

/cc @loudej @glennc @muratg

### DIFF
--- a/src/Microsoft.Framework.Runtime/ApplicationHostContext.cs
+++ b/src/Microsoft.Framework.Runtime/ApplicationHostContext.cs
@@ -32,11 +32,16 @@ namespace Microsoft.Framework.Runtime
             ProjectDirectory = projectDirectory;
             Configuration = configuration;
             RootDirectory = Runtime.ProjectResolver.ResolveRootDirectory(ProjectDirectory);
-            ProjectResolver = new ProjectResolver(ProjectDirectory, RootDirectory);
+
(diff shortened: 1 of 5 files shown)


davidfowl:
Will fix the CI if we think this is a good idea :smile: :

Eilon:
I like it. :shipit:

loudej:
We added the SDK version in the project "debug" settings specifically so you can hop around to test different versions. Doesn't this idea conflict with that feature?

Instead, how about adding that "minimum" SDK version to project.json we talked about? That's where node does its similar thing.

Sent from my Windows Phone

-----Original Message-----

davidfowl:
@loudej Yah I know, the problem is tho minimum won't help in this situation at all and we'll have 2 versions that can conflict (3 actually).

- The SDK version in global.json that tools look at. This version also becomes the default runtime version launched by tools like VS and VS Code.
- The minimum runtime version in project.json, which acutally does nothing for forcing that you want an exact version. We want to say that these beta4 samples require beta4 to run, not >= beta4. Do we support arbitrary ranges?
- Does the minimum runtime version persist into built package or is it a project.json thing only
- When you chose a specific SDK version in the tooling via the launchSettings that one wins but is it confusing that you can potentially pivot all 3?

Issue #1612, opened 2015-04-07
dnu pack fails when project.json is readonly.


phenning:
When running dnu pack on a project where the project.json is a read-only file, the following exception and callstack is generated. This impacts SCC (TFS) scenarios where the project.json is not checked out.

```
1>C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v14.0\DNX\Microsoft.DNX.targets(134,5): error : System.UnauthorizedAccessException: Access to the path 'c:\users\phenning\documents\visual studio 2015\Projects\artifacts\bin\ConsoleApp6\Debug\app\project.json' is denied.

1>C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v14.0\DNX\Microsoft.DNX.targets(134,5): error : at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)

1>C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v14.0\DNX\Microsoft.DNX.targets(134,5): error : at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)

1>C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v14.0\DNX\Microsoft.DNX.targets(134,5): error : at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)

1>C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v14.0\DNX\Microsoft.DNX.targets(134,5): error : at System.IO.StreamWriter.CreateFile(String path, Boolean append, Boolean checkHost)

1>C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v14.0\DNX\Microsoft.DNX.targets(134,5): error : at System.IO.StreamWriter..ctor(String path, Boolean append, Encoding encoding, Int32 bufferSize, Boolean checkHost)

[...]

PR #1850, opened 2015-05-12
Make sure lock file and global install project are not readonly

victorhurdugaci:
Fixes https://github.com/aspnet/dnx/issues/1612

cc @davidfowl

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Restore/LockFileFormat.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/LockFileFormat.cs
@@ -59,6 +59,12 @@ namespace Microsoft.Framework.PackageManager
 
         public void Write(string filePath, LockFile lockFile)
         {
+            // Make sure that if the lock file exists, it is not readonly
+            if (File.Exists(filePath))
+            {
+                FileOperationUtils.MakeWritable(filePath);
+            }
+
             using (var stream = new FileStream(filePath, FileMode.Create, FileAccess.Write, FileShare.None))
             {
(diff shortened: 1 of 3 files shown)


victorhurdugaci:
cc @ChengTian

ChengTian:
:shipit:

src/Microsoft.Framework.PackageManager/Restore/LockFileFormat.cs
```diff
@@ -59,6 +59,17 @@ public LockFile Read(Stream stream)
 
         public void Write(string filePath, LockFile lockFile)
         {
+            // Make sure that if the lock file exists, it is not readonly
+            if (File.Exists(filePath))
+            {
+                FileAttributes attributes = File.GetAttributes(filePath);
```
davidfowl: Any reason this code is duplicated?

moozzyk: :+1: 
(also use `var`?)

davidfowl: Also can't this fail sometimes? Can it ever fail?

victorhurdugaci: It can fail if you don't have permissions but we should let it fail. Let the callers decide if they want to handle the exception or not. In the two cases in which we use it, we want it to fail because there is nothing we can do if we cannot remove the readonly attribute

davidfowl:
:shipit:

Issue #1818, opened 2015-05-08
restore fails to find a wrapped csproj when its folder in in global.json/projects list


abpiskunov:
When adding project references, for csproj we wrap it and add a normal project reference to xproj file, for xproj we add its subfolder to global.json/projects list.

I tried to add reference to csproj file alone and it worked well, but when i added it's subfolder 'foo' in the global.json/projects list restore started to fail to resolve it.

Project.json:
![image](https://cloud.githubusercontent.com/assets/1592091/7538786/8b1f7f20-f570-11e4-8b8f-ee700bd44b02.png)

Global.json
![image](https://cloud.githubusercontent.com/assets/1592091/7538793/9759136e-f570-11e4-8d12-f829918adeaf.png)

my projects:
![image](https://cloud.githubusercontent.com/assets/1592091/7538802/a28c7e24-f570-11e4-8b8b-49a398e04cd1.png)

restore output:

![image](https://cloud.githubusercontent.com/assets/1592091/7538809/aeaa83fe-f570-11e4-82aa-68fe5ee1f15e.png)

ClassLibrary1's wrapped project.json:

{
  "version": "1.0.0-*",
  "frameworks": {
    "net452": {
      "wrappedProject": "../../foo/ClassLibrary1/ClassLibrary1.csproj",
      "bin": {
        "assembly": "../../foo/ClassLibrary1/obj/{configuration}/ClassLibrary1.dll",
        "pdb": "../../foo/ClassLibrary1/obj/{configuration}/ClassLibrary1.pdb"
      }
    }
  }
}

muratg:
@ChengTian

ChengTian:
Hi @abpiskunov , each element in `projects` of `global.json` should be a **parent** folder of your projects.

For example, you have `src/WebApplication5/project.json`, then `src` should be added to `projects`, not `src/WebApplication5`

You can fix it by replacing `foo` with `.` (dot, i.e. current folder).

ChengTian:
@davidfowl , it's actually quite intuitive that I should be able to add `src/MyProject` to `projects` in `global.json`. Is there any specific reason that we don't support it?

davidfowl:
it shouldn't be that big of a perf hit i guess. Verify that it doesn't

abpiskunov:
[...]

PR #1854, opened 2015-05-13
ProjectResolver throws when resolving ambiguous project names

ChengTian:
parent https://github.com/aspnet/dnx/issues/1818

### DIFF
--- a/src/Microsoft.Framework.Runtime/ProjectResolver.cs
+++ b/src/Microsoft.Framework.Runtime/ProjectResolver.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Framework.Runtime
     public class ProjectResolver : IProjectResolver
     {
         private readonly HashSet<string> _searchPaths = new HashSet<string>();
-        private readonly Dictionary<string, ProjectInformation> _projects = new Dictionary<string, ProjectInformation>();
+        private ILookup<string, ProjectInformation> _projects;
 
         public ProjectResolver(string projectPath)
         {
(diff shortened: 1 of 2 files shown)


src/Microsoft.Framework.Runtime/ProjectResolver.cs
```diff
@@ -37,14 +37,21 @@ public bool TryResolveProject(string name, out Project project)
...
+                return false;
             }
 
-            return false;
+            var candidates = _projects[name].Where(p => File.Exists(Path.Combine(p.FullPath, Project.ProjectFileName)));
+            if (candidates.Count() > 1)
+            {
+                var allCandidatePaths = string.Join(Environment.NewLine, candidates.Select(x => x.FullPath).OrderBy(x => x));
+                throw new InvalidOperationException(
+                    $"'{name}' is an ambiguous name resolved to following projects:{Environment.NewLine}{allCandidatePaths}");
```
davidfowl: `The project name {name} is ambiguous between the following projects`

src/Microsoft.Framework.Runtime/ProjectResolver.cs
```diff
@@ -37,14 +37,21 @@ public bool TryResolveProject(string name, out Project project)
...
-            if (_projects.TryGetValue(name, out projectInfo))
+            if (!_projects.Contains(name))
             {
-                project = projectInfo.Project;
-                return project != null;
+                return false;
             }
 
-            return false;
+            var candidates = _projects[name].Where(p => File.Exists(Path.Combine(p.FullPath, Project.ProjectFileName)));
```
davidfowl: This code paths runs every time somebody asks for a project. Checking the existence is why ProjectInformation.Project is lazy and only ever runs once.

ChengTian:
@davidfowl , thanks for the feedback, updated.

ChengTian:
Just came up with an edge case where you have the following `global.json`:
```json
{
  "projects": ["src", "src/../src"]
}
```
This will make `ProjectResolver` throw, even you only have a single project `src/ProjectA`. And the error message can be very confusing:
```
The project name 'ProjectA' is ambiguous between the following projects:
<Root>/src/ProjectA
<Root>/src/ProjectA
```

Need to dedup. Working on it.

ChengTian:
@davidfowl , added dedup of project path and added tests for the case above.

davidfowl:
Good catch! :shipit:

PR #1860, opened 2015-05-13
Use ResetColor and Revert AnsiConsole

BrennanConroy:
@davidfowl @victorhurdugaci @muratg

### DIFF
--- a/src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/AnsiConsole.cs
+++ b/src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/AnsiConsole.cs
@@ -9,33 +9,18 @@ namespace Microsoft.Framework.Runtime.Common.CommandLine
 {
     internal class AnsiConsole
     {
-        private AnsiConsole(TextWriter writer, bool useConsoleColor)
+        private AnsiConsole(TextWriter writer)
         {
             Writer = writer;
-
-            _useConsoleColor = useConsoleColor;
-            if (_useConsoleColor)
-            {
-                OriginalForegroundColor = Console.ForegroundColor;
-            }
         }
 
         private int _boldRecursion;
-        private bool _useConsoleColor;
 
-        public static AnsiConsole GetOutput(bool useConsoleColor)
-        {
-            return new AnsiConsole(Console.Out, useConsoleColor);
-        }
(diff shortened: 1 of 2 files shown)


src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/AnsiConsole.cs
```diff
@@ -9,34 +9,19 @@ namespace Microsoft.Framework.Runtime.Common.CommandLine
...
         }
 
         private int _boldRecursion;
-        private bool _useConsoleColor;
 
-        public static AnsiConsole GetOutput(bool useConsoleColor)
-        {
-            return new AnsiConsole(Console.Out, useConsoleColor);
-        }
+        public static AnsiConsole Output = new AnsiConsole(Console.Out);
```
moozzyk: So now anyone can set it from outside?

BrennanConroy: This is the original code

moozzyk: The original code had a method that created the instance each time it was called. It was changed to a public static field. It should be a public static property with just a getter that creates a new console each time or, if we care, create `private static readonly` (or lazily initialized) fields that would be returned from static properties.

BrennanConroy: The original code: https://github.com/aspnet/dnx/commit/d751821e0ae182e826f33dea3e7520a5f34a0383#diff-70f0f8d78afc1288add29e793af6cf3dR18
We could change it like you said though.

moozzyk: Oh, I only compared against the latest version. Please change the code so that we don't allow setting the static fields - we don't need it and it is ugly...

davidfowl:
:shipit: Assuming you tested it on all platforms. 

You also have to revert all of those changes you made to all the other repos :smile:

BrennanConroy:
Ya, I think I only needed to change 1 other repo. But I'll make sure.

davidfowl:
@BrennanConroy hold off on this a second.

BrennanConroy:
@davidfowl Has the second ended yet?

Issue #1732, opened 2015-04-28
Restore should take multiple project.json paths as command line parameter


PradeepKadubandi:
This is so that tooling can kick off restores for different projects with one dnx process. (Came up as part of restore discussion)

PR #1867, opened 2015-05-14
 Make "dnu restore" accept multiple restore root dirs 

ChengTian:
parent https://github.com/aspnet/dnx/issues/1732

`dnu restore <dir1> <dir2> <dir3> ...` and dnu will restore for each path specified as an argument.

@davidfowl

### DIFF
--- a/src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandArgument.cs
+++ b/src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandArgument.cs
@@ -2,13 +2,28 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 
 namespace Microsoft.Framework.Runtime.Common.CommandLine
 {
     internal class CommandArgument
     {
+        public CommandArgument()
+        {
+            Values = new List<string>();
+        }
+
         public string Name { get; set; }
         public string Description { get; set; }
-        public string Value { get; set; }
+        public List<string> Values { get; private set; }
+        public bool MultipleValues { get; set; }
+        public string Value
+        {
+            get
+            {
+                return Values.FirstOrDefault();
+            }
+        }
     }
 }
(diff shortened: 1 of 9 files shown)


src/Microsoft.Framework.PackageManager/Program.cs
```diff
@@ -61,7 +61,9 @@ public int Main(string[] args)
             {
                 c.Description = "Restore packages";
 
-                var argRoot = c.Argument("[root]", "Root of all projects to restore. It can be a directory, a project.json, or a global.json.");
+                var argRoot = c.Argument("[root]",
+                    "Root of projects to restore. A value can be a path to directory/project.json/global.json. Accepts multiple values.",
+                    multipleValues: true);
```
ChengTian: @Eilon , I bet you have a better idea about help text of the argument here. This argument can have multiple values:
```
dnu restore <path1> <path2> <path3>
```
The three paths are all values of the argument `[root]`

davidfowl:
As part of this change can you make it so that we don't recurse if the current directory has a `project.json`?

src/Microsoft.Framework.PackageManager/Program.cs
Eilon: Can the name of the "argument" be `[projects]` instead?

`List of projects and project folders to restore. The values can be: a path to project.json or global.json file, or a folder containing a project.json or global.json.`

ChengTian: Yeah, `roots` is not a clearly defined concept. It's better to known concept `project`. Will change to `projects`

ChengTian: I am going to make the change mentioned by @davidfowl below. So if there is no `project.json` in given folder, we recursively traverse all subfolders and restore for all `project.json` we see. So we need to revise the help info a little bit:
```
List of projects and project folders to restore. The values can be: a path to project.json or global.json file, or a folder having project.json in subfolders
```

Eilon: Slight tweak (also fixed typo from my earlier string): ` List of projects and project folders to restore. Each value can be: a path to a project.json or global.json file, or a folder to recursively search for project.json files.`

ChengTian:
@davidfowl , I added that part in the last commit.

davidfowl:
@ChengTian why is the build failing?

ChengTian:
@davidfowl , build passed

ChengTian:
@davidfowl , our CI will be fixed today. Please sign off so that Tooling team can integrate this feature and verify it.

ChengTian:
@davidebbo , rebased and `build verify` locally

davidebbo:
@ChengTian wrong David I think ;)

Eilon:
@davidfowl !

davidfowl:
:shipit:

Issue #1842, opened 2015-05-11
dnu pack with multiple configurations will nest output directories


smbecker:
According to the docs on dnu pack, --configuration can be specified multiple times to pack multiple configurations. However, if you do that, then [this line](https://github.com/aspnet/dnx/blob/6f3fc248ab29baf4bbfa0218967fe46ac5c76cc4/src/Microsoft.Framework.PackageManager/Building/BuildManager.cs#L101) will cause the output directories to be nested within the each previous configuration. The following command

```
dnu pack --configuration Debug --configuration --Release
```

will cause the artifacts from Debug to be placed in <out>/Debug and the artifacts from Release to be placed in <out>/Debug/Release.

If desired, I can submit a PR after going through the necessary contribution paperwork.

ChengTian:
Thanks for reporting this and pinpointing the line in code.

@muratg , this should be beta5 bug

PR #1876, opened 2015-05-15
Fix build output path when more than one configuration is specified

ChengTian:
parent https://github.com/aspnet/dnx/issues/1842


davidfowl:
:shipit:

PR #1877, opened 2015-05-15
Path.Combine throws "Illegal character in path"

metadings:
The makefile.shade did throw "Illegal character in path", I assume because there's a path with quotes around.

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -1049,16 +1049,19 @@ functions @{
             return CXX;
         }
 
-        foreach(var dir in Environment.GetEnvironmentVariable("PATH").Split(Path.PathSeparator))
+        foreach(var dir in Environment.GetEnvironmentVariable("PATH").Split(new string[] { ";" }, StringSplitOptions.None))
         {
-            if (File.Exists(Path.Combine(dir, "clang++")))
-            {
-                return Path.Combine(dir, "clang++");
-            }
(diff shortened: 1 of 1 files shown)


makefile.shade
```diff
@@ -1049,16 +1049,19 @@ functions @{
             return CXX;
         }
 
-        foreach(var dir in Environment.GetEnvironmentVariable("PATH").Split(Path.PathSeparator))
+        foreach(var dir in Environment.GetEnvironmentVariable("PATH").Split(new string[] { ";" }, StringSplitOptions.None))
```
davidfowl: This change is wrong. The path separation on non windows oses is ':' thats why we use the Path separator. Can you give an example or where things went wrong.

makefile.shade
```diff
@@ -1049,16 +1049,19 @@ functions @{
...
         {
-            if (File.Exists(Path.Combine(dir, "clang++")))
-            {
-                return Path.Combine(dir, "clang++");
-            }
-            else if (File.Exists(Path.Combine(dir, "clang++-3.5")))
-            {
-                return Path.Combine(dir, "clang++-3.5");
+            try {
```
davidfowl: This is also pretty terrible. Try catch swallow. Better to fix things properly.

metadings: You're right... I'm going to remove this PR :)

PR #1879, opened 2015-05-15
React to CoreCLR update

troydai:
@pranavkm @Tratcher 

Fix build errors show when build locally against volatile feed. I can't repo the build error on Mono CI.


davidfowl:
What happened here?

pranavkm:
This was an attempt to pull in the transitive dependencies in. We could probably close this since we decided not to use this approach.

PR #1886, opened 2015-05-18
Enabling STL in dnx.dll

moozzyk:
With this change I am introducing a new dnx.exe for Windows whose only purpose is to load the right version of the dnx_xxx.dll. The old dnx.exe is compiled as dnx_win32.dll and dnx_onecore.dll. The only difference between dnx_win32.dll and dnx_onecore.dll is how they are built. The result of this change is that it enables using STL in the dnx_xxx.dll. We may need/want to do similar split to dnx.coreclr.dll in the future if we needed to use STL there.

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -61,7 +61,7 @@ var BOOTSTRAPPER_HOST_NAME = 'dnx.host'
 var PROGRAM_FILES_X86 = '${Environment.GetFolderPath(Environment.SpecialFolder.ProgramFilesX86)}'
 var MSBUILD = '${Path.Combine(PROGRAM_FILES_X86, "MSBuild", "14.0", "Bin", "MSBuild.exe")}'
 var VS_REDIST_ROOT = '${Path.Combine(PROGRAM_FILES_X86, @"Microsoft Visual Studio 14.0\VC\redist")}'
-var WIN10_SDK = '${Path.Combine(PROGRAM_FILES_X86, @"Windows Kits\10")}'
+var WIN10_SDK_LIB = '${Path.Combine(PROGRAM_FILES_X86, @"Windows Kits\10\Lib")}'
 var CLANG = '${SearchForClang()}'
 
 var MANAGED_PROJECTS = '${FindAllProjects(
@@ -89,6 +89,8 @@ var WIN_MANAGED_PROJECTS = '${FindAllProjects(
 var RC_FILES = '${FindAllFiles("Resource.rc", BOOTSTRAPPER_FOLDER_NAME, BOOTSTRAPPER_CLR_NAME, BOOTSTRAPPER_CORECLR_NAME)}'
 
 var RUNTIME_TARGETS='${new List<RuntimeTarget>()}'
+
+var CAN_BUILD_ONECORE = '${Directory.Exists(WIN10_SDK_LIB) && Directory.GetFiles(WIN10_SDK_LIB, "onecore.lib", SearchOption.AllDirectories).Any()}'
 
 @{
     // Always build mono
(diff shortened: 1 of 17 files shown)


makefile.shade
```diff
@@ -206,9 +209,9 @@ var RUNTIME_TARGETS='${new List<RuntimeTarget>()}'
         }
     }
 
-#build-dnx-win .ensure-msbuild .update-rc-files target='build-windows'
-    var bootstrapperProj = '${Path.Combine(ROOT, "src", BOOTSTRAPPER_EXE_NAME, BOOTSTRAPPER_EXE_NAME + ".vcxproj")}'
-    @{
+#build-dnx-win-exe .ensure-msbuild .update-rc-files target='build-windows'
+    var bootstrapperProj = '${Path.Combine(ROOT, "src", BOOTSTRAPPER_WINDOWS_FOLDER_NAME, "dnx.windows.vcxproj")}'
+    @{
```
BrennanConroy: Spaces

makefile.shade
```diff
@@ -206,9 +209,9 @@ var RUNTIME_TARGETS='${new List<RuntimeTarget>()}'
         }
     }
 
-#build-dnx-win .ensure-msbuild .update-rc-files target='build-windows'
-    var bootstrapperProj = '${Path.Combine(ROOT, "src", BOOTSTRAPPER_EXE_NAME, BOOTSTRAPPER_EXE_NAME + ".vcxproj")}'
-    @{
+#build-dnx-win-exe .ensure-msbuild .update-rc-files target='build-windows'
+    var bootstrapperProj = '${Path.Combine(ROOT, "src", BOOTSTRAPPER_WINDOWS_FOLDER_NAME, "dnx.windows.vcxproj")}'
```
BrennanConroy: dnx.windows is defined above, use that for vcxproj?

makefile.shade
```diff
@@ -224,8 +227,25 @@ var RUNTIME_TARGETS='${new List<RuntimeTarget>()}'
         Exec(MSBUILD, bootstrapperProj + " /p:Configuration=" + Configuration2 + ";Platform=x64;TargetFramework=dnxcore50;RuntimeType=CORECLR_WIN" + configRuntimeTargetOS);
     }
 
-    directory delete='${Path.Combine(BUILD_DIR2, BOOTSTRAPPER_EXE_NAME)}'
-    copy sourceDir='${Path.Combine(ROOT, "src", BOOTSTRAPPER_EXE_NAME)}' include='bin/**/' outputDir='${Path.Combine(BUILD_DIR2, BOOTSTRAPPER_EXE_NAME)}' overwrite='${true}'
+    directory delete='${Path.Combine(BUILD_DIR2, BOOTSTRAPPER_WINDOWS_FOLDER_NAME)}'
+    copy sourceDir='${Path.Combine(ROOT, "src", BOOTSTRAPPER_WINDOWS_FOLDER_NAME)}' include='bin/**/' outputDir='${Path.Combine(BUILD_DIR2, BOOTSTRAPPER_WINDOWS_FOLDER_NAME)}' overwrite='${true}'
+
+#build-dnx-win-dll .ensure-msbuild .update-rc-files target='build-windows'
+    var bootstrapperProj = '${Path.Combine(ROOT, @"src\dnx\dnx.vcxproj")}'
```
BrennanConroy: nit: We use the predefined variable everywhere else

src/dnx.windows/dnx.cpp
```diff
@@ -0,0 +1,35 @@
...
+    {
+        auto last_error = GetLastError();
+        _tprintf(L"%s could not be loaded. Last error: %d\n", dnx_dll_name, last_error);
+        return -1;
+    }
+
+    auto entry_point = (int (STDAPICALLTYPE*)(int, wchar_t **))GetProcAddress(dnx_dll, "DnxMain");
+    if (!entry_point)
+    {
+        _tprintf(L"Getting entry point failed\n");
```
BrennanConroy: exit here?

src/dnx.windows/dnx.cpp
```diff
@@ -0,0 +1,35 @@
...
+#include "stdafx.h"
+
+int _tmain(int argc, _TCHAR* argv[])
+{
+    OSVERSIONINFO version_info;
+    ZeroMemory(&version_info, sizeof(OSVERSIONINFO));
+    version_info.dwOSVersionInfoSize = sizeof(OSVERSIONINFO);
+
+#pragma warning(disable:4996)
+    GetVersionEx(&version_info);
```
BrennanConroy: Is there a reason we don't use the Version Helper functions, since this is deprecated?

src/dnx.windows/dnx.cpp
moozzyk: Yes. Will fix.

src/dnx.windows/dnx.cpp
moozzyk: Yes. They are not supported in the libs we have to link against.

makefile.shade
moozzyk: Not sure. It is for folder name and that's why I did not want to use it. The assumption that the project name and the folder name are the same does not have to be correct.

makefile.shade
moozzyk: Fixed

makefile.shade
moozzyk: Leaving as is.

makefile.shade
moozzyk: Fixing.

Issue #862, opened 2014-11-06
Avoid NRE for `kpm.cmd help unsupported`


dougbu:
`kpm.cmd unsupported` displays an Exception containing
```
TODO: Error: unrecognized argument 'unsupported'
```
Expect similar output from `kpm.cmd help unsupported` but that command instead displays
```
System.NullReferenceException: Object reference not set to an instance of an object.
```

BrennanConroy:
71cb0ee4c1a01efe4fdae5ef9d58e839604fe1d3

PR #1893, opened 2015-05-18
Fix null-ref on bad arg to help

BrennanConroy:
#862

### DIFF
--- a/src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandLineApplication.cs
+++ b/src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandLineApplication.cs
@@ -332,6 +332,10 @@ namespace Microsoft.Framework.Runtime.Common.CommandLine
             else
             {
                 target = Commands.SingleOrDefault(cmd => string.Equals(cmd.Name, commandName, StringComparison.OrdinalIgnoreCase));
+                if (target == null)
+                {
+                    throw new InvalidOperationException(string.Format("'{0}' is not a command", commandName));
+                }
                 headerBuilder.AppendFormat(" {0}", commandName);
             }
 


src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandLineApplication.cs
```diff
@@ -332,6 +332,10 @@ public void ShowHelp(string commandName = null)
             else
             {
                 target = Commands.SingleOrDefault(cmd => string.Equals(cmd.Name, commandName, StringComparison.OrdinalIgnoreCase));
+                if (target == null)
+                {
+                    throw new Exception(string.Format("TODO: Error: unexpected value '{0}' for option 'help'", commandName));
```
muratg: String interpolation

muratg:
:shipit:

src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandLineApplication.cs
BrennanConroy: Sent from my HTC

----- Reply message -----

moozzyk: Don't throw `Exception`. This should be either `ArgumentException` or `InvalidOperationException`

BrennanConroy:
Updated exception. We can't use c#6 features because when building dnx-mono we use an old .NET compiler version. @davidfowl said there might be a way around this

src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandLineApplication.cs
```diff
@@ -332,6 +332,10 @@ public void ShowHelp(string commandName = null)
             else
             {
                 target = Commands.SingleOrDefault(cmd => string.Equals(cmd.Name, commandName, StringComparison.OrdinalIgnoreCase));
+                if (target == null)
+                {
+                    throw new ArgumentException(string.Format("TODO: Error: unexpected value '{0}' for option 'help'", commandName));
+                }
```
moozzyk: consider passing the name of the parameter as the second parameter to the `ArgumentException` ctor. (Note, I am not feeling strongly about this - this is command line so probably parameter names don't matter)

moozzyk:
:ship::it:

src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandLineApplication.cs
davidfowl: Get rid of the TODO

src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandLineApplication.cs
```diff
@@ -332,6 +332,10 @@ public void ShowHelp(string commandName = null)
             else
             {
                 target = Commands.SingleOrDefault(cmd => string.Equals(cmd.Name, commandName, StringComparison.OrdinalIgnoreCase));
+                if (target == null)
+                {
+                    throw new ArgumentException(string.Format("Error: unexpected value '{0}' for option 'help'", commandName), "commandName");
+                }
```
ChengTian: If we run into this case, it means that the user is trying to do something like
```
dnu help invalid_subcommand
```
We may want to make the message more friendly:
```
$"'{commandName}' is not a subcommand of '{cmd.Parent.Name}'"
```

Issue #1891, opened 2015-05-18
We don't need breadcrumbs without version


victorhurdugaci:
Today we write two breadcrumbs for each package: one with version and one without. We don't need to one without version because a patch is always targeting a particular version

Issue #1889, opened 2015-05-18
Need a breadcrumb for runtime that includes architecture


victorhurdugaci:
Mostly, we don't care about architecture when servicing but for DNX we want both the architecture and the framework type

Issue #1887, opened 2015-05-18
Servicing index is in the wrong folder


victorhurdugaci:
It is under `program files\DNX`. Should be instead under `program files\Microsoft DNX`

moozzyk:
Shouldn't it be in `ProgramData\Microsoft DNX`?

victorhurdugaci:
Nope, only the breadcrumbs are under `Program Data`

moozzyk:
Good to know where to read it from in the native parts of the stack then... ;)

PR #1899, opened 2015-05-19
Breadcrumbs and servicing updates

victorhurdugaci:
Fixes https://github.com/aspnet/dnx/issues/1891, https://github.com/aspnet/dnx/issues/1889, https://github.com/aspnet/dnx/issues/1887

- Write only breadcrumbs with version
- Use the correct folder for the service index
- Include version information in the runtime breadcrumb

### DIFF
--- a/src/Microsoft.Framework.Runtime.Sources/Impl/Constants.cs
+++ b/src/Microsoft.Framework.Runtime.Sources/Impl/Constants.cs
@@ -9,6 +9,7 @@ namespace Microsoft.Framework.Runtime
         public const string BootstrapperFullName = "Microsoft .NET Execution environment";
         public const string DefaultLocalRuntimeHomeDir = ".dnx";
         public const string RuntimeShortName = "dnx";
+        public const string RuntimeLongName = "Microsoft DNX";
         public const string RuntimeNamePrefix = RuntimeShortName + "-";
         public const string WebConfigRuntimeVersion = RuntimeNamePrefix + "version";
         public const string WebConfigRuntimeFlavor = RuntimeNamePrefix + "clr";
--- a/src/Microsoft.Framework.Runtime/DefaultHost.cs
+++ b/src/Microsoft.Framework.Runtime/DefaultHost.cs
@@ -192,22 +192,9 @@ Please make sure the runtime matches a framework specified in {Project.ProjectFi
 
         private void AddRuntimeServiceBreadcrumb()
         {
-#if DNX451
-            var runtimeAssembly = typeof(Servicing.Breadcrumbs).Assembly;
-#else
-            var runtimeAssembly = typeof(Servicing.Breadcrumbs).GetTypeInfo().Assembly;
-#endif
-
-            var version = runtimeAssembly
-                ?.GetCustomAttribute<AssemblyInformationalVersionAttribute>()
-                ?.InformationalVersion;
-
-            if (!string.IsNullOrWhiteSpace(version))
-            {
-                var semanticVersion = new NuGet.SemanticVersion(version);
-                Servicing.Breadcrumbs.Instance.AddBreadcrumb(runtimeAssembly.GetName().Name, semanticVersion);
-            }
+            var env = (IRuntimeEnvironment)ServiceProvider.GetService(typeof(IRuntimeEnvironment));
+            var frameworkBreadcrumbName = $"{Constants.RuntimeShortName}-{env.RuntimeType}-{env.RuntimeArchitecture}-{env.RuntimeVersion}";
+            Servicing.Breadcrumbs.Instance.AddBreadcrumb(frameworkBreadcrumbName);
         }
-
     }
 }
--- a/src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
+++ b/src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Framework.Runtime.Servicing
     public class Breadcrumbs
     {
         // It is recommended that you disable breadcrumbs only for perf runs
-        private const string NoBreadcrumbsEnvironmentVariableName = "DNX_NO_BREADCRUMBS";
+        private static readonly string NoBreadcrumbsEnvironmentVariableName = Constants.RuntimeShortName.ToUpper() + "_NO_BREADCRUMBS";
 
         private static readonly string _logType = typeof(Breadcrumbs).Name;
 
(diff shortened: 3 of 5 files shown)


victorhurdugaci:
cc @moozzyk @joeloff

src/Microsoft.Framework.Runtime.Sources/Impl/Constants.cs
```diff
@@ -9,6 +9,7 @@ internal static class Constants
         public const string BootstrapperFullName = "Microsoft .NET Execution environment";
         public const string DefaultLocalRuntimeHomeDir = ".dnx";
         public const string RuntimeShortName = "dnx";
+        public const string RuntimeLongName = "Microsoft " + RuntimeShortName;
```
davidfowl: bizzare

src/Microsoft.Framework.Runtime/DefaultHost.cs
```diff
@@ -192,22 +192,9 @@ private void Initialize(DefaultHostOptions options, IServiceProvider hostService
...
-            var version = runtimeAssembly
-                ?.GetCustomAttribute<AssemblyInformationalVersionAttribute>()
-                ?.InformationalVersion;
-
-            if (!string.IsNullOrWhiteSpace(version))
-            {
-                var semanticVersion = new NuGet.SemanticVersion(version);
-                Servicing.Breadcrumbs.Instance.AddBreadcrumb(runtimeAssembly.GetName().Name, semanticVersion);
-            }
+            var env = ServiceProvider.GetService(typeof(IRuntimeEnvironment)) as IRuntimeEnvironment;
```
moozzyk: nit: Consider using regular cast instead of `as` casting. `as` serves a different purpose.

src/Microsoft.Framework.Runtime/DefaultHost.cs
```diff
@@ -192,22 +192,9 @@ private void Initialize(DefaultHostOptions options, IServiceProvider hostService
...
+            string frameworkBreadcrumbName = $"{Constants.RuntimeShortName}-{env.RuntimeType}-{env.RuntimeArchitecture}-{env.RuntimeVersion}";
```
moozzyk: nit: `var` (or even do it inline?)

src/Microsoft.Framework.Runtime.Sources/Impl/Constants.cs
moozzyk: Does it matter that we now use lowercase 'dnx' for folder and environment variable names while the actual names are uppercase? Are we going to support bredcrumbs on other platforms (where it does)?

victorhurdugaci: Not now but that's a fair point. I'll change it to uppercase

moozzyk: :+1:

victorhurdugaci:
Updated

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.Framework.Runtime.Servicing
     public class Breadcrumbs
     {
         // It is recommended that you disable breadcrumbs only for perf runs
-        private const string NoBreadcrumbsEnvironmentVariableName = "DNX_NO_BREADCRUMBS";
+        private const string NoBreadcrumbsEnvironmentVariableName = Constants.RuntimeShortName + "_NO_BREADCRUMBS";
```
moozzyk: Nit: this is still `dnx_NO_BREACRUMBS` and not `DNX_NO_BREADCRUMBS`. Consider `Constants.RuntimeShortName.ToUpper() + "_NO_BREADCRUMBS"`. May require using `static readonly` instead of `const`

victorhurdugaci: Env variables are not case sensitive

davidfowl: Are you sure about that? Is that true on other OSes?

davidfowl: Also make sure it's uppercase.

test/Microsoft.Framework.Runtime.FunctionalTests/ServiceBreadcrumbsFacts.cs
```diff
@@ -31,7 +31,6 @@ public void BreadcrumbsAreCreatedSuccessfully()
             breadcrumbs.AddBreadcrumb("Test", new NuGet.SemanticVersion("1.0.0"));
             breadcrumbs.WriteAllBreadcrumbs();
 
-            Assert.True(File.Exists(Path.Combine(TempFolderPath, "Test")));
             Assert.True(File.Exists(Path.Combine(TempFolderPath, "Test.1.0.0")));
```
moozzyk: Is it possible to add a test that verifies that runtime breadcrumb is being written?

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
victorhurdugaci: What other OSs? This is a Windows only feature...

moozzyk:
Other than the nits. :ship::it:

src/Microsoft.Framework.Runtime/Servicing/Breadcrumbs.cs
davidfowl: I think this it will be forever, and that's a bad reason to have inconsistent casing. Please make it uppercase DNX_

moozzyk: @victorhurdugaci - @glennc mentioned it may change.

test/Microsoft.Framework.Runtime.FunctionalTests/ServiceBreadcrumbsFacts.cs
victorhurdugaci: Unfortunately, that's not easy... It could mess up with the files on the user's machine plus it would require admin rights to cleanup. I did a manual verification to see that the breadcrumb is written.

victorhurdugaci:
@moozzyk boat Italy? :p

test/Microsoft.Framework.Runtime.FunctionalTests/ServiceBreadcrumbsFacts.cs
moozzyk: Agreed.

moozzyk:
No, a hidden way of saying what will [happen](http://en.wikipedia.org/wiki/Costa_Concordia_disaster) if did not get it right...

victorhurdugaci:
Ha!

Issue #1469, opened 2015-03-24
Give more error info when lockfile is invalid


ChengTian:
We should report the version mismatch/missing lockfile
Basically the exception can be the same, but it needs more information:
If the lock file dependencies are out of date, it should add that to the error message;
if the lock file version is too new and the runtime can’t understand it, it should fail and say why

Basically, always add extra information if you know why.

ChengTian:
@davidfowl , you mentioned that you don't like using `ICompilationMessage` to surface details of an invalid lock file. Should I just throw an exception in ctor of `ApplicationHostContext` when lock file is invalid?

PR #1900, opened 2015-05-19
Show details of invalid lock file

ChengTian:
parent https://github.com/aspnet/dnx/issues/1469

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Building/BuildContext.cs
+++ b/src/Microsoft.Framework.PackageManager/Building/BuildContext.cs
@@ -58,13 +58,13 @@ namespace Microsoft.Framework.PackageManager
 
             var result = builder.Build(_project.Name, _outputPath);
 
+            var errors = _applicationHostContext.DependencyWalker.GetDependencyDiagnostics(_project.ProjectFilePath);
+            diagnostics.AddRange(errors);
+
             if (result.Diagnostics != null)
             {
                 diagnostics.AddRange(result.Diagnostics);
             }
-
-            var errors = _applicationHostContext.DependencyWalker.GetDependencyDiagnostics(_project.ProjectFilePath);
-            diagnostics.AddRange(errors);
 
             return result.Success && !diagnostics.HasErrors();
         }
--- a/src/Microsoft.Framework.Runtime/ApplicationHostContext.cs
+++ b/src/Microsoft.Framework.Runtime/ApplicationHostContext.cs
@@ -2,7 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Runtime.Versioning;
 using Microsoft.Framework.Runtime.Caching;
 using Microsoft.Framework.Runtime.Common.DependencyInjection;
@@ -38,12 +40,6 @@ namespace Microsoft.Framework.Runtime
 
             PackagesDirectory = packagesDirectory ?? NuGetDependencyResolver.ResolveRepositoryPath(RootDirectory);
 
-            var referenceAssemblyDependencyResolver = new ReferenceAssemblyDependencyResolver(FrameworkReferenceResolver);
-            NuGetDependencyProvider = new NuGetDependencyResolver(new PackageRepository(PackagesDirectory));
-            var gacDependencyResolver = new GacDependencyResolver();
-            ProjectDepencyProvider = new ProjectReferenceDependencyProvider(ProjectResolver);
-            var unresolvedDependencyProvider = new UnresolvedDependencyProvider();
-
             var projectName = PathUtility.GetDirectoryName(ProjectDirectory);
 
             Project project;
@@ -58,40 +54,27 @@ namespace Microsoft.Framework.Runtime
             }
 
             var projectLockJsonPath = Path.Combine(ProjectDirectory, LockFileReader.LockFileName);
-            var lockFileExists = File.Exists(projectLockJsonPath);
-            var validLockFile = false;
+            var lockFileReader = new LockFileReader();
+            var lockFile = lockFileReader.Read(projectLockJsonPath);
 
-            if (lockFileExists)
+            if (!skipLockFileValidation)
             {
-                var lockFileReader = new LockFileReader();
-                var lockFile = lockFileReader.Read(projectLockJsonPath);
-                validLockFile = lockFile.IsValidForProject(Project);
-
-                if (validLockFile || skipLockFileValidation)
-                {
-                    NuGetDependencyProvider.ApplyLockFile(lockFile);
-
-                    DependencyWalker = new DependencyWalker(new IDependencyProvider[] {
-                        ProjectDepencyProvider,
-                        NuGetDependencyProvider,
-                        referenceAssemblyDependencyResolver,
-                        gacDependencyResolver,
-                        unresolvedDependencyProvider
-                    });
-                }
+                lockFile.Validate(project);
             }
 
-            if ((!validLockFile && !skipLockFileValidation) || !lockFileExists)
-            {
-                // We don't add NuGetDependencyProvider to DependencyWalker
-                // It will leave all NuGet packages unresolved and give error message asking users to run "dnu restore"
-                DependencyWalker = new DependencyWalker(new IDependencyProvider[] {
-                    ProjectDepencyProvider,
-                    referenceAssemblyDependencyResolver,
-                    gacDependencyResolver,
-                    unresolvedDependencyProvider
-                });
-            }
+            NuGetDependencyProvider = new NuGetDependencyResolver(new PackageRepository(PackagesDirectory), lockFile);
+            var referenceAssemblyDependencyResolver = new ReferenceAssemblyDependencyResolver(FrameworkReferenceResolver);
+            var gacDependencyResolver = new GacDependencyResolver();
+            ProjectDepencyProvider = new ProjectReferenceDependencyProvider(ProjectResolver);
+            var unresolvedDependencyProvider = new UnresolvedDependencyProvider();
+
+            DependencyWalker = new DependencyWalker(new IDependencyProvider[] {
+                ProjectDepencyProvider,
+                NuGetDependencyProvider,
+                referenceAssemblyDependencyResolver,
+                gacDependencyResolver,
+                unresolvedDependencyProvider
+            });
 
             LibraryExportProvider = new CompositeLibraryExportProvider(new ILibraryExportProvider[] {
                 new ProjectLibraryExportProvider(ProjectResolver, ServiceProvider),
--- a/src/Microsoft.Framework.Runtime/DependencyManagement/LockFile.cs
+++ b/src/Microsoft.Framework.Runtime/DependencyManagement/LockFile.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 
 namespace Microsoft.Framework.Runtime.DependencyManagement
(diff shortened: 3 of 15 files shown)


ChengTian:
Error message when run with out-of-date lock file:
runwithinvalidlockfile

ChengTian:
Error message when dnu restore with an invalid "locked" lock file:

dnurestoreoninvalidlockfile

src/Microsoft.Framework.PackageManager/Building/BuildContext.cs
```diff
@@ -216,6 +216,12 @@ private ApplicationHostContext GetApplicationHostContext(Runtime.Project project
 
                 applicationHostContext.DependencyWalker.Walk(project.Name, project.Version, targetFramework);
 
+                if (applicationHostContext.Diagnostics.Any() &&
+                    applicationHostContext.DependencyWalker.Libraries.Any(l => !l.Resolved))
+                {
+                    throw new InvalidOperationException(applicationHostContext.Diagnostics.First().FormattedMessage);
```
davidfowl: Throwing exceptions during build is no good.

src/Microsoft.Framework.PackageManager/List/DependencyListOperation.cs
```diff
@@ -83,6 +84,11 @@ private ApplicationHostContext CreateApplicationHostContext()
 
             hostContext.DependencyWalker.Walk(hostContext.Project.Name, hostContext.Project.Version, _framework);
 
+            if (hostContext.Diagnostics.Any() && hostContext.DependencyWalker.Libraries.Any(l => !l.Resolved))
+            {
+                throw new InvalidOperationException(hostContext.Diagnostics.First().FormattedMessage);
```
davidfowl: Why would dnu list throw an exception?

src/Microsoft.Framework.Runtime/DependencyManagement/LockFile.cs
```diff
@@ -15,18 +15,24 @@ public class LockFile
...
 
             var actualTargetFrameworks = project.GetTargetFrameworks();
 
             // The lock file should contain dependencies for each framework plus dependencies shared by all frameworks
-            if (ProjectFileDependencyGroups.Count != actualTargetFrameworks.Count() + 1)
+            var frameworkNumDiff = ProjectFileDependencyGroups.Count - (actualTargetFrameworks.Count() + 1);
+            if (frameworkNumDiff != 0)
             {
+                var action = frameworkNumDiff < 0 ? "added to" : "removed from";
+                message = $"One or more frameworks were {action} {Project.ProjectFileName}";
```
moozzyk: If we are going to localize this I believe this will have to be changed to 2 full strings anyways so not sure if using ternary just for a substring makes sense/difference...

src/Microsoft.Framework.Runtime/DependencyManagement/LockFile.cs
```diff
@@ -15,18 +15,24 @@ public class LockFile
...
         public IList<LockFileTarget> Targets { get; set; } = new List<LockFileTarget>();
 
-        public bool IsValidForProject(Project project)
+        public bool IsValidForProject(Project project, out string message)
         {
+            message = null;
+
             if (Version != LockFileReader.Version)
             {
+                message = $"Expected lock file version is {LockFileReader.Version}, while actual version is {Version}";
```
moozzyk: I am sure @Eilon will help you word this message a bit better...

src/Microsoft.Framework.Runtime/ApplicationHostContext.cs
```diff
@@ -65,34 +76,30 @@ public class ApplicationHostContext
...
-                // We don't add NuGetDependencyProvider to DependencyWalker
-                // It will leave all NuGet packages unresolved and give error message asking users to run "dnu restore"
-                DependencyWalker = new DependencyWalker(new IDependencyProvider[] {
-                    ProjectDepencyProvider,
-                    referenceAssemblyDependencyResolver,
-                    gacDependencyResolver,
-                    unresolvedDependencyProvider
-                });
+                // Don't add NuGetDependencyProvider to DependencyWalker
+                dependencyProviders = dependencyProviders.Where(d => d != NuGetDependencyProvider);
```
moozzyk: I just checked and it does work but I would like to understand how... You assign a `WhereIterator` to a reference that is a source for this same iterator. Probably Where captures the original reference somewhere but it is not obvious.

src/Microsoft.Framework.Runtime/ApplicationHostContext.cs
```diff
@@ -65,34 +76,30 @@ public class ApplicationHostContext
...
             {
-                // We don't add NuGetDependencyProvider to DependencyWalker
-                // It will leave all NuGet packages unresolved and give error message asking users to run "dnu restore"
-                DependencyWalker = new DependencyWalker(new IDependencyProvider[] {
-                    ProjectDepencyProvider,
-                    referenceAssemblyDependencyResolver,
-                    gacDependencyResolver,
-                    unresolvedDependencyProvider
-                });
+                // Don't add NuGetDependencyProvider to DependencyWalker
```
moozzyk: This is comment does not add any value in its current form. It's quite obvious from the line below that you are not adding NuGetDependencyProvider but the answer to the "why" is missing. Same above.

src/Microsoft.Framework.Runtime/ApplicationHostContext.cs
```diff
@@ -44,6 +47,14 @@ public class ApplicationHostContext
             ProjectDepencyProvider = new ProjectReferenceDependencyProvider(ProjectResolver);
             var unresolvedDependencyProvider = new UnresolvedDependencyProvider();
 
+            IEnumerable<IDependencyProvider> dependencyProviders = new IDependencyProvider[] {
```
moozzyk: If you want to avoid copying this several times move it to a separate function. This way you will avoid this kind of weird pattern where you override the source for iterator with a computation.

src/Microsoft.Framework.Runtime/DependencyManagement/LockFile.cs
Eilon: How about: `$"The expected lock file version ({LockFileReader.Version}) does not match the actual version ({Version})."` ?

src/Microsoft.Framework.Runtime/DependencyManagement/LockFile.cs
ChengTian: Yeah, this is an interesting topic. @davidfowl 
Does this mean we should always avoid string interpolation and use `string.Format()` instead? Looks like string interpolation is not localization-friendly.

moozzyk: I don't think there is a difference between `string.Format` and string interpolation here. `string.Format` would be as much localization unfriendly in this case as string interpolation. The point is that localization is broken if the parameters are localizable.

ChengTian: Yeah, I see your point. The problem is that, even I fix the `{action}` part, we still have
```
$"One or more frameworks were added to {Project.ProjectFileName}"
```
which is still not localizable. That's why I am asking whether we should use `string.Format()` instead :smile:

Eilon: Eh, we're not localizing anything anytime soon here, so I don't care either way at this time. If it's easy to make it localizable, go ahead, but I wouldn't lose any sleep if it wasn't super localizable right now.

ChengTian: Ok, I'll remove the ternary operation to make future change easier.

moozzyk: :clap:

ChengTian:
Thanks for the feedback. Please check the updated patch.

src/Microsoft.Framework.Runtime/ApplicationHostContext.cs
```diff
@@ -60,39 +71,36 @@ public class ApplicationHostContext
...
-                        ProjectDepencyProvider,
-                        NuGetDependencyProvider,
-                        referenceAssemblyDependencyResolver,
-                        gacDependencyResolver,
-                        unresolvedDependencyProvider
-                    });
+                }
+                else
+                {
+                    dependencyProviders.Remove(NuGetDependencyProvider);
```
moozzyk: :clap: - just add a comment why we are doing this - similarly to what we have before ("We don't add NuGetDependencyProvider to DependencyWalker
// It will leave all NuGet packages unresolved and give error message asking users to run "dnu restore")

src/Microsoft.Framework.Runtime/DependencyManagement/LockFile.cs
```diff
@@ -15,18 +15,31 @@ public class LockFile
...
             }
 
             var actualTargetFrameworks = project.GetTargetFrameworks();
 
             // The lock file should contain dependencies for each framework plus dependencies shared by all frameworks
-            if (ProjectFileDependencyGroups.Count != actualTargetFrameworks.Count() + 1)
+            var frameworkNumDiff = ProjectFileDependencyGroups.Count - (actualTargetFrameworks.Count() + 1);
+            if (frameworkNumDiff != 0)
             {
+                if (frameworkNumDiff < 0)
```
moozzyk: nit: consider ternary

src/Microsoft.Framework.Runtime/DependencyManagement/LockFile.cs
```diff
@@ -47,6 +60,7 @@ public bool IsValidForProject(Project project)
                             string.Equals(f.FrameworkName.ToString(), group.FrameworkName, StringComparison.Ordinal));
                     if (framework == null)
                     {
+                        message = $"The framework '{group.FrameworkName}' was removed from {Project.ProjectFileName}";
```
moozzyk: nit: formatting seems to be off in line 60

moozzyk: Ignore. it is a continuation from the previous line which was not shown.

moozzyk:
Looks good to me but I am not very familiar with this code so I don't feel comfortable to sign off on it.

ChengTian:
@davidfowl , followed your suggestion and here is the new output:

`dnx . run` with invalid lock file:
invalidlockfile_run

`dnu build` with invalid lock file:
invalidlockfile_build

`dnu restore` on a locked invalid lock file:
invalidlockfile_restore

ChengTian:
Latest output based on @davidfowl and @DamianEdwards 's feedback:

`dnx . run` when dependencies in `project.json` were changed:
newinvalidlockfile_run_dependencies

`dnx . run` when lock file version doesn't match:
newinvalidlockfile_run_lockfileversion

`dnx . build` when lock file version doesn't match:
newinvalidlockfile_build_lockfileversion

`dnx . run` when lock file is correct but a dependency is not in packages folder:
newinvalidlockfile_run_deletepackage

Eilon:
Can we like, you know, not have stack traces for expected types of errors? This shouldn't crash dnx... it should just report an error, set the return value of the app to some non-zero value, and exit.

That probably ought to be a separate bug, though.

davidfowl:
@Eilon that's actually trickier than you think. This error message can show up in helios in the white screen of death as well.

Eilon:
Do you agree the current behavior is quite lame? I logged https://github.com/aspnet/dnx/issues/1923

ChengTian:
@davidfowl , improved based on your feedback. Output is still the same as screenshots above.

ChengTian:
This is replaced by https://github.com/aspnet/dnx/pull/2281

PR #1902, opened 2015-05-19
Splitting dnx.coreclr.dll to dnx_win32.coreclr.dll and dnx_onecore.co…

moozzyk:
…reclr.dll

Note that the difference between dnx_win32.coreclr.dll and dnx_onecore.coreclr.dll is only how we build the code and the code itself should remain the same.

- Enables using STL in the dnx_xxx.coreclr.dll
- Opens a way to compile for arm (onecore)

Bonus: Fixing some warnings and errors that started appearing after Warning Level was changed to 4

### DIFF
--- a/src/dnx/dnx.cpp
+++ b/src/dnx/dnx.cpp
@@ -235,7 +235,11 @@ int CallApplicationProcessMain(int argc, LPTSTR argv[])
     bool fSuccess = true;
     HMODULE m_hHostModule = nullptr;
 #if CORECLR_WIN
-    LPCTSTR pwzHostModuleName = _T("dnx.coreclr.dll");
+#if ONECORE
+        LPCTSTR pwzHostModuleName = _T("dnx.onecore.coreclr.dll");
+#else
+        LPCTSTR pwzHostModuleName = _T("dnx.win32.coreclr.dll");
+#endif
 #elif CORECLR_DARWIN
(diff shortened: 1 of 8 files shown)


src/dnx/dnx.cpp
```diff
@@ -235,7 +235,11 @@ int CallApplicationProcessMain(int argc, LPTSTR argv[])
     bool fSuccess = true;
     HMODULE m_hHostModule = nullptr;
 #if CORECLR_WIN
-    LPCTSTR pwzHostModuleName = _T("dnx.coreclr.dll");
+#if ONECORE
+        LPCTSTR pwzHostModuleName = _T("dnx_onecore.coreclr.dll");
```
davidfowl: dnx.onecore.coreclr
dnx.win32.coreclr

BrennanConroy: Also, tabs

moozzyk: I think tabs are good. This is a nested `#if`. VS will automatically move precompiler directives to the left and adding nesting to the actual content signals it is a nested if.

moozzyk: @davidfowl I infer that in this case we also want to have dnx.onecore.dll and dnx.win32.dll instead of dnx_onecore.dll and dnx_win32.dll (the [other](https://github.com/aspnet/dnx/pull/1886/files#diff-4d19362f790908c080d93fa6c5e708adR13) PR)

Issue #1892, opened 2015-05-18
Bad error message for `dnu commands install`


danroth27:
```
C:\Users\danroth27>dnu commands install ConsoleApp123321
  CACHE https://www.myget.org/f/aspnetmaster/api/v2/FindPackagesById()?Id='Conso
leApp123321'
  CACHE https://www.nuget.org/api/v2/FindPackagesById()?Id='ConsoleApp123321'
Unable to locate the package ConsoleApp123321
The name of the package to be installed was not specified.
```
Seems like the name of the package was actually specified, no?

danroth27:
In this case the error was caused by the package not being found on the configured feeds.

muratg:
@victorhurdugaci

PR #1904, opened 2015-05-19
Global Install error message change

BrennanConroy:
#1892

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Install/InstallGlobalCommand.cs
+++ b/src/Microsoft.Framework.PackageManager/Install/InstallGlobalCommand.cs
@@ -47,7 +47,7 @@ namespace Microsoft.Framework.PackageManager
 
             if (packageIdAndVersion == null)
             {
-                WriteError("The name of the package to be installed was not specified.");
+                WriteError($"Failed to install {packageId}");
                 return false;
             }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Framework.PackageManager/Install/InstallGlobalCommand.cs
```diff
@@ -47,7 +47,7 @@ public FeedOptions FeedOptions
 
             if (packageIdAndVersion == null)
             {
-                WriteError("The name of the package to be installed was not specified.");
+                WriteError($"Installing {packageId} failed");
```
moozzyk: are we sure `packageId` is not null?

BrennanConroy: Good catch, should we just say "Installing failed" instead? They should know what they are trying to install

victorhurdugaci: `ResolvePackageIdAndVersion` returns `null` if the package id is null. That means that we cannot figure out from the result of that method if the package is null because the user didn't specify it or because it was not found. I suggest adding a check before we call that method. If the user passes a `null` package, give an appropriate error message and if the package is not found, give a different one

victorhurdugaci: Maybe validate the args even before we get the `Execute`

victorhurdugaci: Suggested error message: "Failed to install .... "

victorhurdugaci:
other than the comment about the error message :shipit: (you might have to rebase)

Issue #1905, opened 2015-05-19
Add PluginHandler support for IPlugin Protocols


NTaylorMullen:
`PluginHandler` should support resolving an active protocol for registered `IPlugin`s. This also involves updating the `IPlugin` signature to contain protocol based information for the `PluginHandler` to use/set.

NTaylorMullen:
25feee326325ae7305181b550de83a098dfb5f85

PR #1909, opened 2015-05-20
Add IPlugin protocol to PluginHandler.

NTaylorMullen:
- When a plugin is registered with the system a client protocol can now be provided. This protocol is then cross checked with the created IPlugin's protocol and the result of Register returns the Protocol that will be used when communicating with the registered Plugin.
- Added tests to validate error conditions and protocol resolutions.

#1905

### DIFF
--- a/src/Microsoft.Framework.DesignTimeHost.Abstractions/IPlugin.cs
+++ b/src/Microsoft.Framework.DesignTimeHost.Abstractions/IPlugin.cs
@@ -9,5 +9,7 @@ namespace Microsoft.Framework.DesignTimeHost
     public interface IPlugin
     {
         void ProcessMessage(JObject data, IAssemblyLoadContext assemblyLoadContext);
+
+        int Protocol { get; set; }
     }
 }
--- a/src/Microsoft.Framework.DesignTimeHost/Models/OutgoingMessages/RegisterPluginResponseMessage.cs
+++ b/src/Microsoft.Framework.DesignTimeHost/Models/OutgoingMessages/RegisterPluginResponseMessage.cs
@@ -1,0 +1,10 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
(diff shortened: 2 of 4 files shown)


src/Microsoft.Framework.DesignTimeHost.Abstractions/IPlugin.cs
```diff
@@ -9,5 +10,7 @@ namespace Microsoft.Framework.DesignTimeHost
     public interface IPlugin
     {
         void ProcessMessage(JObject data, IAssemblyLoadContext assemblyLoadContext);
+
+        Version Protocol { get; set; }
```
davidfowl: Does it really need to be a version instead of being a number?

NTaylorMullen: Doesn't have to be. Thought I'd follow the same approach we used in SignalR since that proved to work well :smile:

NTaylorMullen: Would the alternate be a double? int?? Seems kinda odd as a protocol version.

davidfowl: int

NTaylorMullen:
@davidfowl Updated.

FYI: @ToddGrun

NTaylorMullen:
Going to push this tomorrow if no more feedback comes in.

src/Microsoft.Framework.DesignTimeHost.Abstractions/IPlugin.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
```
BrennanConroy: Can remove

NTaylorMullen: :+1:

Issue #1906, opened 2015-05-19
Enable PluginHandler to return no-op messages when IPlugin can't handle message.


NTaylorMullen:
This should involve modifying the `ProcessMessage` to return a `bool` which indicates if the message was handled or not. In the case that it's not handled AND a correlation ID was provided it should return a no-op message to the client.

NTaylorMullen:
5aa77220fc983ccd69454a1f221eccddc4c70d7a

PR #1917, opened 2015-05-20
[Design] Add no-op support for IPlugins.

NTaylorMullen:
Running this through design prior to adding tests.

- In the case that the PluginHandler cannot handle a message a no-op is returned.
- In the case that a Plugin cannot handle a message a no-op is returned. Handling a message is determined by the result of the ProcessMessages call to the invoked Plugin.

#1906


NTaylorMullen:
This all good?

Issue #1792, opened 2015-05-06
It's possible to get duplicate entries in the project.lock.json file for frameworkAssemblies


davidfowl:
```
"Microsoft.CodeAnalysis.Analyzers/1.0.0-rc3-20150504-02": {
        "frameworkAssemblies": [
          "System",
          "System"
        ]
      },
```

PR #1920, opened 2015-05-21
Fixed duplicate framework assemblies

davidfowl:
- Changed lock file format to be a list of string for framework assemblies

#1792

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Utils/LockFileUtils.cs
+++ b/src/Microsoft.Framework.PackageManager/Utils/LockFileUtils.cs
@@ -201,7 +201,10 @@ namespace Microsoft.Framework.PackageManager.Utils
                     continue;
                 }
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.Framework.PackageManager/Utils/LockFileUtils.cs
```diff
@@ -201,7 +201,10 @@ private static void AddFrameworkReferences(LockFileTargetLibrary lockFileLib, Fr
                     continue;
                 }
 
-                lockFileLib.FrameworkAssemblies.Add(assemblyReference);
+                if (!lockFileLib.FrameworkAssemblies.Contains(assemblyReference.AssemblyName))
```
pranavkm: Why not change this to a `ISet<>` instead?

pranavkm:
:watch:

PR #1922, opened 2015-05-21
Fixing dnx.coreclr.vcxproj so that it builds correctly in VS

moozzyk:
The issue was that the post build step used an absolute path while we started bin placing binaries using a relative path.


moozzyk:
@victorhurdugaci - can you pull and try it? Works on my box but I want to be _doubleplussure_.

davidfowl:
I didn't test it but it looks good so :shipit:

victorhurdugaci:
It works. no more errors :shipit:

moozzyk:
Will commit as soon as I get commit permit from @Tratcher

Issue #1859, opened 2015-05-13
CommandLineUtils: ShowHelp throws NRE


moozzyk:
(Originally opened in the EF repo: https://github.com/aspnet/EntityFramework/issues/2178)

`dnx . ef help add`

throws

```
                     _/\__
               ---==/    \\
         ___  ___   |.    \|\
        | __|| __|  |  )   \\\
        | _| | _|   \_/ |  //|\\
        |___||_|       /   \\\/\\

System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Framework.Runtime.Common.CommandLine.CommandLineApplication.ShowHelp(String commandName)
   at Microsoft.Data.Entity.Commands.Program.ShowHelp(String command)
   at Microsoft.Data.Entity.Commands.Program.<>c__DisplayClass7_8.<Main>b__19()
   at Microsoft.Framework.Runtime.Common.CommandLine.CommandLineApplication.Execute(String[] args)
   at Microsoft.Data.Entity.Commands.Program.Main(String[] args)
```
on
     *    1.0.0-beta4 clr     x86

bricelam:
Same bug is manifest if you do `dnu help blah`.

muratg:
@victorhurdugaci

victorhurdugaci:
Oh no, the unicorn null refs :smile:

moozzyk:
@victorhurdugaci only the head of the unicorn we don't know about the rest of it...

PR #1924, opened 2015-05-21
Prevent help from null refing when the command doesn't exist

victorhurdugaci:
Fixes https://github.com/aspnet/dnx/issues/1859

cc @moozzyk

### DIFF
--- a/src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandLineApplication.cs
+++ b/src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandLineApplication.cs
@@ -332,7 +332,17 @@ namespace Microsoft.Framework.Runtime.Common.CommandLine
             else
             {
                 target = Commands.SingleOrDefault(cmd => string.Equals(cmd.Name, commandName, StringComparison.OrdinalIgnoreCase));
-                headerBuilder.AppendFormat(" {0}", commandName);
+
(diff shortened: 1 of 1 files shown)


src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandLineApplication.cs
```diff
@@ -332,7 +332,17 @@ public void ShowHelp(string commandName = null)
...
-                headerBuilder.AppendFormat(" {0}", commandName);
+
+                if (target != null)
+                {
+                    headerBuilder.AppendFormat(" {0}", commandName);
+                }
+                else
+                {
+                    // The command name is invalid so don't try to show help for something that doesn't exist
+                    target = this;
```
moozzyk: Will this just result in showing the general help?

victorhurdugaci: It will show the help for the last valid command

moozzyk:
:ship::it:

moozzyk:
Actually - wouldn't be better also to tell that the command typed is not valid?

victorhurdugaci:
Oh, whoops, I merged before seeing your second comment. I think what we do now is sufficient because the help already includes available commands

moozzyk:
Sounds good.

PR #1930, opened 2015-05-22
"dnu restore" shouldn't stop on invalid root

ChengTian:
Reaction to https://github.com/aspnet/dnx/commit/4bb7649793806ea0296c108fcc19b7816d14e850#commitcomment-11318212

@davidfowl , invalid root error is also shown at the very end of `dnu restore` output, just like other kinds of restore errors. This makes sure it is not lost in a long output:
capture


BrennanConroy:
:shipit:

PR #1931, opened 2015-05-22
Fix #876, Correct handling of quotation marks in commands

### LINKED ISSUE #876: Script blocks with spaces in arguments will fail.
ScriptExecutor will remove the quotes in script block unconditionally, as a result user can't wrap argument in quotes.

dougbu:
- don't strip quotes surrounding command-line arguments when used in `scripts`
 - prevented command paths and grouping arguments containing spaces
 - pass `preserveSurroundingQuotes` into `CommandGrammar` to control stripping
- support expected shell capabilities such as quoting command and redirections
 - use `cmd /s /c` when executing `scripts` on Windows
 - use `/bin/bash -c` when executing `scripts` on Linux
- add functional tests of commands and scripts (together with `dnu restore`)
 - no `dnu restore` for .NET Core due to extra time required

note:
- `CommandGrammar` may need additional generalizations
 - e.g. unquoted term can't contain more than one escape sequence
 - but it's probably good enough for now
- _not_ adding quotes around variables replacements inserted into commands
 - no way of determining if replacement is already quoted
 - basically, leave it to project.json author to perform appropriate quoting for their platform

nits:
- ~~let VS do its thing with test services~~
- make `CommandGrammar` constructor private; used only in `Process()` method
- update a few `ScriptExecutor` comments, long lines, et cetera

### DIFF
--- a/src/Microsoft.Framework.ApplicationHost/Impl/Syntax/CommandGrammar.cs
+++ b/src/Microsoft.Framework.ApplicationHost/Impl/Syntax/CommandGrammar.cs
@@ -9,12 +9,12 @@ namespace Microsoft.Framework.ApplicationHost.Impl.Syntax
 {
     internal class CommandGrammar : Grammar
     {
-        public CommandGrammar(Func<string, string> variable)
+        private CommandGrammar(Func<string, string> variable, bool preserveSurroundingQuotes)
         {
             var environmentVariablePiece = Ch('%').And(Rep(Ch().Not(Ch('%')))).And(Ch('%')).Left().Down().Str()
                 .Build(key => variable(key) ?? "%" + key + "%");
 
-            var escapeSequencePiece = 
+            var escapeSequencePiece =
                 Ch('%').And(Ch('%')).Build(_=>"%")
                     .Or(Ch('^').And(Ch('^')).Build(_ => "^"))
                     .Or(Ch('\\').And(Ch('\\')).Build(_ => "\\"))
@@ -30,6 +30,12 @@ namespace Microsoft.Framework.ApplicationHost.Impl.Syntax
             var unquotedTerm = Rep1(unquotedPiece.Or(specialPiece)).Str();
 
             var quotedTerm = Ch('\"').And(Rep(quotedPiece.Or(specialPiece)).Str()).And(Ch('\"')).Left().Down();
+            if (preserveSurroundingQuotes)
+            {
+                // Str() value assigned to quotedTerm does not include quotation marks surrounding the quoted or
+                // special piece. Add those quotes back if requested.
+                quotedTerm = quotedTerm.Build(str => "\"" + str + "\"");
+            }
 
             var whitespace = Rep(Ch(' '));
 
@@ -40,9 +46,9 @@ namespace Microsoft.Framework.ApplicationHost.Impl.Syntax
 
         public readonly Parser<IList<string>> Parse;
 
-        public static string[] Process(string text, Func<string, string> variables)
+        public static string[] Process(string text, Func<string, string> variables, bool preserveSurroundingQuotes)
         {
-            var grammer = new CommandGrammar(variables);
+            var grammer = new CommandGrammar(variables, preserveSurroundingQuotes);
             var cursor = new Cursor(text, 0, text.Length);
 
             var result = grammer.Parse(cursor);
--- a/src/Microsoft.Framework.PackageManager/Scripts/ScriptExecutor.cs
+++ b/src/Microsoft.Framework.PackageManager/Scripts/ScriptExecutor.cs
@@ -34,13 +34,15 @@ namespace Microsoft.Framework.PackageManager
 
             foreach (var scriptCommandLine in scriptCommandLines)
             {
+                // Preserve quotation marks around arguments since command is about to be passed to a shell. May need
+                // the quotes to ensure the shell groups arguments correctly.
                 var scriptArguments = CommandGrammar.Process(
                     scriptCommandLine,
-                    GetScriptVariable(project, getVariable));
+                    GetScriptVariable(project, getVariable),
+                    preserveSurroundingQuotes: true);
 
-                // Ensure the array won't be empty and the first element won't be null or empty string.
+                // Ensure the array won't be empty and the elements won't be null or empty strings.
                 scriptArguments = scriptArguments.Where(argument => !string.IsNullOrEmpty(argument)).ToArray();
-
                 if (scriptArguments.Length == 0)
                 {
                     continue;
@@ -48,18 +50,20 @@ namespace Microsoft.Framework.PackageManager
 
                 if (!PlatformHelper.IsMono)
                 {
-                    // Forward-slash is used in script blocked only. Replace them with back-slash to correctly
-                    // locate the script. The directory separator is platform-specific. 
-                    scriptArguments[0] = scriptArguments[0].Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar);
+                    // Only forward slashes are used in script blocks. Replace with backslashes to correctly
+                    // locate the script. The directory separator is platform-specific.
+                    scriptArguments[0] = scriptArguments[0].Replace(
+                        Path.AltDirectorySeparatorChar,
+                        Path.DirectorySeparatorChar);
 
-                    // Command-lines on Windows are executed via "cmd /C" in order
-                    // to support batch files, &&, built-in commands like echo, etc.
+                    // Command-lines on Windows are executed via "cmd /S /C" in order to support batch files, &&,
+                    // built-in commands like echo, et cetera. /S allows quoting the command as well as the arguments.
                     // ComSpec is Windows-specific, and contains the full path to cmd.exe
                     var comSpec = Environment.GetEnvironmentVariable("ComSpec");
                     if (!string.IsNullOrEmpty(comSpec))
                     {
                         scriptArguments =
-                            new[] { comSpec, "/C", "\"" }
+                            new[] { comSpec, "/S", "/C", "\"" }
                             .Concat(scriptArguments)
                             .Concat(new[] { "\"" })
                             .ToArray();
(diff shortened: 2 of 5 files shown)


dougbu:
/cc @davidfowl 

Any suggestions on unit / integration / functional testing of scripts? I don't see anything in this repo at the moment.

dougbu:
@davidfowl since you like to look at individual commits, I rebased and added a couple more :smiley_cat: 

Latest version is more resilient and I've tested it numerous ways -- manually. (Don't seem to have test infrastructure in place to automated anything here.)

dougbu:
Rebased again and added tests.

@davidfowl this bug is for Beta 5.

davidfowl:
@dougbu This is broken on *nix. Look at travis

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuRestoreTests.cs
```diff
@@ -0,0 +1,109 @@
...
+        {
+            get
+            {
+                return TestUtils.GetRuntimeComponentsCombinations();
+            }
+        }
+
+        [ConditionalTheory]
+        [MemberData(nameof(RuntimeComponents))]
+        [OSSkipCondition(OperatingSystems.Linux | OperatingSystems.MacOSX)]
```
BrennanConroy: Why?

dougbu: Because i am positive my `bash`-writing skills are out-of-date and don't have a Linux system to play on. We can follow-up later with additional scripts.

dougbu:
@davidfowl Travis builds now successful because all new tests are disabled on Mac and Linux. AppVeyor build queued; no idea when it will start but local `verify` builds are fine.

dougbu:
Should have mentioned: No regressions on Mac or Linux i.e. I disabled only my new tests there.

test/Microsoft.Framework.ApplicationHost.FunctionalTests/AppHostTests.cs
```diff
@@ -170,5 +171,131 @@ public void AppHostShowsErrorWhenCurrentTargetFrameworkWasNotFoundInProjectJson(
...
+                string error;
+                exitCode = BootstrapperTestUtils.ExecBootstrapper(
+                    runtimeHomePath,
+                    arguments: $@"""{ projectPath }"" { command } extra",
+                    stdOut: out output,
+                    stdErr: out error);
+
+                Assert.Equal(0, exitCode);
+                Assert.Empty(error);
+                Assert.Equal(expectedOutput, output);
```
ChengTian: `DNX_TRACE` is turned on by default in our TeamCity CI and you'll get extra info in output.

Use `Assert.Contains()` or turn off `DNX_TRACE` in the subprocess with something like:
https://github.com/aspnet/dnx/blob/dev/test/Bootstrapper.FunctionalTests/BootstrapperTests.cs#L114

dougbu: :ok: I turned off `DNX_TRACE` for other added tests in this PR. Will do the same here.

src/Microsoft.Framework.Runtime.Sources/Impl/EntryPointExecutor.cs
```diff
@@ -33,6 +33,17 @@ public static Task<int> Execute(Assembly assembly, string[] args, IServiceProvid
...
+                    {
+                        var arg = args[index];
+                        if (arg.StartsWith("\"", StringComparison.Ordinal) &&
+                            arg.EndsWith("\"", StringComparison.Ordinal))
+                        {
+                            // Remove quotes around this argument.
+                            args[index] = arg.Substring(1, arg.Length - 2);
+                        }
+                    }
+
```
dougbu: As @davidfowl and I discussed, will undo this change.

src/Microsoft.Framework.ApplicationHost/Impl/Syntax/CommandGrammar.cs
```diff
@@ -29,7 +29,9 @@ public CommandGrammar(Func<string, string> variable)
 
             var unquotedTerm = Rep1(unquotedPiece.Or(specialPiece)).Str();
 
-            var quotedTerm = Ch('\"').And(Rep(quotedPiece.Or(specialPiece)).Str()).And(Ch('\"')).Left().Down();
+            var quotedTerm = Ch('\"').And(Rep(quotedPiece.Or(specialPiece)).Str()).And(Ch('\"'))
+                .Left().Down()
+                .Build(str => "\"" + str + "\"");
```
dougbu: As @davidfowl and I discussed offline, will make restoration of surrounding quotes conditional. Won't be done when parsing `commands` but will be for `scripts`.

src/Microsoft.Framework.PackageManager/Scripts/ScriptExecutor.cs
```diff
@@ -67,18 +71,44 @@ public bool Execute(Runtime.Project project, string scriptName, Func<string, str
...
+                        scriptArguments[0] = $"{ quote }{ prefix }{ scriptCandidate }{ quote }";
+                    }
+
+                    // Always use /bin/bash -c in order to support redirection and so on; similar to Windows case.
+                    // Unlike Windows, must escape quotation marks within the newly-quoted string.
+                    scriptArguments = new[] { "/bin/bash", "-c", "\"" }
+                        .Concat(scriptArguments.Select(argument => argument.Replace("\"", "\\\"")))
+                        .Concat(new[] { "\"" })
+                        .ToArray();
                 }
```
dougbu: /cc @troydai I fixed the problems in this `else` block that we discussed today. E.g. code no longer checks for the script file in whatever random directory `dnu` was executing.

src/Microsoft.Framework.ApplicationHost/Impl/Syntax/CommandGrammar.cs
```diff
@@ -9,12 +9,12 @@ namespace Microsoft.Framework.ApplicationHost.Impl.Syntax
...
+        private CommandGrammar(Func<string, string> variable, bool preserveSurroundingQuotes)
         {
             var environmentVariablePiece = Ch('%').And(Rep(Ch().Not(Ch('%')))).And(Ch('%')).Left().Down().Str()
                 .Build(key => variable(key) ?? "%" + key + "%");
 
-            var escapeSequencePiece = 
+            var escapeSequencePiece =
                 Ch('%').And(Ch('%')).Build(_=>"%")
                     .Or(Ch('^').And(Ch('^')).Build(_ => "^"))
                     .Or(Ch('\\').And(Ch('\\')).Build(_ => "\\"))
```
dougbu: Didn't change `escapeSequencePiece` but it probably should not remove Windows (`'^'`) or Linux (`'\\'`) escapes when `preserveSurroundingQuotes`. Should be fine for now since double quotes are far more common than escaping a single character.

src/Microsoft.Framework.ApplicationHost/Impl/Syntax/CommandGrammar.cs
```diff
@@ -30,6 +30,12 @@ public CommandGrammar(Func<string, string> variable)
             var unquotedTerm = Rep1(unquotedPiece.Or(specialPiece)).Str();
 
             var quotedTerm = Ch('\"').And(Rep(quotedPiece.Or(specialPiece)).Str()).And(Ch('\"')).Left().Down();
+            if (preserveSurroundingQuotes)
+            {
+                // Str() value assigned to quotedTerm does not include quotation marks surrounding the quoted or
+                // special piece. Add those quotes back if requested.
+                quotedTerm = quotedTerm.Build(str => "\"" + str + "\"");
+            }
```
dougbu: Also likely minor but `quotedTerm` will not keep a (Linux-only) single-quoted argument string together.

dougbu:
@davidfowl rebased, updated, and working on Linux. (Not sure why the Travix-ci/pr build hung but the push build succeeded.)

src/Microsoft.Framework.PackageManager/Scripts/ScriptExecutor.cs
```diff
@@ -67,18 +71,44 @@ public bool Execute(Runtime.Project project, string scriptName, Func<string, str
...
                 else
                 {
-                    var scriptCandiate = scriptArguments[0] + ".sh";
-                    if (File.Exists(scriptCandiate))
+                    // Special-case a script name that, perhaps with added .sh, matches an existing file.
+                    var surroundWithQuotes = false;
+                    var scriptCandidate = scriptArguments[0];
+                    if (scriptCandidate.StartsWith("\"", StringComparison.Ordinal) &&
```
victorhurdugaci: Can you have spaces before the first quote or after the last one?

src/Microsoft.Framework.PackageManager/Scripts/ScriptExecutor.cs
```diff
@@ -67,18 +71,44 @@ public bool Execute(Runtime.Project project, string scriptName, Func<string, str
...
-                        scriptArguments[0] = scriptCandiate;
-                        scriptArguments = new[] { "/bin/bash" }.Concat(scriptArguments).ToArray();
+                        scriptCandidate = scriptCandidate + ".sh";
                     }
+
+                    if (File.Exists(Path.Combine(project.ProjectDirectory, scriptCandidate)))
+                    {
+                        // scriptCandidate may be a path relative to the project root. If so, likely will not be found
+                        // in the $PATH; add ./ to let bash know where to look.
+                        var prefix = Path.IsPathRooted(scriptCandidate) ? string.Empty : "./";
```
victorhurdugaci: I don't think the path can be rooted  because of the if block above. The project directory + a rooted path will not result in a valid file name

src/Microsoft.Framework.PackageManager/Scripts/ScriptExecutor.cs
dougbu: That's possible in the project.json file. But `CommandGrammar` removes spaces between and before all arguments and ignores the (remote) possibility of an argument that contains more than a correctly-quoted string.

victorhurdugaci:
Looks good to me and I only had the two comments above

src/Microsoft.Framework.PackageManager/Scripts/ScriptExecutor.cs
dougbu: `Path.Combine()` returns the right-hand argument if it is rooted.

victorhurdugaci:
:shipit:

dougbu:
90f1f03

Issue #1864, opened 2015-05-13
dnu should always write out its version number as the first line of text


lodejard:
it would really help know exactly what is doing the restore and build when looking at VS output windows

muratg:
@ChengTian

ChengTian:
Good idea. We can show the first line of `dnu --help` at the beginning of restore/build. So you can see
```
Microsoft .NET Development Utility v1.0.0-beta5-11760
<blank line>
<usual output>
```

ChengTian:
Do we want copy right info like the header xUnit always output?
![capture](https://cloud.githubusercontent.com/assets/1383883/7638189/ecfab83e-fa27-11e4-9d7b-dad7e7125ce8.PNG)

muratg:
First line of `dnu --help` seems good for this case. I don't think we need the (c), but adding @Eilon to let us know.

Eilon:
Just product name and version is fine.

ChengTian:
Should be a easy one but can improve user experience a lot. I can finish this quickly.

PR #1932, opened 2015-05-22
Always show full name and version when running a valid dnu subcommand

ChengTian:
parent https://github.com/aspnet/dnx/issues/1864

@lodejard @muratg @davidfowl

### DIFF
--- a/src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandLineApplication.cs
+++ b/src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandLineApplication.cs
@@ -411,15 +411,11 @@ namespace Microsoft.Framework.Runtime.Common.CommandLine
             }
             headerBuilder.AppendLine();
 
-            var versionAndName = new StringBuilder(FullName);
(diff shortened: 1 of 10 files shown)


ChengTian:
Sample output:
showfulltitle

ChengTian:
Show the header info only when a subcommand is actually running. If you provide an invalid subcommand/argument/option, we don't show the header:

showfulltitle_error

src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandLineApplication.cs
```diff
@@ -432,6 +428,23 @@ public void ShowVersion()
             Console.WriteLine(VersionGetter());
         }
 
+        public string GetFullNameAndVersion()
+        {
+            return VersionGetter == null ? FullName : string.Format("{0} v{1}", FullName, VersionGetter());
```
ChengTian: This file will be compiled by Mono msc compiler, which doesn't support string interpolation. So using `string.Format()` here

BrennanConroy:
:shipit:

Issue #1572, opened 2015-04-02
Rename k-crossgen tool


ChengTian:
Reported by @moozzyk 

This is the last k-*.cmd we have. It's mainly an internally used tool, so the renaming is not urgent.

@muratg @davidfowl  Do we want to rename it to `dnx-crossgen`? Or you have some better names for it

davidfowl:
merge it into dnu or make it another global command /cc @loudej

ChengTian:
I guess merging it into dnu is better.

`dnvm` is using `k-crossgen` and assumes `k-corssgen` comes with dnx. Making `k-crossgen` a global command means `dnvm` will need to install this global command first. It's very different from the current approach.

davidfowl:
Ah hah! Yes lets merge it in

ChengTian:
So we all agree on the new name `dnx-crossgen`?

muratg:
dnx-crossgen sounds fine to me.  Unless @davidfowl has any objections, let's go with that name.

anurse:
`dnu optimize` ? Eventually, it could even take parameters such as allowing me to "optimize" specific packages/apps/runtimes/whatever.

I, as a user, have no idea what a `crossgen` is... but regardless, people don't generally run this command manually so I'm not very worried either way.

ChengTian:
I guess `optimize` is a too general term and users may imagine they can do something like
```
dnu optimize MyApp
```
to make their app run faster...

If we don't expect people to use this in most cases, `crossgen` can be a better choice, which means "keep away if you don't know what it is"

troydai:
In CLR we use Ngen.exe as well. I think `crossgen` is a more accurate name.

ChengTian:
Moving forward with `dnx-crossgen` if there is no objection

muratg:
:+1:

ChengTian:
Talked to @joeloff and confirmed that, as long as the command name is in the form of `<any_prefix>-crossgen.cmd`, our installer can always find it. So there is no change needed in installer part.

PR #1935, opened 2015-05-22
Add dnx-crossgen.cmd

ChengTian:
parent https://github.com/aspnet/dnx/issues/1572

@muratg 

Here is the plan: I am going to add `dnx-crossgen.cmd` without removing `k-crossgen.cmd` and change all other repos to use `dnx-crossgen.cmd`.

Then Coherence-Signed should be able to pass because even some part of our building process is still using the `k-crossgen.cmd`, `k-crossgen.cmd` is available.

After Coherence-Signed passes, I'll merge another PR that removes `k-crossgen.cmd`. This ensures smooth transition without breaking anything.

### DIFF
--- a/scripts/dnx-crossgen.cmd
+++ b/scripts/dnx-crossgen.cmd
@@ -1,0 +1,11 @@
+@Echo OFF
+SETLOCAL
+SET ERRORLEVEL=
+
+SET "DNX_RUNTIME_PATH=%~dp0."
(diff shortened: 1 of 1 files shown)


ChengTian:
Since we already agreed on the new name, I'll just check in when all PRs are ready

muratg:
Belated :shipit:

PR #1938, opened 2015-05-22
Remove k-crossgen.cmd

ChengTian:
Follow-up to https://github.com/aspnet/dnx/pull/1935. This should be merged as last step of https://github.com/aspnet/dnx/issues/1572 after Coherence-Signed passes.


moozzyk:
:ship::it:

PR #1939, opened 2015-05-22
Remove Ansi Color Codes from output in tests

BrennanConroy:


test/Microsoft.Framework.CommonTestUtils/TestUtils.cs
```diff
@@ -386,5 +386,30 @@ private static bool IsWindows()
             return PlatformHelper.IsWindows;
 #endif
         }
+
+        private static string RemoveAnsiColorCodes(string text)
+        {
+            int escapeIndex = 0;
```
moozzyk: looks like a great place to use a regex..

moozzyk:
:ship: :it:

test/Microsoft.Framework.CommonTestUtils/TestUtils.cs
```diff
@@ -386,5 +386,30 @@ private static bool IsWindows()
...
+            for (; ;)
+            {
+                escapeIndex = text.IndexOf("\x1b[", escapeIndex);
+                if (escapeIndex != -1)
+                {
+                    int endIndex = escapeIndex + 3;
+                    while (endIndex != text.Length && text[endIndex] != 'm')
```
ChengTian: `endIndex < text.Length`? Otherwise it will throw "index out of bounds" in edge case where text has the value of `"\x1b["`

BrennanConroy: endIndex wont index if it equals the length of the string, so no 'out of bounds' happens

ChengTian: I mean, when `text` is `"\x1b["`, `endIndex` is 3, while `text.Length` is 2

test/Microsoft.Framework.CommonTestUtils/TestUtils.cs
ChengTian: var

test/Microsoft.Framework.CommonTestUtils/TestUtils.cs
BrennanConroy: Oh, good point

ChengTian:
:shipit:

PR #1940, opened 2015-05-22
Consistent versioning of native projects

moozzyk:
Apart from removing magic (building on CI works differently than building outside CI) and moving building fully just to one build system this change makes it possible to access version inside the code which is needed in dnx for servicing.


moozzyk:
:watch:

BrennanConroy:
My favorite kind of PR...
I think it looks good.
:ship: :it: ?

Issue #1764, opened 2015-05-01
dnu restore log should indicate when postrestore and prepare scripts start getting excecuted


balachir:
1. New ASP.NET 5 project from StarterWeb template
2. View package restore log

EXPECTED
Package restore log should contain entries to indicate when postrestore and prepare scripts start getting executed. This will make it more intuitive to the user that these scripts in project.json are part of the package restore, and also give them a clear idea regarding which part of the log belongs to which part of the package restore.

![image](https://cloud.githubusercontent.com/assets/8246794/7437937/8700e078-f011-11e4-9488-87222265cd42.png)

PR #1942, opened 2015-05-23
Indicate the beginning of executed scripts

ChengTian:
parent https://github.com/aspnet/dnx/issues/1764

@balachir @muratg @davidfowl 

Sample output:
capture

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Scripts/ScriptExecutor.cs
+++ b/src/Microsoft.Framework.PackageManager/Scripts/ScriptExecutor.cs
@@ -15,6 +15,13 @@ namespace Microsoft.Framework.PackageManager
     {
         private static readonly string ErrorMessageTemplate = "The '{0}' script failed with status code {1}.";
 
+        public ScriptExecutor(IReport report = null)
+        {
+            Report = report;
+        }
+
+        public IReport Report { get; set; }
+
         public bool Execute(Runtime.Project project, string scriptName, Func<string, string> getVariable)
         {
             IEnumerable<string> scriptCommandLines;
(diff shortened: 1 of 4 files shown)


muratg:
Looks good, though I wonder "in project.json" part is useful or not. I assume there won't be other DNX pieces which will execute scripts.   I think we can check-in like this, and revisit later.

:shipit:

src/Microsoft.Framework.PackageManager/Scripts/ScriptExecutor.cs
```diff
@@ -23,6 +30,8 @@ public bool Execute(Runtime.Project project, string scriptName, Func<string, str
                 return true;
             }
 
+            Report.WriteLine($"Executing script '{scriptName}' in {Runtime.Project.ProjectFileName}");
```
dougbu: `Report` can be `null`. Need to use `Report?.WriteLine(...)`

src/Microsoft.Framework.PackageManager/Scripts/ScriptExecutor.cs
```diff
@@ -15,6 +15,13 @@ public class ScriptExecutor
     {
         private static readonly string ErrorMessageTemplate = "The '{0}' script failed with status code {1}.";
 
+        public ScriptExecutor(IReport report = null)
+        {
+            Report = report;
+        }
+
+        public IReport Report { get; set; }
```
dougbu: Curious: Why does this have a setter, let alone a `public` setter?

ChengTian: To make it work with `RestoreCommand` class, we need this public setter. `RestoreCommand` cannot finalize the value of `Reports` when it is constructing `ScriptExecutor`, so setting `IReport` with constructor is not enough.

We may need to redesign `RestoreCommand` to get rid of this but it's a separate topic.

src/Microsoft.Framework.PackageManager/Scripts/ScriptExecutor.cs
ChengTian: Good point

Issue #1155, opened 2015-02-02
Create 'runtimes' folder in "kpm bundle" output


ChengTian:
Today, when you `kpm bundle` an app with runtime included, runtime is bundled to 
``` 
<OutputRoot>/approot/packages
```
which also contains all other NuGet packages used by the app. It’s not consistent with structure of `$KRE_HOME`, which has
```
a. $KRE_HOME/packages for NuGet packages
b. $KRE_HOME/runtimes for runtime folders
```
To be more consistent, `kpm bundle` output should have the same structure:
```
a. <OutputRoot>/approot/runtimes
b. <OutputRoot>/approot/packages
```

PR #1944, opened 2015-05-23
"dnu publish" puts bundled runtimes to "runtimes" subfolder

ChengTian:
parent https://github.com/aspnet/dnx/issues/1155

Part of verification can only be done after @davidfowl fixes `dnu publish`.


davidfowl:
:shipit:

Issue #1934, opened 2015-05-22
dnu restore says JSON.NET could not be found


borgdylan:
When i restore packages i get this stack trace:
```
dylan@ubuntu-server:/var/www/Code/dylannet/vnext/dylan.NET.K$ dnu restore
Restoring packages for /var/www/Code/dylannet/vnext/dylan.NET.K/project.json
----------
System.IO.FileNotFoundException: Could not load file or assembly 'Newtonsoft.Json' or one of its dependencies
File name: 'Newtonsoft.Json'
  at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () <0xb41f5cc0 + 0x00035> in <filename unknown>:0 
  at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (System.Threading.Tasks.Task task) <0xb41f4d20 + 0x000b7> in <filename unknown>:0 
  at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Threading.Tasks.Task task) <0xb41f4c90 + 0x00084> in <filename unknown>:0 
  at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd (System.Threading.Tasks.Task task) <0xb41f4c40 + 0x0003f> in <filename unknown>:0 
  at System.Runtime.CompilerServices.TaskAwaiter`1[TResult].GetResult () <0xb44e96d0 + 0x0001f> 24246 in <filename unknown>:0 
  at Microsoft.Framework.PackageManager.RestoreCommand+<>c__DisplayClass72_0+<<ExecuteCommand>b__0>d.MoveNext () <0xb39075d0 + 0x00157> in <filename unknown>:0 
--- End of stack trace from previous location where exception was thrown ---
  at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () <0xb41f5cc0 + 0x00035> in <filename unknown>:0 
  at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (System.Threading.Tasks.Task task) <0xb41f4d20 + 0x000b7> in <filename unknown>:0 
  at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Threading.Tasks.Task task) <0xb41f4c90 + 0x00084> in <filename unknown>:0 
[...]

PR #1945, opened 2015-05-23
Stop recursion in AssemblyResolve

davidfowl:
- Short circuit by calling LoadAssemblyImpl directly instead of
recursion via Load
- Changed dnx451 implementation of the LoadContextAccessor to
store LoadContext directly instead of wrapping it in the interface.

#1934

### DIFF
--- a/src/Microsoft.Framework.Runtime.Loader/LoadContextAccessor.cs
+++ b/src/Microsoft.Framework.Runtime.Loader/LoadContextAccessor.cs
@@ -39,7 +39,7 @@ namespace Microsoft.Framework.Runtime.Loader
     {
         private static readonly LoadContextAccessor _instance = new LoadContextAccessor();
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Framework.Runtime.Loader/LoadContextAccessor.cs
```diff
@@ -59,11 +59,16 @@ public IAssemblyLoadContext Default
             }
         }
 
-        public IAssemblyLoadContext GetLoadContext(Assembly assembly)
+        IAssemblyLoadContext IAssemblyLoadContextAccessor.GetLoadContext(Assembly assembly)
```
BrennanConroy: public/private/internal?

davidfowl: implicit implementation

Issue #1849, opened 2015-05-12
"dnu restore" should use overload of SendAsync that does not buffer the response


ChengTian:
https://msdn.microsoft.com/en-us/library/hh193601(v=vs.118).aspx with HttpCompletionOption.ResponseHeadersRead.

This is a good way to avoid memory explosion.

muratg:
:+1:

PR #1959, opened 2015-05-27
"dnu restore" uses correct HTTP request option to avoid memory explosion

ChengTian:
parent https://github.com/aspnet/dnx/issues/1849

@Tratcher 

The goal is to have a seekable stream containing HTTP response content. However, we don't want to have the whole response content into memory.

Use `HttpCompletionOption.ResponseHeadersRead` and copy `response.Content` to a temp file, then open a stream to the temp file. This can meet the goal above.

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
@@ -129,7 +129,7 @@ namespace Microsoft.Framework.PackageManager.Restore.NuGet
             {
                 HttpStatusCode statusCode;
 
-                using (var response = await _client.SendAsync(request))
+                using (var response = await _client.SendAsync(request, HttpCompletionOption.ResponseHeadersRead))
                 {
                     response.EnsureSuccessStatusCode();
                     statusCode = response.StatusCode;


Tratcher:
:shipit:

src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
```diff
@@ -129,7 +129,7 @@ internal class HttpSource
             {
                 HttpStatusCode statusCode;
 
-                using (var response = await _client.SendAsync(request))
+                using (var response = await _client.SendAsync(request, HttpCompletionOption.ResponseHeadersRead))
                 {
                     response.EnsureSuccessStatusCode();
```
kichalla: we seem to be creating a file even when the status code is not successful...not sure if its intentional...

kichalla:
:shipit:

src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
ChengTian: Yeah, I wanted to avoid deeply nested `using` statements here. With `FileOptions.DeleteOnClose` option, the temporary file will always be deleted if status code is not successful or response content is invalid

davidfowl: If we could avoid making the file at all in that case it would be better but not if the code gets nasty :smile:

kichalla: if I am not wrong, following should work without the need for deep nesting of `using`:
1. make request with responseheadersread, ensure success status code...AFAIK we need not dispose the response here as its the response stream in step 3. that needs to be disposed.
2. create a file stream
3. using(response's stream) { copy to the file stream }

@Tratcher is the expert here so he can validate :smile:

Tratcher: The FileStream using extends to the end of this method. Extending the response using around all of that would make your nesting even deeper. Just a style issue.

But honestly there's not much point in optimizing the failure case. Perf primarily matters in the success case.

Issue #1539, opened 2015-03-30
Print out summary after dnu restore


davidfowl:
Here's the kind of information the summary should have

- How many packages were downloaded
- What feeds were considered
- What nuget.config files were used (might be tricky)
- Where packages were install to
- How long everything took (I think we have this already)

BrennanConroy:
549c7753dd6fd13a693dc66e3ef8dd1e8effecf3

PR #1964, opened 2015-05-27
Add some summary to dnu restore

BrennanConroy:
#1539 

summaryrestore


src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
```diff
@@ -93,10 +97,32 @@ public RestoreCommand(FrameworkName fallbackFramework)
...
+                    Reports.Error.WriteLine($"    {message}");
+                }
+            }
+
+            var settings = Settings as Settings;
+            if (settings != null)
+            {
+                var configFiles = settings.GetNuGetConfigFiles();
+
+                Reports.Information.WriteLine($"{Environment.NewLine}NuGet Config files used:");
```
dougbu: Suggest using `Reports.Quiet` so that the extra information goes away when `--quiet` is used.

BrennanConroy: Ya, that might be a good idea.

src/Microsoft.Framework.PackageManager/NuGet/Core/Configuration/Settings.cs
```diff
@@ -601,6 +601,24 @@ public bool DeleteSection(string section)
             return true;
         }
 
+        public IEnumerable<string> GetNuGetConfigFiles()
+        {
+            var next = _next;
```
BrennanConroy: Realized I start on the second element in the list here, will change to start at the beginning.

BrennanConroy:
@davidfowl @troydai @victorhurdugaci

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
```diff
@@ -190,6 +215,12 @@ public RestoreCommand(FrameworkName fallbackFramework)
                     Reports.Information.WriteLine(string.Format("Total time {0}ms", sw.ElapsedMilliseconds));
                 }
 
+                if (_installCount > 0)
+                {
+                    InformationMessages.GetOrAdd("Installed:", _ => new List<string>()).Add($"{_installCount} packages to {packagesDirectory}");
```
victorhurdugaci: "package(s)" because "1 packages" is not correct :)

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
```diff
@@ -93,10 +97,31 @@ public RestoreCommand(FrameworkName fallbackFramework)
...
-                Reports.Error.WriteLine("Errors in {0}".Red().Bold(), category.Key);
+                Reports.Error.WriteLine("{0}Errors in {1}".Red().Bold(), Environment.NewLine, category.Key);
                 foreach (var message in category.Value)
                 {
-                    Reports.Error.WriteLine("    {0}", message);
+                    Reports.Error.WriteLine($"    {message}");
+                }
+            }
+
+            var settings = Settings as Settings;
```
davidfowl: Add it to ISettings or expose Settings as Settings.

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
```diff
@@ -93,10 +97,31 @@ public RestoreCommand(FrameworkName fallbackFramework)
 
             foreach (var category in ErrorMessages)
             {
-                Reports.Error.WriteLine("Errors in {0}".Red().Bold(), category.Key);
+                Reports.Error.WriteLine("{0}Errors in {1}".Red().Bold(), Environment.NewLine, category.Key);
```
victorhurdugaci: Ugh, we really need to unify this coloring and how errors are displayed. It seems that everyone does it in their own way...

victorhurdugaci:
It would be nice to display how many packages were installed and how many were skipped because they were already there. Also, cached ones would be interesting.

But, overall, looks good

davidfowl:
I also want to see effective feeds.

victorhurdugaci:
Oh! Oh! Can we show from which feed each downloaded package came from?

Sent from my phone
________________________________

davidfowl:
@victorhurdugaci might be overkill.

src/Microsoft.Framework.PackageManager/NuGet/Core/Configuration/Settings.cs
```diff
@@ -601,6 +601,24 @@ public bool DeleteSection(string section)
...
         }
 
+        public IEnumerable<string> GetNuGetConfigFiles()
+        {
+            var next = this;
+            while (next != null)
+            {
+                yield return next.ConfigFilePath;
+
+                next = next._next;
```
troydai: this is so unreadable. can you rename this `next`?

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
BrennanConroy: Make an issue or something then!

troydai: Suggestion: String interpolation

troydai:
:watch:

BrennanConroy:
Are effective feeds ones that installed packages came from? Or does it include ones where the package was found but it was cached?

ChengTian:
The `Feed used` section is already displaying effective feeds because it only displays feeds returned by `PackageSourceUtils.GetEffectivePackageSources()`.

I think an effective feed means a feed actually in use, no matter it is cached or not.

src/Microsoft.Framework.PackageManager/NuGet/Core/Configuration/Settings.cs
troydai: Action?

troydai:
:watch:

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
```diff
@@ -93,10 +97,30 @@ public RestoreCommand(FrameworkName fallbackFramework)
...
-                Reports.Error.WriteLine("Errors in {0}".Red().Bold(), category.Key);
+                Reports.Error.WriteLine($"{Environment.NewLine}Errors in {category.Key}".Red().Bold());
                 foreach (var message in category.Value)
                 {
-                    Reports.Error.WriteLine("    {0}", message);
+                    Reports.Error.WriteLine($"    {message}");
+                }
+            }
+
+            if (Settings is Settings)
```
Eilon: You get MAAAAAAAAAAD props for writing this! Not as cool as intern Sonja's code, but still pretty cool.

troydai:
Looks good to me. :shipit:

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
```diff
@@ -93,10 +97,30 @@ public RestoreCommand(FrameworkName fallbackFramework)
...
+            {
+                var configFiles = (Settings as Settings).GetConfigFiles();
```
moozzyk: double cast - the recommended pattern is
```
var settings = Settings as Settings;
if(settings != null)
{
   configFiles = settings.GetConfigFiles();
...
```

PR #1965, opened 2015-05-27
Let VS do its thing

dougbu:
- catch DNX.sln up with recent changes (project addition)
- use test service in one .xproj


troydai:
:shipit:

dougbu:
4cea4cc

PR #1980, opened 2015-05-29
Fix DesignTimeAssemblyLoadContextFactory

troydai:
@davidfowl @NTaylorMullen


troydai:
Use https://github.com/aspnet/dnx/pull/1981. Target release branch.

PR #1981, opened 2015-05-29
Fix DesignTimeAssemblyLoadContextFactory

troydai:
@davidfowl @NTaylorMullen

### DIFF
--- a/src/Microsoft.Framework.DesignTimeHost/DesignTimeAssemblyLoadContextFactory.cs
+++ b/src/Microsoft.Framework.DesignTimeHost/DesignTimeAssemblyLoadContextFactory.cs
@@ -31,23 +31,18 @@ namespace Microsoft.Framework.DesignTimeHost
 
         public IAssemblyLoadContext Create(IServiceProvider serviceProvider)
         {
-            return GetRuntimeFactory(serviceProvider).Create(serviceProvider);
-        }
-
-        private IAssemblyLoadContextFactory GetRuntimeFactory(IServiceProvider serviceProvider)
-        {
-            var cacheKey = Tuple.Create("RuntimeLoadContextFactory", _project.Name, _appEnv.Configuration, _appEnv.RuntimeFramework);
-
-            return _cache.Get<IAssemblyLoadContextFactory>(cacheKey, ctx =>
+            var hostContextKey = Tuple.Create("RuntimeLoadContext", _project.Name, _appEnv.Configuration, _appEnv.RuntimeFramework);
+            
+            var appHostContext = _cache.Get<ApplicationHostContext>(hostContextKey, ctx => 
             {
                 var applicationHostContext = new ApplicationHostContext(serviceProvider,
-                                                                    _project.ProjectDirectory,
-                                                                    packagesDirectory: null,
-                                                                    configuration: _appEnv.Configuration,
-                                                                    targetFramework: _appEnv.RuntimeFramework,
-                                                                    cache: _cache,
-                                                                    cacheContextAccessor: _cacheContextAccessor,
-                                                                    namedCacheDependencyProvider: _namedDependencyProvider);
+                                                                        _project.ProjectDirectory,
+                                                                        packagesDirectory: null,
+                                                                        configuration: _appEnv.Configuration,
+                                                                        targetFramework: _appEnv.RuntimeFramework,
+                                                                        cache: _cache,
+                                                                        cacheContextAccessor: _cacheContextAccessor,
+                                                                        namedCacheDependencyProvider: _namedDependencyProvider);
 
                 applicationHostContext.DependencyWalker.Walk(_project.Name, _project.Version, _appEnv.RuntimeFramework);
 
@@ -62,9 +57,12 @@ namespace Microsoft.Framework.DesignTimeHost
 
                 // Add a cache dependency on restore complete to reevaluate dependencies
                 ctx.Monitor(_namedDependencyProvider.GetNamedDependency(_project.Name + "_Dependencies"));
-
-                return new AssemblyLoadContextFactory(applicationHostContext.ServiceProvider);
+                
+                return applicationHostContext;
             });
+
+            var factory = new AssemblyLoadContextFactory(appHostContext.ServiceProvider);
+            return factory.Create(appHostContext.ServiceProvider);
         }
     }
 }


src/Microsoft.Framework.DesignTimeHost/DesignTimeAssemblyLoadContextFactory.cs
```diff
@@ -31,23 +31,18 @@ internal class DesignTimeAssemblyLoadContextFactory : IAssemblyLoadContextFactor
...
-        }
-
-        private IAssemblyLoadContextFactory GetRuntimeFactory(IServiceProvider serviceProvider)
-        {
-            var cacheKey = Tuple.Create("RuntimeLoadContextFactory", _project.Name, _appEnv.Configuration, _appEnv.RuntimeFramework);
-
-            return _cache.Get<IAssemblyLoadContextFactory>(cacheKey, ctx =>
+            var hostContextKey = Tuple.Create("RuntimeLoadContext", _project.Name, _appEnv.Configuration, _appEnv.RuntimeFramework);
+            
+            var appHostContext = _cache.Get<ApplicationHostContext>(hostContextKey, ctx =>
```
NTaylorMullen: Not too familiar with this part of the code base so forgive my ignorance... I don't see when this cache entry is ever set. It must though because you `appHostContext.ServiceProvider` below. 

For more background: When does it get set, is anyone else expected to "get" it?

src/Microsoft.Framework.DesignTimeHost/DesignTimeAssemblyLoadContextFactory.cs
```diff
@@ -62,9 +57,12 @@ private IAssemblyLoadContextFactory GetRuntimeFactory(IServiceProvider servicePr
 
                 // Add a cache dependency on restore complete to reevaluate dependencies
                 ctx.Monitor(_namedDependencyProvider.GetNamedDependency(_project.Name + "_Dependencies"));
-
-                return new AssemblyLoadContextFactory(applicationHostContext.ServiceProvider);
+
```
NTaylorMullen: no need for trailing whitespace.

src/Microsoft.Framework.DesignTimeHost/DesignTimeAssemblyLoadContextFactory.cs
davidfowl: It's a GetOrAdd style API, that's what the lambda is for.

src/Microsoft.Framework.DesignTimeHost/DesignTimeAssemblyLoadContextFactory.cs
```diff
@@ -62,9 +57,12 @@ private IAssemblyLoadContextFactory GetRuntimeFactory(IServiceProvider servicePr
 
                 // Add a cache dependency on restore complete to reevaluate dependencies
                 ctx.Monitor(_namedDependencyProvider.GetNamedDependency(_project.Name + "_Dependencies"));
-
-                return new AssemblyLoadContextFactory(applicationHostContext.ServiceProvider);
+                
+                return applicationHostContext;
             });
+            
+            var factory = new AssemblyLoadContextFactory(appHostContext.ServiceProvider);
```
davidfowl: This is fine for now. I guess we don't really need to cache the ProjectLoader and NuGetLoader because they are mostly stateless.

davidfowl:
@NTaylorMullen can you help @troydai test this out?

src/Microsoft.Framework.DesignTimeHost/DesignTimeAssemblyLoadContextFactory.cs
NTaylorMullen: For more background: If the `ApplicationHostContext` is cached why did caching the `AssemblyLoadContextFactory` based on the still cached `ApplicationHostContext` break the scenario? Isn't that assuming the `ServiceProvider` gets updated (since it can't be set)? From my understanding of `IServiceProvider` that shouldn't be possible. Definitely missing something here.

src/Microsoft.Framework.DesignTimeHost/DesignTimeAssemblyLoadContextFactory.cs
NTaylorMullen: Oh dur, of course since there's a lambda. Kinda odd API name though `Get` instead of `GetOrAdd`.

src/Microsoft.Framework.DesignTimeHost/DesignTimeAssemblyLoadContextFactory.cs
davidfowl: Caching didn't break it, calling Create with the wrong service provider broke it.

This was the old code:
```
return GetRuntimeFactory(serviceProvider).Create(serviceProvider);
```

The new code calls create with the correct service provider.

NTaylorMullen: Ah, that makes sense. Muchas gracias :smile:

NTaylorMullen:
@davidfowl will do

troydai:
@NTaylorMullen I'll send you a private build, if you have any quick test use it. Otherwise talk to me tomorrow, let's write some tests.

src/Microsoft.Framework.DesignTimeHost/DesignTimeAssemblyLoadContextFactory.cs
troydai: Maybe we should rename it.

src/Microsoft.Framework.DesignTimeHost/DesignTimeAssemblyLoadContextFactory.cs
troydai: Yeah, VS Code sucks at formatting whitespace.

moozzyk: I told you: https://visualstudiogallery.msdn.microsoft.com/bcb3ef8e-a3a4-401e-9e63-fdf00b45268f and you will never miss a trailing whitespace...

troydai: Thanks for the advice again @moozzyk :) These days I rarely start VS.

moozzyk: @troydai OK. Then do this in VS Code (if you have not don this already): 
1) Ctrl + P type Settings
2) Select - "Preferences: Open User Settings"
3) Paste this:

// Place your settings in this file to overwrite the default settings
{
	"editor.renderWhitespace": true,
	"editor.insertSpaces": "auto"
}

The result: You will see white spaces (tabs vs. spaces) as in VS. If you do tab it will insert spaces if the file uses spaces (hence auto).

troydai:
@davidfowl review?

PR #1984, opened 2015-05-29
Removing all dnx.clr.managed.dll dependencies

moozzyk:
dnx.clr.managed.dll will need to be strong-name signed and therefore cannot depend on any assemblies that don't have a strong name.

### DIFF
--- a/src/dnx.clr.managed/RuntimeBootstrapper.cs
+++ b/src/dnx.clr.managed/RuntimeBootstrapper.cs
@@ -1,0 +1,29 @@
+﻿using System;
+using System.Reflection;
+using System.Threading.Tasks;
+
+namespace dnx.clr.managed
+{
+    internal class RuntimeBootstrapper
+    {
+        public static int Execute(string[] argv)
+        {
+            var executeMethodInfo = GetBootstrapperType()
+                .GetMethod("Execute", BindingFlags.Static | BindingFlags.Public, null, new[] { argv.GetType() }, null);
+            return (int)executeMethodInfo.Invoke(null, new object[] { argv });
+        }
+
+        public static Task<int> ExecuteAsync(string[] argv)
+        {
+            var executeMethodInfo = GetBootstrapperType()
+                .GetMethod("ExecuteAsync", BindingFlags.Static | BindingFlags.Public, null, new[] { argv.GetType() }, null);
+            return (Task<int>)executeMethodInfo.Invoke(null, new object[] { argv });
+        }
+
+        private static Type GetBootstrapperType()
+        {
+            var dnxHost = Assembly.Load("dnx.host");
+            return dnxHost.GetType("dnx.host.RuntimeBootstrapper");
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


moozzyk:
Travis failed with:

 Assertion at threadpool.c:1201, condition `threads != NULL' not met

=================================================================
Got a SIGABRT while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries 
used by your application.
=================================================================

src/dnx.clr.managed/dnx.clr.managed.xproj
```diff
@@ -16,5 +16,13 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
     <ProduceOutputsOnBuild>False</ProduceOutputsOnBuild>
   </PropertyGroup>
+  <ItemGroup>
```
davidfowl: Remove this

src/dnx.clr.managed/RuntimeBootstrapper.cs
```diff
@@ -0,0 +1,29 @@
...
+            var executeMethodInfo = GetBootstrapperType()
+                .GetMethod("Execute", BindingFlags.Static | BindingFlags.Public, null, new[] { argv.GetType() }, null);
+            return (int)executeMethodInfo.Invoke(null, new object[] { argv });
+        }
+
+        public static Task<int> ExecuteAsync(string[] argv)
+        {
+            var executeMethodInfo = GetBootstrapperType()
+                .GetMethod("ExecuteAsync", BindingFlags.Static | BindingFlags.Public, null, new[] { argv.GetType() }, null);
+            return Task.FromResult((int)executeMethodInfo.Invoke(null, new object[] { argv }));
```
davidfowl: This doesn't look correct. The method returns a Task<int> already, why not just cast?

moozzyk: :blush: Crap. You are right. I will fix and will check if we have any tests testing async main method. This should have failed.

src/dnx.clr.managed/dnx.clr.managed.xproj
moozzyk: OK. I was not sure whether we keep xproj in sync with project.json.

src/dnx.clr.managed/RuntimeBootstrapper.cs
moozzyk: This does not show up since in Helios we apparently use a handshake mechanism based on a TCS. Since the application started successfully the result of the tcs was set and when the InvalidCastException was thrown afterwards trying to set the exception was a no op since the result of the tcs was already set.

moozzyk:
@davifowl - updated the PR. Tested manually that it is no longer throwing the `InvalidCastException`

davidfowl:
:shipit:

troydai:
:shipit:

PR #1986, opened 2015-05-29
Add Microsoft.Framework.Runtime.Internals.xproj

troydai:
The xproj file is missing.

PR #1993, opened 2015-05-30
Davidfowl/fix transitive load

davidfowl:


NTaylorMullen:
:shipit:

src/Microsoft.Framework.Runtime.Internals/Microsoft.Framework.Runtime.Internals.xproj
```diff
@@ -0,0 +1,20 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="__ToolsVersion__" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
dougbu: Fix this .xproj with one of the scripts before checking it in. Or wait for @troydai's PR to fix the same gap.

troydai:
Update the xproj here https://github.com/aspnet/dnx/pull/1986/files

PR #2002, opened 2015-06-01
patch to make ef portable build work with 'any'

anurse:
I'll clean it up shortly, but the change should be straightforward.

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Program.cs
+++ b/src/Microsoft.Framework.PackageManager/Program.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics;
 using System.Net;
 using System.Threading;
 using Microsoft.Framework.Runtime;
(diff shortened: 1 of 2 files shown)


anurse:
Cleaned up.

/cc @davidfowl @troydai @muratg

src/Microsoft.Framework.PackageManager/Program.cs
```diff
@@ -66,5 +69,45 @@ public int Main(string[] args)
 
             return app.Execute(args);
         }
+
+        [Conditional("DEBUG")]
+        // TODO(anurse): This seems like a broadly useful helper
+        private static void WaitForOrLaunchDebuggerIfRequested(ref string[] args)
```
anurse: This ended up being really useful. DNU "needs" it's own debug helper code because it's difficult to use the dnx one since you generally run dnu through a batch file so it's hard to insert arguments between "dnx" and "Microsoft.Framework.PackageManager". It's possible, but hard. Also, I did add a few extra features to the debug helper.

Happy to remove it if people really don't like it, but I think it's useful. Might even be a useful helper to put in a Sources package

anurse:
Some unit tests broke, I'm fixing those.

anurse:
We're doing this differently.

PR #2004, opened 2015-06-01
Adding native support for strong naming assemblies.

moozzyk:


test/Microsoft.Framework.Runtime.Tests/ProjectFacts.cs
```diff
@@ -373,7 +373,7 @@ public void CompilerOptionsAreSet()
         {
             var project = Project.GetProject(@"
 {
-    ""compilationOptions"": { ""allowUnsafe"": true, ""define"": [""X"", ""y""], ""platform"": ""x86"", ""warningsAsErrors"": true, ""optimize"": true }
+    ""compilationOptions"": { ""allowUnsafe"": true, ""define"": [""X"", ""y""], ""platform"": ""x86"", ""warningsAsErrors"": true, ""optimize"": true, ""strongNameKeyFile"" : ""c:\\keyfile.snk"", ""delaySign"" : true }
```
pranavkm: Can we get a couple of variants for the `delaySign` values ('1`, '-1`, `True`, `""`)?

pranavkm:
Looks good. Would like to see more test cases before :shipit:

test/Microsoft.Framework.Runtime.Tests/ProjectFacts.cs
moozzyk: Not sure if I get the purpose - wouldn't it be just like testing the JSON parser (which I assume is already tested)? We only allow `true/false` `bool` values in the project.json and I assume `rawOptions.ValueAsNullableBoolean("delaySign")` will break on everything else. I left `1`/`true` for environment variables to not break anyone who is already using it (e.g. NuGet).

moozzyk:
I tested the changes manually on the dnx.clr.managed.dll (which is the target for this change)

test/Microsoft.Framework.Runtime.Tests/ProjectFacts.cs
pranavkm: Nevermind - didn't realize the place you were string comparing `1`, `true` is the env variable. Ignore my comment

moozzyk: No worries :)

src/Microsoft.Framework.Runtime.Roslyn/RoslynCompiler.cs
```diff
@@ -192,17 +193,21 @@ public class RoslynCompiler
...
-                    string.Equals(delaySignString, "true", StringComparison.OrdinalIgnoreCase) ||
-                    string.Equals(delaySignString, "1", StringComparison.OrdinalIgnoreCase));
+                var delaySign =
+                    delaySignString == null
+                        ? compilationContext.Compilation.Options.DelaySign
+                        : string.Equals(delaySignString, "true", StringComparison.OrdinalIgnoreCase) ||
+                          string.Equals(delaySignString, "1", StringComparison.OrdinalIgnoreCase);
 
-                var strongNameProvider = new DesktopStrongNameProvider();
+                var strongNameProvider = new DesktopStrongNameProvider(ImmutableArray.Create(Directory.GetCurrentDirectory()));
```
davidfowl: That's the wrong directory. Use the project directory of the compilation context

moozzyk: Will do

davidfowl: This entire strong naming code path also needs to be skipped on mono.

davidfowl: You can fake sign on any platform if no key is provided and delay sign is true. We should also support that.

moozzyk: do you mean we should not throw the error if delay sign is true and the file key is not provided and just skip the code? Why would it be useful?

moozzyk: Fake signing will be separate.

moozzyk:
Updated with:
- changed `strongNameKeyFile` to `keyFile`
- fixed the issue with the snk file search location - using `compilationContext.Project.ProjectDirector` instead of `Directory.GetCurrentDirectory()`
- added a check to fail on Mono

moozzyk:
:watch:

davidfowl:
:shipit:

PR #2005, opened 2015-06-02
Strong naming dnx.clr.managed.dll

moozzyk:


moozzyk:
Closing. We will use fake strong name.

PR #2008, opened 2015-06-02
teach dnx about netcore50 compat with netportable50

anurse:
We're still locking down the netportable name, but this logic is sound no matter the final name :).

/cc @bricelam @troydai @muratg @davidfowl


troydai:
should this be in release as well?

anurse:
It's part of the new PR. It should be in release.

PR #2009, opened 2015-06-02
teach dnx about netcore50 compat with netportable50

anurse:
We're still locking down the netportable name, but this logic is sound no matter the final name :).

/cc @bricelam @troydai @muratg @davidfowl


bricelam:
:shipit:

bricelam:
I was able to build this locally and get it working. Thanks, @anurse

troydai:
:shipit:

bricelam:
Merged as 9b57562fc7f685d0399b1602318d3e8cc5d65770

Issue #1901, opened 2015-05-19
dnu source 


victorhurdugaci:
Just adding a work item for myself for beta6

victorhurdugaci:
Brain dump of the ideas so far:

`project.json` gets a new top level property (example below if for git):

```
"repository": {
    "type": "git",
    "url": "https://github/blah..."
}
```

The only required property is "type". Then the source control provider handles the rest.

For "git" we embed, in the nuspec:

1. "git" as the repo type
1. The repo URL
1. The commit hash
1. The path to the project relative to the repo root (e.g. `src`)

Other thoughts:

- `dnu source <packageName>` will get the sources for `<packageName>`. The version is resolved from the lock file and this command can only run in the context of a project
- By default `dnu source` will clone sources in `./.dnx/source/SOMETHING`. That something is probably something that has to do with the repo and commit id. It cannot be a package name because you can produce multiple packages from the same repo and we don't want to clone multiple times for the same commit id. Maybe you can specify a different path
- `dnu pack` will embed all the info in the nuspec
- If there are multiple packages in the same folder, like we do in our case with the `src`, then all of them will be sourced when one is sourced.
- This command will update `global.json` with the source info so, even though you run it a project level, the changes affect the entire solution.
- There should be a switch to stop using sources like `--no-source` or something.
- For git, I think we should clone with `depth 1` so we don't fill the harddrive with history

Still to decide:

- Can you point to existing an source clone? If yes, what happens if there are already changes there? Do we just use it as is or switch to that commit hash? Currently, I think the best is to just the source as is, meaning that if you pass a repo manually, we assume you know what you are doing.
- How do we store the repos on harddrive to easily find them if they were already cloned? What if they have changes we revert them or simply not care?
- Extensibility? What if I want TFS or... SourceSafe?
- Do we assume that git or the source control CLI must be on the path? Can you pass the path as argument instead?

cc @davidfowl @lodejard @DamianEdwards

PR #2010, opened 2015-06-02
dnu source + tests

victorhurdugaci:
Fixes https://github.com/aspnet/dnx/issues/1901

`Dnu source` to store the repo information and the get the sources for a particular package. For now, the information is stored in the `repo.json` file in the nupkg.

Work cross platform.

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Reports.cs
+++ b/src/Microsoft.Framework.PackageManager/Reports.cs
@@ -10,6 +10,26 @@ namespace Microsoft.Framework.PackageManager
         public IReport Quiet { get; set; }
         public IReport Error { get; set; }
 
+        public void WriteInformation(string message)
+        {
+            Information.WriteLine(message);
+        }
+
+        public void WriteVerbose(string message)
+        {
+            Verbose.WriteLine(message);
+        }
+
+        public void WriteError(string message)
+        {
+            Error.WriteLine(message.Red());
+        }
+
+        public void WriteWarning(string message)
+        {
+            Information.WriteLine(message.Yellow());
+        }
+
         public Reports ShallowCopy()
         {
             return MemberwiseClone() as Reports;
--- a/src/Microsoft.Framework.PackageManager/SourceControl/Constants.cs
+++ b/src/Microsoft.Framework.PackageManager/SourceControl/Constants.cs
@@ -1,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Framework.PackageManager.SourceControl
+{
+    public static class Constants
+    {
+        public const string RepoTypeKey = "type";
+        public const string SnapshotInfoFileName = "repo.json";
+    }
+}
--- a/src/Microsoft.Framework.PackageManager/SourceControl/GitSourceControlProvider.cs
+++ b/src/Microsoft.Framework.PackageManager/SourceControl/GitSourceControlProvider.cs
@@ -1,0 +1,215 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Microsoft.Framework.PackageManager.SourceControl
+{
+    internal class GitSourceControlProvider
+    {
+        private const string Git = "git";
+
+        private const string RepositoryUrlKey = "url";
+        private const string CommitHashKey = "commit";
+        private const string ProjectPathKey = "path";
+
+        private bool? _isInstalled;
+
+        private readonly Reports _buildReports;
+
+        public GitSourceControlProvider(Reports buildReports)
+        {
+            if (buildReports == null)
+            {
+                throw new ArgumentNullException(nameof(buildReports));
+            }
+
+            _buildReports = buildReports;
+        }
+
+        public bool IsInstalled
+        {
+            get
+            {
+                if (!_isInstalled.HasValue)
+                {
+                    _isInstalled = ProcessUtilities.ExecutableExists(Git);
+                }
+
+                return _isInstalled.Value;
+            }
+        }
+
+        public void AddMissingSnapshotInformation(string folderName, IDictionary<string, string> snapshotInformation)
+        {
+            if (!snapshotInformation.ContainsKey(RepositoryUrlKey) || string.IsNullOrEmpty(snapshotInformation[RepositoryUrlKey]))
+            {
+                throw new ArgumentNullException("The repository URL must be specified.");
+            }
+
+            if (!snapshotInformation.ContainsKey(CommitHashKey))
+            {
+                snapshotInformation[CommitHashKey] = GetHeadCommitId(folderName);
+            }
+
+            if (!snapshotInformation.ContainsKey(ProjectPathKey))
+            {
+                var repositoryRoot = GetRepositoryRoot(folderName);
+
+                // Get the path relative to the repo root
+                var pathRelativeToRepositoryRoot = Path.GetDirectoryName(folderName)
+                    .Substring(repositoryRoot.Length)
+                    .Replace('\\', '/')
+                    .TrimStart('/');
+                snapshotInformation[ProjectPathKey] = pathRelativeToRepositoryRoot;
+            }
+        }
+
+        private string GetHeadCommitId(string folderName)
+        {
+            StringBuilder commitId = new StringBuilder();
+            bool hasErrors = false;
+
+            if (ProcessUtilities.Execute(
+                    Git,
+                    arguments: "rev-parse HEAD",
+                    workingDirectory: folderName,
+                    stdOut: (msg) => { commitId.Append(msg); },
+                    stdErr: (msg) => {
+                        hasErrors = true;
+                        _buildReports.WriteError(msg);
+                    })
+                || hasErrors)
+            {
+                return commitId.ToString();
+            }
+
+            throw new InvalidOperationException("Failed to get the commit hash.");
+        }
+
+        private string GetRepositoryRoot(string folderName)
+        {
+            StringBuilder repositoryRoot = new StringBuilder();
+            bool hasErrors = false;
+
+            if (ProcessUtilities.Execute(
+                Git,
+                arguments: "rev-parse --show-toplevel",
+                workingDirectory: folderName, 
+                stdOut: (msg) => { repositoryRoot.Append(msg); },
+                stdErr: (msg) => {
+                    hasErrors = true;
+                    _buildReports.WriteError(msg);
+                })
+                || hasErrors)
+            {
+                return repositoryRoot.ToString();
+            }
+
+            throw new InvalidOperationException("Failed to resolve the repository root.");
+        }
+
+        public string CreateShortFolderName(IDictionary<string, string> snapshotInfo)
+        {
+            if (!ValidateRepositoryType(snapshotInfo) || 
+                !ValidateBuildSnapshotInformation(snapshotInfo))
+            {
+                return null;
+            }
+
+            var repoUrl = snapshotInfo[RepositoryUrlKey];
+            var commitHash = snapshotInfo[CommitHashKey];
+
+            string repoName = Path.GetFileNameWithoutExtension(repoUrl);
+            string shortHash = commitHash.Substring(0, 8);
+
+            return repoName + shortHash;
+        }
+
+        public bool GetSources(string destinationFolder, IDictionary<string, string> snapshotInfo)
+        {
+            if (!ValidateBuildSnapshotInformation(snapshotInfo))
+            {
+                return false;
+            }
+
+            var repoUrl = snapshotInfo[RepositoryUrlKey];
+            var commitHash = snapshotInfo[CommitHashKey];
+
+            _buildReports.WriteInformation($"Cloning from: {repoUrl}");
+
+            // First clone
+            if (!ProcessUtilities.Execute(
+                Git,
+                $"clone --recursive {repoUrl} {destinationFolder}",
+                workingDirectory: null,
+                stdOut: _buildReports.WriteInformation,
+                stdErr: _buildReports.WriteError))
+            {
+                return false;
+            }
+
+            _buildReports.WriteVerbose($"Resetting to commit hash: {repoUrl}");
+
+            // Then sync to that particular commit
+            if (!ProcessUtilities.Execute(
+                Git,
+                $"reset --hard {commitHash}",
+                workingDirectory: destinationFolder,
+                stdOut: _buildReports.WriteInformation,
+                stdErr: _buildReports.WriteError))
+            {
+                return false;
+            }
+
+            return true;
+        }
+
+        public string GetSourceFolderPath(IDictionary<string, string> snapshotInfo)
+        {
+            if (ValidateBuildSnapshotInformation(snapshotInfo))
+            {
+                return snapshotInfo[ProjectPathKey];
+            }
+
+            return null;
+        }
+
+        public bool IsRepository(string folder)
+        {
+            return ProcessUtilities.Execute(Git, "status", workingDirectory: folder);
+        }
+
+        private bool ValidateRepositoryType(IDictionary<string, string> snapshotInfo)
+        {
+            if (!snapshotInfo.ContainsKey(RepositoryUrlKey))
+            {
+                _buildReports.WriteError("The repository information is missing the repository URL.");
+                return false;
+            }
+
+            return true;
+        }
+
+        private bool ValidateBuildSnapshotInformation(IDictionary<string, string> snapshotInfo)
+        {
+            bool noErrors = true;
+
+            if (!snapshotInfo.ContainsKey(CommitHashKey))
+            {
+                _buildReports.WriteError("The repository information is missing the commit hash.");
+                noErrors = false;
+            }
+            if (!snapshotInfo.ContainsKey(ProjectPathKey))
+            {
+                _buildReports.WriteError("The repository information is missing the project path.");
+                noErrors = false;
+            }
+
+            return noErrors;
+        }
+    }
+}
--- a/src/Microsoft.Framework.PackageManager/SourceControl/ProcessUtilities.cs
+++ b/src/Microsoft.Framework.PackageManager/SourceControl/ProcessUtilities.cs
@@ -1,0 +1,108 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using System.Text;
+using Microsoft.Framework.Runtime;
+
+namespace Microsoft.Framework.PackageManager.SourceControl
+{
+    internal static class ProcessUtilities
+    {
+        private const int DefaultProcessTimeout = 5 * 60 * 1000;
+
+        public static bool ExecutableExists(string executableName)
+        {
+            string whereApp;
+            if (PlatformHelper.IsWindows)
+            {
+                whereApp = "where";
+            }
+            else
+            {
+                whereApp = "whereis";
+            }
+
+            return Execute(whereApp, executableName);
+        }
+
+        public static bool Execute(
+            string executable,
+            string arguments = null,
+            string workingDirectory = null,
+            Action<string> stdOut = null,
+            Action<string> stdErr = null,
+            int timeout = DefaultProcessTimeout)
+        {
+            Process proc = null;
+
+            try
+            {
+                ProcessStartInfo startInfo = new ProcessStartInfo()
+                {
+                    FileName = executable,
+                    Arguments = arguments,
+
+                    RedirectStandardOutput = true,
+                    RedirectStandardError = true,
+
+                    UseShellExecute = false,
+                    
+#if DNX451
+                    WindowStyle = ProcessWindowStyle.Hidden,
+#else
+                    CreateNoWindow = true
+#endif
+                };
+
+                if (!string.IsNullOrEmpty(workingDirectory))
+                {
+                    startInfo.WorkingDirectory = workingDirectory;
+                }
+
+                proc = new Process();
+                proc.StartInfo = startInfo;
+                proc.Start();
+
+                proc.EnableRaisingEvents = true;
+
+                if (stdOut != null)
+                {
+                    proc.OutputDataReceived += (sender, e) =>
+                    {
+                        if (e.Data != null)
+                        {
+                            stdOut(e.Data);
+                        }
+                    };
+                    proc.BeginOutputReadLine();
+                }
+
+                if (stdErr != null)
+                {
+                    proc.ErrorDataReceived += (sender, e) =>
+                    {
+                        if (e.Data != null)
+                        {
+                            stdErr(e.Data);
+                        }
+                    };
+                    proc.BeginErrorReadLine();
+                }
+
+                proc.WaitForExit(timeout);
+
+                return proc.ExitCode == 0;
+            }
+            catch (Exception ex)
+            {
+                if (stdErr != null)
+                {
+                    stdErr(ex.ToString());
+                }
+                return false;
+            }
+        }
+    }
+}
--- a/src/Microsoft.Framework.PackageManager/SourceControl/SourceBuilder.cs
+++ b/src/Microsoft.Framework.PackageManager/SourceControl/SourceBuilder.cs
@@ -1,0 +1,106 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using Newtonsoft.Json;
+using NuGet;
+
+namespace Microsoft.Framework.PackageManager.SourceControl
+{
+    internal class SourceBuilder
+    {
+        private readonly Reports _reports;
+        private readonly IPackageBuilder _packageBuilder;
+        private readonly Runtime.Project _project;
+
+        public SourceBuilder(Runtime.Project project, IPackageBuilder packageBuilder, Reports buildReport)
+        {
+            if (project == null)
+            {
+                throw new ArgumentNullException(nameof(project));
+            }
+            if (buildReport == null)
+            {
+                throw new ArgumentNullException(nameof(buildReport));
+            }
+            if (packageBuilder == null)
+            {
+                throw new ArgumentNullException(nameof(packageBuilder));
+            }
+
+            _project = project;
+            _packageBuilder = packageBuilder;
+            _reports = buildReport;
+        }
+
+        public bool Build(string outputPath)
+        {
+            if (_project.Repository == null)
+            {
+                // Nothing to do
+                return true;
+            }
+
+            _reports.Information.WriteLine("Embedding repository information...");
+
+            string repositoryType;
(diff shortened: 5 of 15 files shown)


victorhurdugaci:
Please review: @davidfowl @glennc

src/Microsoft.Framework.PackageManager/ConsoleCommands/SourceConsoleCommand.cs
```diff
@@ -0,0 +1,52 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Framework.PackageManager.SourceControl;
+using Microsoft.Framework.Runtime.Common.CommandLine;
+
+namespace Microsoft.Framework.PackageManager
+{
+    internal class SourceConsoleCommand
+    {
+        public static void Register(CommandLineApplication cmdApp, ReportsFactory reportsFactory)//, IApplicationEnvironment applicationEnvironment)
```
BrennanConroy: Remove comment

src/Microsoft.Framework.PackageManager/SourceControl/GitSourceControlProvider.cs
```diff
@@ -0,0 +1,177 @@
...
+                $"reset --hard {commitHash}",
+                workingDirectory: destinationFolder,
+                stdOut: out stdOut,
+                stdErr: out stdErr))
+            {
+                _buildReports.WriteError(stdErr);
+                return false;
+            }
+
+            _buildReports.WriteError(stdErr);
```
BrennanConroy: We want to write out stdErr when it succeeded?

src/Microsoft.Framework.PackageManager/SourceControl/SourceBuilder.cs
```diff
@@ -0,0 +1,113 @@
...
+                minimal: true))
+            {
+                return false;
+            }
+
+            var projectFolder = Path.GetDirectoryName(_project.ProjectFilePath);
+
+            if (!sourceControlProvider.IsRepo(projectFolder))
+            {
+                WriteWarning($"The project is not under a repository.  The repository information will not be included in the package.");
```
BrennanConroy: Double space after period

BrennanConroy: Also, no need for $

src/Microsoft.Framework.PackageManager/SourceControl/SourceCommand.cs
```diff
@@ -0,0 +1,244 @@
...
+            if (!snapshotInfo.TryGetValue(Constants.RepoTypeKey, out repoType))
+            {
+                _reports.WriteError("Repository type information is missing from the repository information file.");
+                return false;
+            }
+
+            var provider = SourceControlProviderFactory.ResolveProvider(repoType, _reports);
+            if (provider == null)
+            {
+                _reports.WriteError($"Unknown repository type '{provider}'");
```
BrennanConroy: I don't think you meant to print provider (which is null) here

src/Microsoft.Framework.PackageManager/SourceControl/SourceCommand.cs
```diff
@@ -0,0 +1,244 @@
...
+            if (!Directory.Exists(srcFolder))
+            {
+                _reports.WriteError($"The source code folder '{srcFolder}' is missing.");
+                return false;
+            }
+
+            var globalFile = GlobalSettings.GetGlobalFilePath(_solutionRoot);
+            if (!File.Exists(globalFile))
+            {
+                _reports.WriteError($"The '{GlobalSettings.GlobalFileName}' is missing from '{_solutionRoot}'.");
```
BrennanConroy: globalFile?

src/Microsoft.Framework.PackageManager/SourceControl/SourceCommand.cs
```diff
@@ -0,0 +1,244 @@
...
+                return false;
+            }
+
+            _reports.Verbose.WriteLine($"Updating {GlobalSettings.GlobalFileName}...");
```
BrennanConroy: same as above

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuSourceTests.cs
```diff
@@ -0,0 +1,263 @@
...
+using Newtonsoft.Json.Linq;
+using Xunit;
+
+namespace Microsoft.Framework.PackageManager.FunctionalTests
+{
+    [Collection(nameof(PackageManagerFunctionalTestCollection))]
+    public class DnuSourceTests
+    {
+        [Theory]
+        [MemberData("RuntimeComponents")]
```
BrennanConroy: Might as well do nameof here

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuSourceTests.cs
```diff
@@ -0,0 +1,263 @@
...
+            const string fakeRepoName = "https://example.com/example.git";
+
+            var runtimeHomeDir = TestUtils.GetRuntimeHomeDir(flavor, os, architecture);
+
+            using (var testEnv = new DnuTestEnvironment(runtimeHomeDir))
+            {
+                var testProjectFolder = ScafoldEmptyProject(
+                    testEnv.RootDir,
+                    "TestPackage",
+                    _ => @"{
```
BrennanConroy: $@

src/Microsoft.Framework.PackageManager/SourceControl/GitSourceControlProvider.cs
```diff
@@ -0,0 +1,177 @@
...
+                var repoRoot = GetRepoRoot(folderName);
+                var pathRelativeToRepoRoot = Path.GetDirectoryName(folderName)
+                    .Substring(repoRoot.Length)
+                    .Replace('\\', '/')
+                    .TrimStart('/');
+                snapshotInformation[ProjectPathKey] = pathRelativeToRepoRoot;
+            }
+        }
+
+        private string GetHEADCommitId(string folderName)
```
ChengTian: Based on our naming convention, should be `GetHeadCommitId`. (We have `HttpSource` instead of `HTTPSource`)

src/Microsoft.Framework.PackageManager/SourceControl/GitSourceControlProvider.cs
```diff
@@ -0,0 +1,177 @@
...
+            // First clone
+            if (!ProcessUtilities.RunApp(
+                RepoApp, 
+                $"clone {repoUrl} {destinationFolder}",
+                workingDirectory: null,
+                stdOut: out stdOut,
+                stdErr: out stdErr))
+            {
+                _buildReports.WriteError(stdErr);
+                return false;
```
ChengTian: So if I do `dnu source MyPackage` for one project and then do `dnu source MyPackage` for another project that uses exactly same version of `MyPackage`, it will fail with
```
fatal: destination path '%userprofile%/.dnx/source/MyPackage-{CommitHash}' already exists and is not an empty directory.
```
Seems this is inconvenient when you switch between multiple projects.

src/Microsoft.Framework.PackageManager/SourceControl/ProcessUtilities.cs
```diff
@@ -0,0 +1,98 @@
...
+                proc.EnableRaisingEvents = true;
+
+                proc.OutputDataReceived += (sender, e) =>
+                {
+                    output.Append(e.Data);
+                };
+                proc.ErrorDataReceived += (sender, e) =>
+                {
+                    errors.Append(e.Data);
+                };
```
ChengTian: What's the point of having so many lines of code here if you don't need to display the output of subprocess while it is running. Can we just use `proc.StandardError.ReadToEnd()`?

src/Microsoft.Framework.PackageManager/SourceControl/ProcessUtilities.cs
```diff
@@ -0,0 +1,98 @@
...
+        public static bool ExecutableExists(string executableName)
+        {
+            string whereApp;
+            if (PlatformHelper.IsWindows)
+            {
+                whereApp = "where";
+            }
+            else
+            {
+                whereApp = "whereis";
```
ChengTian: Is it guaranteed that `whereis` is always installed? Should we just start a subprocess and check whether it throws certain kind of exception?

src/Microsoft.Framework.PackageManager/SourceControl/SourceBuilder.cs
```diff
@@ -0,0 +1,113 @@
...
+                return true;
+            }
+
+            _reports.Information.WriteLine("Embedding repository information...");
+
+            string repositoryType;
+            if (!_project.Repository.TryGetValue(Constants.RepoTypeKey, out repositoryType) ||
+                string.IsNullOrEmpty(repositoryType))
+            {
+                WriteError("The project file contains a repository section but the repository type is missing.");
```
ChengTian: `repository section` -> `"repository" property`? I guess section is not a common term we use when talking about JSON

src/Microsoft.Framework.PackageManager/SourceControl/SourceCommand.cs
```diff
@@ -0,0 +1,244 @@
...
+using Newtonsoft.Json.Linq;
+using NuGet;
+
+namespace Microsoft.Framework.PackageManager.SourceControl
+{
+    internal class SourceCommand
+    {
+        private const string SourcesFolderName = "sources";
+
+        private const string SourcesEnvironmentVariable = "DNX_SOURCES_DIR";
```
ChengTian: We have `DNX_PACKAGES` instead of `DNX_PACKAGES_DIR`. Let's follow the same naming convention.

src/Microsoft.Framework.PackageManager/SourceControl/GitSourceControlProvider.cs
```diff
@@ -0,0 +1,177 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Microsoft.Framework.PackageManager.SourceControl
+{
+    internal class GitSourceControlProvider: SourceControlProvider
+    {
+        private const string RepoApp = "git";
```
davidfowl: Too many short names here. Repository is ok to type :smile:

src/Microsoft.Framework.PackageManager/SourceControl/GitSourceControlProvider.cs
```diff
@@ -0,0 +1,177 @@
...
+            }
+
+            if (!snapshotInformation.ContainsKey(CommitHashKey))
+            {
+                snapshotInformation[CommitHashKey] = GetHEADCommitId(folderName);
+            }
+
+            if (!snapshotInformation.ContainsKey(ProjectPathKey))
+            {
+                var repoRoot = GetRepoRoot(folderName);
```
davidfowl: Some comments here about why this is trying to manipulate the path would be good.

src/Microsoft.Framework.PackageManager/SourceControl/GitSourceControlProvider.cs
victorhurdugaci: Whoops. Removed

src/Microsoft.Framework.PackageManager/SourceControl/SourceCommand.cs
victorhurdugaci: Fixed

src/Microsoft.Framework.PackageManager/SourceControl/SourceCommand.cs
victorhurdugaci: Nope. I just want "global.json"

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuSourceTests.cs
victorhurdugaci: Nope. It gets too complicated to escape each `{`

src/Microsoft.Framework.PackageManager/SourceControl/GitSourceControlProvider.cs
victorhurdugaci: Sure

src/Microsoft.Framework.PackageManager/SourceControl/ProcessUtilities.cs
victorhurdugaci: Yes. AFAIK `whereis` is on all distributions

victorhurdugaci:
Updated

src/Microsoft.Framework.PackageManager/ConsoleCommands/SourceConsoleCommand.cs
```diff
@@ -0,0 +1,52 @@
...
+                    "[package]",
+                    "The name of the package for which to retrieve the sources. Can only specify packages used by the project.",
+                    multipleValues: true);
+
+                var projectFileOption = c.Option(
+                    "-p|--project",
+                    "Optional. The path to a project.json file. If not specified, the project in the current folder is used.",
+                    CommandOptionType.SingleValue);
+                var packagesFolderOption = c.Option(
+                    "-pa|--packages",
```
troydai: Even if we support `-pa`, it is not a good practice. Choose another single-character name.

src/Microsoft.Framework.PackageManager/Reports.cs
```diff
@@ -10,6 +10,26 @@ public class Reports
...
+        public void WriteError(string message)
+        {
+            Error.WriteLine(message.Red());
+        }
+
+        public void WriteWarning(string message)
+        {
+            Information.WriteLine(message.Yellow());
+        }
+
```
troydai: Why do you add these shortcut? You can call the on the `IReport` directly.

src/Microsoft.Framework.PackageManager/SourceControl/Constants.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Framework.PackageManager.SourceControl
+{
+    public static class Constants
```
troydai: Has to be public?

src/Microsoft.Framework.PackageManager/SourceControl/GitSourceControlProvider.cs
```diff
@@ -0,0 +1,178 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Microsoft.Framework.PackageManager.SourceControl
+{
+    internal class GitSourceControlProvider: SourceControlProvider
+    {
+        private const string RepositoryApp = "git";
+
```
troydai: Why the additional white space?

src/Microsoft.Framework.PackageManager/ConsoleCommands/SourceConsoleCommand.cs
ChengTian: :+1: Or simply do not have short option for this.

src/Microsoft.Framework.PackageManager/ConsoleCommands/SourceConsoleCommand.cs
```diff
@@ -0,0 +1,52 @@
...
+                    "Optional. The local packages folder",
+                    CommandOptionType.SingleValue);
+                var sourceFolderOption = c.Option(
+                    "-s|--sources",
```
victorhurdugaci: Maybe outputtarget?

ChengTian: Or follow the convention used by `dnu publish`, use `-o | --out`

src/Microsoft.Framework.PackageManager/ConsoleCommands/SourceConsoleCommand.cs
troydai: Any action?

src/Microsoft.Framework.PackageManager/SourceControl/GitSourceControlProvider.cs
```diff
@@ -0,0 +1,178 @@
...
+            if (!snapshotInformation.ContainsKey(ProjectPathKey))
+            {
+                var repoRoot = GetRepoRoot(folderName);
+
+                // Get the path relative to the repo root
+                var pathRelativeToRepoRoot = Path.GetDirectoryName(folderName)
+                    .Substring(repoRoot.Length)
+                    .Replace('\\', '/')
+                    .TrimStart('/');
```
troydai: There are similar code in _ScriptExecutor_ for unifying the directory separator. Consolidate the code?

src/Microsoft.Framework.PackageManager/SourceControl/GitSourceControlProvider.cs
```diff
@@ -0,0 +1,178 @@
...
+
+        private string GetHeadCommitId(string folderName)
+        {
+            string stdOut;
+            string stdErr;
+
+            if (ProcessUtilities.RunApp(RepositoryApp, "rev-parse HEAD", folderName, out stdOut, out stdErr) &&
+                string.IsNullOrEmpty(stdErr))
+            {
+                return stdOut;
```
troydai: It worth consider what if __git__ changes their output format how do we patch our tools? And can customer customize the logic to get around it.

src/Microsoft.Framework.PackageManager/SourceControl/GitSourceControlProvider.cs
```diff
@@ -0,0 +1,178 @@
...
+            }
+
+            throw new InvalidOperationException(stdErr);
+        }
+
+        public override string CreateShortFolderName(IDictionary<string, string> snapshotInfo)
+        {
+            var repoUrl = snapshotInfo[RepositoryUrlKey];
+            var commitHash = snapshotInfo[CommitHashKey];
```
troydai: Could these keys be missing?

src/Microsoft.Framework.PackageManager/SourceControl/ProcessUtilities.cs
```diff
@@ -0,0 +1,84 @@
...
+                {
+                    startInfo.WorkingDirectory = workingDirectory;
+                }
+
+                proc = new Process();
+                proc.StartInfo = startInfo;
+                proc.Start();
+
+
+                proc.WaitForExit();
```
troydai: Prepare for timeout

src/Microsoft.Framework.PackageManager/SourceControl/SourceCommand.cs
```diff
@@ -0,0 +1,244 @@
...
+
+            var globalFile = GlobalSettings.GetGlobalFilePath(_solutionRoot);
+            if (!File.Exists(globalFile))
+            {
+                _reports.WriteError($"The '{GlobalSettings.GlobalFileName}' is missing from '{_solutionRoot}'.");
+                return false;
+            }
+
+            _reports.Verbose.WriteLine($"Updating {GlobalSettings.GlobalFileName}...");
+            ModifyJson(globalFile, jObj =>
```
troydai: Why does it need to be in this form? Can the logic be completed in a simple function?

troydai:
I reset my case. :watch:

src/Microsoft.Framework.PackageManager/SourceControl/GitSourceControlProvider.cs
victorhurdugaci: They will not change the format because they would break everyone, not just us.

src/Microsoft.Framework.PackageManager/SourceControl/ProcessUtilities.cs
victorhurdugaci: Good point

src/Microsoft.Framework.PackageManager/SourceControl/GitSourceControlProvider.cs
troydai: Alright, we don't prepare for the doomsday scenario then.

davidfowl:
I think we need to rename this command

victorhurdugaci:
We've decided to keep the name as "source". Unless there are any more comments, I will merge this soon

victorhurdugaci:
Renamed to `sources`. And now the usage is `dnu sources get <package>`. @anurse @davidfowl

src/Microsoft.Framework.PackageManager/Reports.cs
troydai: Ping?

src/Microsoft.Framework.PackageManager/SourceControl/SourceCommand.cs
troydai: Ping?

src/Microsoft.Framework.PackageManager/Reports.cs
victorhurdugaci: We want to use these everywhere, not just here. Today everyone does it in their own way

src/Microsoft.Framework.PackageManager/SourceControl/SourceCommand.cs
victorhurdugaci: Too much code in one single method

src/Microsoft.Framework.PackageManager/SourceControl/GitSourceControlProvider.cs
```diff
@@ -0,0 +1,215 @@
...
+
+            var repoUrl = snapshotInfo[RepositoryUrlKey];
+            var commitHash = snapshotInfo[CommitHashKey];
+
+            _buildReports.WriteInformation($"Cloning from: {repoUrl}");
+
+            // First clone
+            if (!ProcessUtilities.Execute(
+                Git,
+                $"clone {repoUrl} {destinationFolder}",
```
troydai: add `--recursive` for submodules.

troydai:
:watch:

troydai:
:shipit:

src/Microsoft.Framework.PackageManager/ConsoleCommands/SourceConsoleCommand.cs
```diff
@@ -0,0 +1,69 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Framework.PackageManager.SourceControl;
+using Microsoft.Framework.Runtime;
+using Microsoft.Framework.Runtime.Common.CommandLine;
+
+namespace Microsoft.Framework.PackageManager
+{
+    internal class SourceConsoleCommand
```
anurse: nit: Source**s**ConsoleCommand

victorhurdugaci: :+1:

src/Microsoft.Framework.PackageManager/SourceControl/ProcessUtilities.cs
anurse: `whereis` is present on most distributions but is not part of the POSIX spec. The recommended approach is usually [`type -p`](https://en.wikipedia.org/wiki/Type_(Unix)) since that is part of the POSIX spec: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/type.html

I'm mostly OK either way for now, but it might be good to see if you can switch to that since it should be most compatible on POSIX systems.

src/Microsoft.Framework.PackageManager/SourceControl/SourceCommand.cs
```diff
@@ -0,0 +1,253 @@
...
+using Newtonsoft.Json.Linq;
+using NuGet;
+
+namespace Microsoft.Framework.PackageManager.SourceControl
+{
+    internal class SourceCommand
+    {
+        private const string SourcesFolderName = "sources";
+
+        private const string SourcesEnvironmentVariable = "DNX_SOURCES";
```
anurse: There's already an `EnvironmentNames` static class. This should probably go there.

src/Microsoft.Framework.PackageManager/SourceControl/SourceCommand.cs
anurse: This seems like an odd pattern to me. Why not put the lambda body in the `ModifyJson` method (and give it a more appropriate name, like `AddSourcesDirectoryToProjects` or something...)? The `ModifyJson` method is only ever used here and given this lambda body.

src/Microsoft.Framework.PackageManager/SourceControl/SourceControlProviderFactory.cs
```diff
@@ -0,0 +1,28 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Framework.PackageManager.SourceControl
+{
+    internal static class SourceControlProviderFactory
+    {
+        // TODO: When we have a second source control provider extract the common API in a base class/interface
```
anurse: Are there bugs filed for specific providers we're interested in? `hg` ? `tfvc` ?

Also, should we consider allowing users to install custom providers?

[All these are just things to consider for the future, but we should capture them in bugs]

src/Microsoft.Framework.Runtime/GlobalSettings.cs
```diff
@@ -53,7 +53,7 @@ public static bool TryGetGlobalSettings(string path, out GlobalSettings globalSe
                                              jobject.ValueAsStringArray("sources") ??
                                              new string[] { };
 
-                    globalSettings.ProjectSearchPaths = projectSearchPaths;
+                    globalSettings.ProjectSearchPaths = new List<string>(projectSearchPaths);
```
anurse: nit: I've seen `ToList` used in most places to do this. But I don't really care much either way :)

anurse:
Looks good! Small nits and one minor suggestion to use `type` instead of `whereis`. Other than that, :shipit:

src/Microsoft.Framework.PackageManager/SourceControl/SourceCommand.cs
moozzyk: :rant: I only wish we did not use the `CommonPrefix + ` pattern. It is impossible to find all occurrences of an environment variable across all the projects (managed and native) in the dnx if you are not aware of this class. I have been on my quest to understanding what is the benefit of the `CommonPrefix + ` pattern for some time now...

victorhurdugaci: If I remember well, the prefix came when we did the 3rd (?) rename because it was really hard to rename everywhere

moozzyk: I would think 3 renames are not hard you just replace all the `"DNX_` occurrences in the `EnvironmentNames.cs` file with Ctrl+Shift+H. The first time I almost missed that we use DNX_DEFAULTLIB in the managed code at all because I could not find a single occurrence of this string...

anurse: I could agree with removing `CommonPrefix`, **especially** if all the environment variable names are in one place, because then it's just a matter of updating the prefix in one place.

src/Microsoft.Framework.PackageManager/SourceControl/SourceControlProviderFactory.cs
victorhurdugaci: https://github.com/aspnet/dnx/issues/2154

Issue #1956, opened 2015-05-26
Using XDocument instead of XmlSerializer to serialize NuGet manifest


troydai:

muratg:
@troydai is this finished?

troydai:
This is being reviewed in PR #2012

muratg:
:+1:

troydai:
Fixed. https://github.com/aspnet/dnx/commit/103b3a59669fa6f5932b600ed30e9ae64b5e3c00

PR #2012, opened 2015-06-03
[Prototype] Replace XmlSerializer with XDocument in NuGet manifest

troydai:
Replacing XmlSerializer with XDocument in runtime prevent us from being frequently broken by XmlSerializer change. It also benefits us by reducing additional data model for serialization purpose only.

/review @davidfowl @yishaigalatzer @muratg 
/cc @Eilon @BrennanConroy @victorhurdugaci @moozzyk 

Address: #1956


src/Microsoft.Framework.Runtime/NuGet/Authoring/ManifestMetadata.cs
```diff
@@ -27,7 +52,7 @@ public string MinClientVersionString
             set
             {
                 Version version = null;
-                if (!String.IsNullOrEmpty(value) && !System.Version.TryParse(value, out version))
+                if (!string.IsNullOrEmpty(value) && !System.Version.TryParse(value, out version))
```
moozzyk: nit: the first is check seems to be not needed.

src/Microsoft.Framework.Runtime/NuGet/Authoring/ManifestReader.cs
```diff
@@ -50,6 +48,7 @@ private static ManifestMetadata ReadMetadata(XElement xElement)
                 {
                     ReadMetadataValue(manifestMetadata, element, allElements);
                 }
+
```
moozzyk: Why can't this loop just be `foreach(var element in xElement.Elements())` ? Or can it?

src/Microsoft.Framework.Runtime/NuGet/Authoring/ManifestReader.cs
```diff
@@ -25,18 +26,15 @@ public static Manifest ReadManifest(XDocument document)
...
+                ReadMetadata(metadataElement),
+                ReadFilesList(document.Root.ElementsNoNamespace("files").FirstOrDefault()));
         }
 
         private static ManifestMetadata ReadMetadata(XElement xElement)
         {
             var manifestMetadata = new ManifestMetadata();
-            manifestMetadata.DependencySets = new List<ManifestDependencySet>();
-            manifestMetadata.ReferenceSets = new List<ManifestReferenceSet>();
+
```
moozzyk: Ln. 38 should just be:
`manifestMetadata.MinClientVersionString = (string)xElement.Attribute("minClientVersion");

The `GetOptionalAttributeValue` does not add any value - attributes don't inherit namespaces and the `(string)` explicit conversion operator already handle the case if the attribute is not there (i.e. returns null). 

I suggest that the `GetOptionalAttributeValue()` method should be removed.

src/Microsoft.Framework.Runtime/NuGet/Authoring/ManifestReader.cs
```diff
@@ -152,33 +148,36 @@ private static List<ManifestReferenceSet> ReadReferenceSets(XElement referencesE
...
+                    else
+                    {
+                        return new PackageReferenceSet(ReadReference(element, throwIfEmpty: true));
+                    }
+                }).ToList();
             }
         }
 
-        public static List<ManifestReference> ReadReference(XElement referenceElement, bool throwIfEmpty)
+        public static IEnumerable<string> ReadReference(XElement referenceElement, bool throwIfEmpty)
```
moozzyk: So why are we using `ElementsNoNamespace` almost everywhere? It's semantically incorrect because it strips the namespace and it may capture elements that should not be captured (they were put to different namespace to not be captured). It also does not save you any typing because if you don't expect your elements to belong to a namespace `referenceElement.Element("reference")` is shorter than `referenceElement.ElementsNoNamespace("reference")` (if you do `referenceElement.Element(ns + "reference")` is still shorter). On top of that if you use just the `Element()` method you can use explicit conversion operators.

src/Microsoft.Framework.Runtime/NuGet/Authoring/ManifestMetadata.cs
troydai: Yup. You're write.

src/Microsoft.Framework.Runtime/NuGet/Authoring/ManifestReader.cs
troydai: Good suggestion. Let me look into it.

src/Microsoft.Framework.Runtime/NuGet/Authoring/ManifestReader.cs
troydai: I'm sure we can. But I keep the change to barely minimal. This is part of the original code of reading manifest. I lean to not to update it since we're going to refactory NuGet anyway.

troydai: But yeah valid suggestion.

src/Microsoft.Framework.Runtime/NuGet/Authoring/ManifestReader.cs
troydai: Great suggestion. This is the legacy code. I will look into this.

src/Microsoft.Framework.Runtime/NuGet/Authoring/ManifestReader.cs
```diff
@@ -189,28 +188,30 @@ public static List<ManifestReference> ReadReference(XElement referenceElement, b
...
-                        TargetFramework = element.GetOptionalAttributeValue("targetFramework").SafeTrim()
-                    }).ToList();
+                    where assemblyNameAttribute != null && !string.IsNullOrEmpty(assemblyNameAttribute.Value)
+                    select new FrameworkAssemblyReference(
+                        assemblyNameAttribute.Value?.Trim(),
+                        element.GetOptionalAttributeValue("targetFramework")?.Trim()
+                                                                            ?.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
+                                                                            ?.Select(VersionUtility.ParseFrameworkName)
+                               ?? Enumerable.Empty<FrameworkName>()))
+                   .ToList();
```
moozzyk: It does not seem that the `targetFramework` attribute is a coma separated list - at least it was not before. Also if you look at Microsoft.Net.Http.Server.nuspec the frameworkAssembly are repeated for different targetFrameworks.

src/Microsoft.Framework.Runtime/NuGet/Authoring/ManifestReader.cs
```diff
@@ -189,28 +188,30 @@ public static List<ManifestReference> ReadReference(XElement referenceElement, b
...
                     let assemblyNameAttribute = element.Attribute("assemblyName")
-                    where assemblyNameAttribute != null && !String.IsNullOrEmpty(assemblyNameAttribute.Value)
-                    select new ManifestFrameworkAssembly
-                    {
-                        AssemblyName = assemblyNameAttribute.Value.SafeTrim(),
-                        TargetFramework = element.GetOptionalAttributeValue("targetFramework").SafeTrim()
-                    }).ToList();
+                    where assemblyNameAttribute != null && !string.IsNullOrEmpty(assemblyNameAttribute.Value)
+                    select new FrameworkAssemblyReference(
+                        assemblyNameAttribute.Value?.Trim(),
```
moozzyk: This is never null so `?` is not needed.

src/Microsoft.Framework.Runtime/NuGet/Authoring/ManifestReader.cs
```diff
@@ -189,28 +188,30 @@ public static List<ManifestReference> ReadReference(XElement referenceElement, b
...
 
             return (from element in frameworkElement.ElementsNoNamespace("frameworkAssembly")
                     let assemblyNameAttribute = element.Attribute("assemblyName")
-                    where assemblyNameAttribute != null && !String.IsNullOrEmpty(assemblyNameAttribute.Value)
-                    select new ManifestFrameworkAssembly
-                    {
-                        AssemblyName = assemblyNameAttribute.Value.SafeTrim(),
-                        TargetFramework = element.GetOptionalAttributeValue("targetFramework").SafeTrim()
-                    }).ToList();
+                    where assemblyNameAttribute != null && !string.IsNullOrEmpty(assemblyNameAttribute.Value)
```
moozzyk: nit:
`let assemblyNameAttribute = element.Attribute("assemblyName")
where assemblyNameAttribute != null && !string.IsNullOrEmpty(assemblyNameAttribute.Value)`

could be just:

`let assemblyName = (string)element.Attribute("assemblyName")
where !string.IsNullOrEmpty(assemblyName)`

src/Microsoft.Framework.Runtime/NuGet/Authoring/ManifestVersionUtility.cs
```diff
@@ -55,21 +76,20 @@ private static int GetMaxVersionFromMetadata(ManifestMetadata metadata)
...
+            return obj?.GetType()
+                       .GetRuntimeProperties()
+                       .Where(prop => prop.GetMethod != null && prop.GetMethod.IsPublic && !prop.GetMethod.IsStatic)
+                       .Select(prop => GetVersionFromPropertyInfo(obj, prop))
+                       .Max()
+                      ?? DefaultVersion;
         }
-
-        private static int VisitProperty(object obj, PropertyInfo property)
+
```
moozzyk: nit: spaces

src/Microsoft.Framework.Runtime/NuGet/Authoring/ManifestReader.cs
troydai: Yup.

src/Microsoft.Framework.Runtime/NuGet/Authoring/ManifestReader.cs
troydai: Will update.

src/Microsoft.Framework.Runtime/NuGet/Authoring/Manifest.cs
```diff
@@ -57,15 +60,30 @@ public void Save(Stream stream, bool validate)
...
+            var root = new XElement(schemaNamespace + "package");
+
+            var metadataElement = Metadata.ToXElement(schemaNamespace);
+            if (metadataElement == null)
+            {
+                throw new InvalidOperationException("Failed to serialize manifest metadata to xml.");
+            }
+            root.Add(metadataElement);
+
+            if (Files != null && Files.Count != 0)
```
moozzyk: Does not seem `Files` can be null.

src/Microsoft.Framework.Runtime/NuGet/Authoring/Manifest.cs
```diff
@@ -57,15 +60,30 @@ public void Save(Stream stream, bool validate)
...
+                foreach (var file in Files)
+                {
+                    filesElement.Add(file.ToXElement(schemaNamespace));
+                }
 
-            // Define the namespaces to use when serializing
-            var ns = new XmlSerializerNamespaces();
-            ns.Add("", schemaNamespace);
+                root.Add(filesElement);
+            }
```
moozzyk: Consider being more fluent - you should be able to create an Xml document in one go preserving visually the document structure. Something like this:

```
var metadataElement = Metadata.ToXElement(schemaNamespace);
if (metadataElement == null)
{
    throw new InvalidOperationException("Failed to serialize manifest metadata to xml.");
}

new XDocument(
    new XElement(
        schemaNamespace + "package", 
            metadataElement,
            Files.Any() ? 
            new XElement(
                schemaNamespace + files, 
                    Files.Select(f => f.ToXElement)))).Save(stream);
```

(Yeah, not wanting empty "files" element breaks the pattern a little bit - note you can pass null so it still can be inline)

src/Microsoft.Framework.Runtime/NuGet/Authoring/ManifestMetadata.cs
```diff
@@ -37,322 +62,82 @@ public string MinClientVersionString
...
-                if (DependencySets == null || DependencySets.Count == 0)
+                if (value == null)
                 {
-                    return null;
-                }
-
-                if (DependencySets.Any(set => set.TargetFramework != null))
-                {
-                    return DependencySets.Cast<object>().ToList();
+                    _authors = Enumerable.Empty<string>();
```
moozzyk: Not sure why _authors cannot be `null`... Anyways - use the ternary operator.

src/Microsoft.Framework.Runtime/NuGet/Authoring/ManifestMetadata.cs
```diff
@@ -37,322 +62,82 @@ public string MinClientVersionString
...
         {
             get
             {
-                if (ReferenceSets == null || ReferenceSets.Count == 0)
-                {
-                    return null;
-                }
-
-                if (ReferenceSets.Any(set => set.TargetFramework != null))
+                if (_owners == null || _owners.IsEmpty())
```
moozzyk: ternary

src/Microsoft.Framework.Runtime/NuGet/Authoring/ManifestMetadata.cs
```diff
@@ -37,322 +62,82 @@ public string MinClientVersionString
...
-                return new SemanticVersion(Version);
-            }
-        }
-
-        Uri IPackageMetadata.IconUrl
-        {
-            get
-            {
-                if (IconUrl == null)
+                if (value == null)
```
moozzyk: same as above

src/Microsoft.Framework.Runtime/NuGet/Authoring/ManifestVersionUtility.cs
```diff
@@ -41,13 +62,13 @@ private static int GetMaxVersionFromMetadata(ManifestMetadata metadata)
...
                 metadata.DependencySets.Any(d => d.TargetFramework != null);
+
             if (dependencyHasTargetFramework)
             {
                 return TargetFrameworkSupportForDependencyContentsAndToolsVersion;
             }
 
-            SemanticVersion semanticVersion;
-            if (SemanticVersion.TryParse(metadata.Version, out semanticVersion) && !String.IsNullOrEmpty(semanticVersion.SpecialVersion))
+            if (metadata.Version != null && !string.IsNullOrEmpty(metadata.Version.SpecialVersion))
```
moozzyk: `if (!string.IsNullOrEmpty(metadata.Version?.SpecialVersion))` ?

src/Microsoft.Framework.Runtime/NuGet/Authoring/PackageDependencySet.cs
```diff
@@ -11,36 +11,46 @@ namespace NuGet
...
-                return _targetFramework;
+                throw new ArgumentNullException(nameof(dependencies));
             }
+
+            TargetFramework = targetFramework;
+            Dependencies = new ReadOnlyCollection<PackageDependency>(dependencies.ToList());
         }
 
-        public ICollection<PackageDependency> Dependencies
+        public PackageDependencySet(string targetFramework, IEnumerable<PackageDependency> dependencies)
```
moozzyk: Call the other ctor. 
```
public PackageDependencySet(string targetFramework, IEnumerable<PackageDependency> dependencies)
   this(targetFramework != null ? VersionUtility.ParseFrameworkName(targetFramework) : null, dependencies) 
{}
```

src/Microsoft.Framework.Runtime/NuGet/Authoring/PackageDependencySet.cs
```diff
@@ -11,36 +11,46 @@ namespace NuGet
 {
     public class PackageDependencySet : IFrameworkTargetable
     {
-        private readonly FrameworkName _targetFramework;
-        private readonly ReadOnlyCollection<PackageDependency> _dependencies;
-
-        public PackageDependencySet(FrameworkName targetFramework, IEnumerable<PackageDependency> dependencies)
+        public PackageDependencySet(IEnumerable<PackageDependency> dependencies)
```
moozzyk: Call the other ctor:

```
public PackageDependencySet(IEnumerable<PackageDependency> dependencies)
 : this(null, dependencies) 
{}
```
(You may need to cast null to select the right overload)

src/Microsoft.Framework.Runtime/NuGet/Authoring/PackageReferenceSet.cs
```diff
@@ -14,38 +14,30 @@ public class PackageReferenceSet : IFrameworkTargetable
         private readonly FrameworkName _targetFramework;
         private readonly ICollection<string> _references;
 
-        public PackageReferenceSet(FrameworkName targetFramework, IEnumerable<string> references)
+        public PackageReferenceSet(IEnumerable<string> references)
```
moozzyk: Call the other ctor.

src/Microsoft.Framework.Runtime/NuGet/Packages/FrameworkAssemblyReference.cs
```diff
@@ -3,21 +3,15 @@
...
-            : this(assemblyName, Enumerable.Empty<FrameworkName>())
-        {
-        }
-
         public FrameworkAssemblyReference(string assemblyName, IEnumerable<FrameworkName> supportedFrameworks)
         {
-            if (String.IsNullOrEmpty(assemblyName))
+            if (string.IsNullOrEmpty(assemblyName))
             {
                 throw new ArgumentNullException("assemblyName");
```
moozzyk: Should be `ArgumentException` because `string.Empty` is not null

src/Microsoft.Framework.Runtime/NuGet/Packages/PackageDependency.cs
```diff
@@ -7,32 +7,31 @@ namespace NuGet
 {
     public class PackageDependency
     {
-        public PackageDependency(string id)
-            : this(id, versionSpec: null)
+        public PackageDependency(string id, string version)
         {
+            if (string.IsNullOrEmpty(id))
+            {
+                throw new ArgumentNullException(nameof(id));
```
moozzyk: Should be `ArgumentException` because `string.Empty` is not null

src/Microsoft.Framework.Runtime/NuGet/Packages/PackageDependency.cs
```diff
@@ -7,32 +7,31 @@ namespace NuGet
...
+            VersionSpec = string.IsNullOrEmpty(version) ? null : VersionUtility.ParseVersionSpec(version);
         }
 
         public PackageDependency(string id, IVersionSpec versionSpec)
         {
-            if (String.IsNullOrEmpty(id))
+            if (string.IsNullOrEmpty(id))
             {
-                throw new ArgumentNullException("id");
+                throw new ArgumentNullException(nameof(id));
```
moozzyk: Should be `ArgumentException` because `string.Empty` is not null (everywhere)

src/Microsoft.Framework.Runtime/NuGet/SemanticVersion.cs
```diff
@@ -51,12 +49,12 @@ private SemanticVersion(Version version, string specialVersion, string originalS
             }
             Version = NormalizeVersionValue(version);
             SpecialVersion = specialVersion ?? String.Empty;
-            _originalString = String.IsNullOrEmpty(originalString) ? version.ToString() + (!String.IsNullOrEmpty(specialVersion) ? '-' + specialVersion : null) : originalString;
+            OriginalString = String.IsNullOrEmpty(originalString) ? version.ToString() + (!String.IsNullOrEmpty(specialVersion) ? '-' + specialVersion : null) : originalString;
```
moozzyk: nit: lowercase string in `string.IsNullOrEmpty()`

src/Microsoft.Framework.Runtime/NuGet/SemanticVersion.cs
```diff
@@ -78,23 +76,25 @@ public string SpecialVersion
             get;
             private set;
         }
+        
+        public string OriginalString { get; }
 
         public string[] GetOriginalVersionComponents()
         {
-            if (!String.IsNullOrEmpty(_originalString))
+            if (!String.IsNullOrEmpty(OriginalString))
```
moozzyk: nit: casing (String)

src/Microsoft.Framework.Runtime/NuGet/Xml/ManifestFileXmlExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace NuGet
+{
+    public static class ManifestFileXmlExtensions
+    {
+        public static XElement ToXElement(this ManifestFile manifestFile, XNamespace ns)
+        {
+            var elem = new XElement(ns + "file");
+            elem.SetAttributeValue(ns + "src", manifestFile.Source);
+            elem.SetAttributeValue(ns + "target", manifestFile.Target);
+            elem.SetAttributeValue(ns + "exclude", manifestFile.Exclude);
```
moozzyk: Are you sure the attributes should be in the namespace?  You realize that if the namespace is not empty the attribute will have to be prefixed. If the namespace is not bound to a prefix the prefix will be auto generated and you will get this: 
`<file p1:src="file" xmlns:p1="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd" xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd" />`
I think you actually don't want attributes belong to any namespace here - it is quite uncommon in XML to put attributes to a namespace. Typically you would do this only if you want to extend an element that belongs to a non empty namespace and the schema does not allow putting attributes in the empty namespace on this element.

src/Microsoft.Framework.Runtime/NuGet/Xml/ManifestFileXmlExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+
+            return elem;
```
moozzyk: Consider (given that I think the attributes should be in the empty namespace):

```
return `new XElement(ns + "file", new XAttribute("src", manifestFile.Source), new XAttribute("target", manifestFile.Target), new XAttribute("exclude", manifestFile.Exclude");
```

src/Microsoft.Framework.Runtime/NuGet/Xml/PackageMetadataXmlExtensions.cs
```diff
@@ -0,0 +1,197 @@
...
+        {
+            var elem = new XElement(ns + "metadata");
+            elem.SetAttributeValue(ns + "minClientVersion", metadata.MinClientVersionString);
+
+            elem.Add(new XElement(ns + "id", metadata.Id));
+            elem.Add(new XElement(ns + "version", metadata.Version.OriginalString));
+            AddElementIfNotNull(elem, ns, "title", metadata.Title);
+            elem.Add(new XElement(ns + "requireLicenseAcceptance", metadata.RequireLicenseAcceptance));
+            AddElementIfNotNull(elem, ns, "authors", metadata.Authors, authors => string.Join(",", authors));
+            AddElementIfNotNull(elem, ns, "owners", (metadata.Owners != null && metadata.Owners.Any()) ? metadata.Owners : metadata.Authors,
```
moozzyk: Does not `.Owners` already return `Authors` if there are no Owners?

src/Microsoft.Framework.Runtime/NuGet/Xml/PackageMetadataXmlExtensions.cs
```diff
@@ -0,0 +1,197 @@
...
+        private const string TargetFramework = "targetFramework";
+        private const string FrameworkAssemblies = "frameworkAssemblies";
+        private const string FrameworkAssembly = "frameworkAssembly";
+        private const string AssemblyName = "assemblyName";
+        private const string Dependencies = "dependencies";
+
+        public static XElement ToXElement(this ManifestMetadata metadata, XNamespace ns)
+        {
+            var elem = new XElement(ns + "metadata");
+            elem.SetAttributeValue(ns + "minClientVersion", metadata.MinClientVersionString);
```
moozzyk: again - I don't think you want the attribute in the namespace (review everywhere). You could also move attribute creation to the ctor of the owning element (as well as children element).

troydai: Will gix.

src/Microsoft.Framework.Runtime/NuGet/Xml/PackageMetadataXmlExtensions.cs
troydai: Yup, will update.

src/Microsoft.Framework.Runtime/NuGet/Xml/PackageMetadataXmlExtensions.cs
```diff
@@ -0,0 +1,197 @@
...
+        }
+
+        private static XElement GetXElementFromGroupableItemSets<TSet, TItem>(
+            XNamespace ns,
+            IEnumerable<TSet> objectSets,
+            Func<TSet, bool> isGroupable,
+            Func<TSet, string> getGroupIdentifer,
+            Func<TSet, IEnumerable<TItem>> getItems,
+            Func<XNamespace, TItem, XElement> getXElementFromItem,
+            string elementRootName,
```
moozzyk: it's not root. It's parent.

src/Microsoft.Framework.Runtime/NuGet/Xml/ManifestFileXmlExtensions.cs
troydai: This is wrong. Attribute don't need namespace here. Will fix.

src/Microsoft.Framework.Runtime/NuGet/Xml/PackageMetadataXmlExtensions.cs
```diff
@@ -0,0 +1,197 @@
...
+            string identiferAttributeName)
+        {
+            if (objectSets == null || objectSets.IsEmpty())
+            {
+                return null;
+            }
+
+            var result = new XElement(ns + elementRootName);
```
moozzyk: This is not used.

src/Microsoft.Framework.Runtime/NuGet/Xml/PackageMetadataXmlExtensions.cs
```diff
@@ -0,0 +1,197 @@
...
+
+                    childElements.Add(groupElem);
+                }
+            }
+
+            return new XElement(ns + elementRootName, childElements.ToArray());
+        }
+
+        private static XElement GetXElementFromPackageReference(XNamespace ns, string reference)
+        {
```
moozzyk: collapse to single line.

src/Microsoft.Framework.Runtime/NuGet/Xml/PackageMetadataXmlExtensions.cs
```diff
@@ -0,0 +1,197 @@
...
+            var element = new XElement(ns + Reference);
+            element.SetAttributeValue(File, reference);
+
+            return element;
+        }
+
+        private static XElement GetXElementFromPackageDependency(XNamespace ns, PackageDependency dependency)
+        {
```
moozzyk: `return new XElement(ns + "dependency", new XAttribute("id", dpendency.id), dependency.VersionSpec != null ? new XAttribute("version", dependency.VersionSpec.ToString()) : null);`

src/Microsoft.Framework.Runtime/NuGet/Xml/PackageMetadataXmlExtensions.cs
```diff
@@ -0,0 +1,197 @@
...
+
+                    if (reference.SupportedFrameworks != null && reference.SupportedFrameworks.Any())
+                    {
+                        element.SetAttributeValue(
+                            "targetFramework",
+                            string.Join(", ", reference.SupportedFrameworks.Select(VersionUtility.GetFrameworkString)));
+                    }
+
+                    return element;
+                }).ToArray());
```
moozzyk: You should not need `.ToArray()` here.

src/Microsoft.Framework.Runtime/NuGet/Xml/PackageMetadataXmlExtensions.cs
```diff
@@ -0,0 +1,197 @@
...
+            if (references == null || references.IsEmpty())
+            {
+                return null;
+            }
+
+            return new XElement(
+                ns + FrameworkAssemblies,
+                references.Select(reference =>
+                {
+                    var element = new XElement(ns + FrameworkAssembly);
```
moozzyk: you can collapse attribute creation

moozzyk:
Add some tests for non-empty namespace - looks like attributes are not written correctly. 

You seem to be collapsing some elements by grouping attribute values to a comma separated list. Add tests for this - the code to do that is not trivial so it would be good to have some coverage.

troydai:
Ping! @davidfowl

test/Microsoft.Framework.Runtime.Tests/NuGet/Authoring/ManifestTests.cs
```diff
@@ -0,0 +1,86 @@
...
+            var manifest = new Manifest(new ManifestMetadata
+            {
+                Id = id,
+                Version = version,
+                Authors = authors
+            });
+
+            using (var mem = new MemoryStream())
+            {
+                //System.Diagnostics.Debugger.Launch();
```
davidfowl: Remove dead code

src/Microsoft.Framework.Runtime/NuGet/Authoring/Manifest.cs
```diff
@@ -6,34 +6,37 @@
...
 
 namespace NuGet
 {
-    [XmlType("package")]
     public class Manifest
     {
         private const string SchemaVersionAttributeName = "schemaVersion";
 
-        public Manifest()
+        public Manifest(ManifestMetadata metadata, IEnumerable<ManifestFile> files = null)
```
davidfowl: Don't use optional parameters.

src/Microsoft.Framework.Runtime/NuGet/SemanticVersion.cs
```diff
@@ -13,14 +13,12 @@ namespace NuGet
     /// </summary>
     public sealed class SemanticVersion : IComparable, IComparable<SemanticVersion>, IEquatable<SemanticVersion>
     {
-        private readonly string _originalString;
```
davidfowl: Avoid random changes like these, it adds noise to the review.

test/Microsoft.Framework.Runtime.Tests/NuGet/Authoring/ManifestMetadataTests.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using NuGet;
+using Xunit;
+
+namespace Microsoft.Framework.Runtime.Tests.NuGet.Authoring
+{
+    public class ManifestMetadataTests
```
davidfowl: Where are the Fact attributes?

src/Microsoft.Framework.Runtime/NuGet/Authoring/ManifestVersionUtility.cs
```diff
@@ -55,21 +76,20 @@ private static int GetMaxVersionFromMetadata(ManifestMetadata metadata)
             return DefaultVersion;
         }
 
-        private static int VisitObject(object obj)
+        private static int GetVersionFromObject(object obj)
```
davidfowl: Get rid of all of this code.

davidfowl:
Run some tests by building some repositories in universe and consuming them with the new dnx and these changes. We need to make sure we don't regress anywhere we're reading or writing nuspecs.

src/Microsoft.Framework.Runtime/NuGet/SemanticVersion.cs
troydai: No, it's not a random change. The original string is consumed by external code so I expose it as a property.

test/Microsoft.Framework.Runtime.Tests/NuGet/Authoring/ManifestMetadataTests.cs
troydai: On the methods?

troydai: Added.

src/Microsoft.Framework.Runtime/NuGet/Authoring/ManifestVersionUtility.cs
troydai: These codes are still used to decide the manifest schema. However it can be simplified. I filed following issue to track: https://github.com/aspnet/dnx/issues/2199

davidfowl:
:shipit:

troydai:
Running last round of rebase and build verify. I'll merge soon.

PR #2013, opened 2015-06-03
Enabling strong naming without private key

moozzyk:


src/Microsoft.Framework.Runtime/CompilerOptions.cs
```diff
@@ -76,6 +78,11 @@ public static CompilerOptions Combine(params CompilerOptions[] options)
                 {
                     result.DelaySign = option.DelaySign;
                 }
+
+                if(option.StrongName != null)
```
davidfowl: formatting

davidfowl: Why not just always set the strong name?

davidfowl:
:shipit:

src/Microsoft.Framework.Runtime/CompilerOptions.cs
moozzyk: I think it is a bigger discussion. We just talked with @muratg about this and it should be possible and would likely unblock some EF7 scenarios. On the other hand - shouldn't we then allow people to set their own public key? If we don't we would force people to use snk (real strong naming) if they want a different key/public key token. 
We could also fake strong name by default and allow providing own public key.

PR #2015, opened 2015-06-03
teach dnx about dotnet, and unteach it about netportable50

anurse:
`lib/dotnet/Foo.dll` ==> `.NETPlatform, Version=v5.0`

/cc @emgarten @yishaigalatzer @muratg @troydai @davidfowl @bricelam @ericstj

### DIFF
--- a/src/Microsoft.Framework.Runtime/NuGet/Utility/VersionUtility.cs
+++ b/src/Microsoft.Framework.Runtime/NuGet/Utility/VersionUtility.cs
@@ -23,6 +23,7 @@ namespace NuGet
         public static readonly string AspNetCoreFrameworkIdentifier = "Asp.NetCore";
         public static readonly string DnxCoreFrameworkIdentifier = "DNXCore";
         public static readonly string PortableFrameworkIdentifier = ".NETPortable";
+        public static readonly string NetPlatformFrameworkIdentifier = ".NETPlatform";
 
         internal const string NetFrameworkIdentifier = ".NETFramework";
         private const string NetCoreFrameworkIdentifier = ".NETCore";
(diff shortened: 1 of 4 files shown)


src/Microsoft.Framework.Runtime/NuGet/Utility/VersionUtility.cs
```diff
@@ -23,6 +23,7 @@ public static class VersionUtility
         public static readonly string AspNetCoreFrameworkIdentifier = "Asp.NetCore";
         public static readonly string DnxCoreFrameworkIdentifier = "DNXCore";
         public static readonly string PortableFrameworkIdentifier = ".NETPortable";
```
muratg: Should we remove this?

anurse: No, it is still used for profile-based portable (`portable-net45+win8`, etc.)

anurse:
Travis failure seems to be unrelated

src/Microsoft.Framework.Runtime/NuGet/Utility/VersionUtility.cs
```diff
@@ -251,7 +246,7 @@ public static FrameworkName ParseFrameworkName(string frameworkName)
                     return UnsupportedFrameworkName;
                 }
 
-                version = _emptyVersion;
+                version = identifierPart.Equals(NetPlatformFrameworkIdentifier) ? new Version(5, 0) : _emptyVersion;
```
anurse: This is turning out to be a bit weird. So I'm going to drop it. The Version Number for "dotnet" will be "0.0". We can display it as "5.0" in the places we want to, if we want to.

src/Microsoft.Framework.Runtime/NuGet/Utility/VersionUtility.cs
bricelam: I just sent out PRs to remove all these from our repos...

bricelam:
Will the defines flag be `DOTNET` or `DOTNET00`?

src/Microsoft.Framework.Runtime/NuGet/Utility/VersionUtility.cs
anurse: @emgarten came through with a clean way to do this in the new nuget code, so we can keep it! Ignore my previous comment :).

anurse:
Good question, I'll double check but it should be DOTNET since I think we take the short name nuget gives us.

anurse:
To answer your question @bricelam , it will be whatever you typed in "frameworks", but uppercase. So if you typed "dotnet00" it would be DOTNET00. This is true for any framework.

bricelam:
Cool.

PR #2020, opened 2015-06-04
Anurse/fix release

anurse:
/cc @troydai


troydai:
@anurse do we still need this PR?

anurse:
Guh, I don't even remember what it was for... @davidfowl ?

anurse:
I just remembered why we need it :). I'll rebase it and use it to implement the compatibility matrix stuff we need to do as well.

anurse:
Actually, we already have it in release. I'm just so turned around by all this :)

PR #2023, opened 2015-06-05
Making dnx.clr.managed strong named... again

moozzyk:
The appdomain manager has to have a real strong name (as opposed to fake one) otherwise bootstrapping CLR will fail due to strong name verification.


moozzyk:
:watch:

davidfowl:
@jkotas Is there a way around this?

jkotas:
I doubt that there is a way around it.

moozzyk:
@davidfowl meaning it is good to go right? (It's blocking #2024)

PR #2024, opened 2015-06-05
Using full Appdomain manager assembly name when loading.

### LINKED ISSUE #1958: Use `ICLRControl::SetAppDomainManagerType` instead of config file
Currently we specify AppDomainManagerType in the dnx.clr.config. We should use [`ICLRControl::SetAppDomainManagerType`](https://msdn.microsoft.com/en-us/library/ms164355(v=vs.110).aspx) method instead

moozzyk:
Removing dnx.clr.config (fixes #1958 and a potential buffer overrun when creating the config file path)
Removing unused code in CLR bootstrapping


moozzyk:
:watch:

src/dnx.clr/KatanaManager.h
```diff
@@ -105,13 +76,9 @@ class KatanaManager :
...
         _applicationBase = applicationBase;
 
-        _HR(CLRCreateInstance(CLSID_CLRMetaHost, PPV(&_MetaHost)));
         _HR(CLRCreateInstance(CLSID_CLRMetaHostPolicy, PPV(&_MetaHostPolicy)));
 
-        IStreamPtr cfgStream = new ComObject<FileStream>();
-        _HR(static_cast<FileStream*>(cfgStream.GetInterfacePtr())->Open(_clrConfigFilePath));
-
-        WCHAR wzVersion[130] = { 0 };
+        WCHAR wzVersion[130] = L"v4.0.30319";
```
davidfowl: What does this do and what did `{ 0 }` do before?

davidfowl:
:shipit: After those minor comments. Also @troydai reverted the strong name changes so I assume this will fail until we fix it.

src/dnx.clr/KatanaManager.h
moozzyk:
```
wzVersion[0] = L"v";
wzVersion[1] = L"4";
wzVersion[2] = L".";
wzVersion[3] = L"0";
wzVersion[4] = L".";
...
```
Before it was just `wzVersion[0] = 0;` indicating the empty string since the version was passed in the config file.

moozzyk:
@davidfowl this can only go in after #2023

Issue #2016, opened 2015-06-03
Wrapped 4.5 project with NuGet package reference cannot be published


glennc:
* Empty web application referencing a 4.5 csproj
* Add NuGet package (Tried with Moq as well as Microsoft.ServiceFabric.Actor package)

Resulting application can run, and build, but if you try and publish you get the error below. If you go to the wrapped projects project.json and run `dnu restore` to get a lock file then the publish will work.

Potentially related is that when you first create the csproj it targets net452 and initially I see errors resolving the packages. The problem appears to go away without me actually doing anything that should've resolved it if it was a problem. So it might be nothing, but I thought I would mention it.

```
Using Project dependency WrapRepro 1.0.0 for DNX,Version=v4.5.1
  Copying source code from Project dependency WrapRepro
    Source C:\code\WrapRepro\src\WrapRepro\project.json
    Target C:\code\WrapRepro\output\approot\src\WrapRepro

Using Project dependency ClassLibrary1 1.0.0 for .NETFramework,Version=v4.5
  Packing nupkg from Project dependency ClassLibrary1
    Source C:\code\WrapRepro\wrap\ClassLibrary1\project.json
    Target C:\code\WrapRepro\output\approot\packages\ClassLibrary1\1.0.0

Building ClassLibrary1 for .NETFramework,Version=v4.5
  Using Project dependency ClassLibrary1 1.0.0
    Source: C:\code\WrapRepro\wrap\ClassLibrary1\project.json

  Unable to resolve dependency Moq 4.2.1502.0911

  Using Assembly dependency framework/mscorlib 4.0.0.0
    Source: C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.5\mscorlib.dll

  Using Assembly dependency framework/System 4.0.0.0
    Source: C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.5\System.dll

  Using Assembly dependency framework/System.Core 4.0.0.0
    Source: C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.5\System.Core.dll

  Using Assembly dependency framework/Microsoft.CSharp 4.0.0.0
    Source: C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.5\Microsoft.CSharp.dll

[...]

PR #2036, opened 2015-06-09
"dnu wrap" install NuGet dependencies of csproj to DNX packages folder

ChengTian:
parent https://github.com/aspnet/dnx/issues/2016

This combined with fixing from Tooling side (https://github.com/aspnet/Tooling-Internal/issues/362) should fix the bug in parent issue

@davidfowl @glennc @muratg

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Commands/WrapCommand.cs
+++ b/src/Microsoft.Framework.PackageManager/Commands/WrapCommand.cs
@@ -343,7 +343,37 @@ namespace Microsoft.Framework.PackageManager
 
             AddToGlobalJsonSources(rootDir, Path.GetDirectoryName(targetProjectJson));
 
+            // NuGet dependencies of a csproj is installed in <RootDir>/packages,
+            // we need to move them to DNX packages folder
+            InstallCsprojNuGetDependencies(targetProjectJson, rootDir);
+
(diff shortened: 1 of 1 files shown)


src/Microsoft.Framework.PackageManager/Commands/WrapCommand.cs
```diff
@@ -343,9 +343,32 @@ private void EmitProjectWrapper(XElement projectElement)
...
+            feedOptions.IgnoreFailedSources = false;
+            feedOptions.Sources.Add(solutionLocalPackagesDir);
+
+            restoreCommand.SkipRestoreEvents = true;
+            restoreCommand.RestoreDirectories.Add(wrapperProjectFilePath);
+            restoreCommand.FeedOptions = feedOptions;
+            restoreCommand.Reports = Reports;
+
+            // Install all NuGet dependencies to DNX packages folder
+            restoreCommand.Execute().Wait();
```
davidfowl: GetAwaiter().GetResult()

ChengTian:
Thanks for the feedback. Now if you delete some necessary packages from your solution local packages folder and then do `dnu wrap`, you'll see something like:
capture

muratg:
I'm assuming this is tested, changes look good. 

:shipit:

PR #2040, opened 2015-06-09
fix hang in FindRuntimeSpecs

anurse:
Because `imports` was outside the loop, in the case where the last runtime in the list had no imports but the penultimate one DID have imports, the loop became infinite (technically it bounced between two call stack levels infinitely, it didn't actually get stuck in a single infinite loop)

/cc @lodejard @davidfowl

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
@@ -555,9 +555,9 @@ namespace Microsoft.Framework.PackageManager
         {
             effectiveRuntimeSpecs.RemoveAll(spec => spec.Name == runtimeName);
 
+            IEnumerable<string> imports = null;
             foreach (var runtimeFile in runtimeFiles)
             {
-                IEnumerable<string> imports = null;
                 RuntimeSpec runtimeSpec;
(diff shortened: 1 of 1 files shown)


davidfowl:
:shipit:

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
```diff
@@ -555,9 +555,9 @@ public RestoreCommand(FrameworkName fallbackFramework)
         {
             effectiveRuntimeSpecs.RemoveAll(spec => spec.Name == runtimeName);
 
-            IEnumerable<string> imports = null;
             foreach (var runtimeFile in runtimeFiles)
             {
+                IEnumerable<string> imports = null;
```
ericstj: It looks like the subsequent `if (imports != null)` check will always be false.  Are you sure that the fix isn't instead just to close the `foreach` statement before the following `if (imports != null)` check?

anurse: Indeed it should! Fixed now!

ericstj:
works for me :+1:

ericstj:
I just noticed that we still hit this bug in beta5, however it is fixed in beta6.  Was that intentional?

Issue #2026, opened 2015-06-05
DTH sends wrong diagnostics results


troydai:

BillHiebert:
Tracked in TFS by bug 1183897

troydai:
Fixed.

PR #2043, opened 2015-06-10
Send diagnostics once for same group of message

troydai:
Addressing #2026 

/to @BillHiebert @davidfowl 
/review @victorhurdugaci @BrennanConroy 
/cc @muratg @Eilon @danroth27 This is beta5 bug.

Previously the `SendDiagnostics` was ran twice in one "run" of the message process (In a run all messages in the inbox is drained and processed together.) Since the `SendDiagnostics` clear the waiting list of the clients asks for diagnostics information, the second `SendDiagnostics` took no effect essentially. 

It causes problem when following messages fall in one run induce both diagnostics information in _dependency resolve_ and _compilation_. For example: if a `RefreshDependencies` message and a `GetDiagnostics` message combined in one run, the `GetDiagnostics` won't receive any results.

The fix pulls the `SendDiagnostics` away from `SendOutgoingMessages` and executed independently in `DoProcessLoop`. It is executed once for each run.

Here's my test client:
https://github.com/troydai/DesignTimeHostTestser

### DIFF
--- a/src/Microsoft.Framework.DesignTimeHost/ApplicationContext.cs
+++ b/src/Microsoft.Framework.DesignTimeHost/ApplicationContext.cs
@@ -190,15 +190,19 @@ namespace Microsoft.Framework.DesignTimeHost
             {
                 DrainInbox();
 
+                var allDiagnostics = new List<DiagnosticsMessage>();
+
                 if (ResolveDependencies())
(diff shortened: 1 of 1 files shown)


davidfowl:
:shipit:

PR #2049, opened 2015-06-11
Fix Framework References by renaming them

anurse:
This should fix the circular dependency caused by framework assemblies. We do it by actually renaming BOTH the `LibraryDependency` created when a `frameworkAssembly` is specified in a project or package, AND the `LibraryDescription` created when the dependency is resolved. Since the Library names are actually mostly irrelevant (only used for matching with Descriptions), this all works great because the loaders all know to handle these names properly.

We rename a framework assembly dependency `System.Banana` to `fx/System.Banana`. As a result, it can no longer cause a circular dependency with a package or project called `System.Banana`. If a project or package provides BOTH a framework assembly AND a package assembly called `System.Banana`, the framework assembly always wins (because it comes from the GAC)

Still running build verify and tidying up, but go ahead and start reviewing :).

/cc @davidfowl @troydai @muratg


src/Microsoft.Framework.DesignTimeHost/ApplicationContext.cs
```diff
@@ -1233,9 +1233,11 @@ private static string GetProjectRelativeFullPath(Project referencedProject, stri
 
         private static DependencyDescription CreateDependencyDescription(LibraryDescription library)
         {
+            var name = library.Identity.Name;
+            //var name = LibraryRange.GetAssemblyName(library.Identity.Name);
             return new DependencyDescription
             {
-                Name = library.Identity.Name,
```
BrennanConroy: revert this file? Nothing changed.

anurse: Something will change. Once I can get it to build I'm going to be uncommenting line 1237. I left it commented out because I want to see exactly what DTH does without it (it should start showing "fx/System.IO" and friends on the References node, but we want to trim it back to "System.IO"). I don't want to apply the fix until I see the issue though :).

If it ends up being unnecessary, I'll revert the whole file :)

anurse:
Ok, CIs are passing. Ping @davidfowl @troydai

troydai:
Looks good to me. We need it :shipit:

PR #2050, opened 2015-06-11
Fix a warning related to formatted output in dnx.coreclr.cpp

ChengTian:
@moozzyk 

After I upgrade my VS and C++ compiler, building DNX gives me an error:
capture

### DIFF
--- a/src/dnx.coreclr/dnx.coreclr.cpp
+++ b/src/dnx.coreclr/dnx.coreclr.cpp
@@ -506,7 +506,7 @@ extern "C" HRESULT __stdcall CallApplicationMain(PCALL_APPLICATION_MAIN_DATA dat
 
     if (FAILED(hr))
     {
-        wprintf_s(L"TPA      %d %s\n", wcslen(pwszTrustedPlatformAssemblies), pwszTrustedPlatformAssemblies);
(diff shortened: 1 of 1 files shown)


src/dnx.coreclr/dnx.coreclr.cpp
```diff
@@ -506,7 +506,7 @@ extern "C" HRESULT __stdcall CallApplicationMain(PCALL_APPLICATION_MAIN_DATA dat
 
     if (FAILED(hr))
     {
-        wprintf_s(L"TPA      %d %s\n", wcslen(pwszTrustedPlatformAssemblies), pwszTrustedPlatformAssemblies);
+        wprintf_s(L"TPA      %Iu %s\n", wcslen(pwszTrustedPlatformAssemblies), pwszTrustedPlatformAssemblies);
```
moozzyk: does this work on linux/mac?

ChengTian: Looks like we don't compile this file on Linux/mac, which needs `dnx.coreclr.unix/dnx.coreclr.cpp `. This change should be exclusively for windows.
(BTW, Travis CI passes below)

moozzyk:
:ship: :it:

PR #2051, opened 2015-06-11
drop dotnet compatibility bar down to net45

anurse:
This drops the `dotnet` compatibility bar down to `net45`. There are other compatibility things to do for `dotnet` (like supporting arbitrary new frameworks) but DNX doesn't need those for beta5.

/cc @davidfowl @troydai @ChengTian


anurse:
This is in, the branch is just out of date

PR #2052, opened 2015-06-12
Make dnx build 75% faster

victorhurdugaci:
Allow `dnu build` and `dnu pack` to work with globbing patterns and build all projects at once. This is a big perf improvement because each project is compiled only once (caches are shared). Change is 100% backward compatible with the old one-project-one-build.

The more projects you have, the better the results are. MVC has 77 projects in `src` and `test`  combined.

Perf results summary:

```
src:      60s -> 15s
test:     436s -> 106s
src+test: 493s -> 111s
```

Full perf results:

```
Measure-Command { Get-ChildItem -path 'src' -filter project.json -Recurse | Select -expand FullName | Split-Path | %{ dnu build $_ }} | Select -expand TotalSeconds
60.3610104
Measure-Command { dnu build src/** } | Select -expand TotalSeconds
15.2666252
Measure-Command { Get-ChildItem -path 'test' -filter project.json -Recurse | Select -expand FullName | Split-Path | %{ dnu build $_ }} | Select -expand TotalSeconds
436.9794821
Measure-Command { dnu build  test/** } | Select -expand TotalSeconds
103.2918838
Measure-Command { Get-ChildItem -path 'src','test' -filter project.json -Recurse | Select -expand FullName | Split-Path | %{ dnu build $_ }} | Select -expand TotalSeconds
493.5432721
Measure-Command { dnu build  src/** test/** } | Select -expand TotalSeconds
111.036362
```

Please review: @davidfowl @ChengTian

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Building/BuildManager.cs
+++ b/src/Microsoft.Framework.PackageManager/Building/BuildManager.cs
@@ -6,6 +6,8 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using Microsoft.Framework.FileSystemGlobbing;
+using Microsoft.Framework.FileSystemGlobbing.Abstractions;
 using Microsoft.Framework.PackageManager.Utils;
 using Microsoft.Framework.Runtime;
 using Microsoft.Framework.Runtime.Caching;
@@ -19,11 +21,16 @@ namespace Microsoft.Framework.PackageManager
         private readonly IApplicationEnvironment _applicationEnvironment;
         private readonly BuildOptions _buildOptions;
 
+        // These are shared by all projects that will be built by this class
+        private CacheContextAccessor _cacheContextAccessor;
+        private Cache _cache;
+
+        private Runtime.Project _currentProject;
+
         public BuildManager(IServiceProvider hostServices, BuildOptions buildOptions)
         {
             _hostServices = hostServices;
             _buildOptions = buildOptions;
-            _buildOptions.ProjectDir = Normalize(buildOptions.ProjectDir);
 
             _applicationEnvironment = (IApplicationEnvironment)hostServices.GetService(typeof(IApplicationEnvironment));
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.Framework.PackageManager/Building/BuildManager.cs
```diff
@@ -34,9 +41,63 @@ public BuildManager(IServiceProvider hostServices, BuildOptions buildOptions)
...
+                Path.Combine(rootDirectory, file));
+            projectFilesToBuild.AddRange(globbingProjects);
+
+            var sw = Stopwatch.StartNew();
+
+            _cacheContextAccessor = new CacheContextAccessor();
+            _cache = new Cache(_cacheContextAccessor);
+
+            var globalSucess = true;
+            foreach (var project in projectFilesToBuild)
```
pranavkm: Should we build a dependency graph here? If a dependency fails to build, it doesn't make sense to build any of it's dependents

src/Microsoft.Framework.PackageManager/Building/BuildManager.cs
```diff
@@ -34,9 +41,63 @@ public BuildManager(IServiceProvider hostServices, BuildOptions buildOptions)
...
+            _buildOptions.Reports.Information.WriteLine($"Total build time elapsed: { sw.Elapsed }");
+            _buildOptions.Reports.Information.WriteLine($"Total projects built: { projectFilesToBuild.Count }");
+
+            return globalSucess;
+        }
+
+        private static string NormalizeGlobbingPattern(string pattern)
+        {
+            // TODO: use a constant for project.json
+            if (!string.Equals(Path.GetFileName(pattern), "project.json", StringComparison.OrdinalIgnoreCase))
```
pranavkm: Could you pull it into a class-level constant for now? Less refactoring

src/Microsoft.Framework.PackageManager/Building/BuildManager.cs
```diff
@@ -34,9 +41,63 @@ public BuildManager(IServiceProvider hostServices, BuildOptions buildOptions)
...
+            return pattern;
+        }
+
+        private bool BuildInternal(string projectPath)
+        {
             var projectDiagnostics = new List<ICompilationMessage>();
-            Runtime.Project project;
-            if (!Runtime.Project.TryGetProject(_buildOptions.ProjectDir, out project, projectDiagnostics))
+
+            if (!Runtime.Project.TryGetProject(projectPath, out _currentProject, projectDiagnostics))
```
pranavkm: Could this not be a local variable anymore? It only seems to be used in this method.

src/Microsoft.Framework.PackageManager/Building/BuildManager.cs
ChengTian: Use `Microsoft.Framework.Runtime.Project.ProjectFileName`?

pranavkm:
Change looks pretty solid :shipit: 

As a follow up, we should figure out how to make it easy to run tests against packages - since we don't build test projects, simply run them. This would avoid having to recompile src.

src/Microsoft.Framework.PackageManager/Building/BuildManager.cs
victorhurdugaci: That's not a bad suggestion, it could make build even faster when there are a lot of projects but I think that's part of a much bigger change. I want to keep this change as small as possible

src/Microsoft.Framework.PackageManager/Building/BuildManager.cs
victorhurdugaci: It is used in `GetScriptVariable`

victorhurdugaci:
@pranavkm yup, that's on my list. during the lite build I'll generate all the packages (src + test) and then in the second pass run tests using the packages. There might be a few repos that don't support this, tho (MusicStore) is one of them

src/Microsoft.Framework.PackageManager/Building/BuildManager.cs
victorhurdugaci: :+1:

src/Microsoft.Framework.PackageManager/Building/BuildManager.cs
```diff
@@ -19,11 +21,16 @@ public class BuildManager
         private readonly IApplicationEnvironment _applicationEnvironment;
         private readonly BuildOptions _buildOptions;
 
+        // These are shared by all projects that will be built by this class. Fast compile
```
dougbu: "Fast compile" what?

PR #2053, opened 2015-06-12
Fix framework references for .NETCore 5.0 by not allowing them

anurse:
Once they are blocked, the regular assembly references (lib/netcore50 and friends) take over and everything is :sun_with_face: and :sparkles: and :bouquet:

I'm build verifying now and will push to release (with a better commit message) when that's done.

/cc @troydai @davidfowl

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Utils/LockFileUtils.cs
+++ b/src/Microsoft.Framework.PackageManager/Utils/LockFileUtils.cs
@@ -42,9 +42,17 @@ namespace Microsoft.Framework.PackageManager.Utils
                     }
 
                     var assemblyPath = Path.Combine(installPath, filePath);
-                    if (IsAssemblyServiceable(assemblyPath))
+                    try
                     {
-                        lockFileLib.IsServiceable = true;
+                        if (IsAssemblyServiceable(assemblyPath))
+                        {
+                            lockFileLib.IsServiceable = true;
+                            break;
+                        }
+                    }
+                    catch
+                    {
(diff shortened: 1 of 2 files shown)


src/Microsoft.Framework.PackageManager/Utils/LockFileUtils.cs
```diff
@@ -42,9 +42,17 @@ public static LockFileLibrary CreateLockFileLibrary(LockFileLibrary previousLibr
                     }
 
                     var assemblyPath = Path.Combine(installPath, filePath);
-                    if (IsAssemblyServiceable(assemblyPath))
+                    try
                     {
-                        lockFileLib.IsServiceable = true;
+                        if (IsAssemblyServiceable(assemblyPath))
```
anurse: I don't know what caused the exception I got here but it doesn't actually seem wrong to have a guard here. We don't want a broken DLL file causing an exception during restore.

src/Microsoft.Framework.Runtime/NuGet/Utility/VersionUtility.cs
```diff
@@ -131,6 +141,12 @@ public static bool IsDesktop(FrameworkName frameworkName)
                    frameworkName.Identifier == DnxFrameworkIdentifier;
         }
 
+        public static bool HasFrameworkAssemblies(FrameworkName frameworkName)
```
anurse: This isn't needed any more, I'll remove it before merging.

PR #2058, opened 2015-06-12
Add NuGet v3 feed support

ChengTian:
@loudej  @davidfowl  @muratg  @yishaigalatzer

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
@@ -34,7 +34,7 @@ namespace Microsoft.Framework.PackageManager.Restore.NuGet
             string password,
             Reports reports)
         {
-            _baseUri = baseUri + (baseUri.EndsWith("/") ? "" : "/");
+            _baseUri = baseUri + ((baseUri.EndsWith("/") || baseUri.EndsWith("index.json")) ? "" : "/");
             _userName = userName;
             _password = password;
             _reports = reports;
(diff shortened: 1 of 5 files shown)


src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
```diff
@@ -131,13 +139,20 @@ internal class HttpSource
 
                 using (var response = await _client.SendAsync(request, HttpCompletionOption.ResponseHeadersRead))
                 {
+                    if (!throwNotFound && response.StatusCode == HttpStatusCode.NotFound)
```
ChengTian: For v3 feed, FindPackagesById() becomes construct the URI by ourselves and check whether the URI points to a JSON. A 404 page means the packages is not on this feed.

davidfowl:
@ChengTian wrong branch?

src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
yishaigalatzer: Where is the change to change the URL from v3 to v2 in the beta5 branch?

-----Original Message-----

davidfowl: We're doing the real change in release and dev

yishaigalatzer: Cool

-----Original Message-----

ChengTian:
Just wrong target branch, the changes should be correct. Will create a new PR with correct target branch though

PR #2064, opened 2015-06-14
Wei/nuget v3 support release

ChengTian:
@loudej  @davidfowl  @muratg  @yishaigalatzer 

Same changes as https://github.com/aspnet/dnx/pull/2058. Changing target branch to release.

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
@@ -34,7 +34,7 @@ namespace Microsoft.Framework.PackageManager.Restore.NuGet
             string password,
             Reports reports)
         {
-            _baseUri = baseUri + (baseUri.EndsWith("/") ? "" : "/");
+            _baseUri = baseUri + ((baseUri.EndsWith("/") || baseUri.EndsWith("index.json")) ? "" : "/");
             _userName = userName;
             _password = password;
             _reports = reports;
@@ -88,8 +88,16 @@ namespace Microsoft.Framework.PackageManager.Restore.NuGet
             }
         }
 
+        public string BaseUri
+        {
+            get
+            {
+                return _baseUri;
+            }
+        }
+
         internal async Task<HttpSourceResult> GetAsync(string uri, string cacheKey, TimeSpan cacheAgeLimit,
-            Action<Stream> ensureValidContents)
+            Action<Stream> ensureValidContents = null, bool throwNotFound = true)
         {
             Stopwatch sw = new Stopwatch();
             sw.Start();
@@ -101,7 +109,7 @@ namespace Microsoft.Framework.PackageManager.Restore.NuGet
                 return result;
             }
 
-            _reports.Quiet.WriteLine(string.Format("  {0} {1}.", "GET".Yellow(), uri));
+            _reports.Quiet.WriteLine(string.Format("  {0} {1}", "GET".Yellow(), uri));
 
             var request = new HttpRequestMessage(HttpMethod.Get, uri);
             if (_userName != null)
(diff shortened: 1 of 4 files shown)


src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
```diff
@@ -131,13 +139,20 @@ internal class HttpSource
 
                 using (var response = await _client.SendAsync(request, HttpCompletionOption.ResponseHeadersRead))
                 {
+                    if (!throwNotFound && response.StatusCode == HttpStatusCode.NotFound)
```
davidfowl: Why is this special?

ChengTian: For v3 feed, FindPackagesById() becomes constructing the URI by ourselves and check whether the URI points to a JSON. A 404 page means the packages is not on this feed.

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv3Feed.cs
```diff
@@ -0,0 +1,298 @@
...
+
+        internal static bool DetectNuGetV3(HttpSource httpSource, out Uri packageBaseAddress)
+        {
+            try
+            {
+                var result = httpSource.GetAsync(httpSource.BaseUri, "index_json", TimeSpan.FromHours(6)).Result;
+                using (var reader = new StreamReader(result.Stream))
+                {
+                    var content = reader.ReadToEnd();
+                    var indexJson = JObject.Parse(content);
```
JamesNK: You could load the JObject directly from the StreamReader and avoid allocating the string. You're already doing this on line 141.

src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
davidfowl: Meaning you don't want to fail and retry right?

ChengTian: Think about this scenario:
You have two feeds in NuGet.config: FeedA and FeedB
100 packages your app needs are only available on FeedA and 1 package your app needs is only available on FeedB.
Do we want to fail and retry the 100 packages on FeedB even they are not there?

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv3Feed.cs
ChengTian: Thanks. Fixed.

src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
davidfowl: Sure, I was confirming not pushing back.

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv3Feed.cs
```diff
@@ -0,0 +1,297 @@
...
+        public PackageInfo BuildModel(string id, string version)
+        {
+            return new PackageInfo
+            {
+                // If 'Id' element exist, use its value as accurate package Id
+                // Otherwise, use the value of 'title' if it exist
+                // Use the given Id as final fallback if all elements above don't exist
+                Id = id,
+                Version = SemanticVersion.Parse(version),
+                ContentUri = _baseUri + id.ToLowerInvariant() + "/" + version.ToLowerInvariant() + "/" + id.ToLowerInvariant() + "." + version.ToLowerInvariant() + ".nupkg",
```
davidfowl: We to ToLowerInvariant on both id and version twice.

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv3Feed.cs
```diff
@@ -0,0 +1,297 @@
...
+            };
+        }
+
+        public async Task<Stream> OpenNuspecStreamAsync(PackageInfo package)
+        {
+            return await PackageUtilities.OpenNuspecStreamFromNupkgAsync(package, OpenNupkgStreamAsync, _reports.Information);
```
davidfowl: Why is this opening the nupkg stream?

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv3Feed.cs
ChengTian: fixing

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv3Feed.cs
ChengTian: The method `OpenNuspecStreamFromNupkgAsync` will use `OpenNupkgStreamAsync` to open a nupkg stream and find out the nuspec entry and open a stream to that entry.

davidfowl: That's not right for the V3 feed. The Nuspec is downloadable without getting the nupkg.

ChengTian: Sorry we didn't know that. But current approach is not incorrect, it still works. Moreover,  seeking the nuspec entry in nupkg stream is faster than sending a separate nuspec download request.

ChengTian:
Got sign-off from @davidfowl

PR #2072, opened 2015-06-16
Eable loading of an app domain manager from a stream

moozzyk:
Needed to be able to redirect to a different runtime for servicing purposes.

### DIFF
--- a/src/dnx.clr/HostAssemblyStore.cpp
+++ b/src/dnx.clr/HostAssemblyStore.cpp
@@ -1,0 +1,103 @@
+#include "stdafx.h"
+#include "HostAssemblyStore.h"
+#include "FileStream.h"
+#include "ComObject.h"
+#include <string>
+
+const wchar_t* AppDomainManagerAssemblyName = L"dnx.clr.managed, Version=1.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60, ProcessorArchitecture=MSIL";
+
+namespace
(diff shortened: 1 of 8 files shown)


moozzyk:
/cc @davidfowl @lodejard 

@jkotas - I would appreciate if you could also take a look at this

moozzyk:
@blowdart - this will need security review

src/dnx.clr/HostAssemblyStore.cpp
```diff
@@ -0,0 +1,103 @@
+#include "stdafx.h"
+#include "HostAssemblyStore.h"
+#include "FileStream.h"
+#include "ComObject.h"
+#include <string>
+
+const wchar_t* AppDomainManagerTypeName = L"dnx.clr.managed, Version=1.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60, ProcessorArchitecture=MSIL";
```
moozzyk: `AppDomainManagerTypeName` :arrow_right: `AppDomainManagerAssemblyName`

moozzyk:
:watch:

jkotas:
The implementation looks fine to me.

You should watch out for unexpected issues ... HostAssemblyManager is a corner case feature that was build and tested for very specific scenario. It is used it a new mode here that it was not used before.

moozzyk:
Thanks @jkotas!

Issue #2074, opened 2015-06-16
Extend lifetime of HTTP cache for v3 feed detection


ChengTian:
Suggested by @davidfowl , for the HTTP requests used to detect whether given URIs are v3 feeds, we should cache them for one week rather than just 6 hours.

Also we should make `--no-cache` option control the behavior of this kind of cache, too.

PR #2075, opened 2015-06-16
Extend lifetime of v3 feed detection HTTP cache

ChengTian:
parent https://github.com/aspnet/dnx/issues/2074

@davidfowl , doing this for beta6 now.


davidfowl:
:shipit:

PR #2076, opened 2015-06-16
Fx prefix should be stripped when LibraryRange is created

troydai:
@anurse @davidfowl @Eilon @muratg 
/cc @bricelam 

The issue is found by EF when `dnu list -a` failed in their build when targeting `net46`.

The issue is that when a `LibraryRange` is created from a framework assembly library, the loaded assembly name given has the `fx/` prefix. Also when a `LibraryRange` is created from `Library` the prefix is not stripped.

This should be checked in beta5 since it does breaks users.


anurse:
:shipit:

PR #2079, opened 2015-06-17
Servicing CLR based DNX runtimes

moozzyk:
Allows to use a different runtime based on index.txt file

### DIFF
--- a/src/dnx/pal.win32.cpp
+++ b/src/dnx/pal.win32.cpp
@@ -87,20 +87,29 @@ namespace
 {
     std::wstring get_runtime_path(TraceWriter& trace_writer)
     {
-        wchar_t program_files_path_buffer[MAX_PATH];
-        wchar_t* program_files_paths[] =
+        std::vector<wchar_t*> servicing_locations =
         {
-            L"ProgramFiles(x86)",
-#if !defined(_M_IX86)   // in case of 32-bit processes both ProgramFiles and ProgramFiles(x86) point to ProgramFiles(x86)
-            L"ProgramFiles"
-#endif
+            L"DNX_SERVICING",
+            L"ProgramFiles(x86)"
         };
 
-        for (auto i = 0; i < sizeof(program_files_paths)/sizeof(wchar_t*); i++)
+        wchar_t servicing_location_buffer[MAX_PATH];
+        // The servicing index should be always under %ProgramFiles(x86)% however
+        // on 32-bit OSes there is only %ProgramFiles%
+        if (GetEnvironmentVariable(L"ProgramFiles(x86)", servicing_location_buffer, MAX_PATH) == 0)
         {
-            if (GetEnvironmentVariable(program_files_paths[i], program_files_path_buffer, MAX_PATH) != 0)
+            servicing_locations.push_back(L"ProgramFiles");
+        }
+
+        for (auto servicing_location : servicing_locations)
+        {
+            if (GetEnvironmentVariable(servicing_location, servicing_location_buffer, MAX_PATH) != 0)
             {
-                auto runtime_path = dnx::servicing::get_runtime_path(program_files_path_buffer, trace_writer);
+                // %DNX_SERVICING% should point directly to servicing folder. For program files we need to append the
+                // actual servicing folder location to %ProgramFilesXXX%
+                auto append_servicing_folder = servicing_location != servicing_locations.front();
+                auto runtime_path = dnx::servicing::get_runtime_path(servicing_location_buffer, append_servicing_folder, trace_writer);
+
                 if (runtime_path.length() > 0)
                 {
                     return dnx::utils::path_combine(runtime_path, L"bin\\");
--- a/src/dnx/pal.win32.cpp
+++ b/src/dnx/pal.win32.cpp
@@ -6,7 +6,9 @@
 #include "xplat.h"
 #include "TraceWriter.h"
 #include "utils.h"
+#include "servicing.h"
 #include <sstream>
+#include <vector>
 
 std::wstring GetNativeBootstrapperDirectory()
 {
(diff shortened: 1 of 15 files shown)


moozzyk:
/CC: @victorhurdugaci @muratg @davidfowl 

@blowdart - this will need security review

blowdart:
Can you get me some test results, and set up a meeting next week to go through the code once I get back? Thanks

src/dnx/servicing.cpp
```diff
@@ -0,0 +1,94 @@
...
+}
+
+namespace dnx
+{
+    namespace servicing
+    {
+        std::wstring get_runtime_path(const std::wstring& servicing_root_parent, TraceWriter& trace_writer)
+        {
+            auto servicing_root = utils::path_combine(servicing_root_parent, std::wstring(L"Microsoft DNX"));
+            auto servicing_manifest_path = utils::path_combine(servicing_root, std::wstring(L"Servicing\\index.txt"));
```
davidfowl: Shouldn;'t this be multiple path_combine calls?

src/dnx/pal.win32.cpp
```diff
@@ -81,19 +83,56 @@ BOOL GetFullPath(LPCTSTR szPath, LPTSTR pszNormalizedPath)
...
+        wchar_t program_files_path_buffer[MAX_PATH];
+        wchar_t* program_files_paths[] =
+        {
+            L"ProgramFiles(x86)",
+#if !defined(_M_IX86)   // in case of 32-bit processes both ProgramFiles and ProgramFiles(x86) point to ProgramFiles(x86)
+            L"ProgramFiles"
+#endif
+        };
+
+        for (auto i = 0; i < sizeof(program_files_paths)/sizeof(wchar_t*); i++)
```
victorhurdugaci: Do you really need a `for` loop here? There are just 2 elements

src/dnx/servicing.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    std::wstring find_runtime_replacement(std::ifstream& input, TraceWriter& trace_writer)
+    {
+        const std::string runtime_qualifier = std::string{ "dnx|" } + runtime_moniker + "|" + ProductVersionStr + "=";
+
+        for (std::string line; std::getline(input, line); )
+        {
+            if (line.compare(0, runtime_qualifier.length(), runtime_qualifier) == 0)
+            {
+                return dnx::utils::to_wstring(line.substr(runtime_qualifier.length()));
```
victorhurdugaci: This code seems to pick the first matching entry in the index. It must always pick the latest because new patches are added at the end of the file

src/dnx/pal.win32.cpp
```diff
@@ -81,19 +83,56 @@ BOOL GetFullPath(LPCTSTR szPath, LPTSTR pszNormalizedPath)
...
-    HMODULE hostModule;
+    HMODULE hostModule = nullptr;
     try
     {
-        hostModule = LoadLibraryEx(moduleName, NULL, LOAD_LIBRARY_SEARCH_DEFAULT_DIRS);
+        auto runtime_path = get_runtime_path(traceWriter);
+        if (runtime_path.length() > 0)
+        {
+            traceWriter.Write(std::wstring(L"Redirecting runtime to: ").append(runtime_path), true);
+            SetEnvironmentVariable(_T("DNX_DEFAULT_LIB"), runtime_path.c_str());
```
victorhurdugaci: Where is this env variable used?

src/dnx/pal.win32.cpp
```diff
@@ -81,19 +83,56 @@ BOOL GetFullPath(LPCTSTR szPath, LPTSTR pszNormalizedPath)
     return TRUE;
 }
 
+namespace
+{
+    std::wstring get_runtime_path(TraceWriter& trace_writer)
+    {
```
victorhurdugaci: The servicing index can be located in a different folder, other than program files, if you set the `DNX_SERVICING` env var. https://github.com/aspnet/dnx/blob/32358ec9b8970070ff77cbc5beef28c13e3ab09b/src/Microsoft.Framework.Runtime/Servicing/ServicingTable.cs#L31

src/dnx/servicing.cpp
moozzyk: Does it make any difference (apart from copying the string twice)?

src/dnx/servicing.cpp
moozzyk: I think when I talked to @glennc he said there would be only one entry for each source runtime. However if we panning only on appending entries what you are saying makes sense.

src/dnx/pal.win32.cpp
moozzyk: Are you suggesting I should copy the code twice? (If so the answer is yes, I need a loop).

src/dnx/pal.win32.cpp
moozzyk: Must have missed (or forgotten) when reading the spec. Will fix.

src/dnx/servicing.cpp
victorhurdugaci: For nupkgs that's how it work. We always append new lines, never delete

src/dnx/pal.win32.cpp
moozzyk: In a few places but most notably in the app domain manager(s). It's originally set (here)[https://github.com/aspnet/dnx/blob/dev/src/dnx/dnx.cpp#L259] and updated for servicing to force loading managed assemblies from the redirected runtime.

src/dnx/servicing.cpp
moozzyk: OK. I will confirm with @glennc and fix to take the last occurence.

src/dnx/pal.win32.cpp
moozzyk: On 32-bit windows ProgramFiles(x86) does not exist.

moozzyk: We are now always checking `%ProgramFiles(x86)%` and only if it does not exist we will add  `%ProgramFiles%` to the list

src/dnx/servicing.cpp
moozzyk: As per our conversation - there can be multiple entries for the same runtime. We will now take the last entry from the index.txt.

src/dnx/pal.win32.cpp
moozzyk: Added support for DNX_SERVICING. If the file does not exists or no entries are found in this file we will fall back and check the index.txt in the `%ProgramFiles%` if one exists.

moozzyk:
@victorhurdugaci - updated the PR with:
- added support for DNX_SERVICING
- added fallback mechanism to use %ProgramFiles% if no matching entries are found in the `%DNX_SERVICING\%index.txt` file
- we will now use the last matching entry from the index.txt file instead of using the first one

PR #2083, opened 2015-06-17
Produce CommandParsing.Sources on build

rynowak:
This package currently isn't being copied to the build share, so treat it
like the other shared-source packages.


victorhurdugaci:
looks good

victorhurdugaci:
:shipit:

PR #2087, opened 2015-06-17
Servicing CLR based DNX runtimes

moozzyk:
Recreating #2079 which I closed accidentally.

### DIFF
--- a/src/dnx/pal.win32.cpp
+++ b/src/dnx/pal.win32.cpp
@@ -6,7 +6,9 @@
 #include "xplat.h"
 #include "TraceWriter.h"
 #include "utils.h"
+#include "servicing.h"
 #include <sstream>
+#include <vector>
 
 std::wstring GetNativeBootstrapperDirectory()
 {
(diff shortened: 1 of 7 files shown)


src/dnx/servicing.cpp
```diff
@@ -0,0 +1,98 @@
...
+#else
+        "clr"
+#endif
+        "-win-"
+#if defined(AMD64)
+        "x64"
+#elif defined(ARM)
+        "arm"
+#else
+        "x86"
```
muratg: Would it be better to have another elif for "x86" case? and an "else" as an error case (as in "not implemented")?

src/dnx/pal.win32.cpp
```diff
@@ -81,19 +83,65 @@ BOOL GetFullPath(LPCTSTR szPath, LPTSTR pszNormalizedPath)
...
+        std::vector<wchar_t*> servicing_locations =
+        {
+            L"DNX_SERVICING",
+            L"ProgramFiles(x86)"
+        };
+
+        wchar_t servicing_location_buffer[MAX_PATH];
+        // The servicing index should be always under %ProgramFiles(x86)% however
+        // on 32-bit OSes there is only %ProgramFiles%
+        if (GetEnvironmentVariable(L"ProgramFiles(x86)", servicing_location_buffer, MAX_PATH) == 0)
```
muratg: Should strings like "ProgramFiles(x86)" etc. be #define's to ensure we don't have syntax mistakes as code changes?

src/dnx/servicing.cpp
moozzyk: Will think about that. It would have to be separate though since we are building a string literal here.

src/dnx/pal.win32.cpp
moozzyk: I did not do it since it is a bit pain for native code. They shouldn't be `#define`s but `const`s once you make them consts they should not be in .h (one definition rule) so you have to create a *.cpp file and then use `extern` forward declaration to be able to use it. Because we are using these names only in one place and all but one is used just once I did not bother to do that. If you feel strongly about that I can for now define the consts in this file and when it turns out we need to use them elsewhere the can be moved to a separate file.

src/dnx/servicing.cpp
moozzyk: Talked to Murat - we are not adding the assert. Doesn't seem like we are going to add a new runtime or platform anytime soon and even if we did it is not guaranteed that the variables won't be defined so you won't get the assert anyways and still the moniker will be incorrect.

muratg:
:shipit:

PR #2096, opened 2015-06-18
fix issues with installing from a locked lock file

anurse:
Verbally approved by @Eilon and @danroth27 for beta5

@davidfowl says this should work :). Basically this was causing us to add the Lock File Libraries as synthetic "Packages" that already exist on disk meaning we never tried to download them.

/cc @davidfowl @ericstj @muratg


anurse:
I think the failures are due to the purging of the release feed. Running local verify build.

Issue #2097, opened 2015-06-19
project.lock.json locked by some process error after restore


abpiskunov:
![projectlock](https://cloud.githubusercontent.com/assets/1592091/8244935/7198a6ee-15f5-11e5-9dcc-27f6d411f548.png)

I commented out all code in project system that touches project.lock.json in any possible way and could repro it on latest D14Rel + latest WTE. I suppose it is dnu..

Eilon:
Resolved by https://github.com/aspnet/dnx/pull/2103

PR #2103, opened 2015-06-19
Retry creating file stream in LockFileFormat

troydai:
@davidfowl @muratg

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Restore/LockFileFormat.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/LockFileFormat.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Runtime.Versioning;
+using System.Threading;
 using Microsoft.Framework.Runtime.DependencyManagement;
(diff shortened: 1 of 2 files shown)


troydai:
@ChengTian

src/Microsoft.Framework.PackageManager/Restore/LockFileFormat.cs
```diff
@@ -20,12 +21,37 @@ public class LockFileFormat
...
+        private static FileStream OpenFileStream(string filePath)
+        {
+            int retry = 3;
+            while (true)
+            {
+                try
+                {
+                    return new FileStream(filePath, FileMode.Open, FileAccess.Read, FileShare.Read);
+                }
+                catch (Exception ex)
```
ChengTian: We want to resolve the lockfile read/write conflicts in this PR. However, this will retry even when some other kinds of exceptions are thrown. Is it possible that we narrow down to a more specific exception?

ChengTian:
:shipit:

Eilon:
Fixes https://github.com/aspnet/dnx/issues/2097

PR #2113, opened 2015-06-22
Use Mono.4.0.1 on Travis CI to address dnu restore "unknown header" issue

ChengTian:
@davidfowl 

DNX build has been failing with the following error for 3 days
capture

Investigated with @BrennanConroy and we tried using Mono.4.0.1 (default version) instead of 3.12. Turns out it can be fixed by switching to higher Mono version.

@Tratcher , can this error be regression introduced by recent HttpClient changes?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,4 @@
 language: csharp
 sudo: false
-mono:
-  - 3.12.0
 script:
   - ./build.sh --quiet verify


Tratcher:
Discussed offline. That stack trace indicates a regression in Zip processing, not HTTP processing.

ChengTian:
Talked to @Tratcher and @yishaigalatzer , this should be an issue related to format of zip archive and NuGet team saw similar issues. Looks like Mono.3.12 cannot handle the zip archive format while Mono.4.0.1 can.

This PR should be able to fix the issue.

BrennanConroy:
Looks good. :ship: :it:

Issue #1841, opened 2015-05-11
Make "projects" in global.json accept project folder path directly


ChengTian:
Today, each element in  projects  of  `global.json`  should be a parent folder of your projects.

For example, you have  `src/WebApplication5/project.json` , then `src` should be added to  projects , instead of  `src/WebApplication5`. If you add `src/WebApplication5` as an element in `global.json`, we are unable to resolve it.

However, it's actually quite intuitive that I should be able to add `src/MyProject` to `projects` in `global.json` (based on name of the property).

As @davidfowl  suggests, we should measure the performance hit before doing it.

glen-84:
:+1: How else would you reference a single project if src contains multiple?

ChengTian:
@glen-84 , could you please give an example of the scenario you are talking about?

glen-84:
@ChengTian,

If `src` contains multiple projects, then all[1] of those projects are included, when I may just want to include one.

[1] For some reason, right now, it's including three of my projects and excluding MyProj.Tests.

ChengTian:
@glen-84 , thanks, I see your point. Yeah, with this change merged in, you will be able to add a specific project into `global.json`, without adding its peers.

muratg:
Need to do some profiling and finish. Moving to beta6

PR #2116, opened 2015-06-23
Make global.json accept project directories in 'projects' property

ChengTian:
parent https://github.com/aspnet/dnx/issues/1841

Profiled with DotTrace and there is almost no impact on performance.

@davidfowl


anurse:
:shipit:, though I do wonder if this is a place for globbing in the future. Might be too slow though, and I'd rather it be fast than be globbed :P

ChengTian:
Yeah, we have some optimization in `ProjectResolver` so that it is definitely faster than globbing.

PR #2121, opened 2015-06-24
fix #1113 by adding 'dnu feeds list' command

### LINKED ISSUE #1113: Add nuget sources option
Currently trying to list the available NuGet sources using kpm blows up with an exception:

```
C:\Users\danroth27\Documents\Visual Studio 14\Projects\WebApplication7\src\WebApplication7>kpm sources
Specify --help for a list of available options and commands.
System.Exception: TODO: Error: unrecognized argument 'sources'
   at Microsoft.Framework.Runtime.Common.CommandLine.CommandLineApplication.HandleUnexpectedArg(CommandLineApplication command, String[] args, Int32 index, String argTypeName)
   at Microsoft.Framework.Runtime.Common.CommandLine.CommandLineApplication.Execute(String[] args)
   at Microsoft.Framework.PackageManager.Program.Main(String[] args)
```

anurse:
Fixes #1113

Add `dnu sources` command that lists active sources. It looks like this:

image

Disabled sources have `[Disabled]` next to them in yellow to identify them as disabled. I also tried to centralize some of the disjointed code we have to handle NuGet.config.

/cc @davidfowl @victorhurdugaci @troydai

### DIFF
--- a/src/Microsoft.Framework.PackageManager/NuGetConfig.cs
+++ b/src/Microsoft.Framework.PackageManager/NuGetConfig.cs
@@ -1,0 +1,47 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using NuGet;
+
+namespace Microsoft.Framework.PackageManager
+{
+    /// <summary>
+    /// Manages the NuGet.config hierarchy
+    /// </summary>
+    public class NuGetConfig
+    {
+        public NuGetConfig(ISettings settings, PackageSourceProvider sources)
+        {
+            Settings = settings;
+            Sources = sources;
+        }
+
+        public ISettings Settings { get; }
+
+        public IPackageSourceProvider Sources { get; }
+
+        public static NuGetConfig ForSolution(string solutionDirectory)
+        {
+            return ForSolution(solutionDirectory, new PhysicalFileSystem(Directory.GetCurrentDirectory()));
+        }
+
+        public static NuGetConfig ForSolution(string solutionDirectory, IFileSystem fileSystem)
+        {
+            var settings = SettingsUtils.ReadSettings(
+                solutionDirectory, 
+                nugetConfigFile: null, 
+                fileSystem: fileSystem, 
+                machineWideSettings: new CommandLineMachineWideSettings());
+
+            // Recreate the source provider
+            var sources = PackageSourceBuilder.CreateSourceProvider(settings);
+
+            return new NuGetConfig(settings, sources);
+        }
+    }
+}
--- a/src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
@@ -42,7 +42,6 @@ namespace Microsoft.Framework.PackageManager
         {
             FallbackFramework = fallbackFramework;
             FileSystem = new PhysicalFileSystem(Directory.GetCurrentDirectory());
-            MachineWideSettings = new CommandLineMachineWideSettings();
             ScriptExecutor = new ScriptExecutor();
             ErrorMessages = new Dictionary<string, List<string>>(StringComparer.Ordinal);
             InformationMessages = new Dictionary<string, List<string>>(StringComparer.Ordinal);
@@ -58,7 +57,6 @@ namespace Microsoft.Framework.PackageManager
         public FeedOptions FeedOptions { get; set; }
 
         public List<string> RestoreDirectories { get; } = new List<string>();
-        public string NuGetConfigFile { get; set; }
         public bool Lock { get; set; }
         public bool Unlock { get; set; }
 
@@ -66,7 +64,6 @@ namespace Microsoft.Framework.PackageManager
 
         public List<FrameworkName> TargetFrameworks { get; set; } = new List<FrameworkName>();
         public FrameworkName FallbackFramework { get; set; }
-        public IMachineWideSettings MachineWideSettings { get; set; }
         public IFileSystem FileSystem { get; set; }
         public Reports Reports { get; set; }
         public bool CheckHashFile { get; set; } = true;
@@ -77,8 +74,7 @@ namespace Microsoft.Framework.PackageManager
         private Dictionary<string, List<string>> ErrorMessages { get; set; }
         private Dictionary<string, List<string>> InformationMessages { get; set; }
 
-        protected internal ISettings Settings { get; set; }
-        protected internal IPackageSourceProvider SourceProvider { get; set; }
+        protected internal NuGetConfig Config { get; set; }
 
         public async Task<bool> Execute()
         {
@@ -104,7 +100,7 @@ namespace Microsoft.Framework.PackageManager
                 }
             }
 
-            var settings = Settings as Settings;
+            var settings = Config.Settings as Settings;
             if (settings != null)
             {
                 var configFiles = settings.GetConfigFiles();
@@ -310,7 +306,7 @@ namespace Microsoft.Framework.PackageManager
                     new NuGetDependencyResolver(packageRepository)));
 
             var effectiveSources = PackageSourceUtils.GetEffectivePackageSources(
-                SourceProvider,
+                Config.Sources,
                 FeedOptions.Sources,
                 FeedOptions.FallbackSources);
 
@@ -955,12 +951,7 @@ namespace Microsoft.Framework.PackageManager
 
         private void ReadSettings(string solutionDirectory)
         {
-            Settings = SettingsUtils.ReadSettings(solutionDirectory, NuGetConfigFile, FileSystem, MachineWideSettings);
-
-            // Recreate the source provider and credential provider
-            SourceProvider = PackageSourceBuilder.CreateSourceProvider(Settings);
-            //HttpClient.DefaultCredentialProvider = new SettingsCredentialProvider(new ConsoleCredentialProvider(Console), SourceProvider, Console);
-
+            Config = NuGetConfig.ForSolution(solutionDirectory, FileSystem);
         }
 
         private IFileSystem CreateFileSystem(string path)
--- a/src/Microsoft.Framework.PackageManager/Sources/ListSourcesCommand.cs
+++ b/src/Microsoft.Framework.PackageManager/Sources/ListSourcesCommand.cs
@@ -1,0 +1,48 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+
+namespace Microsoft.Framework.PackageManager
+{
+    internal class ListSourcesCommand
+    {
+        public Reports Reports { get; }
+        public string TargetDirectory { get; }
+
+        public ListSourcesCommand(Reports reports, string targetDirectory)
+        {
+            Reports = reports;
+            TargetDirectory = targetDirectory;
+        }
+
+        public int Execute()
+        {
+            // Most of this code is from NuGet, just some refactoring for our system.
+            // Collect config
+            var config = NuGetConfig.ForSolution(TargetDirectory);
+
+            var sources = config.Sources.LoadPackageSources();
+            if (!sources.Any())
+            {
+                Reports.Information.WriteLine("No sources found.");
+            }
+            else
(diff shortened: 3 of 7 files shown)


test/Microsoft.Framework.PackageManager.FunctionalTests/DnuSourcesTests.cs
```diff
@@ -0,0 +1,99 @@
...
+        {
+            get
+            {
+                return TestUtils.GetRuntimeComponentsCombinations();
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(RuntimeComponents))]
+        public void DnuSources_ListsAllSources(string flavor, string os, string architecture)
```
anurse: I'm not 100% happy with this test. It's a little too brittle. I might see if I can tidy it up before pushing.

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
```diff
@@ -63,7 +61,6 @@ public RestoreCommand(FrameworkName fallbackFramework)
 
         public List<FrameworkName> TargetFrameworks { get; set; } = new List<FrameworkName>();
         public FrameworkName FallbackFramework { get; set; }
-        public IMachineWideSettings MachineWideSettings { get; set; }
```
anurse: Nobody ever changes this value...

victorhurdugaci:
Oh no! Please don't add `sources`. Change the name to something else. I already have a PR for `dnu source` which is for bringing source code :)

src/Microsoft.Framework.PackageManager/ConsoleCommands/SourcesConsoleCommand.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Framework.Runtime.Common.CommandLine;
+
+namespace Microsoft.Framework.PackageManager
+{
+    internal static class SourcesCommand
```
victorhurdugaci: `SourcesConsoleCommand`

anurse:
Ok, but `sources` aligns with `nuget` so we might need to consider all the options...

anurse:
I'm not sure I like `source` for the command you're building, it collides with the concept of package sources. I would propose something more like `dnu clone` for that command. It seems clearer what it does and frees this up.

I'm definitely open to renaming this though if people think that would be less confusing. We could do something like `dnu list-sources` (which does start to bring up potential collision with `dnu list`)

src/Microsoft.Framework.PackageManager/NuGetConfig.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
```
troydai: License.

src/Microsoft.Framework.PackageManager/NuGetConfig.cs
```diff
@@ -0,0 +1,44 @@
...
+    internal class NuGetConfig
+    {
+        public NuGetConfig(ISettings settings, PackageSourceProvider sources)
+        {
+            Settings = settings;
+            Sources = sources;
+        }
+
+        public ISettings Settings { get; }
+        public IPackageSourceProvider Sources { get; }
```
troydai: Add a new line to separate the properties.

src/Microsoft.Framework.PackageManager/NuGetConfig.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            var settings = SettingsUtils.ReadSettings(
+                solutionDirectory, 
+                nugetConfigFile: null, 
+                fileSystem: fileSystem, 
+                machineWideSettings: new CommandLineMachineWideSettings());
+
+            // Recreate the source provider and credential provider
+            var sources = PackageSourceBuilder.CreateSourceProvider(settings);
+            //HttpClient.DefaultCredentialProvider = new SettingsCredentialProvider(new ConsoleCredentialProvider(Console), SourceProvider, Console);
```
troydai: Remove the commented code.

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
```diff
@@ -912,12 +909,9 @@ void Display(string indent, IEnumerable<GraphNode> graphs)
...
         {
-            Settings = SettingsUtils.ReadSettings(solutionDirectory, NuGetConfigFile, FileSystem, MachineWideSettings);
-
-            // Recreate the source provider and credential provider
-            SourceProvider = PackageSourceBuilder.CreateSourceProvider(Settings);
-            //HttpClient.DefaultCredentialProvider = new SettingsCredentialProvider(new ConsoleCredentialProvider(Console), SourceProvider, Console);
-
+            var config = NuGetConfig.ForSolution(solutionDirectory, FileSystem);
+            Settings = config.Settings;
+            SourceProvider = config.Sources;
```
troydai: Why not just hold on to the instance of the config?

src/Microsoft.Framework.PackageManager/Sources/ListSourcesCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+        public Reports Reports { get; set; }
+        public string TargetDirectory { get; set; }
+
+        public int Execute()
+        {
+            // Most of this code is from NuGet, just some refactoring for our system.
+            // Collect config
+            var config = NuGetConfig.ForSolution(TargetDirectory);
+
+            var sources = config.Sources.LoadPackageSources().ToList();
```
troydai: Do we need the `ToList()`?

src/Microsoft.Framework.PackageManager/Sources/ListSourcesCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+            {
+                Reports.Information.WriteLine("Registered Sources:");
+                // Iterate over the sources and report them
+                var nameLength = sources.Max(s => s.Name.Length);
+                for (int i = 0; i < sources.Count; i++)
+                {
+                    var source = sources[i];
+                    var indexNumber = i + 1;
+                    var enabledString = source.IsEnabled ? "" : " [Disabled]";
+                    var line = $"{indexNumber.ToString("0\\.").PadRight(3)} {source.Name.PadRight(nameLength)} {source.Source}{enabledString.Yellow().Bold()}";
```
troydai: I would suggest not to format it into a table. The URL source could be long. If user execute the command on a default 80 characters width console may cause it wrap the output.

Or maybe just print it in a list.

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuSourcesTests.cs
```diff
@@ -0,0 +1,99 @@
...
+                    subcommand: "sources",
+                    arguments: "root/sub",
+                    stdOut: out output,
+                    stdErr: out error,
+                    environment: environment,
+                    workingDir: projectPath);
+
+                Assert.Equal(0, exitCode);
+                Assert.Empty(error);
+                Assert.Contains(expectedContent, output);
```
troydai: Errrr, I don't feel good about this.

troydai:
:watch:

src/Microsoft.Framework.PackageManager/Sources/ListSourcesCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+
+            var sources = config.Sources.LoadPackageSources().ToList();
+            if (!sources.Any())
+            {
+                Reports.Information.WriteLine("No sources found.");
+            }
+            else
+            {
+                Reports.Information.WriteLine("Registered Sources:");
+                // Iterate over the sources and report them
```
victorhurdugaci: Nit: add empty line before comment

BrennanConroy:
Fix tests!

anurse:
I'm going to rename this to `list-sources` for now to unblock. I'll also go through the CI failures and remaining code review feedback.

src/Microsoft.Framework.PackageManager/Sources/ListSourcesCommand.cs
anurse: I'm mostly trying to mirror the existing `nuget sources` command, but I'm open to changing it. I'll remove the table formatting for now, we can always tweak it later.

src/Microsoft.Framework.PackageManager/NuGetConfig.cs
anurse: Sure, it's just moved from the old code so I wasn't sure if it was being left in for a reason.

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
anurse: Just trying not to rock the boat too much and leave the code as it was before. I can do that though.

src/Microsoft.Framework.PackageManager/Sources/ListSourcesCommand.cs
anurse: Probably not, I just need to add a counter. I copied this from `nuget sources` so between this and the below comment I'll probably just rework this a bit in general and I'll cover this comment there.

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuSourcesTests.cs
anurse: Yeah, I'm trying to find a balance between a brittle test and testing the actual functionality. I think I can put the whole of the expected output here, but from experience that can lead to a lot of tests breaking from a small change :). Here, I think it'll be fine, at least until we deal with localization :P

anurse:
Ping @troydai @BrennanConroy @victorhurdugaci - Updated to resolve feedback.

src/Microsoft.Framework.PackageManager/Sources/ListSourcesCommand.cs
```diff
@@ -0,0 +1,42 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+
+namespace Microsoft.Framework.PackageManager
+{
+    internal class ListSourcesCommand
+    {
+        public Reports Reports { get; set; }
```
victorhurdugaci: This seems to be required and if it null it will fail. Why don't you pass it as a constructor arg ?

src/Microsoft.Framework.PackageManager/Sources/ListSourcesCommand.cs
```diff
@@ -0,0 +1,42 @@
...
+        public string TargetDirectory { get; set; }
```
victorhurdugaci: Same as above

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuSourcesTests.cs
```diff
@@ -0,0 +1,154 @@
...
+    'root': {
+        'NuGet.Config': """",
+        'sub': {
+            'NuGet.Config': """"
+        }
+    }
+}";
+
+            var expectedContent = @"Registered Sources:
+1.  Source1 https://source1 [Disabled]
```
victorhurdugaci: Is the order deterministic?

src/Microsoft.Framework.PackageManager/Sources/ListSourcesCommand.cs
anurse: Just because everything else did it this way :). I'll fix this one

src/Microsoft.Framework.PackageManager/Sources/ListSourcesCommand.cs
anurse: Same as above :)

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuSourcesTests.cs
anurse: Generally, yes. I do intentionally clear the machine level sources so that should make it deterministic. Changes to our package source code could adjust this order, but those should cause the tests to fail anyway.

anurse:
Ping @victorhurdugaci again :)

troydai:
My concerns are addressed. :shipit:

anurse:
I can't figure out what's wrong with the PR builds. I'm going to give a private TeamCity build a quick shot and if that doesn't work, just YOLO push and be prepared to revert :).

PR #2124, opened 2015-06-24
Creating dnx.common project

moozzyk:
dnx.common is a static library (on Windows) that contains common code used by other projects - this includes tpa.cpp (to avoid having two copies of this file)

Bonus: fixing dnx.clr project to use common build infrastructure (required fixing warnings after bumping warning level to 4)


moozzyk:
:watch:

davidfowl:
:shipit:

PR #2134, opened 2015-06-25
Cleaning coreclr bootstrapper

### LINKED ISSUE #2114: dnx.coreclr.dll will leak if platform assemblies not found
returning here:
https://github.com/aspnet/dnx/blob/dev/src/dnx.coreclr/dnx.coreclr.cpp#L446

will not release memory allocated here:
https://github.com/aspnet/dnx/blob/dev/src/dnx.coreclr/dnx.coreclr.cpp#L433

moozzyk:
Also fixes #2114


src/dnx.common/include/trace_writer.h
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
```
BrennanConroy: Wrong license

src/dnx.common/utils.cpp
```diff
@@ -83,5 +83,19 @@ namespace dnx
...
 #endif
+
+        xstring_t remove_file_from_path(const xstring_t& path)
+        {
+            if (path.back() == _X('\\') || path.back() == _X('/'))
+            {
+                return path;
+            }
+
+            auto last_separator = path.find_last_of(_X("/\\"));
```
BrennanConroy: nit: Use char instead of string literal, looks nicer

moozzyk: can't - it looks for both forward and backward slashes

src/dnx.common/include/trace_writer.h
moozzyk: Looks like some files were not updated - will fix and review other files.

src/dnx.common/include/trace_writer.h
```diff
@@ -0,0 +1,32 @@
...
+
+namespace dnx
+{
+    class trace_writer
+    {
+    public:
+        trace_writer(bool verbose) : m_verbose(verbose)
+        {}
+
+        void write(const dnx::char_t* entry, bool verbose)
```
troydai: how about make `verbose` a default parameter

src/dnx.coreclr/dnx.coreclr.cpp
```diff
@@ -397,126 +230,126 @@ extern "C" HRESULT __stdcall CallApplicationMain(PCALL_APPLICATION_MAIN_DATA dat
...
 
-    cchTrustedPlatformAssemblies = TRUSTED_PLATFORM_ASSEMBLIES_STRING_BUFFER_SIZE_CCH;
-    pwszTrustedPlatformAssemblies = (LPWSTR)calloc(cchTrustedPlatformAssemblies+1, sizeof(WCHAR));
-    if (pwszTrustedPlatformAssemblies == NULL)
-    {
-        goto Finished;
-    }
-    pwszTrustedPlatformAssemblies[0] = L'\0';
+    std::wstring trusted_platform_assemblies;
+    trusted_platform_assemblies.reserve();
```
BrennanConroy: What?

moozzyk: I was looking for this line at the very moment to add a note. Originally we allocated 64KB but I found that typically all the paths are ~8K so I removed `-#define TRUSTED_PLATFORM_ASSEMBLIES_STRING_BUFFER_SIZE_CCH (63 * 1024) //32K WCHARs` but realized that I did not remember setting the size here. Certainly I will fix that.

src/dnx.coreclr/dnx.coreclr.cpp
```diff
@@ -35,344 +27,185 @@ bool GetTrustedPlatformAssembliesList(const wchar_t* szDirectory, bool bNative,
...
+        coreclr_module = LoadCoreClr(coreclr_dir_buffer, trace_writer);
     }
 
-Finished:
-    //Cleanup
-    if (pFnSetProcessMitigationPolicy)
-    {
-        pFnSetProcessMitigationPolicy = nullptr;
-    }
+    coreclr_module = LoadLibraryExW(L"coreclr.dll", NULL, 0);
```
moozzyk: This is broken...

moozzyk: I used to have a return from inside the `if` block but I decided to treat pining as part of loading and did not change the code correctly

src/dnx.common/include/trace_writer.h
moozzyk: Hard to tell if the default should be `true` or `false`. The idea is that errors are always written but debugging messages are _verbose_

moozzyk: Maybe I should change it to `write_error`, `write_message`?

moozzyk:
PR updated

src/dnx.common/include/trace_writer.h
troydai: I don't think it's a big deal here. I'm totally fine if we keep the method as is.

src/dnx.coreclr/dnx.coreclr.cpp
```diff
@@ -397,126 +232,128 @@ extern "C" HRESULT __stdcall CallApplicationMain(PCALL_APPLICATION_MAIN_DATA dat
...
     pCLRRuntimeHost->UnloadAppDomain(domainId, true);
 
-    pCLRRuntimeHost->Stop();
+    return S_OK;
+}
+
+bool IsTracingEnabled()
+{
+    wchar_t buff[2];
+    return GetEnvironmentVariable(L"DNX_TRACE", buff, 2) && buff[0] == L'1';
```
moozzyk: Should be:
```
return GetEnvironmentVariable(L"DNX_TRACE", buff, 2) == 1 && buff[0] == L'1';
```

src/dnx/pal.win32.cpp
```diff
@@ -38,16 +38,8 @@ void WaitForDebuggerToAttach()
...
-    bool m_fVerboseTrace = (nEnvTraceSize == 1);
-    if (m_fVerboseTrace)
-    {
-        szTrace[1] = _T('\0');
-        return _tcsnicmp(szTrace, _T("1"), 1) == 0;
-    }
-
-    return false;
+    wchar_t buff[2];
+    return GetEnvironmentVariable(L"DNX_TRACE", buff, 2) && buff[0] == L'1';
```
moozzyk: Should be:
```
return GetEnvironmentVariable(L"DNX_TRACE", buff, 2) == 1 && buff[0] == L'1';
```

src/dnx.common/utils.cpp
```diff
@@ -83,5 +83,19 @@ namespace dnx
             return attributes != INVALID_FILE_ATTRIBUTES && ((attributes & FILE_ATTRIBUTE_DIRECTORY) == 0);
         }
 #endif
+
+        xstring_t remove_file_from_path(const xstring_t& path)
+        {
+            if (path.back() == _X('\\') || path.back() == _X('/'))
```
davidfowl: back twice? I guess it doesn't really matter but you could store the result.

src/dnx.coreclr/dnx.coreclr.cpp
```diff
@@ -35,344 +27,187 @@ bool GetTrustedPlatformAssembliesList(const wchar_t* szDirectory, bool bNative,
...
-        fSuccess = true;
-
-        // Continue loading as usual
-        hCoreCLRModule = ::LoadLibraryExW(wszClrPath, NULL, 0);
+        trace_writer.write(L"Failed to load loader module", false);
+        return nullptr;
     }
 
-    if (hCoreCLRModule == nullptr)
+    auto AddDllDirectory_function = (FnAddDllDirectory)GetProcAddress(loader_module, "AddDllDirectory");
```
davidfowl: isn't pfn more idomatic?

src/dnx.coreclr/dnx.coreclr.cpp
```diff
@@ -397,126 +232,128 @@ extern "C" HRESULT __stdcall CallApplicationMain(PCALL_APPLICATION_MAIN_DATA dat
...
-    LPCWSTR szEntryPointTypeName = L"DomainManager";
-    LPCWSTR szMainMethodName = L"Execute";
-
-    int nprops = sizeof(property_keys) / sizeof(wchar_t*);
-
-    hr = pCLRRuntimeHost->CreateAppDomainWithManager(
+    HRESULT hr = pCLRRuntimeHost->CreateAppDomainWithManager(
         L"dnx.coreclr.managed",
-        dwFlagsAppDomain,
+        APPDOMAIN_ENABLE_PLATFORM_SPECIFIC_APPS | APPDOMAIN_ENABLE_PINVOKE_AND_CLASSIC_COMINTEROP | APPDOMAIN_DISABLE_TRANSPARENCY_ENFORCEMENT,
```
davidfowl: Not a fan of long lines like this. I prefer to put the flags in a local.

src/dnx.common/utils.cpp
moozzyk: It does not really matter here. But since we already have two checks like this I just created the `ends_with_slash` function where I use a variable. One issue I have not noticed before was that calling `.back()` on an empty string causes UB. In the function I created I check that the string is not empty before calling `.back()`

src/dnx.coreclr/dnx.coreclr.cpp
moozzyk: I don't want to admit it but it is and the replacement I came up with is not even close. I will change it back to `pfn...`

src/dnx.coreclr/dnx.coreclr.cpp
moozzyk: I can do that but it does not really matter - it won't make some of the flags less cryptic (besides, this way it almost looks like `BindingFlags` :smiley:)

davidfowl:
:shipit:

Issue #2127, opened 2015-06-24
Show more friendly error message in "dnu restore" when HTTP requests time out


ChengTian:
Reported by @glennc 

![capture](https://cloud.githubusercontent.com/assets/1383883/8343888/4ac0e7b8-1a8f-11e5-9736-3676157ea98c.PNG)

The message `A task was canceled` doesn't make sense to users. We should display something like
```
HTTP request timed out. Retrying.
```

cnblogs-dudu:
Still get `A task was canceled`.

![screen shot 2015-06-27 at 13 47 38](https://cloud.githubusercontent.com/assets/3196961/8390651/21b38ef2-1cd3-11e5-854a-914cd7da8ec3.png)

BrennanConroy:
You have to wait for the next dnx version to come out.

cnblogs-dudu:
Sorry. I made a mistake.

ciznx:
@cnblogs-dudu  What mistake did you make? I suffered the same problem when restoring my project...why didn't you write more detail about your solution?

cnblogs-dudu:
@ciznx 
It's a natural problem. Up to now, restoring packages from myget.org is always timeout in China.

ciznx:
@cnblogs-dudu Thank you for your reply.
Yes, I found it too that restring packages on a Linux from China will always timeout. (But how a Windows PC works perfectly...)
It's such a pity.

PR #2135, opened 2015-06-25
Give more friendly error message when HTTP requests time out

ChengTian:
parent https://github.com/aspnet/dnx/issues/2127

@davidfowl @glennc 

capture

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Restore/NuGet/ErrorMessageUtils.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/NuGet/ErrorMessageUtils.cs
@@ -1,0 +1,26 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+
+namespace Microsoft.Framework.PackageManager.Restore.NuGet
+{
+    internal static class ErrorMessageUtils
+    {
+        public static string GetFriendlyTimeoutErrorMessage(TaskCanceledException ex, bool isFinalAttempt, bool ignoreFailure)
+        {
+            // "A task was canceled" doesn't make sense to users, use a better error message.
+            var reaction = "Retrying";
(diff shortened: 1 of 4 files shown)


src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
```diff
@@ -146,25 +146,35 @@ public Task<IEnumerable<PackageInfo>> FindPackagesByIdAsync(string id)
...
+                    var message = ex.Message;
+                    if (ex is TaskCanceledException)
+                    {
+                        message = ErrorMessageUtils.GetFriendlyTimeoutErrorMessage(
+                            ex as TaskCanceledException,
+                            isFinalAttempt,
+                            ignoreFailure: false);
+                    }
+
+                    if (isFinalAttempt)
```
BrennanConroy: I don't think this should be around the _ignoreFailure check since we don't want to do those 3 times

src/Microsoft.Framework.PackageManager/Restore/NuGet/ErrorMessageUtils.cs
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+
+namespace Microsoft.Framework.PackageManager.Restore.NuGet
+{
+    internal static class ErrorMessageUtils
+    {
+        public static string GetFriendlyTimeoutErrorMessage(TaskCanceledException ex, bool isFinalAttempt, bool ignoreFailure)
```
BrennanConroy: ignoreFailure is never any value except false

ChengTian: Fixed

ChengTian:
@glennc @Eilon , are you guys happy with the new warning/error message?

Eilon:
:shipit:

Issue #1116, opened 2015-01-21
Improve help for `kpm packages`


dougbu:
- `kpm help packages add` fails with a `System.Exception: TODO: Error: unrecognized argument 'add'`
  - this command is not an alias for the less-common `kpm packages help add` (that one works)
- `kpm packages add -?` oddly describes the target packages directory as the "[source]"
- other `kpm packages` subcommands have similar issues in their help

davidfowl:
This is the same bug as #547

dougbu:
- thought #547 was about placement of options, not the sub-commands themselves?
 - are you saying we internally treat `help` (which looks like a sub-command) as an alias for `--help`?  that's not obvious even from the help
- even the second bullet?

ChengTian:
`kpm help packages`: shows help info of `packages`, which is a subcommand of `kpm`. In this case, `help` is a subcommand of `kpm` command.
`kpm packages help add` shows help info of `add`, which is a subcommand of `kpm packages`. In this case, `help` is a subcommand of `kpm packages` command.

`kpm help packages add` is invalid because the subcommand `help` of `kpm` only accepts one argument, which should be one of its peers (a subcommand of `kpm`).

The behavior is well-defined and consistent. But seems people don't like this consistency because it's not intuitive enough. @davidfowl , do we want to remove `help` subcommand and only use `--help|-h` option?

davidfowl:
Yes

PR #2139, opened 2015-06-26
Remove help subcommand

ChengTian:
parent https://github.com/aspnet/dnx/issues/1116

@davidfowl @muratg @Eilon 

Correct usage of help subcommand is something like
```
dnu packages help add
```
However, people keep doing
```
dnu packages add help
```
or
```
dnu help packages add
```
This PR removes the confusing help subcommand and we only use `-?|-h|--help` option:
```
dnu packages add --help
```


Eilon:
:shipit: from me.

muratg:
:shipit:

Issue #1372, opened 2015-03-11
`kpm build` executes `postBuild` scripts even if the build failed.


lefb766:
This behaviour is not intuitive and may cause unexpected side-effect.

Here's simplest steps to repro:
1. Create project.json described below and a .cs file with random content (which doesn't compile).
2. Run `kpm build`.

project.json:
 ```json
{
    "scripts" : {
        "postBuild" : [
            "echo postBuild executed"
        ]
    }
}
```

Dnx version: 1.0.0-beta4-11244

davidfowl:
Agreed that's unexpected. Do you want to send a PR to fix it?

lefb766:
Yes, I'd love to. I'm not familiar with the code base however, so it might take long to find where to fix.

davidfowl:
Here's the line of code:

https://github.com/aspnet/DNX/blob/dev/src/Microsoft.Framework.PackageManager/Building/BuildManager.cs#L195

There's a variable called success that you can check before executing the scripts

lefb766:
Thanks.

muratg:
@ChengTian can you look at the PR to see if we can get this in?

PR #2144, opened 2015-06-26
"dnu pack" doesn't run postbuild nor postpack scripts when build fails

ChengTian:
parent https://github.com/aspnet/dnx/issues/1372

@davidfowl @BrennanConroy


BrennanConroy:
:ship: :it:

PR #2146, opened 2015-06-27
Implementing servicing for CoreCLR runtime

moozzyk:
@victorhurdugaci @muratg @davidfowl

### DIFF
--- a/src/dnx/servicing.cpp
+++ b/src/dnx/servicing.cpp
@@ -11,7 +11,7 @@
 
 namespace
 {
-    const std::string runtime_moniker =
+    const char* runtime_moniker =
 #if defined(CORECLR_WIN)
         "coreclr"
 #else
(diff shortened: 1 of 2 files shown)


src/dnx/servicing.cpp
```diff
@@ -31,6 +31,10 @@ namespace
     {
         const std::string runtime_qualifier = std::string{ "dnx|" } + runtime_moniker + "|" + ProductVersionStr + "=";
 
+        trace_writer.write(
+            std::wstring(L"Looking for redirections for runtime ")
+                .append(dnx::utils::to_wstring(runtime_qualifier).substr(0, runtime_qualifier.length() - 1)), true);
+
```
muratg: :+1:

muratg:
Code looks good. I assume you already tested it. :)

:shipit:

davidfowl:
I think trace is still super hard to read :smile:. The rest of the code looks good tho

moozzyk:
@muratg - Yes, I debugged through different scenarios

@davidfowl - there is no string interpolation in C++ ;) - I do have an idea which I need to try out but it will be probably after beta6...

Issue #2061, opened 2015-06-13
Install all versions of packages found in the graph during restore but only write the reduced set to lock file


davidfowl:
Before beta5, if I never used floating version dependencies in my project, it would never go online for the same package once it was downloaded. Since we're doing versionn unification during restore, we don't download some versions that were discovered during the graph walk. The side effect of this is calling restore again on the same project will hit all feeds for any package that was removed during the reduction.

We should download all packages but never write them to the lock file if they weren't in the reduced set.

/cc @anurse @lodejard @yishaigalatzer

PR #2147, opened 2015-06-28
Download all packages even the ones that don't end up in the lock file

davidfowl:
- This makes it so that we avoid hitting remote sources in the most
common cases.

#2061

/cc @emgarten @anurse


emgarten:
I've ported this to NuGet3 also. It helped perf since 17 packages hit this scenario while restoring a UWP project.

PR #2156, opened 2015-06-30
fix #1683 by preferring listed packages

### LINKED ISSUE #1683: DNU restore should still restore unlisted packages
If the Published date value is  1900-01-01T00:00:00 , then the package is unlisted and it should no longer be considered in dependency resolution when versions are being selected, or during initial installs of packages where users are discovering versions.

However, these "unlisted" versions _must still be allowed_ when the user is restoring packages so that the version doesn't disappear out from under them and break their build.

This feature allows package authors to unlist packages for two reasons:
1.The package is being staged for an upcoming release announcement and they want to list the package at the time of the announcement.
2.The package has a very bad bug and the author doesn't want any new users to discover the version that has the bug.

The logic may be different for the v3 feed and we will need to account for that when we add support for v3.

anurse:
...but accepting unlisted packages if there are no matches

Fixes #1683 

/cc @davidfowl @ChengTian

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
@@ -16,14 +16,20 @@ namespace Microsoft.Framework.PackageManager.Restore.NuGet
 {
     public class NuGetv2Feed : IPackageFeed
     {
-        private static readonly XNamespace _defaultNamespace = XNamespace.Get("http://www.w3.org/2005/Atom");
-        private static readonly XName _xnameEntry = XName.Get("entry", _defaultNamespace.NamespaceName);
-        private static readonly XName _xnameTitle = XName.Get("title", _defaultNamespace.NamespaceName);
-        private static readonly XName _xnameContent = XName.Get("content", _defaultNamespace.NamespaceName);
-        private static readonly XName _xnameLink = XName.Get("link", _defaultNamespace.NamespaceName);
-        private static readonly XName _xnameProperties = XName.Get("properties", "http://schemas.microsoft.com/ado/2007/08/dataservices/metadata");
-        private static readonly XName _xnameId = XName.Get("Id", "http://schemas.microsoft.com/ado/2007/08/dataservices");
-        private static readonly XName _xnameVersion = XName.Get("Version", "http://schemas.microsoft.com/ado/2007/08/dataservices");
+        private static readonly XNamespace _defaultNamespace = "http://www.w3.org/2005/Atom";
+        private static readonly XNamespace _odataMetadataNamespace = "http://schemas.microsoft.com/ado/2007/08/dataservices/metadata";
+        private static readonly XNamespace _odataNamespace = "http://schemas.microsoft.com/ado/2007/08/dataservices";
+        private static readonly XName _xnameEntry = _defaultNamespace + "entry";
+        private static readonly XName _xnameTitle = _defaultNamespace + "title";
+        private static readonly XName _xnameContent = _defaultNamespace + "content";
+        private static readonly XName _xnameLink = _defaultNamespace + "link";
+        private static readonly XName _xnameProperties = _odataMetadataNamespace + "properties";
+        private static readonly XName _xnameId = _odataNamespace + "Id";
+        private static readonly XName _xnameVersion = _odataNamespace + "Version";
+        private static readonly XName _xnamePublish = _odataNamespace + "Published";
+
+        // An unlisted package's publish time must be 1900-01-01T00:00:00.
+        private static readonly DateTime _unlistedPublishedTime = new DateTime(1900, 1, 1, 0, 0, 0);
 
         private readonly string _baseUri;
         private readonly Reports _reports;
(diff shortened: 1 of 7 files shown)


src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
```diff
@@ -24,6 +24,10 @@ public class NuGetv2Feed : IPackageFeed
         private static readonly XName _xnameProperties = XName.Get("properties", "http://schemas.microsoft.com/ado/2007/08/dataservices/metadata");
         private static readonly XName _xnameId = XName.Get("Id", "http://schemas.microsoft.com/ado/2007/08/dataservices");
         private static readonly XName _xnameVersion = XName.Get("Version", "http://schemas.microsoft.com/ado/2007/08/dataservices");
+        private static readonly XName _xnamePublish = XName.Get("Published", "http://schemas.microsoft.com/ado/2007/08/dataservices");
```
troydai: Extract the string of namespace to a constant value?

moozzyk: Even better - you can just use `+` instead of `XName.Get()`.

moozzyk: create:
```
XNameSpace astoriaNs = "http://schemas.microsoft.com/ado/2007/08/dataservices");
```

use inline:
```
... astoriaNs + "Published" ...
```

src/Microsoft.Framework.PackageManager/Restore/RestoreOperations.cs
```diff
@@ -293,8 +293,14 @@ public Task<GraphItem> FindLibraryCached(RestoreContext context, LibraryRange li
                 return nonHttpMatch;
             }
 
-            // Otherwise try the http sources
-            var httpMatch = await FindLibrary(libraryRange, providers.Where(p => p.IsHttp), provider => provider.FindLibrary(libraryRange, context.FrameworkName));
+            // Otherwise try listed packages on http sources
+            var httpMatch = await FindLibrary(libraryRange, providers.Where(p => p.IsHttp), provider => provider.FindLibrary(libraryRange, context.FrameworkName, includeUnlisted: false));
+
+            // If the http sources failed to find a listed package that matched, try unlisted packages
+            if(httpMatch == null)
```
troydai: Format.

src/Microsoft.Framework.PackageManager/Restore/RemoteWalkProvider.cs
```diff
@@ -27,10 +27,14 @@ public RemoteWalkProvider(IPackageFeed source)
...
-        public async Task<WalkProviderMatch> FindLibrary(LibraryRange libraryRange, FrameworkName targetFramework)
+        public async Task<WalkProviderMatch> FindLibrary(LibraryRange libraryRange, FrameworkName targetFramework, bool includeUnlisted)
         {
             var results = await _source.FindPackagesByIdAsync(libraryRange.Name);
             PackageInfo bestResult = null;
+            if(!includeUnlisted)
+            {
+                results = results.Where(p => p.Listed);
+            }
             foreach (var result in results)
```
troydai: Add a line before the `foreach`

troydai:
:watch:

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
anurse: :+1:

src/Microsoft.Framework.PackageManager/Restore/NuGet/NuGetv2Feed.cs
```diff
@@ -17,13 +17,19 @@ namespace Microsoft.Framework.PackageManager.Restore.NuGet
...
     {
         private static readonly XNamespace _defaultNamespace = XNamespace.Get("http://www.w3.org/2005/Atom");
-        private static readonly XName _xnameEntry = XName.Get("entry", _defaultNamespace.NamespaceName);
-        private static readonly XName _xnameTitle = XName.Get("title", _defaultNamespace.NamespaceName);
-        private static readonly XName _xnameContent = XName.Get("content", _defaultNamespace.NamespaceName);
-        private static readonly XName _xnameLink = XName.Get("link", _defaultNamespace.NamespaceName);
-        private static readonly XName _xnameProperties = XName.Get("properties", "http://schemas.microsoft.com/ado/2007/08/dataservices/metadata");
-        private static readonly XName _xnameId = XName.Get("Id", "http://schemas.microsoft.com/ado/2007/08/dataservices");
-        private static readonly XName _xnameVersion = XName.Get("Version", "http://schemas.microsoft.com/ado/2007/08/dataservices");
+        private static readonly XNamespace _odataMetadataNamespace = XNamespace.Get("http://schemas.microsoft.com/ado/2007/08/dataservices/metadata");
```
moozzyk: nit: you don't need `XNamespace.Get` - just 
```
 private static readonly XNamespace _odataMetadataNamespace = "http://schemas.microsoft.com/ado/2007/08/dataservices/metadata"; 
```
it will know to convert the string to `XNamespace`

anurse: I suppose... though I dislike implicit conversion operators :). I'll do it though, because you asked nicely :P

troydai:
:shipit:

Issue #2132, opened 2015-06-25
Provide an option to 'force' install a global command


kichalla:
I am writing a class library which has a command meant to be used as a global command. I am going through several iterations to test if this global command works as its supposed to. 

Now since I am still developing and the version of this package is not going to change, I am seeing that `dnu commands install` does not replace the package if its already installed(i suppose my assumption is right?).

Can we provide a 'force' option for install which will replace the content even if the version of the package having the command hasn't changed?

Current available options:
```
Usage: dnu commands install [arguments] [options]

Arguments:
  [package]  The name of the application package
  [version]  The version of the application package

Options:
  -o|--overwrite              Overwrites conflicting commands
  -s|--source <FEED>          A list of packages sources to use for this command
  -f|--fallbacksource <FEED>  A list of packages sources to use as a fallback
  -p|--proxy <ADDRESS>        The HTTP proxy to use when retrieving packages
  --no-cache                  Do not use local cache
  --packages                  Path to restore packages
  --ignore-failed-sources     Ignore failed remote sources if there are local packages meeting version requirements
  --quiet                     Do not show output such as HTTP request/cache information
  --parallel                  Restores in parallel when more than one project.json is discovered.
  -?|-h|--help                Show help information
```

Workaround:
Do not use as global command just do `dnx . <global-command-name> <rest of the stuff that the command expects>`

BrennanConroy:
Does --overwrite not work?

kichalla:
@BrennanConroy : --overrwrite is for conflicting commands...like you could have multiple packages with the same command name

muratg:
@kichalla What's the behavior when you use `--overwrite` with the same package?  It should work here too, no?  Perhaps we can make the help for the cmdline switch more helpful for this case too.

kichalla:
@muratg I verified and the `--overwrite` switch is not overriding in this case.

BrennanConroy:
Should we change the behavior of overwrite or add a new command? If new command what should it be called.

muratg:
I think overwrite should just work.  Not sure why it doesn't when the package name is the same...

victorhurdugaci:
Yeah, make overwrite reinstall the package. Thought your build script should increment the build number (the `t-<number>` pattern) so it always generate a newer package.

BrennanConroy:
004e130e3307a7c241789bb043873b718607ae45

PR #2158, opened 2015-06-30
--overwrite will overwrite command package #2132

BrennanConroy:
#2132 
@victorhurdugaci

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Install/InstallGlobalCommand.cs
+++ b/src/Microsoft.Framework.PackageManager/Install/InstallGlobalCommand.cs
@@ -67,6 +67,20 @@ namespace Microsoft.Framework.PackageManager
 
             var temporaryProjectFileFullPath = CreateTemporaryProject(FeedOptions.TargetPackagesFolder, packageId, packageVersion);
 
+            var packageFullPath = Path.Combine(
+                _commandsRepository.PackagesRoot.Root,
+                _commandsRepository.PathResolver.GetPackageDirectory(packageId, new SemanticVersion(packageVersion)));
+
+            if (OverwriteCommands)
+            {
+                if (Directory.Exists(packageFullPath))
+                {
+                    WriteInfo($"Deleting {packageFullPath}");
+                    FileOperationUtils.DeleteFolder(packageFullPath);
+                    Directory.Delete(packageFullPath);
+                }
+            }
+
             try
             {
                 RestoreCommand.RestoreDirectories.Add(temporaryProjectFileFullPath);
@@ -81,10 +95,6 @@ namespace Microsoft.Framework.PackageManager
                 FileOperationUtils.DeleteFolder(folderToDelete);
                 Directory.Delete(folderToDelete);
             }
-
-            var packageFullPath = Path.Combine(
-                _commandsRepository.PackagesRoot.Root,
-                _commandsRepository.PathResolver.GetPackageDirectory(packageId, new SemanticVersion(packageVersion)));
 
             if (!ValidateApplicationPackage(packageFullPath))
             {
--- a/src/Microsoft.Framework.PackageManager/ConsoleCommands/CommandsConsoleCommand.cs
+++ b/src/Microsoft.Framework.PackageManager/ConsoleCommands/CommandsConsoleCommand.cs
@@ -35,7 +35,7 @@ namespace Microsoft.Framework.PackageManager
                 var argPackage = c.Argument("[package]", "The name of the application package");
                 var argVersion = c.Argument("[version]", "The version of the application package");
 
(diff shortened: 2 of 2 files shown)


src/Microsoft.Framework.PackageManager/Install/InstallGlobalCommand.cs
```diff
@@ -67,6 +67,17 @@ public FeedOptions FeedOptions
 
             var temporaryProjectFileFullPath = CreateTemporaryProject(FeedOptions.TargetPackagesFolder, packageId, packageVersion);
 
+            var packageFullPath = Path.Combine(
```
kichalla: probably we should change the help text too? currently it says `Overwrites conflicting commands`

src/Microsoft.Framework.PackageManager/Install/InstallGlobalCommand.cs
```diff
@@ -67,6 +67,17 @@ public FeedOptions FeedOptions
...
+                _commandsRepository.PackagesRoot.Root,
+                _commandsRepository.PathResolver.GetPackageDirectory(packageId, new SemanticVersion(packageVersion)));
+
+            if (OverwriteCommands)
+            {
+                WriteVerbose($"Deleting {packageFullPath}");
+                FileOperationUtils.DeleteFolder(packageFullPath);
+                Directory.Delete(packageFullPath);
```
troydai: Need a meaningful error message when deleting fails.

troydai:
:watch:

src/Microsoft.Framework.PackageManager/Install/InstallGlobalCommand.cs
```diff
@@ -67,6 +67,17 @@ public FeedOptions FeedOptions
...
             var temporaryProjectFileFullPath = CreateTemporaryProject(FeedOptions.TargetPackagesFolder, packageId, packageVersion);
 
+            var packageFullPath = Path.Combine(
+                _commandsRepository.PackagesRoot.Root,
+                _commandsRepository.PathResolver.GetPackageDirectory(packageId, new SemanticVersion(packageVersion)));
+
+            if (OverwriteCommands)
+            {
+                WriteVerbose($"Deleting {packageFullPath}");
+                FileOperationUtils.DeleteFolder(packageFullPath);
```
victorhurdugaci: Um... what happens if the packages is not installed? As in running with `--overwrite` when there is nothing to overwrite

BrennanConroy: Ya... I fixed it, will update

src/Microsoft.Framework.PackageManager/Install/InstallGlobalCommand.cs
BrennanConroy: Suggestions?

BrennanConroy:
Updated

src/Microsoft.Framework.PackageManager/Install/InstallGlobalCommand.cs
```diff
@@ -67,6 +67,20 @@ public FeedOptions FeedOptions
...
 
+            var packageFullPath = Path.Combine(
+                _commandsRepository.PackagesRoot.Root,
+                _commandsRepository.PathResolver.GetPackageDirectory(packageId, new SemanticVersion(packageVersion)));
+
+            if (OverwriteCommands)
+            {
+                if (Directory.Exists(packageFullPath))
+                {
+                    WriteVerbose($"Deleting {packageFullPath}");
```
victorhurdugaci: I think this should be WriteInformation. It is pretty important to know when things get deleted

victorhurdugaci:
The change seems to be okay but I have one scenario that I would like you to test. I'm wondering if upgrade + overwrite works. Basically have package_v1 and packave_v1.1 and upgrade from package_v1 to v1.1 with overwrite.

I don't expect it to break but please check it. 

If it works, then :shipit:

src/Microsoft.Framework.PackageManager/Install/InstallGlobalCommand.cs
```diff
@@ -67,6 +67,20 @@ public FeedOptions FeedOptions
...
+                    FileOperationUtils.DeleteFolder(packageFullPath);
+                    Directory.Delete(packageFullPath);
```
troydai: Still doesn't capture the IOException if the deleting fails.

BrennanConroy: I don't think we need to catch it. If it ever happens then we shouldn't continue the global install and the exception has as much information in it as we know.

PR #2159, opened 2015-06-30
Performance: reuse StringBuilder in JsonBuffer to reduce GC time

troydai:
Through profiling the scenario of opening a huge solution with 75+ project in VS, it is discovered that DTH spends 21% of CPU time on GC. 

This is one of the efforts trying to reduce the GC time. This change avoid `StringBuilder::AddChar` allocating memory repeatedly.

GC time down to 17%. I'd like to have this to be included in beta6.

/cc @davidfowl @muratg @Tratcher

### DIFF
--- a/src/Microsoft.Framework.Runtime/Json/JsonBuffer.cs
+++ b/src/Microsoft.Framework.Runtime/Json/JsonBuffer.cs
@@ -14,6 +14,8 @@ namespace Microsoft.Framework.Runtime.Json
         public const string ValueTrue = "true";
         public const string ValueFalse = "false";
 
+        private readonly StringBuilder _buffer = new StringBuilder();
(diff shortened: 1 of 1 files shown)


src/Microsoft.Framework.Runtime/Json/JsonBuffer.cs
```diff
@@ -14,6 +14,8 @@ internal class JsonBuffer
         public const string ValueTrue = "true";
         public const string ValueFalse = "false";
 
+        private readonly StringBuilder _buf = new StringBuilder();
+        private readonly StringBuilder _unicodesBuf = new StringBuilder(4);
```
muratg: Suggested names: _buffer and _codePointsBuffer

muratg:
Looks like a simple change to reuse the buffer and respectable improvement in GC time.

:shipit:

Issue #1896, opened 2015-05-19
project.json provide an error for an unknown PCL target


JamesNK:
An unknown PCL framework target turns into DNX 4.5.1 under References in Solution Explorer.

If you have multiple framework targets then the unknown PCL framework target completely disappears.

There should be an error when there is an unknown PCL framework. Right now it appears to fail silently.

A related issue to this is whether there should be a way to decouple the framework target from what appears in the NuGet package. e.g. If I have the target "net45+win+wp80+wpa81" and I want to add "MonoAndroid10+Xamarin.iOS10+MonoTouch10" as well to help Mono choose the right package, right now project.json won't know what to do with it.

JamesNK:
Example of this issue:

I wanted to add a framework target of "net40+sl5+win+wp80+wpa81". It wasn't recognized and didn't appear. Because I had no way to figure out what the magic string is I fell back to using ".NETPortable,Version=v4.0,Profile=Profile328"

muratg:
We should raise an error in this case.

PR #2161, opened 2015-07-01
Show errors when find unsupported framework in project.json

ChengTian:
parent https://github.com/aspnet/dnx/issues/1896

@davidfowl @muratg @troydai 

Error in VS:
capture1

Error in command-line tool (`dnu restore`):
capture

Error in command-line tool (`dnu build`):
capture2

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
@@ -243,9 +243,18 @@ namespace Microsoft.Framework.PackageManager
             var projectLockFilePath = Path.Combine(projectFolder, LockFileFormat.LockFileName);
 
             Runtime.Project project;
-            if (!Runtime.Project.TryGetProject(projectJsonPath, out project))
+            var diagnostics = new List<ICompilationMessage>();
+            if (!Runtime.Project.TryGetProject(projectJsonPath, out project, diagnostics))
             {
                 throw new Exception("TODO: project.json parse error");
+            }
+
+            if (diagnostics.HasErrors())
+            {
+                var errorMessages = diagnostics
+                    .Where(x => x.Severity == CompilationMessageSeverity.Error)
+                    .Select(x => x.Message);
+                ErrorMessages.GetOrAdd(projectJsonPath, _ => new List<string>()).AddRange(errorMessages);
             }
 
             var lockFile = await ReadLockFile(projectLockFilePath);
--- a/src/Microsoft.Framework.Runtime/Project.cs
+++ b/src/Microsoft.Framework.Runtime/Project.cs
@@ -312,7 +312,7 @@ namespace Microsoft.Framework.Runtime
                         key => repository.ValueAsString(key).Value);
             }
 
-            project.BuildTargetFrameworksAndConfigurations(rawProject);
+            project.BuildTargetFrameworksAndConfigurations(rawProject, diagnostics);
 
             PopulateDependencies(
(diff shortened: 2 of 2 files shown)


src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
```diff
@@ -243,11 +243,18 @@ public RestoreCommand(FrameworkName fallbackFramework)
...
-            if (!Runtime.Project.TryGetProject(projectJsonPath, out project))
+            var diagnostics = new List<ICompilationMessage>();
+            if (!Runtime.Project.TryGetProject(projectJsonPath, out project, diagnostics))
             {
                 throw new Exception("TODO: project.json parse error");
             }
 
+            if (diagnostics.HasErrors())
+            {
+                var error = diagnostics.First(x => x.Severity == CompilationMessageSeverity.Error);
```
muratg: Could there be multiple errors? What should be the output in that case?

src/Microsoft.Framework.Runtime/Project.cs
```diff
@@ -312,7 +312,7 @@ internal static Project GetProjectFromStream(Stream stream, string projectName,
                         key => repository.ValueAsString(key).Value);
             }
 
-            project.BuildTargetFrameworksAndConfigurations(rawProject);
+            project.BuildTargetFrameworksAndConfigurations(rawProject, diagnostics);
```
muratg: It doesn't look like we're doing anything with the `diagnostics` after this call. Should there be a check before we populate the dependencies below?

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
ChengTian: Good point! It's possible that we have more than one unknown framework. I should turn to `AggregateException` here, in order to represent multiple errors. Fixed in follow-up commit.

src/Microsoft.Framework.Runtime/Project.cs
ChengTian: This `diagnostics` is for output purpose. The check will be done by caller of this method.

muratg: :+1:

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
```diff
@@ -243,11 +243,20 @@ public RestoreCommand(FrameworkName fallbackFramework)
...
             {
                 throw new Exception("TODO: project.json parse error");
             }
 
+            if (diagnostics.HasErrors())
+            {
+                var exceptions = diagnostics
+                    .Where(x => x.Severity == CompilationMessageSeverity.Error)
+                    .Select(x => new Exception(x.FormattedMessage));
+                throw new AggregateException(exceptions);
```
davidfowl: Don't throw exceptions from command line tools. It looks bad :smile: . I'm also not sure if we should fail up front here. Maybe we can just warn about unknown frameworks during restore.

ChengTian: I cannot see any benefit postponing the failure to build/publish. so let's fail fast to eliminate unnecessary confusion. Just pushed the fix.

ChengTian:
Fixed behavior of `dnu restore`, instead of throwing exceptions, we show something like:
capture

davidfowl:
:shipit:

PR #2167, opened 2015-07-01
Restoring multiple times with same RestoreCommand now clears summary …

BrennanConroy:
…messages

### DIFF
--- a/src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
+++ b/src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
@@ -23,8 +23,6 @@ namespace Microsoft.Dnx.Tooling
     {
         private static readonly int MaxDegreesOfConcurrency = Environment.ProcessorCount;
 
-        private int _installCount;
-
         public RestoreCommand() :
             this(fallbackFramework: null)
         {
@@ -40,8 +38,6 @@ namespace Microsoft.Dnx.Tooling
             FallbackFramework = fallbackFramework;
             FileSystem = new PhysicalFileSystem(Directory.GetCurrentDirectory());
             ScriptExecutor = new ScriptExecutor();
-            ErrorMessages = new Dictionary<string, List<string>>(StringComparer.Ordinal);
-            InformationMessages = new Dictionary<string, List<string>>(StringComparer.Ordinal);
             Reports = new Reports
             {
                 Information = new NullReport(),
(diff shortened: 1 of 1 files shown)


davidfowl:
Can you show before and after since there are no tests

BrennanConroy:
Before:
multirestore
After:
Same except remove the 10 package(s) line

Easily reproducible by installing a global command.

davidfowl:
What happens if you pass multiple projects to restore?

BrennanConroy:
Nothing different.

davidfowl:
Here's what @BrennanConroy and I decided, add a small `RestoreMessageContext` that has Errors and Info messages and pass that to each invoke call. That state is different per call to Execute()

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
```diff
@@ -71,8 +67,33 @@ public RestoreCommand(FrameworkName fallbackFramework)
         public bool SkipRestoreEvents { get; set; }
         public bool IgnoreMissingDependencies { get; set; }
 
-        private Dictionary<string, List<string>> ErrorMessages { get; set; }
-        private Dictionary<string, List<string>> InformationMessages { get; set; }
+        private class SummaryContext
```
davidfowl: Private classes always go to the bottom of the file.

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
```diff
@@ -211,10 +216,10 @@ public RestoreCommand(FrameworkName fallbackFramework)
                     Reports.Information.WriteLine(string.Format("Total time {0}ms", sw.ElapsedMilliseconds));
                 }
 
-                if (_installCount > 0)
+                if (summary.InstallCount > 0)
                 {
-                    InformationMessages.GetOrAdd("Installed:", _ => new List<string>()).Add($"{_installCount} package(s) to {packagesDirectory}");
-                    _installCount = 0;
+                    summary.InformationMessages.GetOrAdd("Installed:", _ => new List<string>()).Add($"{summary.InstallCount} package(s) to {packagesDirectory}");
+                    summary.InstallCount = 0;
```
davidfowl: Why do you reset it?

davidfowl:
@BrennanConroy What's the deal with this change?

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
```diff
@@ -85,46 +106,20 @@ public RestoreCommand(FrameworkName fallbackFramework)
                 RestoreDirectories.Add(Directory.GetCurrentDirectory());
             }
 
+            var summary = new SummaryContext();
+
             bool success = true;
             foreach (var dir in RestoreDirectories.Select(Path.GetFullPath).Distinct())
             {
-                success &= await Execute(dir);
+                success &= await Execute(dir, summary);
```
troydai: Could there be a race condition?

troydai: Never mind, saw the `GetOrAdd` extension.

davidfowl:
@BrennanConroy Is this going to make it for beta7?

BrennanConroy:
Yes!

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
BrennanConroy: Removing

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
```diff
@@ -533,6 +509,7 @@ public RestoreCommand(FrameworkName fallbackFramework)
             if (!SkipInstall)
             {
                 await InstallPackages(installItems, packagesDirectory);
+                summary.InstallCount += installItems.Count();
```
davidfowl: `.Count` not `.Count()`

davidfowl:
:shipit: After that small change

BrennanConroy:
f27d11518c78009eb44e6985e343e69efbb430ee

PR #2170, opened 2015-07-01
fix #2168 by updating the lock file version

### LINKED ISSUE #2168: Update Lock File Version to 1
NuGet has updated the version of the project.lock.json to version 1 in https://github.com/NuGet/NuGet3/commit/adc6584261100faf813f7816445ccf853ed7613e - We should do the same to ensure our lock files are compatible.

/cc @muratg for triage. I suggest beta6, it's a trivial change (and I'm happy to do it :)).

anurse:
Fixes #2168 

Pinging pretty much everyone mostly as an FYI. I'll wait until dev opens back up (and of course, for squirrels) before I merge though :).

Nobody should have to change because of this, but DNX will consider any existing lock file with version -9996 out of date and require a re-restore.

/cc @davidfowl @victorhurdugaci @troydai @moozzyk @Tratcher @BrennanConroy @lodejard @Eilon @muratg

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Restore/LockFileFormat.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/LockFileFormat.cs
@@ -16,7 +16,6 @@ namespace Microsoft.Framework.PackageManager
 {
     public class LockFileFormat
     {
-        public const int Version = -9996;
         public const string LockFileName = "project.lock.json";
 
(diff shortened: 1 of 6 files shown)


troydai:
Looks good to me. :shipit: 
Just make sure the old version lock file is invalidated correctly.

anurse:
This went in to dev

Issue #506, opened 2014-08-05
kpm restore does not fail when a dependency does not support the target framework


rowanmiller:
For example, the following project.json says it targets .NET 4.5 and K10 but the EntityFramework.Relational package only supports .NET 4.5.1 (not 4.5). There are no errors during package restore, you just get build errors when you try to build because there is no reference to EntityFramework.Relational.

```JSON
{
  "version": "1.0.0-*",
  "dependencies": {
    "EntityFramework.Relational": "7.0.0-*"
  },
  "frameworks": {
    "net45": {},
    "k10": {}
  }
}
```

BTW, by comparison here is the message generated from NuGet for non-K projects:

```
Install-Package : Could not install package 'EntityFramework.Relational 7.0.0-alpha3-10425'. You are trying to install this package into a project that targets '.NETFramework,Version=v4.5', but the package does not 
contain any assembly references or content files that are compatible with that framework. For more information, contact the package author.
```

ChengTian:
Agree with @davidfowl .

When you find a dependency that doesn't provide the assembly compatible with your app's target framework, it doesn't mean we should always report an error. Think about metadata/tool packages (e.g. `Microsoft.AspNet.Loader.IIS.Interop`) that don't provide any assembly but you still need them as your dependencies.

So the error should only be reported when you actually need to load an assembly but you cannot find it in the package. There is no way for `dnu restore` to tell this.

I'll talk to Tooling team to check whether we can make something show up in VS.

PR #2172, opened 2015-07-01
Mark a nupkg as unresolved when it doesn't provide compatible assembly

ChengTian:
parent https://github.com/aspnet/dnx/issues/506

@davidfowl 

Learned the logic from @anurse . Here is what you get when a nupkg is only compatible with part of `frameworks`:
capture

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Utils/LockFileUtils.cs
+++ b/src/Microsoft.Framework.PackageManager/Utils/LockFileUtils.cs
@@ -13,6 +13,9 @@ namespace Microsoft.Framework.PackageManager.Utils
 {
     internal static class LockFileUtils
     {
+        private static readonly Func<string, object> PlaceholderFileParser =
+            s => string.Equals(s, "_._", StringComparison.Ordinal) ? s : null;
+
         public static LockFileLibrary CreateLockFileLibrary(LockFileLibrary previousLibrary, IPackagePathResolver resolver, IPackage package, string correctedPackageName = null)
         {
             var lockFileLib = new LockFileLibrary();
@@ -399,11 +402,13 @@ namespace Microsoft.Framework.PackageManager.Utils
 
             ContentPropertyDefinition _assembly = new ContentPropertyDefinition
             {
+                Parser = PlaceholderFileParser,
                 FileExtensions = { ".dll", ".exe", ".winmd" }
             };
 
             ContentPropertyDefinition _dynamicLibrary = new ContentPropertyDefinition
             {
+                Parser = PlaceholderFileParser,
                 FileExtensions = { ".dll", ".dylib", ".so" }
             };
 
--- a/src/Microsoft.Framework.Runtime/DependencyManagement/NuGetDependencyResolver.cs
+++ b/src/Microsoft.Framework.Runtime/DependencyManagement/NuGetDependencyResolver.cs
@@ -94,7 +94,8 @@ namespace Microsoft.Framework.Runtime
             if (package != null)
             {
                 IEnumerable<LibraryDependency> dependencies;
-                bool resolved = true;
+                var resolved = true;
+                var compatible = true;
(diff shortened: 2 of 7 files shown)


BrennanConroy:
Fix tests!

davidfowl:
We need to improve the error message

davidfowl:
We need to also test meta packages to make sure we didn't break anything.

src/Microsoft.Framework.Runtime/DependencyManagement/NuGetDependencyResolver.cs
```diff
@@ -106,6 +106,19 @@ public LibraryDescription GetDescription(LibraryRange libraryRange, FrameworkNam
                         resolved = false;
                         dependencies = Enumerable.Empty<LibraryDependency>();
                     }
+
+                    // If a NuGet dependency is supposed to provide assemblies but there is no assembly compatible with
+                    // current target framework, we should mark this dependency as unresolved
+                    if (targetLibrary != null)
+                    {
+                        var containsAssembly = package.LockFileLibrary.Files
+                            .Any(x => x.StartsWith($"ref{Path.DirectorySeparatorChar}") ||
```
anurse: I think the LockFile path format is normalized to `/`, but I may be wrong.

davidfowl: It is

ChengTian: It is in lock file on disk. We convert `/` to `Path.Separator` when we read a lock file into memory:
https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Framework.PackageManager/Restore/LockFileFormat.cs#L367

I agree that we need really good tests. But the code here is correct.

anurse:
Looks good to me, but as discussed in comments, we should get some really good tests :).

I think some unit tests for just this compatibility logic would be useful. The unit tests can just construct the Lock File objects manually and test compatibility. We should try to cover all the possible scenarios as much as possible.

ChengTian:
Added unit tests

davidfowl:
I actually am not a fan of just marking it unresolved. I think we need another diagnostic to say what went wrong.

ChengTian:
New message after improving diagnostics:
capture

ChengTian:
@davidfowl , added logic to handle placeholder packages like `System.Runtime.4.0.20-beta-23104`.

src/Microsoft.Framework.PackageManager/Utils/LockFileUtils.cs
```diff
@@ -173,6 +175,17 @@ public static LockFileTargetLibrary CreateLockFileTargetLibrary(LockFileLibrary
                 }
             }
 
+            if (!hasCompatibleDependencySet &&
```
davidfowl: I wouldn't do this. I think you just need to detect _._ as a dll and then remove it from compile items if it is _._ /cc @anurse

anurse: To be clear, @davidfowl means `_._` (markdown formatting got in the way), but yes, I would actually just detect  the case where `_._` is the sole entry in any of the lists and remove it, resulting in a zero-length array.

ChengTian:
@davidfowl , please check the latest commits

Talked to @anurse and we think it's better to simply accept `_._` as assembly in lock file and filter them out before compilation. This makes the change minimal and also makes it consistent with the logic in NuGet.

ChengTian:
One more advantage of current approach is that, as @anurse confirmed, this is the exactly same approach used in NuGet.

ChengTian:
@davidfowl , added a commit to filter out `_._` when resolving runtime loadable assemblies.

davidfowl:
:shipit:

PR #2173, opened 2015-07-01
fix #1683 by preferring listed packages

### LINKED ISSUE #1683: DNU restore should still restore unlisted packages

anurse:
...but accepting unlisted packages if there are no matches

Clone of #2156 which was already signed off on. I reverted it temporarily to fix the build and I can't seem to re-open the PR.

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Restore/IWalkProvider.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/IWalkProvider.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Framework.PackageManager
     {
         bool IsHttp { get; }
 
-        Task<WalkProviderMatch> FindLibrary(LibraryRange libraryRange, FrameworkName targetFramework);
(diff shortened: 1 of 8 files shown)


davidfowl:
Why is the build failing?

anurse:
I really don't know, it seems unrelated but it repro'd on TeamCity so I reverted. I'll be investigating today.

PR #2178, opened 2015-07-02
Remove a forgotten console writeline

victorhurdugaci:
@davidfowl


davidfowl:
:shipit: Make sure there's nothing else like this.

PR #2183, opened 2015-07-02
Adding fallback to default servicing location (Fixes #2082)

### LINKED ISSUE #2082: Servicing should fall back to program files if no redirects found in under DNX_SERVICING
Currently for servicing nupkgs we check if %DNX_SERVICING% variable is defined and if it is we use it and quit. Instead, if no redirects were found in the index file in the DNX_SERVICING, we should fall back and check the index files located under %PROGRAMFILES(X86)% to prevent the situation where servicing is disabled just because %DNX_SERVICING% variable exists.

moozzyk:
Fixing a bug where servicing would not work on 32-bit OSes because we never assigned the value of the `PROGRAMFILES` environment variable

### DIFF
--- a/src/Microsoft.Framework.Runtime/Servicing/ServicingTable.cs
+++ b/src/Microsoft.Framework.Runtime/Servicing/ServicingTable.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Threading;
 using NuGet;
@@ -10,7 +11,7 @@ namespace Microsoft.Framework.Runtime.Servicing
 {
     public static class ServicingTable
     {
-        private static ServicingIndex _index;
+        private static List<ServicingIndex> _index;
         private static bool _indexInitialized;
         private static object _indexSync;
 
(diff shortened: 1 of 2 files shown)


moozzyk:
@victorhurdugaci @muratg

src/Microsoft.Framework.Runtime/Servicing/ServicingTable.cs
```diff
@@ -20,39 +22,62 @@ public static class ServicingTable
...
 
-                    if (string.IsNullOrEmpty(servicingRoot))
+                foreach (var servicingRoot in GetServicingRoots())
+                {
+                    var index = new ServicingIndex();
+                    if (servicingRoot != null)
                     {
-                        // Nothing to do, we don't have Program Files. Just return the uninitialized index.
-                        return index;
+                        index.Initialize(servicingRoot);
```
victorhurdugaci: Not because of this change but I realized that we have a bug here. If the path that you pass through the environment variable is not a valid path, the `ServicingIndex` throws. Here: https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Framework.Runtime/Servicing/ServicingIndex.cs#L19

We should check that's a valid path.

victorhurdugaci:
Except for my comment above, the change looks good

src/Microsoft.Framework.Runtime/Servicing/ServicingTable.cs
moozzyk: Will look into it.

src/Microsoft.Framework.Runtime/Servicing/ServicingTable.cs
```diff
@@ -5,12 +5,14 @@
 using System.IO;
 using System.Threading;
 using NuGet;
+using System.Diagnostics;
```
BrennanConroy: Order

src/Microsoft.Framework.Runtime/Servicing/ServicingTable.cs
moozzyk: Added a check:
```
if (servicingRoot != null && servicingRoot.IndexOfAny(Path.GetInvalidPathChars()) == -1)
```

src/Microsoft.Framework.Runtime/Servicing/ServicingTable.cs
moozzyk: Done. `System.Diagnostics` was not actually needed anymore.

PR #2186, opened 2015-07-02
Loading coreclr.dll from the full path (fixes #1258)

### LINKED ISSUE #1258: cleanup kre.coreclr.cpp
AMD64 check needs to be removed: https://github.com/aspnet/XRE/blob/dev/src/kre.coreclr/kre.coreclr.cpp#L232

Always pass full path to coreclr don't try and use relative paths or search around.

/@blowdart

moozzyk:


moozzyk:
@glennc @davidfowl

src/dnx.coreclr/dnx.coreclr.cpp
```diff
@@ -115,7 +115,7 @@ HMODULE LoadCoreClr(dnx::trace_writer& trace_writer)
     }
     else
     {
-        coreclr_module = LoadLibraryExW(L"coreclr.dll", NULL, 0);
+        coreclr_module = LoadLibraryExW(dnx::utils::path_combine(GetModuleDirectory(nullptr), L"coreclr.dll").c_str(), NULL, 0);
```
davidfowl: Do we call GetModuleDirectory anywhere else? If we do can we pass it into the method?

moozzyk: Yes, there is one more place where we do it. I can refactor the code to have just one call and pass the parameter around if this is a (performance/startup time?) concern.

davidfowl: Cleaner

davidfowl:
:shipit:

PR #2198, opened 2015-07-07
Fixing how we read environment variables in native code

moozzyk:
We did not check if the buffer was big enough to hold the value. If it wouldn't be *`the contents of lpBuffer are undefined`* so we would crash or have other interesting side effects (like load dlls from wrong location).


davidfowl:
Why did appveyor fail?

moozzyk:
Because of Greece probably... 

(the build was cancelled)

moozzyk:
:watch:

BrennanConroy:
:ship: :it:

Issue #2202, opened 2015-07-07
dnu sources get package, when git is not installed on the machine or not in the %PATH% it just says "file not found"


abpiskunov:
Its not clear what file is not found. Also it would be nice to add git to the path automatically if it is not there since command relies on it..

victorhurdugaci:
There is nothing that `dnu sources` can do to add `git` to the path. That's something that the setup, or someone else, should do.

abpiskunov:
by someone else you mean dnvm? It would be nice if there would be less small things users would need to do manually ...

victorhurdugaci:
No, I don't think `dnvm` should be responsible for installing the `git` client. That would be pretty bad actually because today we support only `git` but in the future we might support other source control providers. The user should install that. 

When you `dnu pack`, if you don't have `git` you get a warning. I just forgot to add a similar check for `dnu sources`

abpiskunov:
ok, sounds good.

PR #2203, opened 2015-07-07
Report if git is not installed

victorhurdugaci:
Fixes https://github.com/aspnet/dnx/issues/2202

@davidfowl @abpiskunov


davidfowl:
:shipit:

PR #2212, opened 2015-07-08
Returning error codes instead of exceptions (Fixes #1825)

### LINKED ISSUE #1825: [CoreCLR] RuntimeBootstrapManager.Execute unhandled exceptions
Exceptions not handled here: https://github.com/aspnet/dnx/blob/dev/src/dnx.host/RuntimeBootstrapper.cs#L35 will be thrown out to the native CoreCLR runtime and their details lost. The CoreCLR does not support marshalling managed exceptions. We need to catch them and return error codes instead. See https://github.com/aspnet/Helios/issues/151#issuecomment-100377549

moozzyk:
Exceptions thrown in managed code cannot be handled in native code and are causing hard crashes. So, if no one handled an exception we will catch it and convert into an error code. We do it in methods that are invoked from native code.


Tratcher:
:shipit:

Issue #1601, opened 2015-04-06
Remove System.Private.* from the runtime


muratg:
Essentially reverting https://github.com/aspnet/dnx/issues/1598 work being done for Beta4.

muratg:
@pranavkm @troydai should we do this now?

muratg:
@pranavkm?

pranavkm:
Yeah, seems like this should be safe to remove now. I'll give this a try.

davidfowl:
Don't remove anything.

muratg:
@davidfowl never or in beta6?

pranavkm:
We should be good to remove these since we're consuming the newer packages

davidfowl:
@muratg Can we assign somebody else to do this?

BrennanConroy:
677d6c3c45a92171cf92a68ee6eefd55d4fe91a3

PR #1870, opened 2015-05-14
Include System.Runtime.Serialization.Primitives in runtime

troydai:
Since the removal of `Newtonsoft.Json` dependency, the `System.Runtime.Serialization.Primitives` is not included. 

/cc @davidfowl

PR #2213, opened 2015-07-08
#1601 Remove System.Private

BrennanConroy:
@davidfowl

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -574,13 +574,7 @@ var CAN_BUILD_ONECORE = '${Directory.Exists(WIN10_SDK_LIB) && Directory.GetFiles
             // Copy required CoreCLR assemblies
             if (target.Flavor == "coreclr")
             {
-                var extraBinaries = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
-                {
-                    "System.Private.DataContractSerialization.dll",
-                    "System.Private.Networking.dll",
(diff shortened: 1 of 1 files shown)


makefile.shade
```diff
@@ -574,9 +574,6 @@ var CAN_BUILD_ONECORE = '${Directory.Exists(WIN10_SDK_LIB) && Directory.GetFiles
             {
                 var extraBinaries = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
                 {
-                    "System.Private.DataContractSerialization.dll",
-                    "System.Private.Networking.dll",
-                    "System.Private.ServiceModel.dll",
                     "System.Runtime.Serialization.Primitives.dll"
```
davidfowl: Do we need this?

BrennanConroy: Not sure, we added it because we broke some projects when we removed NewtonSoft: #1870

BrennanConroy: Just built UserSecrets with the Primitives removed and it failed.

pranavkm: Should we follow up with the serialization team? My guess is we don't want any of these binaries in the runtime if they aren't necessary.

davidfowl: Ye we should nuke this.

davidfowl:
:shipit:

davidfowl:
@BrennanConroy What's the deal with this?

Issue #2192, opened 2015-07-06
BUG: "dnu commands install" fails to find .sha512


giggio:
I have a project.json that depends on `FluentMigrator` version `1.5.1.0`. This package defines its version as `1.5.1.0`. If I try to install my package (FluentMigrator.Runner.Aspnet):

````
dnu commands install FluentMigrator.Runner.Aspnet --no-cache
````

It fails because it can't find the `.sha512` file:

````
Could not find a part of the path 'C:\Users\username\.dnx\bin\packages\FluentMigrator\1.5.1.0\FluentMigrator.1.5.1.0.nupkg.sha512'.
````

It is looking for the `1.5.1.0` directory, but `dnu` is saving it without the last `.0`, so I have on disk only the `$HOME\.dnx\bin\packages\FluentMigrator\1.5.1` directory. And installation fails.

**Comment by giggio** [comment]
The problem also happens on `dnu restore` for the package.

**Comment by davidfowl** [comment]
The problem is that these packages have inconsistent nuspec versions. 

@muratg @ChengTian @anurse  we should fix this ASAP. A bunch of packages are running into this. We need to resolve the inconsistencies between package version in the nuspec and elsewhere.

**Comment by muratg** [comment]
:+1:

**Comment by troydai** [comment]
@giggio Could you let us know what's version of the DNX you're using?

I fail to repo this issue with either beta5 or beta6 latest with this command (targeting alpha3 since you work around the problem in alpha4):
```
dnu commands install FluentMigrator.Runner.AspNet 1.0.0-alpha3 --no-cache
```

davidfowl:
This absolutely repros. I can repro it myself without a commands install.

giggio:
@troydai that is because I have forked the original FluentMigrator.Runner and uploaded a correct version to Nuget (which I plan to remove as soon as the original is fixed).

troydai:
@giggio I can repo the issue now. Working on it.

troydai:
Fixed in cae8aab5b38e0039f03eb87a8d994e3035179206

giggio:
@davidfowl @troydai Any workarounds at this point to work with these packages and `beta5`?

PR #2220, opened 2015-07-09
Use normalized semantic version in package path/name

troydai:
Addressing #2192 

Still running DNX testing at this point.

### DIFF
--- a/src/Microsoft.Framework.Runtime/NuGet/Packages/DefaultPackagePathResolver.cs
+++ b/src/Microsoft.Framework.Runtime/NuGet/Packages/DefaultPackagePathResolver.cs
@@ -56,7 +56,7 @@ namespace NuGet
         public string GetHashPath(string packageId, SemanticVersion version)
         {
             return Path.Combine(GetInstallPath(packageId, version),
-                                $"{packageId}.{version}{Constants.HashFileExtension}");
+                                $"{packageId}.{version.GetNormalizedVersionString()}{Constants.HashFileExtension}");
         }
 
         public virtual string GetPackageDirectory(string packageId, SemanticVersion version)
@@ -64,7 +64,7 @@ namespace NuGet
             string directory = packageId;
             if (_useSideBySidePaths)
             {
-                directory = Path.Combine(directory, version.ToString());
+                directory = Path.Combine(directory, version.GetNormalizedVersionString());
             }
             return directory;
         }
@@ -74,7 +74,7 @@ namespace NuGet
             string fileNameBase = packageId;
             if (_useSideBySidePaths)
             {
-                fileNameBase += "." + version;
+                fileNameBase += "." + version.GetNormalizedVersionString();
             }
             return fileNameBase + Constants.PackageExtension;
         }
--- a/src/Microsoft.Framework.Runtime/NuGet/SemanticVersion.cs
+++ b/src/Microsoft.Framework.Runtime/NuGet/SemanticVersion.cs
@@ -79,6 +79,15 @@ namespace NuGet
 
         public string OriginalString { get; }
 
+        public string GetNormalizedVersionString()
+        {
+            var revision = Version.Revision > 0 ? ("." + Version.Revision.ToString(CultureInfo.InvariantCulture)) : string.Empty;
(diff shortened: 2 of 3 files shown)


src/Microsoft.Framework.Runtime/NuGet/SemanticVersion.cs
```diff
@@ -79,6 +79,21 @@ public string SpecialVersion
 
         public string OriginalString { get; }
 
+        public string GetNormalizedVersionString()
```
davidfowl: This was copied from NuGet right?

src/Microsoft.Framework.Runtime/NuGet/Packages/DefaultPackagePathResolver.cs
```diff
@@ -64,7 +64,7 @@ public virtual string GetPackageDirectory(string packageId, SemanticVersion vers
             string directory = packageId;
             if (_useSideBySidePaths)
             {
-                directory = Path.Combine(directory, version.ToString());
```
troydai: I'll try to eliminate the usage of `ToString` in scenarios other than debugging.

src/Microsoft.Framework.Runtime/NuGet/SemanticVersion.cs
```diff
@@ -79,6 +79,21 @@ public string SpecialVersion
...
+        {
+            var revision = Version.Revision > 0 ? ("." + Version.Revision.ToString(CultureInfo.InvariantCulture)) : string.Empty;
+            var specialVer = !string.IsNullOrEmpty(SpecialVersion) ? ("-" + SpecialVersion) : string.Empty;
+
+            // SemanticVersion normalizes the missing components to 0.
+            return string.Format(CultureInfo.InvariantCulture,
```
davidfowl: Any reason you don't use string interpolation here?

src/Microsoft.Framework.Runtime/NuGet/SemanticVersion.cs
troydai: Yes, it is.

src/Microsoft.Framework.Runtime/NuGet/SemanticVersion.cs
troydai: I believe we can. No particular reason not to.

davidfowl:
Where is the change that always writes the normalized version folder? Does this code affect both reads and writes? Does it work with v2 and v3 feed formats? Local and remote?

troydai:
_Where is the change that always writes the normalized version folder?_
-It is done implicitly. The `NuGetPackageUtils.InstallFromStream` resolve the target path using updatede code. https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Framework.PackageManager/Utils/NuGetPackageUtils.cs#L27. 

_Does this code affect both reads and writes?_
- Base on the code I read so far, yes. Places where NuGet package paths are required, `IPackagePathResolver` is used to resolve the paths. I'll consult @ChengTian and @anurse tomorrow.

_Does it work with v2 and v3 feed formats?_
- Yes, in terms of the scenarios of #2192 . There may be broader scenarios be impacted, I'll seek advices from @ChengTian and @anurse and test tomorrow.

_ Local and remote?_
- As far as I can see, `DefaultPackagePathResolver` is the where the paths are resolved. I'll take a deeper look tomorrow to see if there is other functions resolving the path. If so these functions should be replaced by `IPackagePathResolver`

davidfowl:
Sounds great! I think this fix will work well then. Please add a basic test

anurse:
Looks good to me, agreed a test would be good.

ChengTian:
Yes,all package install paths should be resolved with `DefaultPackagePathResolver`. Looks good to me.

troydai:
Adding a few tests for `SemanticVersion`

davidfowl:
Add a restore test.

troydai:
We don't have test infrastructure for and actual dun restore test yet. Size of the change seems not to fit in this PR. I file this issue to track the work: https://github.com/aspnet/dnx/issues/2231.

troydai:
Ping :watch:

davidfowl:
:shipit:

PR #2232, opened 2015-07-09
"dnu restore" checks corrupted packages

ChengTian:
parent https://github.com/aspnet/dnx/issues/1243

Without this change, when `{id}.nuspec` is missing, `dnu restore` fails instead of reinstalling.

@davidfowl 
@BrennanConroy (I modified some of your tests)

### DIFF
--- a/src/Microsoft.Framework.Runtime/NuGet/Repositories/PackageRepository.cs
+++ b/src/Microsoft.Framework.Runtime/NuGet/Repositories/PackageRepository.cs
@@ -108,6 +108,14 @@ namespace NuGet
                         continue;
                     }
 
+                    var manifestFilePath = _repositoryRoot.GetFiles(versionDir, "*" + Constants.ManifestExtension)
+                        .FirstOrDefault();
+                    if (string.IsNullOrEmpty(manifestFilePath))
+                    {
+                        // This is a corrupted packages because {id}.nupsec is missing
+                        continue;
+                    }
+
                     if (CheckHashFile && !_repositoryRoot.GetFiles(versionDir, "*" + Constants.HashFileExtension).Any())
                     {
                         // Writing the marker file is the last operation performed by NuGetPackageUtils.InstallFromStream. We'll use the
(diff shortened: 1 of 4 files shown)


test/Microsoft.Framework.PackageManager.FunctionalTests/DnuRestoreTests.cs
```diff
@@ -6,6 +6,7 @@
 using Microsoft.Framework.CommonTestUtils;
 using Microsoft.Framework.PackageManager.FunctionalTests;
 using Xunit;
+using NuGet;
```
BrennanConroy: Order

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuRestoreTests.cs
```diff
@@ -146,5 +147,45 @@ goto argumentStart
...
+                var exitCode = DnuTestUtils.ExecDnu(
+                    runtimeHomePath,
+                    subcommand: "restore",
+                    arguments: $"{projectDir} -s {_fixture.PackageSource} --packages {packagesDir}",
+                    stdOut: out stdOut,
+                    stdErr: out stdErr);
+
+                Assert.Equal(0, exitCode);
+                Assert.Empty(stdErr);
+                Assert.Contains($"Installing alpha.0.1.0", stdOut);
```
BrennanConroy: Check nuspec is back?

ChengTian: Good point

ChengTian:
Thanks for the feedback. Fixed in the latest commit.

src/Microsoft.Framework.Runtime/NuGet/Repositories/PackageRepository.cs
```diff
@@ -108,6 +108,14 @@ public IEnumerable<PackageInfo> FindPackagesById(string packageId)
                         continue;
                     }
 
+                    var manifestFilePath = _repositoryRoot.GetFiles(versionDir, "*" + Constants.ManifestExtension)
```
davidfowl: Do we really need this check at runtime?

ChengTian: After introducing lock file, this logic is never executed during runtime of your apps. It only happens during `dnu restore`.

davidfowl:
:shipit:

PR #2234, opened 2015-07-09
fix #2117 by adding config information to feeds list

### LINKED ISSUE #2117: Memory leak if native images requested for CoreCLR on non-unix
If native images for TPAs are requested on non-windows we will unnecessarily allocate and fill an array for TPA names and then exit here:
https://github.com/aspnet/dnx/blob/dev/src/dnx.coreclr.unix/dnx.coreclr.cpp#L82
without ever releasing the memory.

### LINKED ISSUE #2177: Show NuGet.Config files in use in dnu feeds list command
https://github.com/aspnet/dnx/commit/36eab9fd3ae7de1890118c80dc3d82b2c98f7fa4#commitcomment-11968013

We should show the config files we used when calculating the active feeds.

anurse:
Fixes #2177 

The list of config files is easy, but showing the config file used for each individual source required some changes to our forked NuGet code. It wasn't as invasive as I'd feared, but it would be something to be aware of when we integrate NuGetv3 Code here.

/cc @davidfowl @kichalla 

image

### DIFF
--- a/src/Microsoft.Framework.PackageManager/ConsoleCommands/FeedsConsoleCommand.cs
+++ b/src/Microsoft.Framework.PackageManager/ConsoleCommands/FeedsConsoleCommand.cs
@@ -33,7 +33,7 @@ namespace Microsoft.Framework.PackageManager
 
                 c.OnExecute(() =>
                 {
-                    var command = new ListSourcesCommand(
+                    var command = new ListFeedsCommand(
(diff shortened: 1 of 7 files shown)


kichalla:
:shipit:

davidfowl:
Nice!

davidfowl:
:shipit:

PR #2237, opened 2015-07-10
Adding gtest and a project for tests for native code

moozzyk:


troydai:
I try to review, but the browser just can't catch up with 200 files.

davidfowl:
+1

moozzyk:
Yeah, I saw that but unfortunately there is not much I can do about that. The change is limited to gtest and a test project with just one "test" that only passes (the test is empty). Both are plugged in to the solution and I made sure that they compile fine (from VS and command line with MSBuild). 
Real tests (i have a prototype where I have a test testing real code in a separate branch), integration with makefile, integration with CI will come later. 
I think you need to pull the branch and use diff/codeflow if you really want to review this (but I am not planning to address comments related to gtest)

davidfowl:
You may as well merge this because it's too big to review :smile:

PR #2244, opened 2015-07-10
Adding a test testing actual dnx code.

moozzyk:
- seperating out the entry point (DnxMain) to avoid having two `main` methods in tests
- moving string compare to a common location since it's now being used in multiple files
- moving CALL_APPLICATION_MAIN_DATA to dnx common - it was already being used by multiple projects
- diabling precompiled headers in main since we are including file from other projects which include different "stdafx.h" file

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -358,6 +358,7 @@ var CAN_BUILD_ONECORE = '${Directory.Exists(WIN10_SDK_LIB) && Directory.GetFiles
     @{
         var sourceFiles = new string[] 
         {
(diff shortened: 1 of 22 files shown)


davidfowl:
@moozzyk Are you planning to move more of this code to STL?

:shipit:

moozzyk:
@davidfowl - some of it. I concluded that in parameter expansion it does not make too much sense to move to stl - source is buffers, target is buffers and we don't do much operations in between. Still needs clean up though which I am gradually doing when touching this code.

PR #2247, opened 2015-07-10
Fixing a bug where appbase was not expanded correctly

moozzyk:
Before when looking for --appbase we would look at all the parameters instead of stopping after reaching the first non-bootstrapper parameter. As a result if --appbase appeared after the first non-bootstrapper parameter we would not expand the path and the app would fail.


moozzyk:
This is not the right fix - it would break if `--appbase` was after a parameter with an argument (e.g. `--lib`).

Issue #2238, opened 2015-07-10
Add `--framework` flag to `dnu publish` command


friism:
This would be useful to me, at least: I'm building for a known framework, and I don't want to spend time and space creating dlls for other frameworks

anurse:
Makes some sense to me. Similarly, if for some reason the app has both `dnx46` and `dnx451` in the project.json, this could allow the user to choose the version we will write in to `web.config`.

Thoughts @davidfowl ? It seems like it's basically just the same as letting you constrain `dnu build` by framework.

davidfowl:
Was added

PR #2248, opened 2015-07-10
fix #2246 by using the best match from the project when publishing

### LINKED ISSUE #2246: Support publishing any 'dnx' framework with a CLR runtime
Right now, if you `dnu publish --runtime [some CLR variant]` and your project has `dnx452` and `dnx46` as the only target framework, you will get this error:

```
'DNX,Version=v4.5.1' is not a target framework of the project being published
```

We should change this code so that publish will accept `dnx452` and `dnx46` as well. I almost have a fix ready for this. It is blocking some Helios work.

anurse:
Fixes #2246 

Now, we scan the project.json to find the best matching framework for a runtime rather than just assuming it will be `dnx451`. We take the highest matching framework. This means that if the project has `dnx46` and `dnx451`, we'll publish as 4.6. If the user specifies a framework version in web.config we'll honour it, and adding #2238 could help with this.

While I was here I also noticed some code that only ran when `frameworks` was missing for a project, which is unsupported. So I added a better error instead of assuming a value.

/cc @ChengTian @davidfowl

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Publish/DependencyContext.cs
+++ b/src/Microsoft.Framework.PackageManager/Publish/DependencyContext.cs
@@ -1,10 +1,13 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.Versioning;
 using Microsoft.Framework.Runtime;
 using Microsoft.Framework.Runtime.Caching;
+using Microsoft.Framework.Runtime.Common.Impl;
 using NuGet;
 
 namespace Microsoft.Framework.PackageManager.Publish
@@ -48,7 +51,7 @@ namespace Microsoft.Framework.PackageManager.Publish
             PackageAssemblies = NuGetDependencyResolver.PackageAssemblyLookup.Values.ToLookup(a => a.Library.Identity.Name);
         }
 
-        public static FrameworkName GetFrameworkNameForRuntime(string runtime)
+        public static FrameworkName SelectFrameworkNameForRuntime(IEnumerable<FrameworkName> availableFrameworks, string runtime)
         {
             var parts = runtime.Split(new[] { '.' }, 2);
             if (parts.Length != 2)
@@ -68,9 +71,13 @@ namespace Microsoft.Framework.PackageManager.Publish
             {
                 case "mono":
                 case "clr":
-                    return VersionUtility.ParseFrameworkName("dnx451");
+                    // CLR currently supports anything <= dnx46
+                    return availableFrameworks
+                        .Where(fx => fx.Identifier.Equals(FrameworkNames.LongNames.Dnx, StringComparison.Ordinal) && fx.Version <= new Version(4, 6))
+                        .OrderByDescending(fx => fx.Version)
+                        .FirstOrDefault();
                 case "coreclr":
-                    return VersionUtility.ParseFrameworkName("dnxcore50");
+                    return availableFrameworks.FirstOrDefault(fx => fx.Equals(VersionUtility.ParseFrameworkName("dnxcore50")));
             }
             return null;
         }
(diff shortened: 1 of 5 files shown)


src/Microsoft.Framework.PackageManager/Publish/DependencyContext.cs
```diff
@@ -68,9 +71,13 @@ public static FrameworkName GetFrameworkNameForRuntime(string runtime)
             {
                 case "mono":
                 case "clr":
-                    return VersionUtility.ParseFrameworkName("dnx451");
+                    // CLR currently supports anything <= dnx46
+                    return availableFrameworks
```
davidfowl: How many different places do we have this logic

src/Microsoft.Framework.PackageManager/Publish/PublishManager.cs
```diff
@@ -112,7 +113,19 @@ public bool Publish()
...
             {
-                var frameworkName = DependencyContext.GetFrameworkNameForRuntime(Path.GetFileName(runtime));
+                var frameworkName = DependencyContext.SelectFrameworkNameForRuntime(
+                    project.GetTargetFrameworks().Select(x => x.FrameworkName), 
+                    Path.GetFileName(runtime));
+
+                if (frameworkName == null)
+                {
+                    _options.Reports.Error.WriteLine(
+                        string.Format("The project being published does not support the runtime '{0}'",
```
ChengTian: We don't need `string.Format` here, `IReport.WriteLine` supports formatted output.

src/Microsoft.Framework.PackageManager/Publish/DependencyContext.cs
anurse: One. Right here. But I get your point. I think it might be necessary to centralize the logic that determines the set of allowed frameworks for a specific DNX.

src/Microsoft.Framework.PackageManager/Publish/PublishManager.cs
anurse: :+1:

src/Microsoft.Framework.PackageManager/Publish/PublishManager.cs
```diff
@@ -112,7 +113,19 @@ public bool Publish()
 
             foreach (var runtime in _options.Runtimes)
             {
-                var frameworkName = DependencyContext.GetFrameworkNameForRuntime(Path.GetFileName(runtime));
+                var frameworkName = DependencyContext.SelectFrameworkNameForRuntime(
+                    project.GetTargetFrameworks().Select(x => x.FrameworkName), 
+                    Path.GetFileName(runtime));
```
ChengTian: Can we fix a small bug here BTW? `runtime` can be a path to a runtime folder like
```
dnu publish --runtime C:\Users\wwa\.dnx\runtimes\dnx-clr-win-x86.1.0.0-beta6-12208\
```
Pay attention to the trailing slash : )

test/Microsoft.Framework.PackageManager.Tests/project.json
```diff
@@ -2,7 +2,6 @@
     "dependencies": {
         "Microsoft.Framework.CommonTestUtils": "1.0.0-*",
         "Microsoft.Framework.PackageManager": "1.0.0-*",
-        "Microsoft.Framework.Runtime.Sources": "1.0.0-*",
```
anurse: I'm not sure why my change revealed this but this was conflicting with the versions used in Microsoft.Framework.PackageManager which were visible because that has InternalsVisibleTo this assembly

anurse:
Last call now that builds are passing.

src/Microsoft.Framework.PackageManager/Publish/PublishManager.cs
```diff
@@ -112,7 +113,23 @@ public bool Publish()
 
             foreach (var runtime in _options.Runtimes)
             {
-                var frameworkName = DependencyContext.GetFrameworkNameForRuntime(Path.GetFileName(runtime));
+                // Calculate the runtime name by taking the last path segment (since it could be a path),
+                // but first strip off any trailing '\' or '/' in case the path provided was something like
+                //  "C:\Foo\Bar\dnx-clr-win-x64...\"
+                var runtimeName = Path.GetFileName(runtime.TrimEnd(Path.DirectorySeparatorChar));
```
ChengTian: I would suggest `new DirectoryInfo(runtime).Name` instead of this. It is possible that we are on Windows but a user provides a path with trailing forward slash.

ChengTian:
:shipit: after reaction to my last comment above.

Issue #2240, opened 2015-07-10
Regression: DNU restore tries to load the current project from the package cache


borgdylan:
As of the latest DNX, running ```dnu restore``` requires the nupkg for the current project to be in the package cache, something that looks pretty buggy.
```
dylan@ubuntu-server:/var/www/Code/dylannet/vnext/dylan.NET.K$ dnu restore
Microsoft .NET Development Utility Mono-x86-1.0.0-beta6-12207

Restoring packages for /var/www/Code/dylannet/vnext/dylan.NET.K/project.json
  CACHE http://www.myget.org/F/aspnetwebstacknightly/
  CACHE https://nuget.org/api/v2/curated-feeds/dotnetframework/
  CACHE https://www.myget.org/F/aspnetvnext/api/v2/
  CACHE  https://api.nuget.org/v3/index.json
  CACHE https://www.myget.org/F/npgsql-unstable/api/v2/
  CACHE http://www.myget.org/F/aspnetwebstacknightly/FindPackagesById()?id='Microsoft.Framework.Runtime.Abstractions'
  CACHE https://nuget.org/api/v2/curated-feeds/dotnetframework/FindPackagesById()?id='Microsoft.Framework.Runtime.Abstractions'
  CACHE https://www.myget.org/F/aspnetvnext/api/v2/FindPackagesById()?id='Microsoft.Framework.Runtime.Abstractions'
  CACHE https://nugetapiapps.blob.core.windows.net/ngx/microsoft.framework.runtime.abstractions/index.json
  CACHE https://www.myget.org/F/npgsql-unstable/api/v2/FindPackagesById()?id='Microsoft.Framework.Runtime.Abstractions'
Writing lock file /var/www/Code/dylannet/vnext/dylan.NET.K/project.lock.json
----------
System.IO.DirectoryNotFoundException: Could not find a part of the path "/home/dylan/.dnx/packages/dylan.NET.K/1.3.6/dylan.NET.K.1.3.6.nupkg.sha512".
  at System.IO.FileStream..ctor (System.String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, Boolean anonymous, FileOptions options) <0xb5380b20 + 0x00217> in <filename unknown>:0 
[...]

PR #2251, opened 2015-07-12
Ignore pacakge with unnormalized version

troydai:
Intend to address #2240.

It just test the water. I'll add more test cases.

### DIFF
--- a/src/Microsoft.Framework.Runtime/NuGet/Repositories/PackageRepository.cs
+++ b/src/Microsoft.Framework.Runtime/NuGet/Repositories/PackageRepository.cs
@@ -108,6 +108,15 @@ namespace NuGet
                         continue;
                     }
 
+                    if (!version.IsOriginalStringNormalized())
+                    {
+                        // For a non-http match, if the OriginalVersion string is not normalized that means name of the folder which contains
+                        // the package is not a normalized string. It will cause trouble for file searching in later stage. By invalidating this 
+                        // match, it ensures the package will be reinstalled under a correct folder. This change ensures a package installed 
+                        // by older version of DNX won't prevent new DNX to install correct package.
+                        continue;
+                    }
+
                     var manifestFilePath = _repositoryRoot.GetFiles(versionDir, "*" + Constants.ManifestExtension)
                         .FirstOrDefault();
                     if (string.IsNullOrEmpty(manifestFilePath))
--- a/test/Microsoft.Framework.PackageManager.FunctionalTests/DnuRestoreTests.cs
+++ b/test/Microsoft.Framework.PackageManager.FunctionalTests/DnuRestoreTests.cs
@@ -189,5 +189,51 @@ $@"{{
                 Assert.True(File.Exists(nuspecPath));
             }
         }
+
+        [Theory]
+        [MemberData(nameof(RuntimeComponents))]
+        public void DnuRestore_ReinstallsPackageWithNormalizedVersion(string flavor, string os, string architecture)
+        {
+            var runtimeHomePath = _fixture.GetRuntimeHomeDir(flavor, os, architecture);
+            using (var tempDir = new DisposableDir())
+            {
+                var projectDir = Path.Combine(tempDir, "project");
+                var packagesDir = Path.Combine(tempDir, "packages");
+                var projectJson = Path.Combine(projectDir, Runtime.Project.ProjectFileName);
+
+                Directory.CreateDirectory(projectDir);
+                File.WriteAllText(projectJson, @"
+{
+  ""dependencies"": {
+    ""alpha"": ""0.1.0""
+  }
+}");
+                DnuTestUtils.ExecDnu(
+                    runtimeHomePath,
+                    subcommand: "restore",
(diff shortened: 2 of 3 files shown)


davidfowl:
Can't we do this check here: https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Framework.Runtime/NuGet/Repositories/PackageRepository.cs#L106

troydai:
Yes, change in the PackageRepository was the first though. But then we agreed on Friday that we'd better limit the scope of this change to the dnu restore only. There are more things depends on PackageRepository.

davidfowl:
@troydai I prefer that change. It should only affect restore because PackageRepository has 2 code paths. With and without the lock file.

troydai:
All right.

troydai:
@davidfowl better?

davidfowl:
@troydai Much better! The comment needs some work

davidfowl:
Can you add a test? We do have restore tests ping @ChengTian

troydai:
Yes, tests will be added. I'll work with @ChengTian

src/Microsoft.Framework.Runtime/NuGet/Repositories/PackageRepository.cs
```diff
@@ -108,6 +108,15 @@ public IEnumerable<PackageInfo> FindPackagesById(string packageId)
                         continue;
                     }
 
+                    if (!version.IsOriginalStringNormalized())
+                    {
+                        // For a non-http match, if the OriginalVersion string is not normalized that means name of the folder which contains
+                        // the package is not a normalized string. It will cause trouble for file searching in later stage. By invalidate this 
+                        // match, it ensures the package will be reinstalled under a correct folder. This change ensures a package installed
```
davidfowl: By invalidating this

ChengTian:
Use this as an example of `dnu restore` tests: https://github.com/aspnet/dnx/blob/dev/test/Microsoft.Framework.PackageManager.FunctionalTests/DnuRestoreTests.cs#L153

troydai:
@davidfowl @ChengTian 

Updated. Test Added.

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuRestoreTests.cs
```diff
@@ -189,5 +189,51 @@ public void DnuRestore_ReinstallsCorruptedPackage(string flavor, string os, stri
                 Assert.True(File.Exists(nuspecPath));
             }
         }
+
+        [Theory]
+        [MemberData(nameof(RuntimeComponents))]
+        public void DnuRestore_ReinstallsPackageWithNormalizedVersion(string flavor, string os, string architecture)
+        {
+            var runtimeHomePath = _fixture.GetRuntimeHomeDir(flavor, os, architecture);
+            using (var tempDir = new DisposableDir("/tmp/troy.temp", false))
```
davidfowl: You're not actually hardcoding a temp directory right?  Why isn't this random?

troydai: Ah, shouldn't have it here. It's for local testing. Gonna remove it.

troydai:
:watch:

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuRestoreTests.cs
```diff
@@ -189,5 +189,51 @@ public void DnuRestore_ReinstallsCorruptedPackage(string flavor, string os, stri
...
+    ""alpha"": ""0.1.0""
+  }
+}");
+                DnuTestUtils.ExecDnu(
+                    runtimeHomePath,
+                    subcommand: "restore",
+                    arguments: $"{projectDir} -s {_fixture.PackageSource} --packages {packagesDir}");
+
+                // rename package folder to an unnormalized string
+                Directory.Move(Path.Combine(packagesDir, "alpha", "0.1.0"),
```
davidfowl: Is it possible that the folder can be normalized but the sha file and nupkg inside isn't or that that not possible

troydai: It's actually not possible. Either both normalized or not. https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Framework.PackageManager/Utils/NuGetPackageUtils.cs#L30

But the simplified model used by test is sufficient because the version of a package on disk is parsed from it's folder name: https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Framework.Runtime/NuGet/Repositories/PackageRepository.cs#L102.

troydai:
:watch:

davidfowl:
:shipit: does this also fix the issue for publish? An issue we just filed on this

troydai:
I'll try out

troydai:
This won't fix that issue though. It needs to normalize the version in PublishCommand. It doesn't use `DefaultPackagePathResolver`.

I'll merge this one first then address issue #2267

PR #2252, opened 2015-07-13

### LINKED ISSUE #2223: Add Get/SetData to IApplicationEnvironment

anurse:
Fixes #2223 

/cc @davidfowl @ChengTian

### DIFF
--- a/src/Microsoft.Framework.Runtime.Sources/Impl/ApplicationGlobalData.cs
+++ b/src/Microsoft.Framework.Runtime.Sources/Impl/ApplicationGlobalData.cs
@@ -1,0 +1,84 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Framework.Runtime.Sources.Impl
+{
+    // Helper class to completely centralize all of the application global data logic and avoid #if statements elsewhere.
+    internal class ApplicationGlobalData
+    {
+        private readonly IApplicationEnvironment _hostEnvironment;
+#if DNXCORE50
+        private readonly object _lock;
+        private readonly Dictionary<string, object> _appGlobalData;
+#endif
+
+        public ApplicationGlobalData(IApplicationEnvironment hostEnvironment)
+        {
+            _hostEnvironment = hostEnvironment;
+#if DNXCORE50
+            // If there is no host environment that we are delegating to, we need a place to store this data in CoreCLR
+            if (hostEnvironment == null)
+            {
+                // Initialize a new place to store global data on CoreCLR
+                _lock = new object();
+                _appGlobalData = new Dictionary<string, object>(StringComparer.Ordinal); // Case-sensitive, just like AppDomain.Get/SetData
+            }
+#endif
+        }
+
+        public object GetData(string name)
+        {
+            return _hostEnvironment == null ?
+                GetDataCore(name) :
+                _hostEnvironment.GetData(name);
+        }
+
+        public void SetData(string name, object value)
+        {
+            if (_hostEnvironment == null)
+            {
+                SetDataCore(name, value);
+            }
+            else
+            {
+                _hostEnvironment.SetData(name, value);
+            }
+        }
+
+#if DNX451
+        private object GetDataCore(string name)
+        {
+            return AppDomain.CurrentDomain.GetData(name);
+        }
+
+        private void SetDataCore(string name, object value)
+        {
+            AppDomain.CurrentDomain.SetData(name, value);
+        }
+#else
+        // NOTE(anurse): ConcurrentDictionary seems overkill here. This data is rarely used, and a global lock seems safer.
+        private object GetDataCore(string name)
+        {
+            lock (_lock)
+            {
+                object val;
+                if (_appGlobalData.TryGetValue(name, out val))
+                {
+                    return val;
+                }
+                return null;
+            }
+        }
+
+        private void SetDataCore(string name, object value)
+        {
+            lock (_lock)
+            {
+                _appGlobalData[name] = value;
+            }
+        }
+#endif
+    }
+}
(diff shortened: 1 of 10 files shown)


anurse:
Ping @davidfowl @ChengTian , this is now building on my machine and ready to review

ChengTian:
Build is failing. Looks like we should also implement the newly added interfaces in `dnx.host/ApplicationEnvironment.cs`

src/Microsoft.Framework.Runtime/ApplicationEnvironment.cs
```diff
@@ -49,5 +49,41 @@ public FrameworkName RuntimeFramework
...
+            return AppDomain.CurrentDomain.GetData(name);
+        }
+
+        public void SetData(string name, object value)
+        {
+            AppDomain.CurrentDomain.SetData(name, value);
+        }
+#else
+        // NOTE(anurse): ConcurrentDictionary seems overkill here. This data is rarely used, and a global lock seems safer.
+        private object _lock = new object();
```
davidfowl: readonly

src/Microsoft.Framework.Runtime/ApplicationEnvironment.cs
```diff
@@ -49,5 +49,41 @@ public FrameworkName RuntimeFramework
...
+        private Dictionary<string, object> _appGlobalData = new Dictionary<string, object>();
```
davidfowl: readonly

davidfowl: Case sensitive or not?

anurse: I'll mirror what Get/SetData on AppDomain do (which I don't know and will test first :))

anurse:
@ChengTian will do

src/Microsoft.Framework.Runtime/ApplicationEnvironment.cs
anurse: :+1:

anurse:
The fact that there are two `IApplicationEnvironment` instances in play when running an app is concerning to me and leads me to think we actually should put Get/SetData on their own interface in the service provider (`IApplicationGlobalData`). When DNX boots, it creates an `IApplicationEnvironment` for the physical DLL being executed. In most cases, this is `ApplicationHost.dll`, which then creates a second `IApplicationEnvironment` that overrides the other. If we take this change as-is, anything put into the first `IApplicationEnvironment` will be lost forever when the application host starts up because it's new `IApplicationEnvironment` will overwrite it. (Technically, on Desktop CLR they will share the same data source, but that's an implementation detail). If the goal here is to have a single dictionary of global items, it should be truly global and the only way to achieve that here is to create a new interface/class

src/Microsoft.Framework.Runtime/ApplicationEnvironment.cs
anurse: Get/SetData is case-sensitive

davidfowl:
I don't think that matters, we can wrap the original one

test/Microsoft.Framework.PackageManager.FunctionalTests/Microsoft.Framework.PackageManager.FunctionalTests.xproj
```diff
@@ -13,5 +13,8 @@
   <PropertyGroup>
     <SchemaVersion>2.0</SchemaVersion>
   </PropertyGroup>
+  <ItemGroup>
+    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
+  </ItemGroup>
```
anurse: I don't know what this is but I'll remove it

anurse:
Ping @davidfowl @ChengTian - I've updated the code to allow an Application Environment to delegate to another for data storage. This is only used in the one created by ApplicationHostContext to ensure it shares global data storage with the host.

anurse:
I'd like to try and get this in before the branching today if possible, so please review :). /cc @muratg

anurse:
Looks like I won't make the branching anyway, but I'd still like a sign-off soon :)

src/Microsoft.Framework.Runtime.Sources/Impl/ApplicationGlobalData.cs
```diff
@@ -0,0 +1,84 @@
...
+
+namespace Microsoft.Framework.Runtime.Sources.Impl
+{
+    // Helper class to completely centralize all of the application global data logic and avoid #if statements elsewhere.
+    internal class ApplicationGlobalData
+    {
+        private readonly IApplicationEnvironment _hostEnvironment;
+#if DNXCORE50
+        private readonly object _lock;
+        private readonly Dictionary<string, object> _appGlobalData;
```
halter73: Why not use this on the desktop CLR too? If someone actually depends on this being backed by the current AppDomain, I think they should use the AppDomain directly.

anurse: Seems legit to me. The goal here is to create an application-global data store, not to provide access to AppDomain.CurrentDomain.Get/SetData.

@davidfowl ?

davidfowl: In general, we try to expose a consistent facade over whatever the runtime provides. In this case I think it's fine to use AppDomain.CurrentDomain.Get/SetData, similarly to how we use LogicalCallContext to implement certain features on .NET Framework. Last I heard, the CLR team were thinking about adding a general bag of things just like AppDomain.Get/SetData and I think if that were available, we'd use it.

/cc @jkotas

anurse:
Ping. I'll be merging into release but rather than start a new PR it'd be faster to just get sign-off here and then rebase and merge.

test/Microsoft.Framework.PackageManager.FunctionalTests/Microsoft.Framework.PackageManager.FunctionalTests.xproj
anurse: Per @dougbu I'll unrevert this (unvert? devert?). It's something VS adds to test projects to avoid having to rescan them to determine if they are test projects and the sooner we get it in, the less churn we have later :)

Issue #2254, opened 2015-07-13
Dnu install fails


victorhurdugaci:
Run this on a project:

`dnu install Microsoft.AspNet.Hosting 1.0.0-beta5 --source "https://www.myget.org/F/aspnetvnext/api/v2" --source "https://nuget.org/api/v2/" --no-cache`

Result:

```
Microsoft .NET Development Utility CLR-x86-1.0.0-beta6-12208

  GET https://www.myget.org/F/aspnetvnext/api/v2/
  OK https://www.myget.org/F/aspnetvnext/api/v2/ 843ms
  GET https://nuget.org/api/v2/
  OK https://nuget.org/api/v2/ 691ms
  GET https://www.myget.org/F/aspnetvnext/api/v2/FindPackagesById()?id='Microsoft.AspNet.Hosting'
  GET https://nuget.org/api/v2/FindPackagesById()?id='Microsoft.AspNet.Hosting'
  OK https://nuget.org/api/v2/FindPackagesById()?id='Microsoft.AspNet.Hosting' 218ms
  OK https://www.myget.org/F/aspnetvnext/api/v2/FindPackagesById()?id='Microsoft.AspNet.Hosting' 892ms
Microsoft.AspNet.Hosting.1.0.0-beta5 was added to project.json.
System.ArgumentNullException: Value cannot be null.
Parameter name: path
   at System.IO.Path.GetFullPathInternal(String path)
   at System.IO.Path.GetFullPath(String path)
   at System.Linq.Enumerable.WhereSelectListIterator`2.MoveNext()
   at System.Linq.Enumerable.<DistinctIterator>d__1`1.MoveNext()
   at Microsoft.Framework.PackageManager.RestoreCommand.<Execute>d__68.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task)
   at Microsoft.Framework.PackageManager.InstallCommand.<ExecuteCommand>d__7.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

PR #2258, opened 2015-07-14
Fix regression in "dnu install"

ChengTian:
parent https://github.com/aspnet/dnx/issues/2254

Without this fix, `dnu install` throws `NullReferenceException` when it should use default directory.

@davidfowl @muratg @victorhurdugaci

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
+++ b/src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
@@ -80,13 +80,15 @@ namespace Microsoft.Framework.PackageManager
         {
             ScriptExecutor.Report = Reports.Information;
 
-            if (!RestoreDirectories.Any())
+            var effectiveRestoreDirs = RestoreDirectories.Where(x => !string.IsNullOrEmpty(x));
+
+            if (!effectiveRestoreDirs.Any())
             {
-                RestoreDirectories.Add(Directory.GetCurrentDirectory());
+                effectiveRestoreDirs = new[] { Directory.GetCurrentDirectory() };
             }
 
             bool success = true;
-            foreach (var dir in RestoreDirectories.Select(Path.GetFullPath).Distinct())
+            foreach (var dir in effectiveRestoreDirs.Select(Path.GetFullPath).Distinct())
             {
                 success &= await Execute(dir);
             }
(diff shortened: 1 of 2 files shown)


ChengTian:
Seems we don't have any functional test for `dnu install`. Adding some now and I will ping again after that.

davidfowl:
:shipit:

davidfowl:
@ChengTian Can we add a test

src/Microsoft.Framework.PackageManager/Restore/RestoreCommand.cs
```diff
@@ -80,13 +80,15 @@ public RestoreCommand(FrameworkName fallbackFramework)
         {
             ScriptExecutor.Report = Reports.Information;
 
-            if (!RestoreDirectories.Any())
+            var effectiveRestoreDirs = RestoreDirectories.Where(x => !string.IsNullOrEmpty(x));
```
victorhurdugaci: Can RestoreDirectories be null?

victorhurdugaci: nevermind, I just noticed that it is initialized

ChengTian:
Added tests

test/Microsoft.Framework.PackageManager.FunctionalTests/DnuInstallTests.cs
```diff
@@ -0,0 +1,140 @@
...
+                var projectDir = Path.Combine(tempDir, "project");
+                Directory.CreateDirectory(projectDir);
+                var projectJsonPath = Path.Combine(projectDir, Runtime.Project.ProjectFileName);
+                File.WriteAllText(projectJsonPath, @"{
+  ""dependencies"": { }
+}");
+
+                VerifyDnuInstall(
+                    runtimeHomePath,
+                    packageName: "alpha",
```
victorhurdugaci: Ah!!!! This is the package that ends up in my packages folder. I think there is a test that leaves that package on the disk. I always have it the global packages folder

ChengTian: It is possible that some `dnu commands install` tests forget to redirect `DNX_PACKAGES`/`--packages` to a temporary package folder.

For my newly added tests, I always install to temp folders.

victorhurdugaci: @BrennanConroy ?

Issue #1813, opened 2015-05-07
DNU doesn't get correct local app data folder on *nix with CoreCLR


ChengTian:
https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs#L193
(Search for `LocalAppData` in the file if it is updated)

To make it cross-plat, we can use similar approach as here: https://github.com/NuGet/Home/issues/472

@BrennanConroy

ChengTian:
Talked to @davidfowl and @BrennanConroy . This cross-plat-related issue should be moved to next milestone.

muratg:
Let's do this in Beta 6. 

cc @BrennanConroy

BrennanConroy:
I believe this is waiting on #1692 which we will hopefully get in soon.

ChengTian:
Yeah, I have a branch for this: https://github.com/aspnet/dnx/tree/wei/correct-local-appdata
and the final PR needs something from https://github.com/aspnet/dnx/pull/1692

Issue #1793, opened 2015-05-06
DNU doesn't run on coreclr on *nix/OSX


davidfowl:
This is a work item to track the progress on that feature.

- [x] URI class doesn't handle file paths correctly (DNX PR: #1691 or corefx issue: https://github.com/dotnet/corefx/issues/1745)
- [x] Need proper check for OS and Mono (#1692)
- [x] Restore hangs with more than a couple concurrent requests (https://github.com/dotnet/corefx/issues?q=is%3Aissue+is%3Aopen+Implement+Unix+label%3ASystem.Net)
- [x] DNU doesn't get correct local app data folder on *nix with CoreCLR (https://github.com/aspnet/dnx/issues/1813)
- [x] Don't use Console.GetColor (#1724)
- [ ] NuGet-related paths/files should be case-sensitively consistent (#2270)
- [x] CoreCLR HttpClient is not functional on *nix (https://github.com/dotnet/corefx/issues/2480)

colemickens:
1.0.0-beta5-11722 status for `dnu restore` under Ubuntu:

```
----------
System.ArgumentNullException: Value cannot be null.
Parameter name: path1
   at System.IO.Path.Combine(String path1, String path2, String path3)
   at NuGet.Settings.LoadDefaultSettings(IFileSystem fileSystem, String configFileName, IMachineWideSettings machineWideSettings)
   at Microsoft.Framework.PackageManager.SettingsUtils.ReadSettings(String solutionDir, String nugetConfigFile, IFileSystem fileSystem, IMachineWideSettings machineWideSettings)
   at Microsoft.Framework.PackageManager.RestoreCommand.ReadSettings(String solutionDirectory)
   at Microsoft.Framework.PackageManager.RestoreCommand.<ExecuteCommand>d__68.MoveNext()
----------
Restore failed
Value cannot be null.
Parameter name: path1
```

BrennanConroy:
The above error is because NuGet uses Environment Variables that don't have values on non-windows OS's. Tracked by: https://github.com/NuGet/Home/issues/472

muratg:
@BrennanConroy Can you update the checklist to reflect the current state?

akamud:
Any update on this?

overeemm:
Because of this, the whole getting started on macosx (http://dotnet.github.io/core/getting-started/) is broken, right?

colemickens:
[...]

PR #2262, opened 2015-07-15
Resolve correct app data paths on *nix with CoreCLR/Mono

ChengTian:
parent https://github.com/aspnet/dnx/issues/1813

Part of cross-plat effort.
@davidfowl @BrennanConroy @anurse

### DIFF
--- a/src/Microsoft.Framework.PackageManager/NuGet/CommandLine/CommandLineMachineWideSettings.cs
+++ b/src/Microsoft.Framework.PackageManager/NuGet/CommandLine/CommandLineMachineWideSettings.cs
@@ -3,9 +3,7 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
+using Microsoft.Framework.PackageManager;
 
 namespace NuGet
 {
@@ -15,14 +13,11 @@ namespace NuGet
 
         public CommandLineMachineWideSettings()
         {
-#if DNX451
-            var baseDirectory = Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData);
-#else
-            var baseDirectory = Environment.GetEnvironmentVariable("ProgramData");
-#endif
+            string machineWideConfigDir = DnuEnvironment.GetFolderPath(DnuFolderPath.MachineWideConfigDirectory);
+
             _settings = new Lazy<IEnumerable<NuGet.Settings>>(
                 () => NuGet.Settings.LoadMachineWideSettings(
-                    new PhysicalFileSystem(baseDirectory)));
+                    new PhysicalFileSystem(machineWideConfigDir)));
         }
 
         public IEnumerable<Settings> Settings
--- a/src/Microsoft.Framework.PackageManager/NuGet/Core/Configuration/ConfigurationDefaults.cs
+++ b/src/Microsoft.Framework.PackageManager/NuGet/Core/Configuration/ConfigurationDefaults.cs
@@ -5,7 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Xml.Linq;
+using Microsoft.Framework.PackageManager;
 
 namespace NuGet
 {
@@ -21,13 +21,8 @@ namespace NuGet
 
         private static ConfigurationDefaults InitializeInstance()
         {
-#if DNX451
-            var commonAppData = Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData);
-#else
-            var commonAppData = Environment.GetEnvironmentVariable("ProgramData");
-#endif
-            var baseDirectory = Path.Combine(commonAppData, "NuGet");
-            PhysicalFileSystem fileSystem = new PhysicalFileSystem(baseDirectory);
+            var machineWideSettingsDir = DnuEnvironment.GetFolderPath(DnuFolderPath.MachineWideSettingsBaseDirectory);
+            PhysicalFileSystem fileSystem = new PhysicalFileSystem(machineWideSettingsDir);
             return new ConfigurationDefaults(fileSystem, ConfigurationDefaultsFile);
         }
 
--- a/src/Microsoft.Framework.PackageManager/NuGet/Core/Configuration/Settings.cs
+++ b/src/Microsoft.Framework.PackageManager/NuGet/Core/Configuration/Settings.cs
@@ -10,6 +10,7 @@ using System.Threading;
 using System.Xml;
 using System.Xml.Linq;
 using NuGet.Resources;
+using Microsoft.Framework.PackageManager;
 
 namespace NuGet
 {
@@ -104,13 +105,8 @@ namespace NuGet
         {
             // Walk up the tree to find a config file; also look in .nuget subdirectories
             var validSettingFiles = new List<Settings>();
-#if DNX451
-            var appData = Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData);
-#else
-            var appData = Environment.GetEnvironmentVariable("APPDATA");
-#endif
-            var redirectSettingsPath = Path.Combine(appData,
-                                                    "nuget",
+            var userSettingsDir = DnuEnvironment.GetFolderPath(DnuFolderPath.UserSettingsDirectory);
+            var redirectSettingsPath = Path.Combine(userSettingsDir,
                                                     "nuget.redirect.config");
 
             Settings redirectSettings = null;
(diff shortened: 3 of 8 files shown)


src/Microsoft.Framework.PackageManager/Utils/Envrionment.cs
```diff
@@ -0,0 +1,112 @@
...
+
+using System;
+using System.IO;
+using Microsoft.Framework.Runtime;
+
+// TODO: delete this after System.Environment.GetFolderPath() is available on CoreCLR
+#if DNXCORE50
+namespace Microsoft.Framework.PackageManager.Internal
+{
+    internal static class Environment
```
anurse: This is spelled right, but the file name isn't :)

ChengTian: :+1: good catch. Fixed

src/Microsoft.Framework.PackageManager/Utils/Environment.cs
```diff
@@ -0,0 +1,112 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using Microsoft.Framework.Runtime;
+
+// TODO: delete this after System.Environment.GetFolderPath() is available on CoreCLR
```
davidfowl: I don't think this is coming back.

davidfowl:
I'd prefer to make a higher level abstraction. Something that abstractions paths used in dnu/nuget itself. As an example:

```C#
public class DnuEnvironment
{
    public string GetFolderPath(DnuFolderPath type)
    {
         // Switch here on the types of folders and use the Environment wrapper to detect
    }
}

public enum DnuFolderPath
{
   HttpCacheDirectory,
   MachineWideSettingsDirectory,
   GlobalPackageConfigurationDirectory,
   GlobalPackagesDirectory,
   DnxHome
}
```

src/Microsoft.Framework.PackageManager/Utils/Environment.cs
ChengTian: Ok, removing comments here

ChengTian:
@davidfowl , added higher level of abstraction

src/Microsoft.Framework.PackageManager/NuGet/CommandLine/CommandLineMachineWideSettings.cs
```diff
@@ -3,9 +3,10 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
+using Microsoft.Framework.PackageManager.Internal;
+#if DNXCORE50
+using Environment = Microsoft.Framework.PackageManager.Internal.Environment;
```
BrennanConroy: Environment isn't used in this file anymore

src/Microsoft.Framework.PackageManager/NuGet/Core/Configuration/ConfigurationDefaults.cs
```diff
@@ -21,12 +21,9 @@ public class ConfigurationDefaults
...
         {
-#if DNX451
-            var commonAppData = Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData);
-#else
-            var commonAppData = Environment.GetEnvironmentVariable("ProgramData");
-#endif
-            var baseDirectory = Path.Combine(commonAppData, "NuGet");
+            var machineWideSettingsDir = DnuEnvironment.GetFolderPath(DnuFolderPath.MachineWideSettingsDirectory);
+
+            var baseDirectory = Path.Combine(machineWideSettingsDir, "NuGet");
```
moozzyk: below "nuget" is all lower case - does it matter that it is not here?

ChengTian: It matters on case-sensitive filesystem. I filed a separate bug for it here: https://github.com/aspnet/dnx/issues/2270

ChengTian: Also added one more checkbox in https://github.com/aspnet/dnx/issues/1793

src/Microsoft.Framework.PackageManager/Publish/PublishManager.cs
```diff
@@ -141,12 +145,8 @@ public bool Publish()
                     if (string.IsNullOrEmpty(runtimeHome))
                     {
                         var runtimeGlobalPath = Environment.GetEnvironmentVariable(EnvironmentNames.GlobalPath);
```
davidfowl: What is this and why isn't it a DnuFolderPath?

src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
```diff
@@ -11,7 +11,11 @@
 using System.Security.Cryptography;
 using System.Text;
 using System.Threading.Tasks;
+using Microsoft.Framework.PackageManager.Internal;
 using Microsoft.Framework.Runtime.Internal;
+#if DNXCORE50
+using Environment = Microsoft.Framework.PackageManager.Internal.Environment;
```
davidfowl: Where is this used?

src/Microsoft.Framework.PackageManager/Publish/PublishManager.cs
ChengTian: This is an environment variable `DNX_GLOBAL_PATH`, looks like some global DNX home set by MSI installer. Will make this a DnuFolderPath.

src/Microsoft.Framework.PackageManager/Restore/NuGet/HttpSource.cs
ChengTian: Yeah, it should be removed

src/Microsoft.Framework.PackageManager/NuGet/Core/Configuration/Settings.cs
```diff
@@ -172,15 +169,11 @@ public string ConfigFilePath
             if (configFileName == null)
             {
                 // load %AppData%\NuGet\NuGet.config
-#if DNX451
-                var appDataPath = Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData);
-#else
-                var appDataPath = Environment.GetEnvironmentVariable("APPDATA");
-#endif
-                if (!String.IsNullOrEmpty(appDataPath))
+                var userSettingsDir = DnuEnvironment.GetFolderPath(DnuFolderPath.UserSettingsDirectory);
```
davidfowl: This doesn't look right. Why isn't it directory returning the NuGet/Config folder.

ChengTian: `NuGet.Settings.LoadMachineWideSettings(root)` requires you pass `CommonAppData` as its `root` argument. Note that it requires `CommonAppData`, not `CommonAppData/NuGet/Config` because this method append `/NuGet/Config` to root by itself. I am following the same convention here.

ChengTian: But yes, I can make `DnuFolderPath.UserSettingsDirectory` return `AppData/NuGet`. Making the change now

ChengTian: It is `AppData/NuGet`, instead of `AppData/NuGet/Config` because the comments say so

ChengTian:
@davidfowl , using `DnuEnvironment.GetFolderPath()` at all possible places except one:
I'll break some tests and there is no easy way to fix them if I use `DnuEnvironment.GetFolderPath()` there. I filed a bug for it: https://github.com/aspnet/dnx/issues/2279

src/Microsoft.Framework.PackageManager/Publish/PublishManager.cs
```diff
@@ -140,13 +143,9 @@ public bool Publish()
                     var runtimeHome = Environment.GetEnvironmentVariable(EnvironmentNames.Home);
```
ChengTian: The logic here requires that we use environment variable here (if `DNX_HOME` is set do ..., otherwise ...). So there is no `DnuEnvironment.GetFolderPath()` for this.

src/Microsoft.Framework.PackageManager/Utils/Environment.cs
```diff
@@ -0,0 +1,111 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using Microsoft.Framework.Runtime;
+
+#if DNXCORE50
+namespace Microsoft.Framework.PackageManager.Internal
+{
+    internal static class Environment
```
anurse: Couldn't we merge this in to DnuEnvironment and just use DnuEnvironment everywhere (even to get things like Program Files which are available through System.Environment) to ensure we have one central location for these things?

anurse:
Looks good, but I would like to consider the comment I posted about merging Environment and DnuEnvironment so we can avoid shadowing System.Environment. :watch:

ChengTian:
@anurse , thanks for the feedback. Fixed and Travis just passed.

ChengTian:
Will retarget to release branch before merging

PR #2263, opened 2015-07-15
Enabling implicit appbase (`dnx . run` -> `dnx run`)

### LINKED ISSUE #1403: dnx requiring "run" can be improved a bit
`dnx path-to-project.json run` is a bit redundant - if you name the project file or folder, and there are no arguments, it can be implied.

On the other hand, if `dnx run` can use the current directory as the project then that does make sense, and looks cleaner than `dnx . run ` or `dnx project.json run`

moozzyk:
Before providing appbase folder was necessary. Now if the appbase folder is not provided we will use the current folder (`.`) automatically.

Note that there might be an ambiguity between a folder and command which is resolved by checking if a folder with the provided name physically exists and if it does the parameter is treated as folder path otherwise it is treated as a command.

Fixes #1403


src/dnx/pal.unix.cpp
```diff
@@ -28,28 +29,34 @@ void SetConsoleHost()
...
     }
 
-    return TRUE;
+    return true;
+}
+
+bool DirectoryExists(const char* path)
+{
+    struct stat s;
+    return stat(path, &s) == 0 && S_ISDIR(s.st_mode);
```
moozzyk: Need to verify this manually.

davidfowl:
Where are the mono changes?

moozzyk:
Mono will be done separately. This is already too big...

anurse:
I thought we weren't going to do the folder check and only support filenames (replacing the command, effectively). Selecting a different folder would only be supported by `dnx -p /path/to/project run`

src/dnx/dnx.cpp
```diff
@@ -2,106 +2,81 @@
...
     }
 
-    size_t nOffset = nStrLen - nSuffixLen;
-
-    return ::_tcsnicmp(pszStr + nOffset, pszSuffix, MAX_PATH - nOffset) == 0;
+    return strings_equal_ignore_case(s + str_len - suffix_len, suffix);
 }
 
-bool LastPathSeparatorIndex(LPCTSTR pszPath, size_t* pIndex)
+int split_path(const dnx::char_t* path)
```
BrennanConroy: Not a fan of this function name

src/dnx/dnx.cpp
```diff
@@ -2,106 +2,81 @@
...
 {
-    size_t nLastSlashIndex;
-    size_t nLastBackSlashIndex;
-
-    bool hasLastSlashIndex = LastIndexOfCharInPath(pszPath, _T('/'), &nLastSlashIndex);
-    bool hasLastBackSlashIndex = LastIndexOfCharInPath(pszPath, _T('\\'), &nLastBackSlashIndex);
-
-    if (hasLastSlashIndex && hasLastBackSlashIndex)
+    for (int i = static_cast<int>(x_strlen(path)) - 1; i >= 0; i--)
```
ChengTian: auto?

moozzyk: Can be auto indeed because it has the `static_cast<int>`

src/dnx/dnx.cpp
moozzyk: Couldn't come up with a better name but if you have a proposal let me know... I am also still considering changing it to return two std::strings - directory and filename - these buffer operations are so fragile.

davidfowl:
> I thought we weren't going to do the folder check and only support filenames (replacing the command, effectively). Selecting a different folder would only be supported by dnx -p /path/to/project run

I thought that was the plan? VS uses the full --appbase expansion. The only thing we need to support is `.`

moozzyk:
Even better - I did not like going to the disk to check if the folder exists.

moozzyk:
Updated the PR so that we no longer check if an argument is a folder except for `.` when it always is a folder.

moozzyk:
(Travis fails because the fix for Mono is separate - #2269 - I will push them together or can squash once reviewed)

moozzyk:
:watch:

src/dnx/dnx.cpp
```diff
@@ -110,125 +85,139 @@ int BootstrapperOptionValueNum(LPCTSTR pszCandidate)
...
-    // Allocate memory before doing expansion
-    for (int i = nPathArgIndex; i < nExpandedArgc; ++i)
+void ExpandArgument(const dnx::char_t* value, std::vector<const dnx::char_t*>& expanded_args)
+{
+    if (string_ends_with_ignore_case(value, _X(".dll")) || string_ends_with_ignore_case(value, _X(".exe")))
     {
-        ppszExpandedArgv[i] = new TCHAR[MAX_PATH];
+        // "dnx /path/App.dll arg1" --> "dnx --appbase /path/ /path/App.dll arg1"
+        // "dnx /path/App.exe arg1" --> "dnx --appbase /path/ /path/App.exe arg1"
+        // "dnx App.exe arg1" --> "dnx --appbase --appbase . App.exe arg1"
```
BrennanConroy: Only 1 --appbase right?

src/dnx/dnx.cpp
```diff
@@ -110,125 +85,139 @@ int BootstrapperOptionValueNum(LPCTSTR pszCandidate)
...
 
-bool ExpandCommandLineArguments(int nArgc, dnx::char_t** ppszArgv, int& nExpandedArgc, dnx::char_t**& ppszExpandedArgv)
+dnx::char_t* GetAppBaseParameterValue(size_t argc, dnx::char_t* argv[])
 {
-    // If no args or '--appbase' is already given and it has a value
-    if (nArgc == 0 || GetAppBaseParameterValue(nArgc, ppszArgv))
+    auto index = FindAppBaseOrNonHostOption(argc, argv);
+
+    // no parameters or '--appbase' is the last value in the array or `--appbase` not found
+    if (argc == 0 || index >= argc - 1 || argv[index][0] != _X('-'))
```
BrennanConroy: What happens if some does ```dnx /path/App.dll --appbase something``` where --appbase could be an argument to the dll

moozzyk: It will now work because `/path/App.dll` is a "NonHostOption" and therefore we will create appbase from this argument (you will get `--appbase /path/ /path/App.dll --appbase something`)

src/dnx/dnx.cpp
moozzyk: Yeah. I will fix this.

src/dnx/dnx.cpp
BrennanConroy: But the FindAppBaseOrNonoHostOption returns when it finds a --appbase anywhere in the args.

moozzyk: No. When `BootstrapperOptionValueNum` reaches the .dll argument it will return -1 and then we will break the loop in `FindAppBaseOrNonoHostOption` and return index to this argument. This scenario is also covered by tests - `ExpandCommandLineArguments_should_ignore_appbase_after_bootstrapper_commands` (except for we break the loop on `.` instead of the dll but it does not matter from the perspective of the `BootstrapperOptionValueNum` function since it breaks on anything it can't understand so `.` and path falls into the same category).

BrennanConroy: Oh! I didn't see that BootStrapperOptionValueNum was in the loop.

test/Bootstrapper.FunctionalTests/BootstrapperTests.cs
```diff
@@ -114,7 +114,7 @@ public void BootstrapperShowsVersionAndReturnsZeroExitCodeWhenVersionOptionWasGi
 
         [Theory]
         [MemberData(nameof(RuntimeComponents))]
-        public void BootstrapperInvokesApplicationHostWithInferredAppBase_ProjectDirAsArgument(string flavor, string os, string architecture)
+        public void BootstrapperInvokesApplicationHostWithExplicitAppBase(string flavor, string os, string architecture)
```
davidfowl: We need to cover more cases here:

- Current path:
```
dnx run
```
- Explcit `--appbase`:
```
dnx --appbase {path} Microsoft.Framework.ApplicationHost
```

- Specified path implicit ApplicationHost

```
dnx --project {path} run
```

anurse: What is `--project`?

moozzyk: We have these cases:

`dnx run` - ln. 148
`dnx --appbase` - ln. 128
`dnx . run` - https://github.com/aspnet/dnx/blob/dev/test/Bootstrapper.FunctionalTests/BootstrapperTests.cs#L264

Let me repeat after @anurse - 'What is `--project`?" (and how my change is affecting it - btw. all the tests passed so I assume I did not break it)

anurse:
:shipit: - I don't know what `--project` is, but whatever it is, it sounds new... I'd rather just get this in so we can start seeing what this does.

davidfowl:
@anurse Wasn't that part of the suggestion? `-p` to replace `dnx {path} {command}`

anurse:
Yes but `-p` was just a short form for `--appbase`, not a new argument.

moozzyk:
@anurse Can't be. If `--appbase` is provided we don't do any expansion (and never have) so the user would have to provide the host. We will create a new `-p` argument that will expand to `--appbase {path} Microsoft.Framework.ApplicationHost`

anurse:
My original proposal was to always use `Microsoft.Framework.ApplicationHost` **unless** the first argument is a dll or exe. So `dnx --appbase Foo` would expand to `dnx --appbase Foo Microsoft.Framework.ApplicationHost.dll`. If we want to do it as a different argument, that's OK with me. I just don't see a use case for `dnx --appbase Foo SomethingThatIsNotADll` without using AppHost...

moozzyk:
If we always expanded as you propose how would you do `dnx --appbase AppFolder MyHost.dll run` ?

PR #2269, opened 2015-07-16
Enabling implicit appbase (`dnx . run` -> `dnx run`) on Mono

### LINKED ISSUE #1403: dnx requiring "run" can be improved a bit

moozzyk:
The user no longer has to provide folder (or `.`) when running dnx commands.

Note that there might be an ambiguity between a folder and command which is resolved by checking if a folder with the provided name physically exists and if it does the parameter is treated as folder path otherwise it is treated as a command.

Fixes #1403


moozzyk:
Same caveat as in #2263 - for backwards compatibility it allows specifying path to a folder (or '.').

src/dnx.mono.managed/EntryPoint.cs
```diff
@@ -69,82 +68,92 @@ public static int Main(string[] arguments)
...
             new FrameworkName("DNX", new Version(4, 5, 1)));
     }
 
     private static string[] ExpandCommandLineArguments(string[] arguments)
     {
-        // If '--appbase' is already given and it has a value, don't need to expand
-        var appbaseIndex = arguments.ToList().FindIndex(arg =>
-            string.Equals(arg, "--appbase", StringComparison.OrdinalIgnoreCase));
-        if (appbaseIndex >= 0 && (appbaseIndex < arguments.Length - 1))
+        int parameter_idx = FindAppBaseOrNonHostOption(arguments);
```
davidfowl: c++? Since when do we use _

src/dnx.mono.managed/EntryPoint.cs
```diff
@@ -69,82 +68,92 @@ public static int Main(string[] arguments)
...
+                expandedArgs.Add(arguments[i]);
             }
         }
 
-        // No path argument was found, no expansion is needed
-        if (pathArgIndex >= arguments.Length)
-        {
-            return arguments;
-        }
+        return expandedArgs.ToArray();
```
davidfowl: We should be avoiding array allocation where possible.

src/dnx.mono.managed/EntryPoint.cs
moozzyk: LOL - will fix.

src/dnx.mono.managed/EntryPoint.cs
moozzyk: `Bootstrapper.Execute` takes parameters as an array of strings so we have to allocate it somewhere...

moozzyk:
Removed checking if directory exists - now we only support `.`.

src/dnx.mono.managed/EntryPoint.cs
```diff
@@ -69,82 +68,92 @@ public static int Main(string[] arguments)
...
             new FrameworkName("DNX", new Version(4, 5, 1)));
     }
 
     private static string[] ExpandCommandLineArguments(string[] arguments)
     {
-        // If '--appbase' is already given and it has a value, don't need to expand
-        var appbaseIndex = arguments.ToList().FindIndex(arg =>
-            string.Equals(arg, "--appbase", StringComparison.OrdinalIgnoreCase));
-        if (appbaseIndex >= 0 && (appbaseIndex < arguments.Length - 1))
+        int parameterIdx = FindAppBaseOrNonHostOption(arguments);
```
troydai: `var` instead of `int`?

moozzyk: yes - fixed.

moozzyk:
:watch:

src/dnx.mono.managed/EntryPoint.cs
```diff
@@ -69,82 +68,92 @@ public static int Main(string[] arguments)
...
+        expandedArgs.Add(".");
+        expandedArgs.Add("Microsoft.Framework.ApplicationHost");
+        expandedArgs.Add(argument);
+    }
+
+    private static int FindAppBaseOrNonHostOption(string[] arguments)
+    {
+        for (var i = 0; i < arguments.Length; i++)
+        {
+            if (string.Equals(arguments[i], "--appbase", StringComparison.OrdinalIgnoreCase))
```
troydai: It seems that in `CommandUtil` the options names are compared in case sensitive way. 
https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandLineApplication.cs#L125

moozzyk: I am keeping the behavior we had before: https://github.com/aspnet/dnx/pull/2269/files#diff-71b95628dab21cdd062a2d2518c8ac05L81

@davidfowl should the option names be case sensitive? (dnx.exe does the same case insensitive check)

muratg: Let's keep the current behavior for now.

PR #2275, opened 2015-07-16

### LINKED ISSUE #2223: Add Get/SetData to IApplicationEnvironment

anurse:
Fixes #2223 

Moved to release

/cc @muratg @davidfowl @halter73 @ChengTian


davidfowl:
This isn't going into beta6, it's beta7

anurse:
Why wasn't the bug updated? Anyway, I'll move it.

muratg:
Hmm, this doesn't look very risky to me. Why are we moving it?

PR #2276, opened 2015-07-16

### LINKED ISSUE #2223: Add Get/SetData to IApplicationEnvironment

anurse:
Fixes #2223 

Someone just give me a squirrel already...

/cc @muratg @davidfowl @halter73 @ChengTian


src/Microsoft.Framework.Runtime.Sources/Impl/ApplicationGlobalData.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            return AppDomain.CurrentDomain.GetData(name);
+        }
+
+        private void SetDataCore(string name, object value)
+        {
+            AppDomain.CurrentDomain.SetData(name, value);
+        }
+#else
+        // NOTE(anurse): ConcurrentDictionary seems overkill here. This data is rarely used, and a global lock seems safer.
```
muratg: Is `ConcurrentDictionary` concern because of the performance?  Memory overhead?

test/Microsoft.Framework.Runtime.Tests/ApplicationEnvironmentFacts.cs
```diff
@@ -0,0 +1,69 @@
...
+            var appEnv2 = new ApplicationEnvironment(null, null, null, null);
+            var obj = new object();
+            var key = "testGetSetGlobal:" + Guid.NewGuid().ToString("N");
+            appEnv1.SetData(key, obj);
+            Assert.Same(obj, appEnv2.GetData(key));
+            Assert.Same(obj, AppDomain.CurrentDomain.GetData(key));
+        }
+#else
+        [Fact]
+        public void GetSetDataPersistDataWithinEachApplicationEnvironment()
```
muratg: Don't we need this test in CLR as well?

src/Microsoft.Framework.Runtime.Sources/Impl/ApplicationGlobalData.cs
davidfowl: It's overkill. This thing is rarely used. ConcurrentDictionary only makes sense in specific scenarios.

muratg:
Looks good to me, but since @davidfowl wanted to push it to beta 7, I'd like to him to me to give the shipit.

test/Microsoft.Framework.Runtime.Tests/ApplicationEnvironmentFacts.cs
anurse: No because it's not true in Desktop CLR. All instances of ApplicationEnvironment share data.

src/Microsoft.Framework.Runtime.Sources/Impl/ApplicationGlobalData.cs
```diff
@@ -0,0 +1,84 @@
+﻿using System;
```
troydai: License

anurse: license police

anurse:
Last call, I'm shipping it if I don't hear back soon :)

Issue #2273, opened 2015-07-16
Dnu publish should use IPackagePathResolver to resolve package path


troydai:
https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Framework.PackageManager/Publish/PublishPackage.cs#L28

davidfowl:
@troydai Why isn't this issue closed and tagged and put in a milestone?

davidfowl:
@troydai problems here:

- No bug or enhancement tag
- No assigned person
- Fix not linked to commit

/cc @muratg

Issue #2267, opened 2015-07-15
FileNotFoundException when publishing


sklose:
Since a couple of days I am getting a FileNotFoundException when I run 'dnu publish'. It seems to be related to commit https://github.com/aspnet/dnx/commit/cae8aab5b38e0039f03eb87a8d994e3035179206 where the file paths were canonicalized. 

```
info: Exec
info:   program: dnu
info:   commandline: publish /app/src/Scheduler/Andasa.Scheduler -o /app/artifacts/publish/Andasa.Scheduler --no-source
info:   workingdir: /app

Microsoft .NET Development Utility Mono-x64-1.0.0-beta6-12232


Copying to output path /app/artifacts/publish/Andasa.Scheduler

Using Package dependency AWSSDK 2.3.47.0

Adding NuGet package /root/.dnx/packages/AWSSDK/2.3.47/AWSSDK.2.3.47.0.nupkg to /app/artifacts/publish/Andasa.Scheduler/approot/packages

System.IO.FileNotFoundException: Could not find file "/root/.dnx/packages/AWSSDK/2.3.47/AWSSDK.2.3.47.0.nupkg".

File name: '/root/.dnx/packages/AWSSDK/2.3.47/AWSSDK.2.3.47.0.nupkg'
```

The package acutally exists on disk, but the path is /root/.dnx/packages/AWSSDK/2.3.47/AWSSDK.2.3.47.nupkg (without the .0)

anurse:
Looks related to https://github.com/aspnet/dnx/issues/2192 (cc @troydai)

We recently did some work to clean up the way we convert version numbers to strings (in order to make them more uniform). Can you try re-running `dnu restore` on your project to write the packages into the correct paths and see if you can publish after that?

sklose:
This actually happens in a clean docker container I use for builds on my CI system. All packages are restored from scratch on each build. Restoring, compiling, running unit tests works fine, just the publish step is failing.

sklose:
I might have found the issue:
https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Framework.PackageManager/Publish/PublishPackage.cs line 28 ... there is no call to GetNormalizedVersionString() like in https://github.com/aspnet/dnx/commit/cae8aab5b38e0039f03eb87a8d994e3035179206

troydai:
Hi @sklose. You code you point out may cause the issue, I need to take a look. In the meantime, I believe if you clean your packages folder and restore again this issue will go away.

[...]

PR #2277, opened 2015-07-16
Use normalized version in publish package search

troydai:
Addressing issue #2267 

This is not a perfect fix. Ideally an `IPackagePathResolver` should be used to locate the package. I'd like to push this in beta-6 so I limit the scope of the change.

I track the remaining work in https://github.com/aspnet/dnx/issues/2273.

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Publish/PublishPackage.cs
+++ b/src/Microsoft.Framework.PackageManager/Publish/PublishPackage.cs
@@ -1,8 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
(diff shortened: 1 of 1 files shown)


troydai:
Working on a test now.

/cc @davidfowl @ChengTian @muratg 
/cc customer @sklose

davidfowl:
@troydai Instead of this change, new up a PackageResolver with the right path and call GetPackageDirectory

troydai:
Ping @davidfowl

ChengTian:
Looks good to me :shipit:

muratg:
:shipit:

davidfowl:
This obvious was meant to be in release right?

troydai:
It's checked to release first then merged to dev. Yes.

PR #2280, opened 2015-07-16
Resolve correct app data paths on *nix with CoreCLR/Mono

ChengTian:
Replacing https://github.com/aspnet/dnx/pull/2262, which has wrong targeting branch

@anurse


anurse:
:shipit:

PR #2281, opened 2015-07-16
Show more details for lock file errors

ChengTian:
parent https://github.com/aspnet/dnx/issues/1469

@davidfowl @anurse @troydai 

Show more detailed errors when lock file is invalid:
* Lock file version mismatch at runtime:
capture

* Lock file version mismatch when building:
capture1


test/Microsoft.Framework.ApplicationHost.FunctionalTests/AppHostTests.cs
```diff
@@ -157,14 +157,8 @@ public void AppHostShowsErrorWhenCurrentTargetFrameworkWasNotFoundInProjectJson(
...
 
-                var expectedErrorMsg =$@"The current runtime target framework is not compatible with '{projectName}'.
-
-Current runtime Target Framework: '{runtimeTargetFramework} ({runtimeTargetFrameworkShortName})'
-  Type: {runtimeType}
-  Architecture: {architecture ?? TestUtils.CurrentRuntimeEnvironment.RuntimeArchitecture}
-  Version: {TestUtils.GetRuntimeVersion()}
-
-Please make sure the runtime matches a framework specified in {Project.ProjectFileName}";
+                // TODO: add complete error message check when OS version information is consistent on CoreCLR and CLR
```
ChengTian: On my Windows 8.1, `RuntimeEnvironment.OperatingSystemVersion` returns different values for dnx451 and dnxcore50
For dnx451, it returns `6.2.9200.0`
For dnxcore50, it returns `6.2`

ChengTian:
Previously, if your app doesn't need any NuGet package dependency, it can run/build without a lock file. Now we decide to enforce the existence of lock file, so it always fails if your app doesn't have a lock file.

That's why I am fixing some tests here by calling `dnu restore` to generate lock files.

src/Microsoft.Framework.Runtime/ApplicationHostContext.cs
```diff
@@ -173,5 +176,35 @@ public IServiceProvider ServiceProvider
...
+                }
+                else
+                {
+                    diagnostics.AddRange(_lockFile.GetDiagnostics(Project));
+                }
+                return diagnostics;
+            }
+        }
+
+        public IEnumerable<ICompilationMessage> AllDiagnostics
```
davidfowl: Make this a method

src/Microsoft.Framework.Runtime/ApplicationHostContext.cs
```diff
@@ -173,5 +176,35 @@ public IServiceProvider ServiceProvider
         public string ProjectDirectory { get; private set; }
 
         public string PackagesDirectory { get; private set; }
+
+
+
+        public IEnumerable<ICompilationMessage> LockFileDiagnostics
```
davidfowl: Why is this exposed as a property?

ChengTian: Because sometimes we only need diagnostics for lock file. Check the case below:
capture

In this case, we should only include lock file diagnostics to the error message, instead of all diagnostics. Otherwise, you'll see "unable to resolve Newtonsoft.Json" twice

src/Microsoft.Framework.Runtime/ApplicationHostContext.cs
```diff
@@ -17,6 +19,7 @@ namespace Microsoft.Framework.Runtime
     public class ApplicationHostContext
     {
         private readonly ServiceProvider _serviceProvider;
+        private LockFile _lockFile;
```
troydai: make it `readonly`?

src/Microsoft.Framework.Runtime/ApplicationHostContext.cs
```diff
@@ -173,5 +176,35 @@ public IServiceProvider ServiceProvider
         public string ProjectDirectory { get; private set; }
 
         public string PackagesDirectory { get; private set; }
+
+
```
troydai: Remove empty lines.

src/Microsoft.Framework.Runtime/ApplicationHostContext.cs
```diff
@@ -173,5 +176,35 @@ public IServiceProvider ServiceProvider
...
+                    diagnostics.Add(new FileFormatMessage(
+                        $"The expected lock file doesn't exist. Please run \"dnu restore\" to generate a new lock file.",
+                        Path.Combine(Project.ProjectDirectory, LockFileReader.LockFileName),
+                        CompilationMessageSeverity.Error));
+                }
+                else
+                {
+                    diagnostics.AddRange(_lockFile.GetDiagnostics(Project));
+                }
+                return diagnostics;
```
troydai: Can we cache this collection instead of allocate it repeatedly?

src/Microsoft.Framework.Runtime/DependencyManagement/LockFile.cs
```diff
@@ -15,13 +16,34 @@ public class LockFile
...
+                    $"{message}. Please run \"dnu restore\" to generate a new lock file.",
+                    Path.Combine(project.ProjectDirectory, LockFileReader.LockFileName),
+                    CompilationMessageSeverity.Error);
+            }
+        }
+
         public bool IsValidForProject(Project project)
         {
+            string message;
+            return IsValidForProject(project, out message);
```
troydai: I'd like to share my concerns here. It allocate a string but thrown away. It is understandable that the overload helps to reuse code, but from the perspective of performance it's not ideal.

src/Microsoft.Framework.Runtime/ApplicationHostContext.cs
ChengTian: Here is the output when you don't have a individual property for lock file diagnostics:
capture

Note that it says
```
System.InvalidOperationException: Failed to resolve the following dependencies for target framework 'DNX,Version=v4.5.1':
   Newtonsoft.Json 6.0.0
```
but it also says
```
C:\AspNet\Home\samples\latest\ConsoleApp\project.json(3,28): error: The dependency Newtonsoft.Json >= 6.0.0 could not be resolved.
```
So I am trying to get rid of the redundancy here.

troydai:
:watch:

anurse:
Looks good to me, I concur with most of David and Troy's comments :)

src/Microsoft.Framework.Runtime/DependencyManagement/LockFile.cs
anurse: We'll generally want the string, at least in the hot path (runtime startup). Plus, the big string isn't actually allocated until an error occurs, this is just a pointer to hold it :). So I'm not too worried about this piece of code.

davidfowl:
Where is the DTH change? Also can you make GetDependencyDiagnostics internal?

ChengTian:
@davidfowl , added DTH change. `GetDependencyDiagnostics` is in `Microsoft.Framework.Runtime` and `Microsoft.Framework.DesignTimeHost` is calling it, so we cannot make it internal.

ChengTian:
Ping @davidfowl

ChengTian:
@anurse , @davidfowl was talking about this PR

troydai:
Looks good to me, too.

troydai:
:shipit:

anurse:
:shipit:

ChengTian:
This actually stales after renaming. Will fix it and merge soon.

ChengTian:
Already merged to dev branch. This didn't close automatically.

Issue #2265, opened 2015-07-15
First Chance Exception on Windows when finding which OS user is on


BrennanConroy:
Throwing line: https://github.com/aspnet/dnx/blob/dev/src/dnx.host/NativeMethods.cs#L16

@muratg @davidfowl

davidfowl:
This is the cause https://github.com/aspnet/dnx/blob/dev/src/dnx.host/RuntimeEnvironment.cs#L28

muratg:
@BrennanConroy were you already looking into this one?

BrennanConroy:
1d23ce503d5a2b305320cd1c7c607d5e5e4efc03

Issue #1970, opened 2015-05-28
Don't use environment variables to pass data from native to managed code


davidfowl:
It flows to child contexts and generally messes things up.

muratg:
@moozzyk

blowdart:
Bonus potential security issues here as well :)

moozzyk:
Fixing AppDomainManager may get _interesting_

moozzyk:
When doing this we should also pass OS, Architecture etc. from the native layer to managed code.

anurse:
Also the framework name selected using the `--framework` switch specified in #2333. Right now I'm going to be using `DNX_FRAMEWORK` to do this.

This one might be tricky since it needs to get to the Domain Manager. Possibly a P/Invoke callback could be used?

davidfowl:
Related to #2544. When you make this change, can you make it so that https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Host/Bootstrapper.cs has a richer object model that represents the things it takes as input? Basically a higher level hosting API that doesn't assume the caller is a command line application?

moozzyk:
Fixed in commits: e87f53dd 9a75ce07 c4bb3092 e466998c f4ec7582 7168520e c68fbfb0 a8f5cd44 85f82f9d 777019b6 255999f7 ac0e8301 f3599e46 af50ad10 4904307c 5689e883 05767f40

davidfowl:
This must have the record for most commits associated with bugs

PR #2282, opened 2015-07-16
Pass OS from native to managed

BrennanConroy:
Prevents us from having first-chance exception when checking for OS on Windows.
#2265
@davidfowl @moozzyk

### DIFF
--- a/src/dnx.host/RuntimeEnvironment.cs
+++ b/src/dnx.host/RuntimeEnvironment.cs
@@ -23,6 +23,12 @@ namespace Microsoft.Framework.Runtime
             RuntimeType = Type.GetType("Mono.Runtime") == null ? "CLR" : "Mono";
             RuntimeArchitecture = Environment.Is64BitProcess ? "x64" : "x86";
 #endif
+
+            // This is a temporary workaround until we pass a struct with OS information from native code
+            if (Environment.GetEnvironmentVariable(EnvironmentNames.DnxIsWindows) == "1")
+            {
+                _osName = "Windows";
+            }
         }
 
         public string OperatingSystem
--- a/src/dnx.mono.managed/EntryPoint.cs
+++ b/src/dnx.mono.managed/EntryPoint.cs
@@ -58,6 +58,12 @@ public class EntryPoint
         // Set the default lib path to be next to the entry point location
         Environment.SetEnvironmentVariable(EnvironmentNames.DefaultLib, Path.GetDirectoryName(typeof(EntryPoint).Assembly.Location));
         Environment.SetEnvironmentVariable(EnvironmentNames.ConsoleHost, "1");
+        
+        var p = Environment.OSVersion.Platform;
+        if (p != PlatformID.MacOSX && p != PlatformID.Unix)
+        {
+            Environment.SetEnvironmentVariable(EnvironmentNames.DnxIsWindows, "1");
+        }
 
         arguments = ExpandCommandLineArguments(arguments);
 
--- a/src/dnx/dnx.cpp
+++ b/src/dnx/dnx.cpp
@@ -281,6 +281,7 @@ int CallApplicationProcessMain(int argc, dnx::char_t* argv[], dnx::trace_writer&
         _X("dnx.coreclr.so");
 #else
         _X("dnx.clr.dll");
+        SetEnvironmentVariable(_X("DNX_IS_WINDOWS"), _X("1"));
 #endif
 
         // Note: need to keep as ASCII as GetProcAddress function takes ASCII params
--- a/src/Microsoft.Framework.Runtime.Sources/Impl/EnvironmentNames.cs
+++ b/src/Microsoft.Framework.Runtime.Sources/Impl/EnvironmentNames.cs
@@ -21,5 +21,6 @@ namespace Microsoft.Framework.Runtime
         public const string BuildKeyFile = "DNX_BUILD_KEY_FILE";
         public const string BuildDelaySign = "DNX_BUILD_DELAY_SIGN";
         public const string Sources = "DNX_SOURCES";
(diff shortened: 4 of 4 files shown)


src/dnx.mono.managed/EntryPoint.cs
```diff
@@ -58,6 +58,19 @@ public static int Main(string[] arguments)
         // Set the default lib path to be next to the entry point location
         Environment.SetEnvironmentVariable(EnvironmentNames.DefaultLib, Path.GetDirectoryName(typeof(EntryPoint).Assembly.Location));
         Environment.SetEnvironmentVariable(EnvironmentNames.ConsoleHost, "1");
+        
+        var p = (int)Environment.OSVersion.Platform;
+        string os;
+        if ((p != 4) && (p != 6) && (p != 128))
```
moozzyk: What is 128? And why we can't just use the [PlatformId enum](https://msdn.microsoft.com/en-us/library/3a8hyw88(v=vs.110).aspx)?

src/dnx.mono.managed/EntryPoint.cs
```diff
@@ -58,6 +58,19 @@ public static int Main(string[] arguments)
...
+        {
+            os = "Windows";
+        } else if (p == 6)
```
moozzyk: Fix formatting. (Same below)

moozzyk:
Create a bug to fix that properly (or update #1970 with the discussion we had today)
I assume you will send a change for Helios soon.

src/dnx.mono.managed/EntryPoint.cs
BrennanConroy: I was using the old way we did it, https://github.com/aspnet/dnx/commit/79b8ee24e8ea5a241687756a28c0be1a057ecb1b#diff-c889083e2a55b6b0e4ad6cc6e256f8c4L404

moozzyk: Yeah - I still don't know why we could not use something that is readable...

victorhurdugaci: From [here](http://www.mono-project.com/docs/faq/technical/):

> The execution platform can be detected by using the System.Environment.OSVersion.Platform value. However correctly detecting Unix platforms, in every cases, requires a little more work. The first versions of the framework (1.0 and 1.1) didn’t include any PlatformID value for Unix, so Mono used the value 128. The newer framework 2.0 added Unix to the PlatformID enum but, sadly, with a different value: 4 and newer versions of .NET distinguished between Unix and MacOS X, introducing yet another value 6 for MacOS X.

moozzyk: I found: `It means Unix under the 1.x .NET profile;` - so we don't have to check against 128

BrennanConroy:
@moozzyk 
Talked with Fowler, and we doing the passing Environment variable

src/dnx/dnx.cpp
```diff
@@ -281,6 +281,7 @@ int CallApplicationProcessMain(int argc, dnx::char_t* argv[], dnx::trace_writer&
         _X("dnx.coreclr.so");
 #else
         _X("dnx.clr.dll");
+        SetEnvironmentVariable(_X("DNX_RUNTIME_OS"), _X("1"));
```
anurse: Since the flag indicates that the OS is Windows, perhaps a better name would be `DNX_IS_WINDOWS` or something? I spent a few minutes trying to figure out how `1` could be the name of an OS :).

anurse:
Environment variable seems really gross (though I understand the short-term fix). Why are we not planning to support a cleaner way to pass this data @davidfowl ? As @moozzyk said, we already have #1970 to cover this.

moozzyk:
@anurse there is more than just OS to be passed around. In addition to OS we should be passing architecture and DNX_DEFAULT_LIB. We want to do this in one go and also scope this change as much as possbile because it is for beta6

moozzyk:
:ship: :it:

I suppose that because we pass OS from native code in the final change we will actually pass Windows/Darwin/Linux (something you had before) instead of just "IsWindows". Scoping to just to "IsWindows" does not have any advantage over returning actual OS. We would initially only check for Windows but we could check for other OSes if we needed.

src/dnx/dnx.cpp
davidfowl: Yes please :smile:

davidfowl:
@BrennanConroy don't tag PRs just the issues

anurse:
@moozzyk I totally agree, I was referring to @BrennanConroy 's comment that @davidfowl said to just use the environment variable. If that was just as a temporary measure until #1970 can be fixed, I'm totally :+1: to that :)

src/dnx.host/RuntimeEnvironment.cs
```diff
@@ -23,6 +23,12 @@ public RuntimeEnvironment()
             RuntimeType = Type.GetType("Mono.Runtime") == null ? "CLR" : "Mono";
             RuntimeArchitecture = Environment.Is64BitProcess ? "x64" : "x86";
 #endif
+
+            // This is a temporary workaround until we pass a struct with OS information from native code
+            if (Environment.GetEnvironmentVariable(EnvironmentNames.OperatingSystemName) == "1")
```
victorhurdugaci: This is confusing. "if the operating system **name** is '1' then ..." . One as in "one core?"

anurse: See my comment above. It's actually just used to indicate that the OS is Windows so `uname` should **not** be called. I suggest a new name like `DNX_IS_WINDOWS`.

victorhurdugaci: Sorry, I should have used the troll face for the second part of my comment :)

Issue #2235, opened 2015-07-09
We need a mechanism to pass exception details to native code


moozzyk:
To "fix" #1825 and prevent from Access Violations we catch the exception before it escapes managed code and turn into an error/HRESULT. This way we are loosing all exception details. In addition native parts of the code just write to the stdout in case of errors but in some scenarios (like Helios) it's useless because there isn't really stdout. 
The idea is to create another dll which would contain an error sink. It will be used by both managed Managed (P/Invoke) and native code to report errors that occurred during application startup. Users should also be able to plug-in to the error sink to be able to observe these errors.

Some additional context here: https://github.com/aspnet/Helios/pull/179

davidfowl:
This is pretty important to do. Any exception that happens before the application boots up shows:

![image](https://cloud.githubusercontent.com/assets/95136/9022279/b28c7d14-3824-11e5-8270-b14b513a7c5d.png)

PR #2283, opened 2015-07-16
fix publishing issues with dnx452/dnx46 and Helios

anurse:
* select correct framework when restoring other projects
* allow overriding the location of AspNet.Loader to simplify Helios build
* also, log errors in the bootstrapper to avoid losing them when transitioning back to native code in hosts like Helios.

/cc @davidfowl @ChengTian @muratg 

aspnet/Helios#185 is blocked on this

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Publish/DependencyContext.cs
+++ b/src/Microsoft.Framework.PackageManager/Publish/DependencyContext.cs
@@ -51,6 +51,14 @@ namespace Microsoft.Framework.PackageManager.Publish
             PackageAssemblies = NuGetDependencyResolver.PackageAssemblyLookup.Values.ToLookup(a => a.Library.Identity.Name);
         }
 
+        public static FrameworkName SelectFrameworkNameForRuntime(IEnumerable<FrameworkName> availableFrameworks, FrameworkName currentFramework, string runtime)
+        {
+            // Filter out frameworks incompatible with the current framework before selecting
+            return SelectFrameworkNameForRuntime(
+                availableFrameworks.Where(f => VersionUtility.IsCompatible(currentFramework, f)), 
+                runtime);
+        }
+
         public static FrameworkName SelectFrameworkNameForRuntime(IEnumerable<FrameworkName> availableFrameworks, string runtime)
         {
             var parts = runtime.Split(new[] { '.' }, 2);
--- a/src/Microsoft.Framework.PackageManager/Publish/PublishProject.cs
+++ b/src/Microsoft.Framework.PackageManager/Publish/PublishProject.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Runtime.Versioning;
 using System.Threading.Tasks;
 using System.Xml;
 using System.Xml.Linq;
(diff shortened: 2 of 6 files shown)


src/Microsoft.Framework.PackageManager/Publish/PublishProject.cs
```diff
@@ -597,20 +604,27 @@ private static void CopyAspNetLoaderDll(PublishRoot root, string wwwRootOutPath)
...
-                return;
+                // Copy Microsoft.AspNet.Loader.IIS.Interop/tools/*.dll into bin to support AspNet.Loader.dll
+                var package = root.Packages.SingleOrDefault(
+                    x => string.Equals(x.Library.Name, "Microsoft.AspNet.Loader.IIS.Interop"));
+                if (package == null)
+                {
+                    return;
+                }
+
+                var resolver = new DefaultPackagePathResolver(root.SourcePackagesPath);
```
troydai: :+1:

anurse:
Looking in to why Travis is sad.

src/Microsoft.Framework.PackageManager/Publish/DependencyContext.cs
```diff
@@ -51,6 +51,14 @@ public void Walk(string projectName, SemanticVersion projectVersion)
             PackageAssemblies = NuGetDependencyResolver.PackageAssemblyLookup.Values.ToLookup(a => a.Library.Identity.Name);
         }
 
+        public static FrameworkName SelectFrameworkNameForRuntime(IEnumerable<FrameworkName> availableFrameworks, FrameworkName currentFramework, string runtime)
+        {
+            // Filter out frameworks incompatible with the current framework before selecting
+            return SelectFrameworkNameForRuntime(
```
davidfowl: Again

src/dnx.host/RuntimeBootstrapper.cs
```diff
@@ -58,90 +58,97 @@ private static void PrintErrors(Exception ex)
 
         public static Task<int> ExecuteAsync(string[] args, FrameworkName targetFramework)
         {
-            var app = new CommandLineApplication(throwOnUnexpectedArg: false);
```
davidfowl: Why are there changes here?

src/Microsoft.Framework.PackageManager/Publish/DependencyContext.cs
anurse: It is unified. It's calling the same function.

src/dnx.host/RuntimeBootstrapper.cs
anurse: Partly because it was how I was able to actually figure out what was wrong, since the Exception was normally being lost in the ether. I think a useful thing to keep in though so I left it. Without this, if an exception is thrown while booting DNX/Helios, it is lost to the ether. Now, at least, if there is a Console attached it can go there. It's not perfect, but it's something to help debugging when you're running `iisexpress` from the console.

src/Microsoft.Framework.PackageManager/Publish/DependencyContext.cs
anurse: Files #2285 to help unify this further though.

anurse:
Travis was sad because Mono only supports one framework but I was running the multi-framework publish test on it. I disabled it and it should get happy again.

Any further comments? I'd like to close on this and give the CI time to settle before the noon deadline for release :). :watch: @davidfowl @troydai

src/dnx.host/RuntimeBootstrapper.cs
moozzyk: It is related to aspnet/dnx#2235 and https://github.com/aspnet/Helios/issues/180

Issue #2291, opened 2015-07-20
"dnu restore" caching behavior regression


ChengTian:
`dnu restore` doesn't cache anything after you delete `%userprofile%/AppData/Local/dnu/cache`

![capture](https://cloud.githubusercontent.com/assets/1383883/8784126/2b132a84-2ed5-11e5-9247-4e01de305447.PNG)

PR #2292, opened 2015-07-20
Fix dnu caching behavior regression

ChengTian:
parent https://github.com/aspnet/dnx/issues/2291

@davidfowl , we must fix it in beta 6


ChengTian:
Simple fix and I already verified it. I am going to merge it @muratg

ChengTian:
Checked into release and merged back to dev

Issue #1501, opened 2015-03-26
Kpm bundle with configuration as 'release' is not honored when built from source.


vijayrkn:
Runtime does not honor the configuration passed to kpm bundle when built from the source.

Repro:
kpm.cmd bundle "C:\WebApplication2" --out "C:\Temp\AspNetPublish\WebApplication2-89" --configuration Release --runtime dnx-clr-win-x86.1.0.0-beta4-11296 --wwwroot-out "wwwroot" --quiet

troydai:
To keep up with the progress check out branch https://github.com/aspnet/dnx/tree/trdai/1501

troydai:
Fixed in 33eccd43e5fa250c2650744b6d28140cd42096a7

PR #2293, opened 2015-07-20
Adding configuration option into the bash and batch script generated by dnu publish

troydai:
Addressing: https://github.com/aspnet/dnx/issues/1501

### DIFF
--- a/src/Microsoft.Framework.PackageManager/Publish/PublishRoot.cs
+++ b/src/Microsoft.Framework.PackageManager/Publish/PublishRoot.cs
@@ -106,7 +106,7 @@ namespace Microsoft.Framework.PackageManager.Publish
             }
 
             const string template = @"
-@""{0}{1}.exe"" --appbase ""%~dp0{2}"" Microsoft.Framework.ApplicationHost {3} %*
+@""{0}{1}.exe"" --appbase ""%~dp0{2}"" Microsoft.Framework.ApplicationHost --configuration {3} {4} %*
 ";
 
             foreach (var commandName in _project.Commands.Keys)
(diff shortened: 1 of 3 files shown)


BrennanConroy:
:shipit:

anurse:
:shipit:

src/Microsoft.Framework.PackageManager/Publish/PublishRoot.cs
```diff
@@ -106,7 +106,7 @@ private void GenerateBatchFiles()
             }
 
             const string template = @"
-@""{0}{1}.exe"" --appbase ""%~dp0{2}"" Microsoft.Framework.ApplicationHost {3} %*
+@""{0}{1}.exe"" --appbase ""%~dp0{2}"" Microsoft.Framework.ApplicationHost --configuration {3} {4} %*
```
davidfowl: string interpolation

troydai: I don't want to update it in this PR so as to limit the scope of the change.

davidfowl: You totally should though

troydai: Do you mind if I merge this first then follow up the change to use string interpolation?

davidfowl:
Why is appveyor failing?

troydai:
AppVeyor timeout:
```
Build execution time has reached the maximum allowed time for your plan (40 minutes).
```
According to log, when it timeout the last test was running,

Issue #1548, opened 2015-03-30
Need a more descriptive message when an incorrect file globbing pattern is entered


Praburaj:
For example if I accidentally put a `*` in a globbing pattern that does not allow `*` e.g:
```
"compileFiles": [ "*.cs" ],
```

All we show in the error is this without any line number information, or pointers on what's wrong etc. We should print out a more descriptive error with some information on what exactly is wrong. 

![image](https://cloud.githubusercontent.com/assets/2294732/6908138/2215bcae-d6f0-11e4-9d0e-a8488583aa19.png)

Expected: `compileFile` cannot contain wildcard patterns. 

@troydai

troydai:
We may exam the 'literal' patterns more carefully while parsing project.json.

https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Framework.Runtime.Hosting/PatternGroup.cs#L80

troydai:
No, we didn't take action. The error is expected. We can have a better error message through.

troydai:
Fixed in https://github.com/aspnet/dnx/commit/a486da1669280f1f46ea3e11cf5e6f42c9a00133

PR #2313, opened 2015-07-22
Add a more descriptive message when wildcard is used in literal search pattern

troydai:
Addressing #1548 

image

### DIFF
--- a/src/Microsoft.Framework.Runtime/PatternsCollectionHelper.cs
+++ b/src/Microsoft.Framework.Runtime/PatternsCollectionHelper.cs
@@ -13,7 +13,12 @@ namespace Microsoft.Framework.Runtime
     {
         private static readonly char[] PatternSeparator = new[] { ';' };
 
-        public static IEnumerable<string> GetPatternsCollection(JsonObject rawProject, string projectDirectory, string projectFilePath, string propertyName, IEnumerable<string> defaultPatterns = null)
+        public static IEnumerable<string> GetPatternsCollection(JsonObject rawProject,
+                                                                string projectDirectory,
+                                                                string projectFilePath,
+                                                                string propertyName,
+                                                                IEnumerable<string> defaultPatterns = null,
+                                                                bool literalPath = false)
         {
             defaultPatterns = defaultPatterns ?? Enumerable.Empty<string>();
 
@@ -21,19 +26,19 @@ namespace Microsoft.Framework.Runtime
             {
                 if (!rawProject.Keys.Contains(propertyName))
                 {
-                    return CreateCollection(projectDirectory, defaultPatterns.ToArray());
+                    return CreateCollection(projectDirectory, propertyName, defaultPatterns, literalPath);
                 }
 
                 var valueInString = rawProject.ValueAsString(propertyName);
                 if (valueInString != null)
                 {
-                    return CreateCollection(projectDirectory, valueInString);
+                    return CreateCollection(projectDirectory, propertyName, new string[] { valueInString }, literalPath);
                 }
 
                 var valuesInArray = rawProject.ValueAsStringArray(propertyName);
                 if (valuesInArray != null)
                 {
-                    return CreateCollection(projectDirectory, valuesInArray.Select(s => s.ToString()).ToArray());
+                    return CreateCollection(projectDirectory, propertyName, valuesInArray.Select(s => s.ToString()), literalPath);
                 }
             }
             catch (Exception ex)
@@ -44,7 +49,7 @@ namespace Microsoft.Framework.Runtime
             throw FileFormatException.Create("Value must be either string or array.", rawProject.Value(propertyName), projectFilePath);
         }
 
-        private static IEnumerable<string> CreateCollection(string projectDirectory, params string[] patternsStrings)
+        private static IEnumerable<string> CreateCollection(string projectDirectory, string propertyName, IEnumerable<string> patternsStrings, bool literalPath)
         {
             var patterns = patternsStrings.SelectMany(patternsString => GetSourcesSplit(patternsString))
                                           .Select(patternString => patternString.Replace('/', Path.DirectorySeparatorChar).Replace('\\', Path.DirectorySeparatorChar));
@@ -53,7 +58,12 @@ namespace Microsoft.Framework.Runtime
             {
                 if (Path.IsPathRooted(pattern))
                 {
-                    throw new InvalidOperationException(string.Format("Patten {0} is a rooted path, which is not supported.", pattern));
+                    throw new InvalidOperationException($"The '{propertyName}' property cannot be a rooted path.");
+                }
+
+                if (literalPath && pattern.Contains('*'))
+                {
+                    throw new InvalidOperationException($"The '{propertyName}' property cannot contain wildcard characters.");
                 }
             }
 
--- a/test/Microsoft.Framework.Runtime.Tests/FileGlobbing/ProjectFilesCollectionFacts.cs
+++ b/test/Microsoft.Framework.Runtime.Tests/FileGlobbing/ProjectFilesCollectionFacts.cs
@@ -69,7 +69,7 @@ namespace Microsoft.Framework.Runtime.Tests.FileGlobbing
              ""compileFiles"": ""signle.cs"",
              ""shared"": ""shared/**/*.cs"",
              ""sharedExclude"": ""excludeShared*.cs"",
-             ""sharedFiles"": ""**/*.cs"",
+             ""sharedFiles"": ""shared.cs"",
              ""contentBuiltIn"": """",
              ""content"": ""**/*.content"",
              ""contentExclude"": ""excludecontent"",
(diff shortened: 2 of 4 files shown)


troydai:
Working on adding test.

troydai:
/review @ChengTian @victorhurdugaci @BrennanConroy 
/cc @davidfowl @muratg 

Tested in latest VS and VS Code on Omnisharp.

test/Microsoft.Framework.Runtime.Tests/FileGlobbing/ProjectFilesCollectionFacts.cs
```diff
@@ -111,6 +111,25 @@ public void RewriteCompileBuiltIn()
...
+         }");
+
+            var exception = Assert.Throws<FileFormatException>(() =>
+            {
+                var target = new ProjectFilesCollection(rawProject, string.Empty, string.Empty);
+            });
+
+            Assert.Equal(
+                exception.InnerException.Message,
+                $"Pattern *.cs is not expected to contain wild-card.");
```
moozzyk: nit: remove '$'

test/Microsoft.Framework.Runtime.Tests/FileGlobbing/ProjectFilesCollectionFacts.cs
```diff
@@ -111,6 +111,25 @@ public void RewriteCompileBuiltIn()
...
+        public void ExceptionThrowWhenWildcardPresentsInLiteralPath()
+        {
+            var rawProject = Deserialize(@"
+         {
+             ""compileFiles"": ""*.cs""
+         }");
+
+            var exception = Assert.Throws<FileFormatException>(() =>
+            {
+                var target = new ProjectFilesCollection(rawProject, string.Empty, string.Empty);
```
moozzyk: nit: is `target` needed at all?

test/Microsoft.Framework.Runtime.Tests/FileGlobbing/ProjectFilesCollectionFacts.cs
```diff
@@ -111,6 +111,25 @@ public void RewriteCompileBuiltIn()
             Assert.Equal(NormalizePatterns("*.cs", "../*.cs", "**/*.cpp", "**/*.h"), target.CompilePatternsGroup.IncludePatterns);
         }
 
+
+        [Fact]
+        public void ExceptionThrowWhenWildcardPresentsInLiteralPath()
+        {
+            var rawProject = Deserialize(@"
+         {
```
moozzyk: formatting seems to be off

ChengTian:
Is there any specific reason that we disallow wildcard globbing for `compileFiles`? How do users know when they can use wildcard and when they cannot? I remember you can use * in `exclude` patterns and the inconsistency here seems confusing

src/Microsoft.Framework.Runtime/PatternsCollectionHelper.cs
```diff
@@ -53,7 +58,12 @@ private static IEnumerable<string> CreateCollection(string projectDirectory, par
             {
                 if (Path.IsPathRooted(pattern))
                 {
-                    throw new InvalidOperationException(string.Format("Patten {0} is a rooted path, which is not supported.", pattern));
+                    throw new InvalidOperationException($"Pattern {pattern} is a rooted path, which is not supported.");
+                }
+
+                if (literalPath && pattern.Contains('*'))
+                {
+                    throw new InvalidOperationException($"Pattern {pattern} is not expected to contain wild-card.");
```
moozzyk: I think it shold be ".. to contain (a wildcard|wildcard characters)."

src/Microsoft.Framework.Runtime/PatternsCollectionHelper.cs
```diff
@@ -21,19 +26,19 @@ public static IEnumerable<string> GetPatternsCollection(JsonObject rawProject, s
...
                 {
-                    return CreateCollection(projectDirectory, defaultPatterns.ToArray());
+                    return CreateCollection(projectDirectory, defaultPatterns, literalPath);
                 }
 
                 var valueInString = rawProject.ValueAsString(propertyName);
                 if (valueInString != null)
                 {
-                    return CreateCollection(projectDirectory, valueInString);
+                    return CreateCollection(projectDirectory, new string[] { valueInString }, literalPath);
```
moozzyk: nit: try just: `new[] { valueInString }`

src/Microsoft.Framework.Runtime/PatternsCollectionHelper.cs
```diff
@@ -13,7 +13,12 @@ internal static class PatternsCollectionHelper
     {
         private static readonly char[] PatternSeparator = new[] { ';' };
 
-        public static IEnumerable<string> GetPatternsCollection(JsonObject rawProject, string projectDirectory, string projectFilePath, string propertyName, IEnumerable<string> defaultPatterns = null)
+        public static IEnumerable<string> GetPatternsCollection(JsonObject rawProject,
+                                                                string projectDirectory,
+                                                                string projectFilePath,
+                                                                string propertyName,
+                                                                IEnumerable<string> defaultPatterns = null,
+                                                                bool literalPath = false)
```
moozzyk: the default value is a bit gross here... how many places does it affect?

src/Microsoft.Framework.Runtime/PatternsCollectionHelper.cs
troydai: Will update.

src/Microsoft.Framework.Runtime/PatternsCollectionHelper.cs
troydai: There're a few places. Plus the parameters list is already long and in most cases it is not a literal path.

troydai:
@ChengTian `compileFiles` is different from other patterns. It accepts relative paths to files only. Wildcards are not expected to present.

moozzyk:
_Wildcards are not expected to present. It's documented here: https://github.com/aspnet/Home/wiki/Project.json-file#rules_ Is it possible to update the project.json schema so that it shows this in the corresponding help tooltip (if it is not there already) .

src/Microsoft.Framework.Runtime/PatternsCollectionHelper.cs
troydai: No. The type of `valueInString` is not `string`. It's and `JsonString`. Here implicit type conversion is involved.

BrennanConroy:
Add test for rooted path

troydai:
Root path exception is not new.

troydai:
Ping @moozzyk @ChengTian @BrennanConroy :watch:

src/Microsoft.Framework.Runtime/PatternsCollectionHelper.cs
```diff
@@ -53,7 +58,12 @@ private static IEnumerable<string> CreateCollection(string projectDirectory, par
             {
                 if (Path.IsPathRooted(pattern))
                 {
-                    throw new InvalidOperationException(string.Format("Patten {0} is a rooted path, which is not supported.", pattern));
+                    throw new InvalidOperationException($"Pattern {pattern} is a rooted path, which is not supported.");
+                }
+
+                if (literalPath && pattern.Contains('*'))
+                {
+                    throw new InvalidOperationException($"Pattern {pattern} is not expected to contain wildcard characters.");
```
anurse: I think a message like this is a little clearer: 

```
The '{propertyName}' property cannot contain wildcard characters, use '{wildcardPropertyName}' to specify wildcard patterns.
```

For example:

```
The 'compileFiles' property cannot contain wildcard characters, use 'compile' to specify wildcard patterns.
```

It does mean you need to pass down `propertyName` and add a `wildcardPropertyName` parameter, but I think the clearer error message is worth it.

anurse:
Just because it isn't new, doesn't mean this isn't a great opportunity to improve test coverage ;P.

src/Microsoft.Framework.Runtime/PatternsCollectionHelper.cs
ChengTian: :+1:

troydai:
Test for rooted path exception was already there.
https://github.com/aspnet/dnx/blob/dev/test/Microsoft.Framework.Runtime.FunctionalTests/ProjectFileGlobbing/ProjectFilesCollectionTests.cs#L389

src/Microsoft.Framework.Runtime/PatternsCollectionHelper.cs
troydai: It's not needed.

The exception comes with line and column information. When error is shown in the VS the error list will direct you to the exactly line of code. In addition, the line and column are also shown when application is running from command line.

Given that, I think the property name is important but not significant enough to change the signature.

anurse: It's a private static method and you have access to the value in all the immediate callers. It seems like a small amount of effort for a much clearer error. We can drop the second part (showing the wildcard property name) since that will require a more dramatic amount of refactoring.

Regardless of how well we illustrate where the error is, we still need to have a clear description of the problem.

anurse:
Ah, then ignore my comment :) :+1:

src/Microsoft.Framework.Runtime/PatternsCollectionHelper.cs
troydai: deal. checkout the latest commit.

test/Microsoft.Framework.Runtime.Tests/FileGlobbing/ProjectFilesCollectionFacts.cs
```diff
@@ -111,6 +111,22 @@ public void RewriteCompileBuiltIn()
...
+        {
+            var rawProject = Deserialize(@"{""compileFiles"": ""*.cs""}");
+
+            var exception = Assert.Throws<FileFormatException>(() =>
+            {
+                var target = new ProjectFilesCollection(rawProject, string.Empty, string.Empty);
+            });
+
+            Assert.Equal(
+                "Pattern *.cs is not expected to contain wildcard characters.",
```
victorhurdugaci: Isn't this exception message supposed to be the same as the one on line 65 in `PatternsCollectionHelper` ?

troydai: yeah ...

troydai:
:watch:

src/Microsoft.Framework.Runtime/PatternsCollectionHelper.cs
anurse: There's no handshake emoji. That makes me sad.

Anyway, great! :+1:

anurse:
:shipit:

PR #2316, opened 2015-07-23
Adding --project|-p, removing `.` from dnx

moozzyk:
`dnx {path} run` becomes `dnx -p {path} run` or `dnx --project {path} run`
`dnx . run` is no longer supported and is replaced by `dnx run` or `dnx -p . run`
`dnx {path}\project.json run` is no longer supported and is replaced by `dnx -p {path}\project.json`

### DIFF
--- a/src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
+++ b/src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
@@ -89,12 +89,29 @@ public class EntryPoint
             return arguments;
         }
 
+        var argExpanded = false;
         var expandedArgs = new List<string>();
         for (var i = 0; i < arguments.Count(); i++)
         {
-            if (i == parameterIdx)
+            if (!argExpanded)
             {
-                ExpandArgument(arguments[i], expandedArgs);
+                if (string.Equals(arguments[i], "-p", StringComparison.OrdinalIgnoreCase) ||
+                    string.Equals(arguments[i], "--project", StringComparison.OrdinalIgnoreCase))
+                {
+                    // Note that ++i is safe here since if we had a trailing -p/--project we would have exited
+                    // before entering the loop because we wouldn't have found any non-host option
+                    ExpandProject(arguments[++i], expandedArgs);
+                    argExpanded = true;
+                }
+                else if (i == parameterIdx)
+                {
+                    ExpandNonHostArgument(arguments[i], expandedArgs);
+                    argExpanded = true;
+                }
+                else
+                {
+                    expandedArgs.Add(arguments[i]);
+                }
             }
             else
             {
@@ -105,7 +122,22 @@ public class EntryPoint
         return expandedArgs.ToArray();
     }
 
-    private static void ExpandArgument(string argument, List<string> expandedArgs)
+    private static void ExpandProject(string projectPath, List<string> expandedArgs)
+    {
+        expandedArgs.Add("--appbase");
+        if (string.Equals(Path.GetFileName(projectPath), "project.json", StringComparison.OrdinalIgnoreCase))
+        {
+            expandedArgs.Add(Path.GetDirectoryName(Path.GetFullPath(projectPath)));
+        }
+        else
+        {
+            expandedArgs.Add(Path.GetFullPath(projectPath));
+        }
+
+        expandedArgs.Add("Microsoft.Dnx.ApplicationHost");
+    }
+
+    private static void ExpandNonHostArgument(string argument, List<string> expandedArgs)
     {
         expandedArgs.Add("--appbase");
 
@@ -117,21 +149,6 @@ public class EntryPoint
             expandedArgs.Add(Path.GetDirectoryName(Path.GetFullPath(argument)));
             expandedArgs.Add(argument);
 
-            return;
-        }
-
-        if (argument.Equals(".", StringComparison.Ordinal))
-        {
-            // "dnx . run" --> "dnx --appbase . Microsoft.Dnx.ApplicationHost run"
-            expandedArgs.Add(argument);
-            expandedArgs.Add("Microsoft.Dnx.ApplicationHost");
-            return;
-        }
-
-        if (string.Equals(Path.GetFileName(argument), "project.json", StringComparison.OrdinalIgnoreCase))
-        {
-            expandedArgs.Add(Path.GetDirectoryName(Path.GetFullPath(argument)));
-            expandedArgs.Add("Microsoft.Dnx.ApplicationHost");
             return;
         }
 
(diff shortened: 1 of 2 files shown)


moozzyk:
All checks will fail since Universe has to be updated for this (https://github.com/aspnet/Universe/pull/252)

src/dnx.host/RuntimeBootstrapper.cs
```diff
@@ -64,6 +64,8 @@ public static Task<int> ExecuteAsync(string[] args, FrameworkName targetFramewor
 
             // These options were handled in the native code, but got passed through here.
             // We just need to capture them and clean them up.
+            var optionProject = app.Option("--project|-p", "Path to the project.json file or the application folder. Defaults to the current folder if not provided.",
+                CommandOptionType.NoValue);
```
troydai: I found the option `CommandOptionType.NoValue` misleading. Will change it to `SingleValue` make this option mandatory?

src/dnx.mono.managed/EntryPoint.cs
```diff
@@ -89,12 +89,28 @@ private static string[] ExpandCommandLineArguments(string[] arguments)
...
 
+        var argExpanded = false;
         var expandedArgs = new List<string>();
         for (var i = 0; i < arguments.Count(); i++)
         {
-            if (i == parameterIdx)
+            if (!argExpanded)
             {
-                ExpandArgument(arguments[i], expandedArgs);
+                if (arguments[i] == "-p" || arguments[i] == "--project")
```
troydai: So we go strictly case sensitive in terms of option names?

src/dnx/dnx.cpp
```diff
@@ -66,7 +66,9 @@ int BootstrapperOptionValueNum(const dnx::char_t* pszCandidate)
         strings_equal_ignore_case(pszCandidate, _X("--lib")) ||
         strings_equal_ignore_case(pszCandidate, _X("--packages")) ||
         strings_equal_ignore_case(pszCandidate, _X("--configuration")) ||
-        strings_equal_ignore_case(pszCandidate, _X("--port")))
+        strings_equal_ignore_case(pszCandidate, _X("--port")) ||
+        strings_equal_ignore_case(pszCandidate, _X("--project")) ||
```
troydai: But we case insensitive here ...

troydai:
Sorry for mention the case-sensitive issue of option names again. I understand we need a further discussion on how to unify the method. So it is not a blocking issue in this PR.

src/dnx.host/RuntimeBootstrapper.cs
moozzyk: I merely wanted this option to be displayed in help. This is because -p should not get out of dnx  since it is converted to --appbase. However there are degenerate cases where --appbase and -p is provided or -p is provided twice where -p can actually get here. I will look a bit more into this code because NoValue feels wrong indeed. Thanks for pointing this out.

src/dnx.mono.managed/EntryPoint.cs
moozzyk: No, this should be aligned with what we already have. I will fix this to be case insensitive.

moozzyk:
That was actually a good catch - even though it is a new option we should be consistent and treat it the same way as other options.

src/dnx/dnx.cpp
```diff
@@ -85,6 +87,8 @@ int BootstrapperOptionValueNum(const dnx::char_t* pszCandidate)
     return -1;
 }
 
+
```
BrennanConroy: Newlines

src/dnx/dnx.cpp
```diff
@@ -135,35 +139,34 @@ void AppendAppbaseFromFile(const dnx::char_t* path, std::vector<const dnx::char_
...
-        AppendAppbaseFromFile(value, expanded_args);
-        expanded_args.push_back(allocate_and_copy(value));
+    auto split_idx = split_path(project_path);
 
-        return;
-    }
-
-    if (strings_equal_ignore_case(value, _X(".")))
+    // note that we split the path first and check the file name to handle paths like c:\MyApp\my_project.json
+    if (string_ends_with_ignore_case(project_path + (split_idx < 0 ? 0 : split_idx), _X("project.json")))
```
BrennanConroy: Wouldn't this pass in the case of my_project.json?

moozzyk: LOL given the comment just above... Should be `strings_equal_ignore_case` and not `ends with`.  Also need to add a test for this.

BrennanConroy:
Looks like some tests are broken

src/dnx/dnx.cpp
```diff
@@ -135,35 +137,34 @@ void AppendAppbaseFromFile(const dnx::char_t* path, std::vector<const dnx::char_
...
-        AppendAppbaseFromFile(value, expanded_args);
-        expanded_args.push_back(allocate_and_copy(value));
+    auto split_idx = split_path(project_path);
 
-        return;
-    }
-
-    if (strings_equal_ignore_case(value, _X(".")))
+    // note that we split the path first and check the file name to handle paths like c:\MyApp\my_project.json
+    if (strings_equal_ignore_case(project_path + (split_idx < 0 ? 0 : split_idx + 1), _X("project.json")))
```
moozzyk: `project_path + (split_idx < 0 ? 0 : split_idx + 1)` could be just `project_path + split_idx + 1` but requires a comment

moozzyk:
Rebased on latest renames.

src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
```diff
@@ -105,7 +122,22 @@ private static string[] ExpandCommandLineArguments(string[] arguments)
...
+        if (string.Equals(Path.GetFileName(projectPath), "project.json", StringComparison.OrdinalIgnoreCase))
+        {
+            expandedArgs.Add(Path.GetDirectoryName(Path.GetFullPath(projectPath)));
+        }
+        else
+        {
+            expandedArgs.Add(Path.GetFullPath(projectPath));
+        }
+
+        expandedArgs.Add("Microsoft.Framework.ApplicationHost");
```
victorhurdugaci: Wrong name. Should be `Microsoft.Dnx.ApplicationHost`

victorhurdugaci:
Other than my comment, it looks shippable :shipit:

src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
moozzyk: Good catch! Will fix this.

davidfowl:
@moozzyk What's the state of this change?

moozzyk:
@davidfowl I was waiting for a good build after all the renames and this cannot be checked in in the afternoon. Just pushed this.

Issue #2162, opened 2015-07-01
Impossible is possible so we need a message


moozzyk:
Depspite the "this should never happen" comment this code:
https://github.com/aspnet/dnx/blob/f1b13799a4e165147827ebecb8dff00f6f48e677/src/Microsoft.Framework.Runtime.Compilation.DesignTime/DesignTimeHostCompiler.cs#L79-L80

may throw which results in a YSOD that is not giving any hints to the user what went wrong. There are no errors in VS whatsoever as well but you just can not make the application work/run. This is what you will see if the exception is thrown:

```
Server Error in '/' Application.

Operation is not valid due to the current state of the object.

Description: An unhandled exception occurred during the execution of the current web request. Please review the stack trace for more information about the error and where it originated in the code. 

Exception Details: System.InvalidOperationException: Operation is not valid due to the current state of the object.

Source Error: 

An unhandled exception was generated during the execution of the current web request. Information regarding the origin and location of the exception can be identified using the exception stack trace below.

Stack Trace: 

[InvalidOperationException: Operation is not valid due to the current state of the object.]

[AggregateException: One or more errors occurred.]

[FileLoadException: Could not load file or assembly 'Microsoft.AspNet.Loader.IIS' or one of its dependencies. General Exception (Exception from HRESULT: 0x80131500)]

[TargetInvocationException: Exception has been thrown by the target of an invocation.]
   System.Web.HttpRuntime.HostingInit(HostingEnvironmentFlags hostingFlags, PolicyLevel policyLevel, Exception appDomainCreationException) +303

[HttpException (0x80004005): Exception has been thrown by the target of an invocation.]
   System.Web.HttpRuntime.FirstRequestInit(HttpContext context) +9924184
   System.Web.HttpRuntime.EnsureFirstRequestInit(HttpContext context) +101
   System.Web.HttpRuntime.ProcessRequestNotificationPrivate(IIS7WorkerRequest wr, HttpContext context) +261
```

TL;DR;

[...]

PR #2323, opened 2015-07-23
Add meaningful message for inconsistent sdk versions issue

troydai:
Address #2162

### DIFF
--- a/src/Microsoft.Framework.Runtime.Compilation.DesignTime/DesignTimeHostCompiler.cs
+++ b/src/Microsoft.Framework.Runtime.Compilation.DesignTime/DesignTimeHostCompiler.cs
@@ -76,8 +76,7 @@ namespace Microsoft.Framework.Runtime.Compilation.DesignTime
             int contextId;
             if (!contexts.TryGetValue(projectPath, out contextId))
             {
-                // This should never happen
-                throw new InvalidOperationException();
+                throw new InvalidOperationException($"The context of the project at {projectPath} could not be determined. It could be caused by multiple global.json files specifying different versions of the SDK for the same project.");
             }
 
             _queue.Send(new GetCompiledAssemblyMessage


src/Microsoft.Framework.Runtime.Compilation.DesignTime/DesignTimeHostCompiler.cs
```diff
@@ -77,7 +77,7 @@ private void OnError(int? contextId, string error)
             if (!contexts.TryGetValue(projectPath, out contextId))
             {
                 // This should never happen
```
moozzyk: Remove this comment.

troydai:
:watch:

src/Microsoft.Framework.Runtime.Compilation.DesignTime/DesignTimeHostCompiler.cs
```diff
@@ -76,8 +76,7 @@ private void OnError(int? contextId, string error)
             int contextId;
             if (!contexts.TryGetValue(projectPath, out contextId))
             {
-                // This should never happen
-                throw new InvalidOperationException();
+                throw new InvalidOperationException($"Failed to identify the context of a project at {projectPath}. It could be caused by multiple global.json introduced different version of sdk for the same project.");
```
moozzyk: I am thinking about a better message. This "Failed to identify the context" is quite cryptic.

Eilon: `$"The context of the project at {projectPath} could not be determined. It could be caused by multiple global.json files specifying different versions of the SDK for the same project.`

troydai: This is perfect ... copying ...

troydai:
Example: 
image

troydai:
:watch:  /cc @Eilon @moozzyk 
image

src/Microsoft.Framework.Runtime.Compilation.DesignTime/DesignTimeHostCompiler.cs
```diff
@@ -76,8 +76,7 @@ private void OnError(int? contextId, string error)
             int contextId;
             if (!contexts.TryGetValue(projectPath, out contextId))
             {
-                // This should never happen
-                throw new InvalidOperationException();
+                throw new InvalidOperationException($"The context of the project at {projectPath} could not be determined. It could be caused by multiple global.json files specifying different versions of the SDK for the same project.");
```
moozzyk: @Eilon - I am wondering if it is worth specifying that the case with multiple global.json files is when the user is using source references. If this is not worth mentioning then :ship::it:

Eilon: I don't actually know what this error is at all - I was just making it make more sense :smile: But given what you're saying I think this is fine for now - a big improvement over no error message at all :smile:

moozzyk: The problem happens if you have multiple global.json (which I understand can only happen if you are using source references) and the sdk versions in these are different. In this case VS will start multiple dnx.exe processes (one for each sdk version). When building the project VS will use the dnx.exe corresponding to the sdk version but since it is a single project building will fail because the part of the project built in one of the dnx.exe will not be able to find references to part of the project built in other dnx.exe

Eilon: How about: `$"The context of the project at {projectPath} could not be determined. This can happen if a project references other projects by source, and those projects have a global.json file specifying a different version of the SDK.`

troydai: Saw this after the merge. Will update the message.

Issue #1855, opened 2015-05-13
Rename and shuffle some of the core projects/assemblies


davidfowl:
Proposal:

merge:
- dnx.coreclr.managed
- dnx.clr.managed

into dnx.host with the appropriate ifdefs.

Renames
dnx.host -> Microsoft.Dnx.Host
Microsoft.Framework.ApplicationHost -> Microsoft.Dnx.ApplicationHost
Microsoft.Framework.DesignTimeHost -> Microsoft.Dnx.DesignTimeHost
Microsoft.Framework.DesignTimeHost.Abstractions -> Microsoft.Dnx.DesignTimeHost.Abstractions
Microsoft.Framework.Runtime -> Microsoft.Dnx.Runtime.Hosting
Microsoft.Framework.Runtime.Abstractions -> Microsoft.Dnx.Abstractions
Microsoft.Framework.Runtime.Caching -> Microsoft.Dnx.Compilation.Caching
Microsoft.Framework.Runtime.Loader -> Microsoft.Dnx.Loader

Compilation
Microsoft.Framework.Runtime.Roslyn -> Microsoft.Dnx.Compilation.CSharp
Microsoft.Framework.Runtime.Roslyn.Abstractions -> Microsoft.Dnx.Compilation.CSharp.Abstractions?
Microsoft.Framework.Runtime.Compilation -> Microsoft.Dnx.Compilation
Microsoft.Framework.Runtime.Compilation.DesignTime -> Microsoft.Dnx.Compilation.DesignTime

Package Manager
Make Microsoft.Framework.Project a separate global tool.
Microsoft.Framework.PackageManager -> Microsoft.Framework.PackageManager??

/cc @anurse @glennc @muratg @lodejard

muratg:
"Microsoft.Framework.Runtime.Roslyn.Abstractions -> Microsoft.Dnx.Compilation.**CSharp**.Abstractions?"

How about VB?  I propose we keep Roslyn in the name if we are going with these changes.

Also /cc @Eilon

Eilon:
These names are all fine by me.

Eilon:
For the Roslyn thing, we might want to not say "Roslyn" at all because that's more of a codename.

So, maybe like:

* `Microsoft.Dnx.Compilation.Abstractions`
* `Microsoft.Dnx.Compilation.Compiler.Abstractions`

?

muratg:
[...]

PR #2346, opened 2015-07-25
Microsoft.Framework -> Microsoft.Dnx renames

victorhurdugaci:
Fixes https://github.com/aspnet/dnx/issues/1855 
Fixes https://github.com/aspnet/Testing/issues/90 
Fixes https://github.com/aspnet/Testing/issues/125

- Latest WTE works with the new renames (including test discovery)
- All tests pass (with the changes in testing and aspnet.xunit)
- Travis/Appveyor will not pass because of bootstrapping issues (will hack the CI when I merge)

- Renames:

image

Please review: @anurse @davidfowl @troydai @muratg


davidfowl:
You missed `dnx.mono.managed`

anurse:
Break early, fix early :P. I say :shipit: with the `dnx.mono.managed` rename. Can't do a ton of reviewing because of the scale of the change, but the longer we wait, the more backed up things get.

davidfowl:
Don't do that. Fix the name. It's super easy.

troydai:
Didn't and can't possibly go through the changes. Given that runtime tests and WTE pass let's get this in as early as possible so that potential issues will surface earlier.

:shipit:

davidfowl:
Please rename the mono assembly in this change. It's localized to the dnx repository and will take 5 minutes.

muratg:
Let's do the change: `dnx.mono.managed` -> `Microsoft.Dnx.DomainManager.Mono` and :shipit: and find issues as early as possible.

anurse:
To be fair, I meant :shipit: AFTER fixing the Mono rename. Sorry if that wasn't clear.

davidfowl:
Lets do that and then rename again. I don't know where DomainManager comes from but only CLR has a domain manager and I don't even remember that name in the original rename list.

muratg:
It was discussed in the issue (https://github.com/aspnet/dnx/issues/1855) last week. Original list suggested merging those into `dnx.host`.  Someone (Pawel?) had a good reason why that's not a good idea.  I agree DomainManager is not exactly true in all of them.  

Any suggestions?  

Can you also fully review the table on the top to make sure you agree with all the names?

anurse:
To be fair, the CoreCLR has a "DomainManager" but it's not an App Domain Manager :).

I still like merging `dnx.mono.managed`, `dnx.coreclr.managed` and `dnx.host` into one assembly and only keeping `dnx.clr.managed` separate. The reason the CLR one has to be separate is that it has to have a strong-name. We want to keep it separate so that it can have a strong-name without requiring that every other assembly have a strong-name (due to virality). So the CLR entry point is an exception to the merge, but only because of that specific reason.

My full proposal is as follows:

`dnx.coreclr.managed` + `dnx.host` (+ `dnx.mono.managed` if possible) -> `Microsoft.Dnx.Host`
`dnx.mono.managed` (if merging is not possible) -> `Microsoft.Dnx.Host.Mono`
`dnx.clr.managed` -> `Microsoft.Dnx.Host.DomainManager`

victorhurdugaci:
PR updated:
dnx.mono.managed -> Microsoft.Dnx.DomainManager.Mono for consistency.

victorhurdugaci:
Updated table:

image

victorhurdugaci:
Now that I'm looking at the names maybe all Microsoft.Dnx.DomainManager.* should be Microsoft.Dnx.Host.* ...

anurse:
Agreed, do that rename and get this in. We can debate merging the Hosts later, it won't affect anything else.

victorhurdugaci:
Updated. Will have to rebase again because someone (looking at you @ChengTian ) merged a PR that conflicts with this one.

victorhurdugaci:
Okay. Rebased and ready for merge. Will update the CI drops to have the new dnx and then I'll merge everything. Expect breaks

Issue #2351, opened 2015-07-27
"Unable to resolve dependency" errors are duplicated during build


ChengTian:
https://github.com/aspnet/dnx/commit/2fb96f80d5b0c12059b54aa529a65afcfcffbd80#commitcomment-12356822

PR #2353, opened 2015-07-27
Remove duplicated build errors

ChengTian:
parent https://github.com/aspnet/dnx/issues/2351

@davidfowl 

The diagnostics returned by `_applicationHostContext.DependencyWalker.GetDependencyDiagnostics(_project.ProjectFilePath)` is a subset of
`_applicationHostContext.GetAllDiagnostics()`, which is already added here: https://github.com/aspnet/dnx/blob/1b0f5975777093e267f2d45d3bf5149d1bed5155/src/Microsoft.Framework.PackageManager/Building/BuildContext.cs#L62

### DIFF
--- a/src/Microsoft.Dnx.Tooling/Building/BuildContext.cs
+++ b/src/Microsoft.Dnx.Tooling/Building/BuildContext.cs
@@ -66,9 +66,6 @@ namespace Microsoft.Dnx.Tooling
                 diagnostics.AddRange(result.Diagnostics);
             }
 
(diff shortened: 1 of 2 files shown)


davidfowl:
:shipit: I think we're going to need a unit test or something for this

ChengTian:
Yeah, I should add it

ChengTian:
Added a functional test to prevent similar regression.

ChengTian:
Rebased after runtime renaming

davidfowl:
:shipit:

PR #2359, opened 2015-07-27
fix #2290 by using the HashCodeCombiner from Common

### LINKED ISSUE #2290: Move HashCodeCombiner to Common repo
We should move the `HashCodeCombiner` class to the Common repo as a Sources package to avoid duplication.

anurse:
Now that HashCodeCombiner has been moved, this changes our code to use the shared sources version.

@victorhurdugaci can this go in before your renames as well?

/cc @troydai @BrennanConroy


BrennanConroy:
Looks good. Are you doing the same for other repos?

anurse:
@NTaylorMullen said he'd take Razor and that's the only other place this occurs.

troydai:
:shipit:

src/Microsoft.Framework.Runtime.Abstractions/project.json
```diff
@@ -2,6 +2,12 @@
     "version": "1.0.0-*",
     "description": "ASP.NET 5 interfaces to take advantage of capabilities exposed by the runtime.",
     "compilationOptions": { "warningsAsErrors": true },
+    "dependencies": {
+        "Microsoft.Framework.HashCodeCombiner.Sources": {
```
davidfowl: Ugh, lets not invent new formatting. We don't do this anywhere else.

anurse: VS did it to me automatically...

NTaylorMullen: :+1: for having whatever VS chooses here. Less work when having to undo.

anurse: See also: https://github.com/aspnet/Razor/pull/463/files#diff-3ca14b98c54139a0defc8a35940fe07e

Eilon: As ugly as it is, fighting the VS defaults is an uphill battle.

davidfowl: We need to fix this. It's ugly as crap /cc @madskristensen

anurse: Yeah, we shouldn't reformat objects that are already on a single line. Much like C# does with properties:

```csharp
public string Foo { get { ... } set { ... } }
```

madskristensen: @davidfowl What needs to be fixed? VS auto-formats C# or JSON?

davidfowl: JSON

madskristensen: I suggest you take a look at using .editorconfig files to control tab vs. spaces and indent size

davidfowl: I think you missed the point of this thing. Andrew pretty much summed it up https://github.com/aspnet/dnx/pull/2359#discussion_r35613141

PR #2374, opened 2015-07-30
Port more compatibility check from NuGet

ChengTian:
parent https://github.com/aspnet/dnx/issues/2356

@anurse @davidfowl 

Will send PRs to fix compatibility problems in Tooling templates and other repos.

### DIFF
--- a/samples/HelloShared/project.json
+++ b/samples/HelloShared/project.json
@@ -6,8 +6,13 @@
         "dnx451": { },
         "dnxcore50": {
             "dependencies": {
+                "System.Globalization": "4.0.10-beta-*",
+                "System.IO": "4.0.10-beta-*",
                 "System.Linq": "4.0.0-beta-*",
-                "System.Runtime": "4.0.20-beta-*"
+                "System.Reflection": "4.0.10-beta-*",
+                "System.Runtime": "4.0.20-beta-*",
+                "System.Text.Encoding": "4.0.10-beta-*",
+                "System.Threading.Tasks": "4.0.10-beta-*"
             }
         }
     }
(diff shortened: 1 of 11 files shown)


ChengTian:
Sample output when we find an incompatible package during restore:
capture

anurse:
:shipit:

samples/HelloShared/project.json
```diff
@@ -6,8 +6,13 @@
...
         "dnxcore50": {
             "dependencies": {
+                "System.Globalization": "4.0.10-beta-*",
+                "System.IO": "4.0.10-beta-*",
                 "System.Linq": "4.0.0-beta-*",
-                "System.Runtime": "4.0.20-beta-*"
+                "System.Reflection": "4.0.10-beta-*",
+                "System.Runtime": "4.0.20-beta-*",
+                "System.Text.Encoding": "4.0.10-beta-*",
+                "System.Threading.Tasks": "4.0.10-beta-*"
```
anurse: We need to find a better solution to this than just bumping up individual packages... But that's the right thing to do for now.

src/Microsoft.Dnx.ApplicationHost/project.json
```diff
@@ -18,6 +18,7 @@
                 "System.Console": "4.0.0-beta-*",
                 "System.Diagnostics.Debug": "4.0.10-beta-*",
                 "System.Linq": "4.0.0-beta-*",
+                "System.ObjectModel": "4.0.10-beta-*",
```
davidfowl: ?

davidfowl:
My problem with the way this is implemented is that it doesn't play with any of the other diagnostics that we have today in the DNX. That means it won't show up in VS or during compilation. The fact that we don't store restore diagnostics anywhere is actually a problem. I'd like to see a solution that ties into our existing diagnostics infrastructure.

ChengTian:
Have a replacement PR sent. Closing this one.

PR #2377, opened 2015-07-30
remove HashCodeCombiner from TPA list

anurse:
HashCodeCombiner.Sources ended up in the TPA list. As far as I can tell, this is the easiest way to remove it :).

/cc @troydai @davidebbo @BrennanConroy


makefile.shade
```diff
@@ -49,7 +49,6 @@ var RUNTIME_CLR_TARGETS = '${new [] {RUNTIME_MONO_BIN, RUNTIME_CLR_WIN_x86_BIN,
 var RUNTIME_CORECLR_TARGETS = '${new [] {RUNTIME_CORECLR_WIN_x86_BIN, RUNTIME_CORECLR_WIN_x64_BIN, RUNTIME_CORECLR_WIN_arm_BIN, RUNTIME_CORECLR_DARWIN_x64_BIN, RUNTIME_CORECLR_LINUX_x64_BIN }}'
 var ALL_TARGETS = '${RUNTIME_CLR_TARGETS.Concat(RUNTIME_CORECLR_TARGETS)}'
 
-var RUNTIME_COMMON_NAME = 'DNX'
```
anurse: This wasn't being used.

makefile.shade
```diff
@@ -1064,6 +1063,10 @@ macro name='DoRestore'
     }
 
 functions @{
+    private static readonly string[] TPA_IGNORED_ASSEMBLIES = new [] {
```
anurse: This has to go here because it is referenced from a function. The `var`s at the top are scoped to a specific method in the generated class and thus not available to the other methods.

anurse:
Going to push this now because it's so tiny and a big nuisance. Will react to comments, if any, post-commit.

Issue #2380, opened 2015-07-30
DNU should default to V3 feed instead of V2 if it can't find any feed


muratg:

ChengTian:
Updating title. Following previous behavior, the newly created NuGet.Config should always be empty. We should fallback to v3 feed directly, without writing it to disk first.

muratg:
:+1:

PR #2381, opened 2015-07-31
Default to official NuGet v3 feed if no feed can be found

ChengTian:
parent https://github.com/aspnet/dnx/issues/2380

@anurse  @muratg @DamianEdwards @davidebbo @davidfowl


anurse:
:shipit:

PR #2382, opened 2015-07-31
[Prototype] Run roslyn analyzer at build time

troydai:
Automation is on the way.

/cc @anurse

### DIFF
--- a/src/Microsoft.Dnx.Compilation.Abstractions/LibraryExport.cs
+++ b/src/Microsoft.Dnx.Compilation.Abstractions/LibraryExport.cs
@@ -1,22 +1,35 @@
-﻿using System.Collections.Generic;
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
 
 namespace Microsoft.Dnx.Compilation
 {
     public class LibraryExport
     {
-        public static readonly LibraryExport Empty = new LibraryExport();
+        public static readonly LibraryExport Empty = new LibraryExport(metadataReference: null);
 
-        private LibraryExport() : this(null, null) { }
-        public LibraryExport(IMetadataReference metadataReference) : this(new List<IMetadataReference>() { metadataReference }) { }
-        public LibraryExport(IList<IMetadataReference> metadataReferences) : this(metadataReferences, null) { }
-        public LibraryExport(IList<ISourceReference> sourceReferences) : this(null, sourceReferences) { }
+        public LibraryExport(IMetadataReference metadataReference)
+            : this(new List<IMetadataReference>() { metadataReference }, sourceReferences: null, analyzerReferences: null)
+        {
+        }
+
         public LibraryExport(IList<IMetadataReference> metadataReferences, IList<ISourceReference> sourceReferences)
+            : this(metadataReferences, sourceReferences, analyzerReferences: null)
+        {
+        }
+
+        public LibraryExport(IList<IMetadataReference> metadataReferences, IList<ISourceReference> sourceReferences, IList<IAnalyzerReference> analyzerReferences)
         {
             MetadataReferences = metadataReferences ?? new List<IMetadataReference>();
             SourceReferences = sourceReferences ?? new List<ISourceReference>();
+            AnalyzerReferences = analyzerReferences ?? new List<IAnalyzerReference>();
         }
 
         public IList<IMetadataReference> MetadataReferences { get; }
+
         public IList<ISourceReference> SourceReferences { get; }
+
+        public IList<IAnalyzerReference> AnalyzerReferences { get; }
     }
 }
--- a/src/Microsoft.Dnx.Compilation.CSharp/CompilationContext.cs
+++ b/src/Microsoft.Dnx.Compilation.CSharp/CompilationContext.cs
@@ -4,9 +4,12 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Diagnostics;
+using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.Dnx.Runtime;
 
 namespace Microsoft.Dnx.Compilation.CSharp
@@ -52,6 +55,8 @@ namespace Microsoft.Dnx.Compilation.CSharp
 
         public IList<ICompileModule> Modules { get; }
 
+        public ImmutableArray<DiagnosticAnalyzer> DiagnosticAnalyzers { get; set; }
+
         public CSharpCompilation Compilation
         {
             get { return _beforeCompileContext.Compilation; }
@@ -76,6 +81,21 @@ namespace Microsoft.Dnx.Compilation.CSharp
         public BeforeCompileContext BeforeCompileContext
         {
             get { return _beforeCompileContext; }
+        }
+
+        public async Task<ImmutableArray<Diagnostic>> GetAnalyzerDiagnsoticsAsnyc()
+        {
+            if (DiagnosticAnalyzers != null && DiagnosticAnalyzers.Length > 0)
+            {
+                var diagnostics = await Compilation.WithAnalyzers(DiagnosticAnalyzers)
+                                                   .GetAllDiagnosticsAsync();
+
+                return diagnostics;
+            }
+            else
+            {
+                return ImmutableArray<Diagnostic>.Empty;
+            }
         }
 
         private IList<ResourceDescription> ResolveResources()
--- a/src/Microsoft.Dnx.Compilation.CSharp/RoslynCompiler.cs
+++ b/src/Microsoft.Dnx.Compilation.CSharp/RoslynCompiler.cs
@@ -51,6 +51,7 @@ namespace Microsoft.Dnx.Compilation.CSharp
             CompilationProjectContext projectContext,
             IEnumerable<IMetadataReference> incomingReferences,
             IEnumerable<ISourceReference> incomingSourceReferences,
+            IEnumerable<IAnalyzerReference> analyzerReferences,
             Func<IList<ResourceDescriptor>> resourcesResolver)
         {
             var path = projectContext.ProjectDirectory;
@@ -86,6 +87,7 @@ namespace Microsoft.Dnx.Compilation.CSharp
             var exportedReferences = incomingReferences.Select(ConvertMetadataReference);
 
             Logger.TraceInformation("[{0}]: Compiling '{1}'", GetType().Name, name);
+
             var sw = Stopwatch.StartNew();
 
             var compilationSettings = projectContext.CompilerOptions.ToCompilationSettings(
@@ -171,6 +173,26 @@ namespace Microsoft.Dnx.Compilation.CSharp
                 }
             }
 
+            if (isMainAspect && analyzerReferences.Any())
+            {
+                try
+                {
+                    var analyzerSet = GetAnalyzerSet(projectContext.Target, analyzerReferences);
+                    if (analyzerSet?.DiagnosticAnalyzers.Any() ?? false)
+                    {
+                        compilationContext.DiagnosticAnalyzers = analyzerSet.DiagnosticAnalyzers;
+                        foreach (var analyzer in analyzerSet.DiagnosticAnalyzers)
+                        {
+                            Logger.TraceInformation($"[{nameof(RoslynCompiler)}]: Found diagnostic analyzer {analyzer}");
+                        }
+                    }
+                }
+                catch (Exception ex)
+                {
+                    Logger.TraceError($"[{nameof(RoslynCompiler)}]: Failed to load analyzer assembly. Exception: {ex.Message}.");
+                }
+            }
+
             if (compilationContext.Modules.Count > 0)
             {
                 var precompSw = Stopwatch.StartNew();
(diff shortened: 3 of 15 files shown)


src/Microsoft.Dnx.Compilation.Abstractions/LibraryExport.cs
```diff
@@ -1,22 +1,35 @@
...
             MetadataReferences = metadataReferences ?? new List<IMetadataReference>();
             SourceReferences = sourceReferences ?? new List<ISourceReference>();
+            AnalyzerReferences = analyzerReferences ?? new List<IAnalyzerReference>();
         }
 
         public IList<IMetadataReference> MetadataReferences { get; }
+
         public IList<ISourceReference> SourceReferences { get; }
+
+        public IList<IAnalyzerReference> AnalyzerReferences { get; }
```
pranavkm: Should this be `IReadOnlyList`?

troydai: Try to keep it inline with other references.

troydai: I personally prefer `IEnumerable`

troydai:
DTH output:
image

Command line output:
image

src/Microsoft.Dnx.Compilation.Abstractions/LibraryExport.cs
anurse: I'll look at cleaning this up in #2226. I prefer Read Only List where it makes sense because then you can get the count without enumerating.

troydai: Yup. I'll keep it as is in this PR.

src/Microsoft.Dnx.Compilation.CSharp/CompilationContext.cs
```diff
@@ -52,6 +55,8 @@ public class CompilationContext
 
         public IList<ICompileModule> Modules { get; }
 
+        public ImmutableArray<DiagnosticAnalyzer> DiagnosticAnalyzers { get; set; }
```
anurse: We don't use ImmutableArray anywhere but I'm vaguely interested in doing it more. Do you know anything about the performance impact?

src/Microsoft.Dnx.Compilation.CSharp/RoslynAnalyzationModule.cs
```diff
@@ -0,0 +1,73 @@
...
+using System.IO;
+using System.Reflection;
+using System.Runtime.Versioning;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.Dnx.Runtime;
+
+namespace Microsoft.Dnx.Compilation.CSharp
+{
+    public class RoslynAnalyzationModule
```
anurse: I'd call this something like `AnalyzerSet`. The Roslyn prefix is redundant (and we should be removing it from other places) and `Module` is only used for `ICompileModule` to match up with that.

src/Microsoft.Dnx.Compilation.CSharp/RoslynCompiler.cs
```diff
@@ -172,6 +173,26 @@ public class RoslynCompiler
                 }
             }
 
+            if (isMainAspect && analyzerReferences.Any())
+            {
+                try
+                {
+                    var analyzationModule = GetAnalyzationModule(projectContext.Target, analyzerReferences);
```
anurse: `var analyzerSet = GetAnalyzerSet(...)`

src/Microsoft.Dnx.Compilation.CSharp/RoslynProjectReference.cs
```diff
@@ -202,11 +211,29 @@ public DiagnosticResult EmitAssembly(string outputPath)
...
+                var diagnostics = new List<Diagnostic>(emitResult.Diagnostics);
+
+                if (CompilationContext.DiagnosticAnalyzers != null && CompilationContext.DiagnosticAnalyzers.Length > 0)
+                {
+                    sw.Start();
+
+                    var analyzerDiagnostics = CompilationContext.Compilation
+                        .WithAnalyzers(CompilationContext.DiagnosticAnalyzers)
+                        .GetAllDiagnosticsAsync()
+                        .Result     // TODO: make the caller an async method?
```
anurse: Can you either a) try making it async or b) file a bug to investigate this further?

src/Microsoft.Dnx.Runtime/Helpers/RoslynAnalyzerHelper.cs
```diff
@@ -0,0 +1,81 @@
...
+                    continue;
+                }
+
+                if (!string.Equals(parts[2], AnalyzerLanguage, StringComparison.OrdinalIgnoreCase))
+                {
+                    continue;
+                }
+
+                var frameworkName = VersionUtility.ParseFrameworkFolderName(parts[1]);
+                if (VersionUtility.GetNearest(runtimeFramework, new[] { frameworkName }) != frameworkName)
```
anurse: I don't think this is right. Instead, you should be able to wrap each file starting `analyzers/` in a class implementing `IFrameworkTargettable` and then pass the whole set of files into GetNearest. It will return to you all the files from one folder, having selected the most appropriate folder. Then you can filter the resulting files by language.

As far as I can tell, this will return analyzers from multiple different folders as long as they are compatible with the runtime folder, which is bad. If there is an `analyzers/net451` and an `analyzers/dotnet` folder, compiling on `dnx451` should **only** pick the analyzers from `analyzers/net451` and **not** any of the analyzers from `analyzers/dotnet`. `GetNearest` should handle that for you.

One other way you could do this is to get the set of folder names directly under `analyzers/` and pass **all** of them in to `GetNearest`. The one returned is the best match and you can scoop up all the analyzers from that folder only.

anurse: Regardless of how you fix it, add a unit test for this selection algorithm to ensure it selects the appropriate folders. It would also help discussion if you did that so I can point out concrete cases that might be missed :P.

anurse:
Looking pretty good so far, but some issues. Structurally I think it looks ok. I'm not a huge fan of using GetNearest in the runtime since we spent some time removing it, but I know there isn't really an easier way to do it unless we put them in the lock file which gets ugly.

src/Microsoft.Dnx.Compilation.CSharp/CompilationContext.cs
troydai: Not aware of performance impact, but I'll research. The reason why this method returns an `ImmutableArray` is because Roslyn requires this type to operate. Either it returns an `ImmutableArray` here or the consumer need to wrap the collection into `ImmutableArray`.

src/Microsoft.Dnx.Compilation.CSharp/RoslynAnalyzationModule.cs
troydai: Sure thing.

src/Microsoft.Dnx.Compilation.CSharp/RoslynCompiler.cs
troydai: will rename together.

src/Microsoft.Dnx.Runtime/Helpers/RoslynAnalyzerHelper.cs
troydai: I see. will update.

src/Microsoft.Dnx.Compilation.CSharp/CompilationContext.cs
anurse: :+1:

src/Microsoft.Dnx.Compilation.CSharp/RoslynProjectReference.cs
troydai: File this issue: https://github.com/aspnet/dnx/issues/2386

I think we should keep the interface as-is for now.

src/Microsoft.Dnx.Compilation.CSharp/RoslynCompiler.cs
```diff
@@ -267,6 +288,27 @@ private CompilationModules GetCompileModules(CompilationTarget target)
...
 
+        private AnalyzerSet GetAnalyzerSet(CompilationTarget target, IEnumerable<IAnalyzerReference> references)
+        {
+            Logger.TraceInformation($"[{nameof(RoslynCompiler)}] Load analyzers.");
+
+            var key = Tuple.Create(
+                target.Name,
+                _environment.RuntimeFramework,
+                _environment.Configuration,
+                "analyzationModules");
```
victorhurdugaci: Name?

src/Microsoft.Dnx.Runtime/DependencyManagement/NuGetDependencyResolver.cs
```diff
@@ -324,6 +330,15 @@ public void Initialize(IEnumerable<LibraryDescription> packages, FrameworkName t
                 }
 
                 dependency.LoadableAssemblies = assemblies;
+
+                var analyzerReferences = RoslynAnalyzerHelper.GetAnalyerReferences(packageInfo, packagePath, _runtimeFramework);
+                foreach (var analyzerAssembly in analyzerReferences?.SelectMany(reference => reference.Files) ?? Enumerable.Empty<string>())
+                {
+                    var assemblyPath = Path.Combine(packagePath, analyzerAssembly).Replace('/', Path.DirectorySeparatorChar);
```
victorhurdugaci: don't we have a helper for this replacement? It's all over the place

src/Microsoft.Dnx.Compilation.CSharp/RoslynProjectReference.cs
anurse: :+1:

src/Microsoft.Dnx.Runtime/Helpers/RoslynAnalyzerHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Runtime.Versioning;
+using Microsoft.Dnx.Compilation;
+using NuGet;
+
+namespace Microsoft.Dnx.Runtime.Helpers
+{
+    internal static class RoslynAnalyzerHelper
+    {
+        private static readonly string AnalyzerBaseFolder = "analyzers";
+        private static readonly string AnalyzerLanguage = "cs";
```
victorhurdugaci: Should this helper go in some CSharp compilation helper, in a different package? It is really c# specific

src/Microsoft.Dnx.Runtime/Helpers/RoslynAnalyzerHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+        private static readonly string AnalyzerExtension = ".dll";
+
+        public static IEnumerable<IAnalyzerReference> GetAnalyerReferences(PackageInfo packageInfo,
+                                                                           string packagePath,
+                                                                           FrameworkName runtimeFramework)
+        {
+            // assumption: analyzers' paths are in following format
+            // analyzer/{framework}/{cs|vb}/{filename}.dll
+            var files = packageInfo.LockFileLibrary.Files.Select(CreateFileReference)
```
anurse: Can we do the filtering of files before the Select so we can get rid of the "Where-not-null" check?

src/Microsoft.Dnx.Runtime/Helpers/RoslynAnalyzerHelper.cs
anurse: Yeah, we may need to look at the layering of this. I'd prefer to avoid having specific Roslyn knowledge in the runtime, but it might be tricky given that we really need to use the NuGet loader/dependency resolver.

src/Microsoft.Dnx.Runtime/Helpers/RoslynAnalyzerHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+                return null;
+            }
+
+            if (!relativePath.EndsWith(AnalyzerExtension, StringComparison.OrdinalIgnoreCase))
+            {
+                return null;
+            }
+
+            var parts = relativePath.Split(new[] { Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar });
+            if (parts.Length != 4)
```
victorhurdugaci: 4 because 3 was taken?

src/Microsoft.Dnx.Compilation.CSharp/RoslynProjectReference.cs
```diff
@@ -202,11 +211,29 @@ public DiagnosticResult EmitAssembly(string outputPath)
...
+                        .ToArray();
+
+                    sw.Stop();
+                    Logger.TraceInformation($"[{nameof(RoslynProjectReference)}]: Analyzing {Name} in {sw.ElapsedMilliseconds}ms");
```
victorhurdugaci: "Analyzed"

davidfowl:
I'm not a fan of the implementation so far. It's beginning to bleed a little into the mainline compilation API. I was hoping it would be an implementation detail of the roslyn compiler. The only place where it needs to bleed is DTH where we need to return analyzer file references directly.

anurse:
We'll have to see what we can make work. Right now, in order for the load context to work we have to use Load instead of LoadFile, which means the runtime loaders need to know how to find the assembly. I'm not a fan either, and it might be possible for us to work around it.

troydai:
@davidfowl would you please be more specific about what's the compilation API you think should avoid the change? The changed interfaces are:
* `LibraryExport`: Here's the analyzer references are carried. 
* `CompilationContext`: It's the return value of `RoslynCompiler`. If the analyzing happens there, it needs to carries the result. Now it carries loaded analyzer. I'll take a look if the changes to it can be removed. But it means the process of creating a separate load context and load analyzer will be pushed further later.
* `NuGetDependencyProvider`: It seems unavoidable. It needs to know the runtime framework so as to filter the analyzer. Of course as of now all the analyzers are wrapped in `dotnet` but it doesn't seem to be guaranteed.

anurse:
We're pushing this out to the backlog. /cc @Eilon @glennc @davidfowl @lodejard

PR #2399, opened 2015-08-03
Fix issue where TargetPackagesFolder not being set correctly

ajaybhargavb:
This fixes the issue where this `The specified package is not an application. The package was added but no commands were installed.` error shows up when trying to install a command from a package with specifying the target folder with `--packages` switch.

cc' @victorhurdugaci @BrennanConroy


victorhurdugaci:
Does global install still work if you don't specify the packages argument?

ajaybhargavb:
Yes. It works.

victorhurdugaci:
Then :shipit:

Issue #2073, opened 2015-06-16
"dnu restore" should download nuspec individually when working with v3 feeds


ChengTian:
In current implementation, we extract nuspecs from nupkgs. However, v3 feeds support that we download nuspec individually.

We should make `dnu` take advantage of this v3 feed feature.

ChengTian:
Punting this because it actually hurts restore performance

PR #2400, opened 2015-08-03
Download nuspecs individually when using NuGet v3 feeds

ChengTian:
parent https://github.com/aspnet/dnx/issues/2073

@davidfowl @lodejard  @anurse

### DIFF
--- a/src/Microsoft.Dnx.Tooling/Restore/NuGet/NuGetv3Feed.cs
+++ b/src/Microsoft.Dnx.Tooling/Restore/NuGet/NuGetv3Feed.cs
@@ -205,6 +205,7 @@ namespace Microsoft.Dnx.Tooling.Restore.NuGet
                 Id = id,
                 Version = SemanticVersion.Parse(version),
                 ContentUri = $"{_baseUri}{lowerInvariantId}/{lowerInvariantVersion}/{lowerInvariantId}.{lowerInvariantVersion}{Constants.PackageExtension}",
+                ManifestUri = $"{_baseUri}{lowerInvariantId}/{lowerInvariantVersion}/{lowerInvariantId}{Constants.ManifestExtension}",
 
                 // v3 feed doesn't indicate if listed?
                 Listed = true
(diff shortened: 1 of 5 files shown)


src/Microsoft.Dnx.Tooling/Restore/NuGet/NuGetv3Feed.cs
```diff
@@ -213,7 +214,41 @@ public PackageInfo BuildModel(string id, string version)
...
+                    if (isFinalAttempt)
+                    {
+                        _reports.Error.WriteLine(
+                            $"Error occurred when downloading manifest: {package.ManifestUri}{Environment.NewLine}  {message}".Red().Bold());
+                        throw;
+                    }
+                    else
+                    {
+                        _reports.Information.WriteLine(
+                            $"Warning occurred when downloading manifest: {package.ManifestUri}{Environment.NewLine}  {message}".Yellow().Bold());
```
victorhurdugaci: Can you please add another part to this message saying that it will retry. Or even better "retry x of y"

ChengTian: We already tell users that we are retrying in error message part:
capture

Moreover, we have 3 or 4 similar retry logic in DNX, we should modify all of them if we want to change this one. That's beyond the scope of this PR. Can you file an issue and label it with suggestion?

anurse:
:shipit:

src/Microsoft.Dnx.Tooling/Restore/NuGet/PackageUtilities.cs
```diff
@@ -4,7 +4,10 @@
 using System;
 using System.IO;
 using System.IO.Compression;
+using System.Linq;
 using System.Threading.Tasks;
+using System.Xml.Linq;
+using NuGet;
```
ChengTian: Yeah, only `NuGet` is needed. Will delete useless ones

davidfowl:
@ChengTian says this is twice as slow. We should punt this change if that is the case.

lodejard:
Was afraid that would be the case. Downloading both nuspec and nupkg will double the round-trips per restore. The only time you'll have a net improvement is when the package graph has enough cousin-dependency version upgrades that the extra time spent downloading all the nuspecs is less than the time spent downloading unused (older versioned cousin) nupkg candidates. And those are likely less common.

+1 for punting the change

ChengTian:
I tested this PR by restoring for MusicStore several times and calculating average restore time. Turns out this "improvement" doubles the restore time.

Issue #2336, opened 2015-07-24
Servicing follow up items


moozzyk:
- [x] we should not fallback to ProgramFiles if DNX_SERVICING is set
    - [x] +Helios
- [x] when DNX_SERVICING is set it must point to a location that contains a readable index.txt file or the application will not run
    - [x] +Helios
- [x] do not allow using network shares in DNX_SERVICING variable
    - [x] +Helios
- [x] if reading the index.txt fails (sharing violation etc.) we should retry and then fail if it is not successful
    - [x] +Helios
- [x] if Servicing folder exists but there is no index.txt file we should fail
    - [x] +Helios
- [ ] fuzz testing for contents of the index.txt (empty file, binary file, short file - shorter than source framework etc.) file and DNX_SERVICING variable name
- [ ] End-to-end testing
- [x] ~~detailed documentation (folder structure, including index.txt format, recommended ACLs on the files etc.)~~ - tracked in the [Docs repo](https://github.com/aspnet/Docs/issues/374)
- [x] ~~enable preparing machines for servicing~~ - tracked as #2526

muratg:
@victorhurdugaci will be helping with some of the items.

@joeloff we'll need to create an MSI to test servicing E2E in the Beta 7 milestone.

moozzyk:
#2488

moozzyk:
#1122

muratg:
@victorhurdugaci E2E testing & coordination assigned to you in RC1.

victorhurdugaci:
E2E validated. I was able to deliver an update in a test environment.

moozzyk:
:clap:

PR #2404, opened 2015-08-04
Removing fallback to %ProgramFiles(x86)% if DNX_SERVICING is specified

moozzyk:


src/Microsoft.Dnx.Runtime/Servicing/ServicingTable.cs
```diff
@@ -21,51 +21,28 @@ public static class ServicingTable
...
+        private static ServicingIndex LoadIndex()
         {
             return LazyInitializer.EnsureInitialized(ref _index, ref _indexInitialized, ref _indexSync, () =>
             {
-                var compositeIndex = new List<ServicingIndex>();
-
-                foreach (var servicingRoot in GetServicingRoots())
+                var index = new ServicingIndex();
+                var servicingRoot = GetServicingRoot();
+                if (servicingRoot != null && servicingRoot.IndexOfAny(Path.GetInvalidPathChars()) == -1)
```
davidfowl: I know the invalid path check isn;t new but why are we doing it?

davidfowl:
No bug associated with this change?

moozzyk:
There is a bug but it's a checklist #2336

src/Microsoft.Dnx.Runtime/Servicing/ServicingTable.cs
moozzyk: It was added to address this comment: https://github.com/aspnet/dnx/pull/2183#discussion_r33820594. However with the current changes it may no longer be necessary since dnx will fail if it can't read the index.txt file and will never actually get here.

davidfowl: Ok lets remove it.

moozzyk: Will fix

moozzyk:
:watch:

davidfowl:
Why is the ci sad? The code looks fine besides that :shipit:

moozzyk:
Apparently some packages (Unable to locate Microsoft.CodeAnalysis.CSharp >= 2.0.0-beta1-*) were removed from the feed. @troydai is fixing this and I will push after the fix is in.

PR #2406, opened 2015-08-04
Ensuring readability of servicing index

moozzyk:
The application now will fail if:
- the %DNX_SERVICING%\index.txt cannot be open
- %PROGRAMFILES(X86)%\Microsoft DNX\Servicing exists but does not contain index.txt file


davidfowl:
Where's the bug for this change?

moozzyk:
#2336

moozzyk:
:watch:

PR #2413, opened 2015-08-05
WIP - Polish "dnu publish" output

### LINKED ISSUE #2376: Polish "dnu publish" output
Reported by @DamianEdwards 

capture

The information in red rectangles should be removed, to make it less noisy and less confusing.

@muratg , I can work on this issue.

MisterJames:
@ChengTian - I'm just starting, locating the code required for this fix. I haven't found the code that outputs "Installing ____" but found the corresponding resource string; however, the resource string doesn't seem to be used.  I'll have a look this evening.

Resolves #2376

### DIFF
--- a/src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
+++ b/src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
@@ -108,6 +108,7 @@ namespace Microsoft.Dnx.Tooling
             {
                 var configFiles = settings.GetConfigFiles();
 
+                // target for PR
                 Reports.Quiet.WriteLine($"{Environment.NewLine}NuGet Config files used:");
                 foreach (var file in configFiles)
                 {
@@ -905,6 +906,7 @@ namespace Microsoft.Dnx.Tooling
                 if (feed != null)
                 {
                     remoteProviders.Add(new RemoteWalkProvider(feed));
+                    // target for PR
                     var list = InformationMessages.GetOrAdd("Feeds used:", _ => new List<string>());
                     if (!list.Contains(feed.Source))
                     {
--- a/test/Microsoft.Dnx.Tooling.FunctionalTests/DnuInstallTests.cs
+++ b/test/Microsoft.Dnx.Tooling.FunctionalTests/DnuInstallTests.cs
@@ -132,6 +132,7 @@ namespace Microsoft.Dnx.Tooling
 
             Assert.Equal(0, exitCode);
             Assert.Empty(stdErr);
+            // possible target for PR
             Assert.Contains($"Installing {packageName}.{packageVersion}", stdOut);
             Assert.Equal(expectedProjectJson, File.ReadAllText(projectJsonPath));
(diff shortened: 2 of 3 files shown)


MisterJames:
Just as an FYI, the dev branch fails locally and appears to be failing on the main repo as well. I haven't introduced any changes yet.

ChengTian:
I think you did a good job finding locations to be modified :+1: 

`dnu publish` is calling `dnu packages add` and `dnu restore` internally and this is the only scenario where we want to hide some output. It means that, when you use `dnu restore`/`dnu packages add` individually, the output shouldn't be changed.

A rough idea can be adding some special internal flag/option, like this: https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs#L428
Just my two cents

ChengTian:
And the main repo build failure should be fixed sometime later today, after we push some new packages to package feeds.

MisterJames:
Thanks @ChengTian. Okay, so after reading your comments above and reviewing the code in a little more detail, I think I understand what you're suggesting.

 - [] Add a flag similar to `SkipRestoreEvents` as part of the `restore` and `package add` commands (restore is already hosting the `SkipRestoreEvents`)
 - [] Test for this flag being set; if it's off, don't log out those details during invocation

Please let me know if that's not on track.

My other question would be then if this is along the lines of "`SkipPackageSummaryDetail`" and/or "`SkipPerformanceDetail`" as flags - it is basically logging out info about config, feeds and some minor perf details. If there's more context to this scenario, it may help with an appropriate name, or perhaps let me know if there are other places such a flag might be used.

Also, looks like I'm still hurting here for restoring packages/building, so I'm going to hold off again here. I want to have a build running cleanly so I can isolate & execute the appropriate tests (and know they are running).

ChengTian:
As far as I know, this is the only place we want to use the two internal flags. One small suggestion is: `SkipPackageSummaryDetail` -> `SkipFeedSummary`. All other part looks good to me.

We do have some issues pushing latest packages to feeds. Yes, please hold off on sending PR until you see a green dnx/dev branch.

ChengTian:
DNX/dev branch build is passing now: https://github.com/aspnet/dnx/branches

MisterJames:
Great, thanks @ChengTian, I'll pull and get to work this weekend. Cheers!

ChengTian:
@MisterJames , sorry I didn't mention that we want to have this in our beta7 release. So the PR needs to be merged before Aug. 14th (Friday).

Feel free to let me know if you don't have time to do it before Friday. I have enough time and can work on this issue : )

MisterJames:
@ChengTian You'd just noted on #2376 that you were working on this, did you want me to put this one down then? I can pick up another if you prefer. I got the build working locally and was just starting...only created a test and have the props stubbed in, so no big deal.

ChengTian:
@MisterJames , yeah, we find it's better to do this as part of another PR, so I'll take care of it.

Feel free to pick another issue you can handle and let the assignee know (or reply to that thread if there is no assignee). We really appreciate your contribution :smile:

PR #2414, opened 2015-08-05
Begin consuming Roslyn 2.0.0-beta1

troydai:
/cc @pranavkm @anurse @davidfowl @Eilon 

This change contains a work around to mitigate a package flawed in Roslyn package: The System.Reflection.Metadata is directly referenced.

I won't wait for sign off so as to unblock the build. Tested locally.

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -2,6 +2,8 @@
 <configuration>
   <packageSources>
     <add key="AspNetVNext" value="https://www.myget.org/F/aspnetvnext/api/v2" />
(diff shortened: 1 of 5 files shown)


Eilon:
Should we stop our own Roslyn build at some point?

troydai:
It's already stopped since Roslyn RTMed.

Eilon:
Ah ok. We'll need to figure out with them what to do once we need to split between dev/release - assuming they at some point have a release as well that we want to stabilize onto.

troydai:
Yup.

PR #2425, opened 2015-08-06
Wei/compatibility check

ChengTian:
parent https://github.com/aspnet/dnx/issues/2356

replacing https://github.com/aspnet/dnx/pull/2374

@davidfowl @anurse

### DIFF
--- a/src/Microsoft.Dnx.Runtime/DependencyManagement/CompatibilityChecker.cs
+++ b/src/Microsoft.Dnx.Runtime/DependencyManagement/CompatibilityChecker.cs
@@ -1,0 +1,110 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Runtime.Versioning;
+using Microsoft.Dnx.Runtime.DependencyManagement;
+
+namespace Microsoft.Dnx.Runtime
+{
+    public class CompatibilityChecker
+    {
+        private readonly LockFile _lockFile;
+        private readonly Dictionary<string, LockFileLibrary> _lockFileLibraries;
+        private readonly Dictionary<FrameworkName, HashSet<string>> _frameworkRuntimeAssemblies;
+
+        public CompatibilityChecker(LockFile lockFile)
+        {
+            _lockFile = lockFile;
+            _lockFileLibraries = _lockFile.Libraries.ToDictionary(l => l.Name);
+            _frameworkRuntimeAssemblies = lockFile.Targets.ToDictionary(
+                keySelector: t => t.TargetFramework,
+                elementSelector: GetRuntimeAssemblySet);
+        }
+
+        public IEnumerable<CompatibilityIssue> GetAllCompatibilityIssues()
+        {
+            foreach (var target in _lockFile.Targets)
+            {
+                foreach (var library in target.Libraries)
+                {
+                    var issue = CheckTargetLibrary(library, target.TargetFramework);
+                    if (issue != null)
+                    {
+                        yield return issue;
+                    }
+                }
+            }
+        }
+
+        public CompatibilityIssue CheckTargetLibrary(LockFileTargetLibrary targetLibrary, FrameworkName framework)
+        {
+            var lockFileLibrary = _lockFileLibraries[targetLibrary.Name];
+            var containsAssembly = lockFileLibrary.Files
+                .Any(x => x.StartsWith($"ref{Path.DirectorySeparatorChar}") ||
+                    x.StartsWith($"lib{Path.DirectorySeparatorChar}"));
+            if (containsAssembly &&
+                !targetLibrary.FrameworkAssemblies.Any() &&
+                !targetLibrary.CompileTimeAssemblies.Any() &&
+                !targetLibrary.RuntimeAssemblies.Any())
+            {
+                return new CompatibilityIssue
+                {
+                    LibraryName = lockFileLibrary.Name,
+                    LibraryVersion = lockFileLibrary.Version,
+                    Framework = framework,
+                    Type = CompatibilityIssue.IssueType.UnsupportedFramework
+                };
+            }
+
+            // Get run-time assemblies provided by current library
+            var runtimeAssemblies = _frameworkRuntimeAssemblies[framework];
+
+            // Get compile-time assembies provided by current library
+            var compileAssemblies = targetLibrary.CompileTimeAssemblies
+                .Where(x => x.Path.EndsWith(".dll", StringComparison.OrdinalIgnoreCase))
+                .Select(x => Path.GetFileNameWithoutExtension(x.Path));
+
+            if (!runtimeAssemblies.IsSupersetOf(compileAssemblies))
+            {
+                // The library provids compile-time reference assembly
+                // but there is no compatible run-time assembly provided by any library in dependency closure
+                return new CompatibilityIssue
+                {
+                    LibraryName = lockFileLibrary.Name,
+                    LibraryVersion = lockFileLibrary.Version,
+                    Framework = framework,
+                    Type = CompatibilityIssue.IssueType.MissingRuntimeAssembly
+                };
+            }
+
+            // There is no compatibility issue found for current library
+            return null;
+        }
+
+        private static HashSet<string> GetRuntimeAssemblySet(LockFileTarget target)
+        {
+            var set = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            foreach (var runtime in target.Libraries
+                .SelectMany(l => l.RuntimeAssemblies)
+                .Where(x => x.Path.EndsWith(".dll", StringComparison.OrdinalIgnoreCase)))
+            {
+                var name = Path.GetFileNameWithoutExtension(runtime.Path);
+
+                // Fix for NuGet/Home#752 - Consider ".ni.dll" (native image/ngen) files matches for ref/ assemblies
+                if (name.EndsWith(".ni"))
+                {
+                    name = name.Substring(0, name.Length - 3);
+                }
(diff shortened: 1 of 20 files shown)


ChengTian:
Compatibility check during build in command-line:
compatibilitycheck_build

Compatibility check during restore in command-line:
compatibilitycheck_restore

Compatibility check during build in VS:
compatibilitycheck_tooling

src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
```diff
@@ -94,7 +94,7 @@ public RestoreCommand(FrameworkName fallbackFramework)
                 success &= await Execute(dir);
             }
 
-            foreach (var category in ErrorMessages)
+            foreach (var category in ErrorMessages.Where(x => x.Value.Any()))
```
davidfowl: Remove the restore check. We'll do an all up diagnostic change in restore.

ChengTian:
@davidfowl , removed check during restore and reused a lookup dictionary to save memory. Please check the last commit.

src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
davidfowl: Remove this?

ChengTian: Removed

src/Microsoft.Dnx.Runtime/DependencyManagement/CompatibilityChecker.cs
```diff
@@ -0,0 +1,122 @@
...
+                    Framework = framework,
+                    Type = CompatibilityIssue.IssueType.UnsupportedFramework
+                };
+            }
+
+            if (string.Equals(framework.Identifier, VersionUtility.NetPlatformFrameworkIdentifier, StringComparison.OrdinalIgnoreCase))
+            {
+                // If target framework is dotnet, we skip the ref-lib matching check
+                return null;
+            }
```
ChengTian: @davidfowl @anurse , added special case check

src/Microsoft.Dnx.Runtime/DependencyManagement/ProjectReferenceDependencyProvider.cs
```diff
@@ -97,7 +97,6 @@ public LibraryDescription GetDescription(LibraryRange libraryRange, FrameworkNam
                 Framework = targetFrameworkInfo.FrameworkName,
                 Dependencies = dependencies,
                 LoadableAssemblies = loadableAssemblies,
-                Compatible = !unresolved,
```
davidfowl: What?

src/Microsoft.Dnx.Runtime/DependencyManagement/NuGetDependencyResolver.cs
```diff
@@ -42,9 +43,15 @@ public NuGetDependencyResolver(PackageRepository repository)
 
         public void ApplyLockFile(LockFile lockFile)
         {
+            ApplyOptimizedLockFile(new OptimizedLockFile(lockFile));
+        }
+
+        public void ApplyOptimizedLockFile(OptimizedLockFile optimizedLockFile)
+        {
+            _compatibilityChecker = new CompatibilityChecker(optimizedLockFile);
```
davidfowl: Null this out in initialize so the memory is freed.

src/Microsoft.Dnx.Runtime/DependencyManagement/CompatibilityChecker.cs
```diff
@@ -0,0 +1,122 @@
...
+namespace Microsoft.Dnx.Runtime
+{
+    public class CompatibilityChecker
+    {
+        private readonly LockFile _lockFile;
+        private readonly ILookup<string, LockFileLibrary> _lockFileLibraries;
+        private readonly Dictionary<FrameworkName, HashSet<string>> _frameworkRuntimeAssemblies;
+
+        public CompatibilityChecker(LockFile lockFile)
+            : this(new OptimizedLockFile(lockFile))
```
davidfowl: Why do we need both? I don't trust that this won't be misused at runtime reducing the performance.

src/Microsoft.Dnx.Runtime/DependencyManagement/CompatibilityChecker.cs
```diff
@@ -0,0 +1,122 @@
...
+                    {
+                        yield return issue;
+                    }
+                }
+            }
+        }
+
+        public CompatibilityIssue CheckTargetLibrary(LockFileTargetLibrary targetLibrary, FrameworkName framework)
+        {
+            var lockFileLibrary = _lockFileLibraries[targetLibrary.Name].First(l => l.Version == targetLibrary.Version);
```
davidfowl: Why not just make the lookup based on id and version?

src/Microsoft.Dnx.Runtime/DependencyManagement/ProjectReferenceDependencyProvider.cs
ChengTian: Yeah, we should mark it as incompatible with `UnsupportedFramework` issue type

src/Microsoft.Dnx.Runtime/DependencyManagement/NuGetDependencyResolver.cs
ChengTian: We cannot null `_compatibilityChecker` out because it will be needed in another member method `GetDescription`

src/Microsoft.Dnx.Runtime/DependencyManagement/CompatibilityChecker.cs
ChengTian: Ok, remove this and only leave the one taking `OptimizedLockFile` in

src/Microsoft.Dnx.Runtime/DependencyManagement/CompatibilityChecker.cs
ChengTian: Because I wanted to save memory here by having a reusable lookup in `OptimizedLockFile`. If we make the lookup based on Tuple(id, version), then `PackageRepository.FindPackagesById()` cannot reuse this. Need to do profiling to make the tradeoff here.

src/Microsoft.Dnx.Runtime/DependencyManagement/NuGetDependencyResolver.cs
ChengTian: My bad, didn't understand what you meant. Will do.

ChengTian:
@davidfowl , I measured performance with profiler and here is statistics (5 times of running for each branch):

**Program under profiling**:
HelloWorld console app which has exactly same `project.json` as MusicStore: https://github.com/aspnet/MusicStore/blob/dev/src/MusicStore/project.json
It makes sure the number of dependencies is big enough to make dependency resolution time easy to observe.

**Current dev branch (dnx before this change)**:
Average total runtime: 2483ms
Profiling information of key part:
capture

**PR branch (dnx after this change)**:
Average total runtime: 2496.33ms
Profiling information of key part:
capture

The conclusion is that this compatibility check doesn't slow down `dnx run` significantly and the performance is acceptable.

src/Microsoft.Dnx.Runtime/DependencyManagement/CompatibilityChecker.cs
```diff
@@ -0,0 +1,117 @@
...
+            }
+
+            if (string.Equals(framework.Identifier, VersionUtility.NetPlatformFrameworkIdentifier, StringComparison.OrdinalIgnoreCase))
+            {
+                // If target framework is dotnet, we skip the ref-lib matching check
+                return null;
+            }
+
+            // Get run-time assemblies provided by current library
+            var runtimeAssemblies = _frameworkRuntimeAssemblies[framework];
```
davidfowl: Can this ever throw?

src/Microsoft.Dnx.Runtime/DependencyManagement/NuGetDependencyResolver.cs
```diff
@@ -42,9 +43,15 @@ public NuGetDependencyResolver(PackageRepository repository)
 
         public void ApplyLockFile(LockFile lockFile)
         {
+            ApplyOptimizedLockFile(new OptimizedLockFile(lockFile));
```
davidfowl: Who calls this?

src/Microsoft.Dnx.Runtime/DependencyManagement/CompatibilityChecker.cs
ChengTian: Yes, if you check compatibility for a framework that isn't present in lockfile. I can add check here

src/Microsoft.Dnx.Runtime/DependencyManagement/NuGetDependencyResolver.cs
ChengTian: ApplicationHostContext
https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Runtime/ApplicationHostContext.cs#L83

src/Microsoft.Dnx.Runtime/DependencyManagement/CompatibilityChecker.cs
ChengTian: This case only happens when you check compatibility against a framework that is not present in your `project.json`. Following current logic in DNX, this should never happen. I added check with `Debug.Assert`, which is for impossible case.

ChengTian:
Improved error message and added compatibility check during `dnu restore`

ChengTian:
Stale again...

@davidfowl , do we want to punt it?

ChengTian:
This change is too aggressive and we decide to punt it for now. Closing this PR and keeping the branch for future reference.

Issue #2343, opened 2015-07-25
Failed to initalize runtime (80131522) on Beta6/Beta7 x64


NoralK:
I received this error on a new clean installation.
Windows 7 Enterprise x64 - SP1 VM with UAC turned off (NO PATCHES)
.Net v4.5.2
PowerShell v4.0 x64
Internet Explorer 11 x64 with the patches it installs
Node.js v0.12.7 x64
VS Code v0.5.0

Environment Variable
```
DNX_UNSTABLE_FEED=https://www.myget.org/F/aspnetbeta6/api/v2
```
Installed DNVM
```
@powershell -NoProfile -ExecutionPolicy unrestricted -Command "&{$Branch='dev';iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/aspnet/Home/dev/dnvminstall.ps1'))}"
```
Installed Beta6 x64 package by:
```
dnvm upgrade -arch x64 -u -r clr
```
DNVM List
```
*    1.0.0-beta6    clr    x64    C:\Users\Noral\.dnx\runtimes    default
```
closed command prompt
opened command prompt
```
dnu restore
Failed to initalize runtime (80131522)
```
However, if I do the following:
```
dnvm upgrade -arch x86 -u -r clr
```
DNVM List
```
     1.0.0-beta6    clr    x64    C:\Users\Noral\.dnx\runtimes
*    1.0.0-beta6    clr    x86    C:\Users\Noral\.dnx\runtimes    default
```
I can run dnu for x86.  Did I miss something?

NoralK:
I think this might be related.  I build a simple MVC project, compiles fine for Beta5, and I get this error:
```
E:\Projects\HelloMvc>dnu build --quiet
Microsoft .NET Development Utility CLR-x86-1.0.0-beta7-12274


Building HelloMvc for DNX,Version=v4.5.1
System.TypeLoadException: Could not load type 'Microsoft.Framework.Runtime.ICompilerOptionsProvider' from assembly 'Microsoft.Framework.Runtime.Abstra
ctions, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'.
   at Microsoft.AspNet.Mvc.RazorPreCompileModule.BeforeCompile(BeforeCompileContext context)
   at Microsoft.Framework.Runtime.Roslyn.RoslynCompiler.CompileProject(CompilationProjectContext projectContext, IEnumerable`1 incomingReferences, IEn
[...]

PR #2431, opened 2015-08-06
Assigning id to domain manager assembly loaded from stream

moozzyk:
Not assigning the assembly id may result in failing to bootstrap CLR with HR 0x80131522

### DIFF
--- a/src/dnx.clr/HostAssemblyStore.cpp
+++ b/src/dnx.clr/HostAssemblyStore.cpp
@@ -8,7 +8,7 @@
 const wchar_t* AppDomainManagerAssemblyName = L"Microsoft.Dnx.Host.Clr, Version=1.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60, ProcessorArchitecture=MSIL";
 
 // IHostAssemblyStore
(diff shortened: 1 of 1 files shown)


jkotas:
LGTM

src/dnx.clr/HostAssemblyStore.cpp
```diff
@@ -8,7 +8,7 @@
 const wchar_t* AppDomainManagerAssemblyName = L"Microsoft.Dnx.Host.Clr, Version=1.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60, ProcessorArchitecture=MSIL";
 
 // IHostAssemblyStore
-HRESULT STDMETHODCALLTYPE HostAssemblyStore::ProvideAssembly(AssemblyBindInfo *pBindInfo, UINT64* /*pAssemblyId*/, UINT64* /*pContext*/,
+HRESULT STDMETHODCALLTYPE HostAssemblyStore::ProvideAssembly(AssemblyBindInfo *pBindInfo, UINT64* pAssemblyId, UINT64* /*pContext*/,
```
jkotas: Maybe initialize *pContext as well just in case it suffers from the same problem?

davidfowl:
:shipit:

src/dnx.clr/HostAssemblyStore.cpp
moozzyk: Will do.

moozzyk:
#2343

PR #2438, opened 2015-08-07
Add new CoreCLR dependency

BrennanConroy:
@davidfowl @troydai


davidfowl:
:shipit:

BrennanConroy:
They just emailed me, and the binary has been renamed. They also suggested taking any System.*.so/dylib for the upcoming beta. Thoughts?

davidfowl:
Sure, if that's the suggestion then modify the exclude logic a bit.

BrennanConroy:
@davidfowl

BrennanConroy:
7fd246c0b5e9c134786f3eeac7ff6774cdc498a2

PR #2439, opened 2015-08-07
Create a .Sources package for the Runtime JSON parser

rynowak:
/cc @davidfowl


DNX.sln
```diff
@@ -63,13 +63,13 @@ Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.Dnx.Compilation.C
...
 EndProject
-Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "dnx.win32", "src\dnx\dnx.vcxproj", "{D0E2FB09-0FEA-478A-9068-D6AA420C6DED}"
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "dnx", "src\dnx\dnx.vcxproj", "{D0E2FB09-0FEA-478A-9068-D6AA420C6DED}"
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "dnx.clr", "src\dnx.clr\dnx.clr.vcxproj", "{0E4CB542-5E58-4D24-9CAE-DAC83D932DBE}"
 EndProject
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.Dnx.Host.Clr", "src\Microsoft.Dnx.Host.Clr\Microsoft.Dnx.Host.Clr.xproj", "{2D50D502-E10A-4B1C-8A49-DBC52B5CE9CB}"
 EndProject
-Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "dnx.win32.coreclr", "src\dnx.coreclr\dnx.coreclr.vcxproj", "{1F31D08E-DE21-48F6-ADE3-315EA6268DD1}"
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "dnx.coreclr", "src\dnx.coreclr\dnx.coreclr.vcxproj", "{1F31D08E-DE21-48F6-ADE3-315EA6268DD1}"
```
rynowak: VS did this on its own.

Eilon:
From where do we intend to start consuming this?

rynowak:
TestHost

DNX.sln
moozzyk: Looks like VS started using the name of vcxproj instead of $(ProjectName)... I think this is actually a good changed (if it is not breaking) since both dnx.vcxproj and dnx.coreclr.vcxproj can create win32 and onecore binary.

src/Microsoft.Dnx.Host.Clr/project.json
```diff
@@ -1,8 +1,14 @@
...
-    },
+  "version": "1.0.0-*",
+  "compilationOptions": {
+    "define": [ "TRACE" ],
+    "allowUnsafe": true,
+    "warningsAsErrors": true,
+    "keyFile": "KeyFile.snk"
+  },
+  "dependencies": {
+    "Microsoft.Dnx.Runtime.Json.Sources": { "version": "1.0.0-*", "type": "build" }
```
moozzyk: Why do we need it here? Does it work at all? Microsoft.Dnx.Host.Clr has a real strong name and cannot depend on non-signed assemblies (it may fail in the second build). This is the reason why it does not have any dependencies.

moozzyk: I know why we need it - recent changes to set .NET Framework version. I wonder if you can fake strong name `Microsoft.Dnx.Runtime.Json.Sources` to make it work. I believe the plan is to fake strong name all assemblies anyways.

Eilon: Hmm.... it's not an assembly, it's shared source...

Eilon:
And TestHost can't use Json.NET? Remember that this is a super simple parser with minimal features. And it can only deserialize - there's no serializer (yet).

src/Microsoft.Dnx.Host.Clr/project.json
moozzyk: Then it should work, indeed.

davidfowl:
@Eilon it could but it's better if the test infrastructure has as little dependencies as possible. It's pretty similar to being a runtime itself.

Eilon:
@davidfowl I certainly agree in principle, but wondering if it's really worth it. That is, what problem does it solve?

davidfowl:
@Eilon it lets James test JSON.NET without conflicting with the version the test runner is using :smile:

Eilon:
Fine, do it :smile:

rynowak:
Let's be real - I hoping to have a like-for-like replacement of the code that's currently being used to read project files. I'm not allowed to use the `Project` class directly, so this is the next best thing.

The TestHost still has a dependency on JSON.NET, because it also needs to serialize data. I'd love to remove that dependency entirely, but right now it's needed for serialization to VS. Maybe we should discuss more.

davidfowl:
:shipit:

davidfowl:
@rynowak Are you going to merge this for beta7?

victorhurdugaci:
:shipit: after rebase

Issue #643, opened 2014-09-18
PackageManager should use System.Net.Http


pranavkm:
The System.Net.Http package has a `HttpClient` which would be the replacement for `Microsoft.Net.HttpClient`.

muratg:
@ChengTian @pranavkm what's the status of this issue?

ChengTian:
Buggy HttpClient from CoreCLR was preventing this from being done. I will try integrating the latest CoreCLR HttpClient to see whether it usable now.

muratg:
@glennc do we need this in RC1?

ChengTian:
@danroth27 talked to me and we should keep trying to integrate HttpClient from CoreCLR and filing bugs if we can catch any with `dnu restore`. Putting this to `beta6`.

ChengTian:
Still waiting for bug fix from CoreCLR team: https://github.com/dotnet/corefx/issues/2168

The purpose of this task is testing HttpClient on CoreCLR and it is not an urgent issue. Moving to next milestone. @muratg

PR #2443, opened 2015-08-08
Use CoreCLR HttpClient in "dnu restore"

ChengTian:
parent https://github.com/aspnet/dnx/issues/643

Finally we can get this work with CoreCLR on Windows.

I tested restoring for MusicStore from v2/v3 feeds with/without authenticated proxy. All basic scenarios work.

@davidfowl @Tratcher @Eilon @danroth27


Eilon:
:shipit: !

davidfowl:
:shipit:

Issue #1752, opened 2015-04-30
Add timeout values to all regex expressions


blowdart:
To avoid regex attacks add a time-out value to every regular expression in the code base.

ChengTian:
I remember @blowdart finds that currently we are unable to add this to CoreCLR version of runtime.

Should be a post-beta5 issue.

@muratg , do we want a `Security` label for this kind of issue, or just use `Bug`?

blowdart:
I'd stick with bug.

As for CoreCLR this is not true. CoreCLR does not provide a global default. You can, and should, in both desktop and core use the regex constructor which takes the Timeout everywhere.

PR #2457, opened 2015-08-11
Add regex match timeout

ChengTian:
parent https://github.com/aspnet/dnx/issues/1752

@davidfowl @Eilon 

Looks like the owners of other repos are reacting to similar issues (for example https://github.com/aspnet/Identity/issues/456), so I'll just react in DNX and FileSystem, which is a submodule of DNX: https://github.com/aspnet/FileSystem/pull/119


Eilon:
@blowdart to review.

This looks fine to me.

Issue #2447, opened 2015-08-10
DNU should allow absence of "versions" property in v3 index.json


ChengTian:
That means we also consider unlisted packages during restore. It can also fix the exception thrown by `dnu restore` when restoring package like https://az320820.vo.msecnd.net/v3-flatcontainer/classlibrary1/index.json

@davidfowl  agrees that this can be assigned to me.

pranavkm:
cc @yishaigalatzer Might be something that needs to be ported to nuget.

yishaigalatzer:
This seems wrong. The plan is to fix the server, although we can revisit the discussion worth a short f2f tomorrow

ChengTian:
@yishaigalatzer , to fix at server side, do you want to return `index.json` that always contain `versions` property, even it is empty?

The current problem is that if all versions of a package are unlisted, we get `index.json` like
```json
{
  "unlistedVersions": [
    "1.0.0"
  ]
}
```
The absence of `versions` property makes `dnu` think it is an invalid response.

/cc @davidfowl @lodejard

ChengTian:
Talked to @davidfowl and we want to leave the current implementation for now. I filed an issue for fixing at server side: https://github.com/NuGet/NuGetGallery/issues/2617

ChengTian:
Need a discussion to decide whether we still want this change. Putting to backlog.

lodejard:
Seems reasonable to treat missing property the same as an empty array

ChengTian:
@lodejard , makes sense, I have a PR for it here: https://github.com/aspnet/dnx/pull/2459. Please help to review.

ChengTian:
The work at server side is tracked at https://github.com/NuGet/NuGetGallery/issues/2617 and NuGet team is deploying the changes now. The fix at DNX side is merged. Closing.

PR #2459, opened 2015-08-11
Allow absence of "versions" array in v3 feed response

ChengTian:
parent https://github.com/aspnet/dnx/issues/2447#issuecomment-130057419

@davidfowl @lodejard

### DIFF
--- a/src/Microsoft.Dnx.Tooling/Restore/NuGet/NuGetv3Feed.cs
+++ b/src/Microsoft.Dnx.Tooling/Restore/NuGet/NuGetv3Feed.cs
@@ -140,7 +140,14 @@ namespace Microsoft.Dnx.Tooling.Restore.NuGet
                                 doc = JObject.Load(new JsonTextReader(reader));
                             }
 
-                            var result = doc["versions"]
+                            var versions = doc["versions"];
+                            if (versions == null)
+                            {
+                                // Absence of "versions" property is equivalent to an empty "versions" array
+                                return Enumerable.Empty<PackageInfo>();
+                            }
+
+                            var result = versions
                                 .Select(x => BuildModel(id, x.ToString()))
                                 .Where(x => x != null);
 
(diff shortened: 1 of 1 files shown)


ChengTian:
Ping

src/Microsoft.Dnx.Tooling/Restore/NuGet/NuGetv3Feed.cs
```diff
@@ -256,8 +262,9 @@ private static void EnsureValidFindPackagesResponse(Stream stream, string uri)
                 var json = JObject.Load(new JsonTextReader(new StreamReader(stream)));
                 if (json["versions"] == null)
```
pranavkm: Maybe avoid doing multiple lookups for the same key?

```C#
var versions = json["versions'];
if (versions != null && versions.Type != JTokenType.Array)
{
   throw new InvalidDataException(
       $"{message} The value of 'versions' property is not an array.");
}
```

Maybe a future refactoring, but it would be useful to not reparse the json multiple times over (once here and once in the body of `FindPackagesByIdAsyncCore`). Perhaps this could stash the parsed json somewhere that flows via `data`?

src/Microsoft.Dnx.Tooling/Restore/NuGet/NuGetv3Feed.cs
```diff
@@ -140,6 +140,12 @@ public Task<IEnumerable<PackageInfo>> FindPackagesByIdAsync(string id)
                                 doc = JObject.Load(new JsonTextReader(reader));
                             }
 
+                            if (doc["versions"] == null)
```
pranavkm: Same deal with not doing multiple lookups?

ChengTian:
Thanks for the comments. Just fixed.

pranavkm:
:shipit:

anurse:
:shipit:

Issue #2460, opened 2015-08-11
Make build error message bold


ChengTian:
Current error message is in dark red, which makes it hard to read in black background:

![](https://cloud.githubusercontent.com/assets/1383883/9118557/0dc38948-3c26-11e5-83b7-305d4478e224.PNG)

Making the message bold should make it lighter, similar to
![](https://cloud.githubusercontent.com/assets/1383883/9118562/1aef417a-3c26-11e5-84a3-53938eeea1c1.PNG)

PR #2461, opened 2015-08-11
Make build error message bold

ChengTian:
parent https://github.com/aspnet/dnx/issues/2460

@davidfowl 

Before the change:
capture_before

After the change:
capture_after


davidfowl:
:shipit:

PR #2464, opened 2015-08-12
Ensuring readability of servicing index

moozzyk:
The application now will fail if:
- the %DNX_SERVICING%\index.txt cannot be open
- %PROGRAMFILES(X86)%\Microsoft DNX\Servicing exists but does not contain index.txt file

### DIFF
--- a/src/dnx/servicing.cpp
+++ b/src/dnx/servicing.cpp
@@ -70,7 +70,18 @@ namespace dnx
                 ? utils::path_combine(servicing_root_parent, std::wstring(L"Microsoft DNX\\Servicing"))
                 : servicing_root_parent;
 
+            if (is_default_servicing_location && !utils::directory_exists(servicing_root))
+            {
+                trace_writer.write(L"The default servicing root does not exist.", true);
+                return std::wstring{};
+            }
+
             auto servicing_manifest_path = utils::path_combine(servicing_root, std::wstring(L"index.txt"));
+
+            if (!utils::file_exists(servicing_manifest_path))
+            {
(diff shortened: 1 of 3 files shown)


moozzyk:
@victorhurdugaci

moozzyk:
#2406

moozzyk:
#2336

moozzyk:
(AppVeyor failure: Build execution time has reached the maximum allowed time for your plan (40 minutes))

src/dnx/servicing.cpp
```diff
@@ -70,8 +70,21 @@ namespace dnx
...
+                trace_writer.write(L"The default servicing root does not exist.", true);
+                return std::wstring{};
+            }
+
             auto servicing_manifest_path = utils::path_combine(servicing_root, std::wstring(L"index.txt"));
 
+            if (!utils::file_exists(servicing_manifest_path))
+            {
+                throw std::runtime_error(std::string("The servicing index file at: ")
+                    .append(dnx::utils::to_string(servicing_manifest_path))
```
moozzyk: @blowdart - is it OK to return the path in the exception? (Note that I am doing a similar change in Helios and there it could potentially show on the error page)

blowdart: You're not giving it away in the previous error, I'd rather you didn't do it here, just so Helios doesn't give it away.

moozzyk: The previous error is about the default servicing root which is a well known location. Here we would potentially write the contents of the %DNX_SERVICING%. I will change the code to omit this information.

anurse:
:shipit: given @blowdart 's comment is resolved

Issue #2312, opened 2015-07-22
Project references: Allow referencing with a name different than the containing folder


muratg:

anurse:
Thoughts (just a bit of a brain dump to facilitate discussion/design):
* We could allow a `path` property under the dependency. That starts to smell a lot like `HintPath` though...
* We could allow a project.json to have an **optional** `name` property to change the name of the project. This would mean someone has to scan all the project.json files in scope in order to determine the name of each project. Perhaps we could do that during `dnu restore` and put it in the lock file?

troydai:
If the scope of the problem is limited to "Allow a different project reference name to be used other than folder name". The solutions are simple:

__Solution 1.1__
	
_Move the folder name to a "localName" property and assign the new name as the key of the reference._

```JSON
{
    "dependencies": {
        "NewAlpha": { "version": "1.0.0", "localName": "Alpha" }
    }
}
```

__Solution 1.2__
	
_Load in dependency project's project.json and use "id" or other property to determine the actual name. Downside of this solution is that project.json of dependencies need to be actively_

__However when we expand the boundry of the problem to include solving the duplicate project references name among the search paths issue, there are more aspects and solution to consider.__
 
_Repo: https://github.com/troydai/dnx-2312/tree/master/samples/Zero_

Error:
```
System.InvalidOperationException: The project name 'Alpha' is ambiguous between the following projects:
/Users/troydai/code/playground/dnx2312/samples/GroupA/Alpha
/Users/troydai/code/playground/dnx2312/samples/GroupB/Alpha
```

__Solution 2.1__
_Automatic adding suffix to the project name when duplicate names are found._

```
For example
	GroupA/Alpha => Alpha.1
	GroupB/Alpha => Alpha.2
```
Pro: 
  a. Project.json schema keeps intact;
  b. ProjectInformation is still evaluated lazily
  c. Automatic
Con:
  a. Confusing project names;
  b. Make it hard to track error base on output;
  c. Can't guarantee the name's persistence
	
__Solution 2.2__
_Read in "name", "id" or other properties from dependency's project.json to determine a name._
Pro:
  a. Intuitive;
  b. Achievable without changing project.json schema
Con:
  a. Performance: project.json need to be actively load (Mitigation: optimized the JsonDeserializer to load partial data.);
  b. If the properties are missing from dependency's project.json, need to fall back solution.

__Solution 2.3__
_Add a "hint" path in project.json_
```		
{
    "dependencies": {
[...]

PR #2466, opened 2015-08-12
Add Id-property to `Project` to override NuGet package-id

billboga:
Previous behavior exclusively uses project name.
New behavior allows project config. to provide override via `id`-key.

Note: could not find any existing tests for validating other properties on `Project` related to `BuildManager`.


khalidabuhakmeh:
:+1:

davidfowl:
Yea, nope this change can't go in without other related changes. See https://github.com/aspnet/dnx/issues/2312

Issue #2417, opened 2015-08-05
Ensure DNU sends a User-Agent HTTP header


xavierdecoster:
Moved from https://github.com/aspnet/Home/issues/801

PR #2468, opened 2015-08-12
"dnu restore" sends User-Agent header

ChengTian:
parent https://github.com/aspnet/dnx/issues/2417

Here is the User-Agent header being sent:
capture1

I used the header sent by nuget.exe as a reference:
capture

The agent name comes from the title of help info:
capture

Although the full name of DNU is `DNX Utility` according to our wiki: https://github.com/aspnet/Home/wiki/DNX-utility

@davidfowl @lodejard @Eilon @glennc


Eilon:
:shipit:

PR #2470, opened 2015-08-13
Using the name defined in project.json as the Project's actual name.

troydai:
Address: #2312 

If the name is missing, fall back to the folder name. Tested in VS.

Question: Fall back the path search when lock file is missing runtime? Or let it fail.

Following: Going to add more tests.


src/Microsoft.Dnx.Runtime/PathSearchBasedProjectResolver.cs
```diff
@@ -0,0 +1,138 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.Dnx.Runtime
+{
+    public class PathSearchBasedProjectResolver : IProjectResolver
```
davidfowl: Why has this code changed at all?

src/Microsoft.Dnx.Runtime/LockFileBasedProjectResolver.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        private readonly string _baseDirectory;
+        private readonly LockFile _lockFile;
+        private readonly ConcurrentDictionary<string, Project> _projectsCache;
+        private readonly List<string> _searchPaths;
+
+        public LockFileBasedProjectResolver(LockFile lockFile, string projectDirectory, Project root)
+        {
+            _lockFile = lockFile;
+            _projectsCache = new ConcurrentDictionary<string, Project>();
```
davidfowl: Why is this dictionary concurrent?

src/Microsoft.Dnx.Runtime/LockFileBasedProjectResolver.cs
```diff
@@ -0,0 +1,64 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.Dnx.Runtime
+{
+    public class LockFileBasedProjectResolver : IProjectResolver
```
davidfowl: This implementation just needs to create the dictionary up front mapping names to paths and nothing more. Do the computation in the ctor.

src/Microsoft.Dnx.Runtime/PathSearchBasedProjectResolver.cs
troydai: Because project name is not determined by folder name alone. Project needs to be load in to parse the name. It can't be indexed when project.json files are collected.

troydai:
Need more reviewer:
/cc @victorhurdugaci @anurse @BrennanConroy @lodejard @loudej

src/Microsoft.Dnx.Runtime/PathSearchBasedProjectResolver.cs
anurse: Why is it still in the runtime? Isn't the runtime only using the LockFile table now?

troydai: Because `ApplicationHostContext` need to fall back to it in DTH. VS could open a project which is not restored. Under that situation lock file doesn't exist, project resolving will relies on path search.

victorhurdugaci:
Please check that projects with ids work with `dnu sources`

src/Microsoft.Dnx.Runtime/PathSearchBasedProjectResolver.cs
anurse: Sounds like it should be in DTH then :)

src/Microsoft.Dnx.Runtime/ApplicationHostContext.cs
```diff
@@ -76,6 +74,9 @@ public class ApplicationHostContext
                 {
                     NuGetDependencyProvider.ApplyLockFile(_lockFile);
 
+                    ProjectResolver = new LockFileBasedProjectResolver(_lockFile, ProjectDirectory, Project);
```
ChengTian: You don't need to pass `ProjectDirectory` in because you have `Project.ProjectDirectory`

src/Microsoft.Dnx.Runtime/LockFileBasedProjectResolver.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            _lockFile = lockFile;
+
+            _baseDirectory = projectDirectory;
+
+            _projects = GetSearchPaths(lockFile, projectDirectory);
+            _projects[root.Name] = new ProjectInformation(root);
+        }
+
+        public IEnumerable<string> SearchPaths
```
davidfowl: This is missing the `global.json` paths.

src/Microsoft.Dnx.Runtime/LockFileBasedProjectResolver.cs
```diff
@@ -0,0 +1,85 @@
...
+
+            public Project Project
+            {
+                get { return _project.Value; }
+            }
+
+            public string ProjectDirectory
+            {
+                get;
+            }
```
ChengTian: nit: follow convention and make it a one-liner

test/Bootstrapper.FunctionalTests/BootstrapperTests.cs
```diff
@@ -118,7 +117,7 @@ public void BootstrapperInvokesApplicationHostWithExplicitAppBase(string flavor,
         {
             var runtimeHomeDir = _fixture.GetRuntimeHomeDir(flavor, os, architecture);
 
-            using (var tempSamplesDir = TestUtils.PrepareTemporarySamplesFolder(runtimeHomeDir))
+            using (var tempSamplesDir = TestUtils.PrepareTemporarySamplesFolder(runtimeHomeDir, globalJson: false))
```
ChengTian: Why? How do you determine solution root if you don't have a global.json? I believe after you restore for `HelloWorld`, its dependency project `HelloShared` should be present in HelloWorld/project.lock.json. How does `dnu restore` find this dependency when you don't have a solution root?

src/Microsoft.Dnx.Runtime/LockFileBasedProjectResolver.cs
troydai: will do.

test/Bootstrapper.FunctionalTests/BootstrapperTests.cs
troydai: The tradition `ProjectResolver` not only look at one project but also it's sibling folder therefore `HelloShare` is resolved. The problem with adding a global.json here is that it's adding unnecessary paths to the `DNX` source code path (https://github.com/aspnet/dnx/blob/victorhu/require-servicing/test/Microsoft.Dnx.CommonTestUtils/TestUtils.cs#L246). As a result when restore the sample application dependences like `Microsoft.Dnx.Compilation.Abstraction` will be resolved as project instead of package.

src/Microsoft.Dnx.Runtime/LockFileBasedProjectResolver.cs
troydai: Path to the `global.json` was not added to the `_searchPaths`. New? https://github.com/aspnet/dnx/blob/victorhu/require-servicing/src/Microsoft.Dnx.Runtime/ProjectResolver.cs#L71

src/Microsoft.Dnx.Runtime/PathSearchBasedProjectResolver.cs
troydai: It is needed in runtime as fallback, per discussion with @davidfowl

src/Microsoft.Dnx.Runtime/LockFileBasedProjectResolver.cs
davidfowl: Not the paths to global.json, the paths declared inside of global.json

troydai:
:watch:

test/Bootstrapper.FunctionalTests/BootstrapperTests.cs
ChengTian: I see. Sounds good.

troydai:
@davidfowl

ChengTian:
Looks good to me :shipit:

davidfowl:
We've decided to delay this

PR #2473, opened 2015-08-13
Remove System.Reflection.Metadata dependency

troydai:
Since Roslyn package now correctly references its dependencies, it is not necessary to reference `System.Reflection.Metadata` specifically.


troydai:
I'll merge once travis pass.

anurse:
:shipit:

PR #2474, opened 2015-08-13
fix #2194 (for DNX) by setting Config File correctly

### LINKED ISSUE #2194: Set configuration file for .NET Framework to App.config next to the appbase
This will improve the compatibility with ConfigurationManager for .NET 4.x libraries running on DNX 45x

anurse:
Fixes #2194 

I'll be checking Helios separately and submitting a bug/PR for that if a change needs to be made. As far as I can tell, Helios is doing the right thing though.

/cc @davidfowl @jods4

### DIFF
--- a/src/Microsoft.Dnx.Runtime.Sources/Impl/Constants.cs
+++ b/src/Microsoft.Dnx.Runtime.Sources/Impl/Constants.cs
@@ -20,11 +20,10 @@ namespace Microsoft.Dnx.Runtime
         public const string WebConfigRuntimePath = "runtime-path";
         public const string BootstrapperHostName = RuntimeShortName + ".host";
         public const string BootstrapperClrName = RuntimeShortName + ".clr";
-        public const string ProjectJsonAppDomainDataKey = ".dnx.clr.DomainManager.projectJson";
-        public const string ProjectJsonPathAppDomainDataKey = ".dnx.clr.DomainManager.projectJsonPath";
 
         public const int LockFileVersion = 1;
 
         public static readonly TimeSpan RegexMatchTimeout = TimeSpan.FromSeconds(10);
+        public static readonly string AppConfigurationFileName = "app.config";
     }
 }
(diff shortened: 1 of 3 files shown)


src/Microsoft.Dnx.Runtime.Sources/Impl/Constants.cs
```diff
@@ -20,11 +20,10 @@ internal static class Constants
         public const string WebConfigRuntimePath = "runtime-path";
         public const string BootstrapperHostName = RuntimeShortName + ".host";
         public const string BootstrapperClrName = RuntimeShortName + ".clr";
-        public const string ProjectJsonAppDomainDataKey = ".dnx.clr.DomainManager.projectJson";
-        public const string ProjectJsonPathAppDomainDataKey = ".dnx.clr.DomainManager.projectJsonPath";
 
         public const int LockFileVersion = 1;
 
         public static readonly TimeSpan RegexMatchTimeout = TimeSpan.FromSeconds(10);
+        public static readonly string AppConfigurationFileName = "App.Config";
```
moozzyk: Shouldn't it actually be app.config (lower case?)

moozzyk:
Why do we need the files in the `misc/XreTestApps/ConfigTest/*` when we are creating all the files in the test itself?

anurse:
Those are for ad-hoc testing. I can remove them though.

src/Microsoft.Dnx.Runtime.Sources/Impl/Constants.cs
anurse: :+1: though I don't actually know what it should be. Regardless, this is only used on case-insensitive file systems (because it's Windows-only). I'll change it to lower-case for consistency though.

davidfowl:
:shipit:

src/Microsoft.Dnx.Runtime.Sources/Impl/Constants.cs
moozzyk: :clap:

moozzyk:
:ship: :it:

PR #2477, opened 2015-08-13
Parallelize package install during publish

ChengTian:
parent: https://github.com/aspnet/dnx/issues/2429 and https://github.com/aspnet/dnx/issues/2376

@davidfowl @Eilon 

Before this change, `dnu publish` takes 30.7 sec to bundle a Starter Web App:
capture

After this change, `dnu publish` takes 13.5 sec to bundle a Starter Web App:
capture1

Note that I also polished output of `dnu publish`, otherwise the parallelization will mess it up.


src/Microsoft.Dnx.Tooling/Packages/Add/AddOptions.cs
```diff
@@ -9,5 +9,9 @@ namespace Microsoft.Dnx.Tooling.Packages
     public class AddOptions : PackagesOptions
     {
         public string NuGetPackage { get; set; }
+
+        // When "dnu packages add" is invoked as a subcommand of other commands
+        // this option can be set to true to have trimmed output
+        public bool ConciseOutput { get; set; }
```
davidfowl: Concise? What about SuppressOutput?

src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
```diff
@@ -66,6 +66,7 @@ public RestoreCommand(FrameworkName fallbackFramework)
         public bool CheckHashFile { get; set; } = true;
         public bool SkipInstall { get; set; }
         public bool SkipRestoreEvents { get; set; }
+        public bool SkipFeedSummary { get; set; }
```
davidfowl: SuppressOutput? We don't really want any output for restore during publish

troydai:
It looks good to me. :shipit:

src/Microsoft.Dnx.Tooling/Packages/Add/AddOptions.cs
ChengTian: I am blocking part of output here and that's why I don't name it suppress. Sounds like you are OK with blocking all output. Then I'll block all output by making `packages add` quiet. There is even no new option needed here.

src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
ChengTian: Will block all output.

src/Microsoft.Dnx.Tooling/Publish/PublishPackage.cs
```diff
@@ -30,11 +31,13 @@ public void Emit(PublishRoot root)
             {
                 NuGetPackage = srcNupkgPath,
                 SourcePackages = root.TargetPackagesPath,
-                Reports = root.Reports
             };
 
+            // Mute "packages add" subcommand
+            options.Reports = Reports.Constants.NullReports;
```
ChengTian: When publishing a starter web app, there are 160+ packages get published and it means that this method is going to be called many many times. That's why I create a constant to avoid allocating/deallocating numerous `NullReport` objects.

ChengTian:
@davidfowl , updated

ChengTian:
Merged https://github.com/aspnet/dnx/commit/0cd694b54d39da2b4fbbab14e74cf3bf5e520076

Issue #2060, opened 2015-06-13
DTH doesn't send us new "Error" message with the fixed version for global.json change


PradeepKadubandi:
This is logged from TFS issue : http://bung/#/DevDiv/search/table?q=CID%3DDevDiv.1156067 

The repro steps are open any project, open it's global.json, introduce some syntax error (remove an existing comma), see that References node says to see error list, correct the error in global.json, notice that references node is still like that.

I debugged through the issue and this is what I found.

We used to send 'FilesChanged' for any global.json changes and the above used to work then (I tried that change that fixes the issue though regresses a scenario).

That means, we send a FilesChanged first, DTH sends us 'Error' message with error information from global.json parsing, user fixes global.json, we send FilesChanged again and DTH sends us new 'Error' message which does not contain an error (values being null and -1).

We are currently sending 'RefreshDependencies' for global.json change (for a valid reason, changes to global.json can change the dependency type from package to project).

The above sequence doesn't seem to work in that order.

We send 'RefreshDependencies', DTH sends us 'Error' message with the error information from global.json parsing, user fixes global.json, we send 'RefreshDependencies' again and DTH does not send us new 'Error' message without the error (values being null and -1).

davidfowl:
@PradeepKadubandi was this supposed to be for beta5?

troydai:
Fixed in 1033f3103a2bb5c6af3d325f6444604ef3a54c59

PR #2480, opened 2015-08-14
Clear the error message after global.json is fixed

troydai:
Addressing #2060 

/cc @davidfowl @PradeepKadubandi

### DIFF
--- a/src/Microsoft.Dnx.DesignTimeHost/ApplicationContext.cs
+++ b/src/Microsoft.Dnx.DesignTimeHost/ApplicationContext.cs
@@ -140,6 +140,7 @@ namespace Microsoft.Dnx.DesignTimeHost
                 };
 
                 _initializedContext.Transmit(message);
+                _remote.GlobalErrorMessage = error;
 
                 // Notify anyone waiting for diagnostics
                 foreach (var connection in _waitingForDiagnostics)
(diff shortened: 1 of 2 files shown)


src/Microsoft.Dnx.DesignTimeHost/ApplicationContext.cs
```diff
@@ -52,6 +52,9 @@ public class ApplicationContext
         private readonly CompilationEngineFactory _compilationEngineFactory;
         private int? _contextProtocolVersion;
 
+        private readonly Message _emptyCleanMessage;
```
davidfowl: _noErrorMessage

troydai: will change

davidfowl:
So this change isn't what I would expect. I'd expect us to keep track of the last error in the World and update the remote and local world and compare them appropriately like the other things.

troydai:
The only place we send out "Error" message is in the "catch all" in the processing loop. It is not associated or computed during update the "World".

troydai:
Ok to sign off? Or have to associate the error with World?

troydai:
Updated. Error is registered in remove `World`.

troydai:
:watch:

davidfowl:
:shipit:

PR #2486, opened 2015-08-14
fix DTH memory explosion

### LINKED ISSUE #2483: Extremely high memory usage in DTH for Mvc.NoFun.sln
I'm now regularly seeing memory usage climb to around 1.2gb in DTH on the smaller MVC solution (~30 projects). This makes VS slow to crawl and every operation feels extremely sluggish. 

Using 1.0.0-beta7-12356

I can easily repro this, and it seems to occur most frequently when doing a build in VS resulting in compilation errors.

I have a memory dump available, here's what's using all of the memory and how it's rooted:
image

anurse:
/cc @rynowak @davidfowl @lodejard 

Figured it out!

Fixes #2483

### DIFF
--- a/src/Microsoft.Dnx.DesignTimeHost/ConnectionContext.cs
+++ b/src/Microsoft.Dnx.DesignTimeHost/ConnectionContext.cs
@@ -17,6 +17,7 @@ namespace Microsoft.Dnx.DesignTimeHost
         private readonly IDictionary<int, ApplicationContext> _contexts;
         private readonly IServiceProvider _services;
         private readonly ProtocolManager _protocolManager;
+        private readonly CompilationCache _cache;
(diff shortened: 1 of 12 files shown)


src/Microsoft.Dnx.DesignTimeHost/ConnectionContext.cs
```diff
@@ -63,7 +65,7 @@ public void OnReceive(Message message)
 
                     applicationContext = new ApplicationContext(_services,
                                                                 _protocolManager,
-                                                                new CompilationEngineFactory(NoopWatcher.Instance, new CompilationCache()),
+                                                                new CompilationEngineFactory(_cache),
```
anurse: The main culprit.

PR #2488, opened 2015-08-14
Servicing:

moozzyk:
- enabling retrying reading of the servicing index file in case of failures
- preventing from using network drives for custom servicing root

Bonus:
- removing a path to the requested servicing location from an exception message

### DIFF
--- a/src/dnx/servicing.cpp
+++ b/src/dnx/servicing.cpp
@@ -7,7 +7,15 @@
 #include <string>
 #include <fstream>
 #include <algorithm>
+#include <thread>
+#include <chrono>
 #include "trace_writer.h"
+
+#if defined(ONECORE)
+#include <Pathcch.h>
+#else
+#include <Shlwapi.h>
+#endif
 
 namespace
 {
@@ -27,8 +35,9 @@ namespace
 #endif
         ;
 
-    std::wstring find_runtime_replacement(std::ifstream& input, dnx::trace_writer& trace_writer)
+    std::wstring find_runtime_replacement(std::ifstream& input, dnx::trace_writer& trace_writer, bool& completed_successfully)
     {
+        completed_successfully = true;
         const std::string runtime_qualifier = std::string{ "dnx|" } + runtime_moniker + "|" + ProductVersionStr + "=";
 
         trace_writer.write(
@@ -46,6 +55,7 @@ namespace
 
         if (!input.eof())
         {
+            completed_successfully = false;
             trace_writer.write(L"Error occured while reading contents of servicing index file.", false);
         }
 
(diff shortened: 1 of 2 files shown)


src/dnx/servicing.cpp
```diff
@@ -77,6 +88,18 @@ namespace dnx
             return require_servicing_buffer_length != 0 && require_servicing_buffer[0] == L'1';
         }
 
+        bool is_network_path(const std::wstring& path)
+        {
+            wchar_t buff[MAX_PATH];
+            return
+#if defined(ONECORE)
+                PathIsUNCEx(path.c_str(), nullptr)
```
victorhurdugaci: Use a macro instead?

moozzyk: No - I hate this thing... (and parameters are different)

src/dnx/servicing.cpp
```diff
@@ -84,15 +107,12 @@ namespace dnx
...
 
             if (!servicing_root_exists)
             {
-                if (require_servicing)
+                if (require_servicing())
                 {
-                    throw std::runtime_error(std::string("Servicing is required for the application to run but the servicing folder '")
-                        .append(dnx::utils::to_string(servicing_root))
-                        .append("' does not exist. The application will now exit."));
+                    throw std::runtime_error("Servicing is required for the application to run but the servicing folder does not exist.");
```
victorhurdugaci: I don't think that printing the folder location is a security issue. It simply makes the message more actionable

src/dnx/servicing.cpp
```diff
@@ -109,26 +134,38 @@ namespace dnx
...
             }
 
-            // index.txt is ASCII
-            std::ifstream servicing_manifest;
-            servicing_manifest.open(servicing_manifest_path, std::ifstream::in);
-            if (servicing_manifest.is_open())
+            for (auto i = 0; i < 3; i++)
             {
-                trace_writer.write(std::wstring(L"Found servicing index file at: ").append(servicing_manifest_path), true);
+                trace_writer.write(std::wstring(L"Reading servicing index file - attempt: ")
```
victorhurdugaci: "attempt x of y". instead of "attempt x"

src/dnx/servicing.cpp
moozzyk: Because it is in the exception you never know where it will be displayed. We discussed the same thing with @blowdart a few days ago and ended up removing the path. That's why I did this here too.

src/dnx/servicing.cpp
moozzyk: will update

victorhurdugaci:
Looks good to me :shipit:

moozzyk:
#2336

Issue #2489, opened 2015-08-14
More perf improvement for "dnu publish"


ChengTian:
A significant perf improvement was made in beta7 (https://github.com/aspnet/dnx/issues/2429 https://github.com/aspnet/dnx/pull/2477) and @davidfowl said we need to explore more about publish perf in later milestones.

ChengTian:
One suggestion from @davidfowl : for the package installation part, we may want to xcopy packages, instead of extracting from nupkgs

ChengTian:
Publishing a Starter Web App with command-line is about 5 seconds now.

PR #2491, opened 2015-08-14
Reuse hash from .nupkg.sha512 files

ChengTian:
parent https://github.com/aspnet/dnx/issues/2489

@davidfowl , during `dnu publish`, `SHA512.ComputeHash()` occupies 96% of time used by `packages add` subcommand:
publish_perf

We can get rid of hash computation by reading `.nupkg.sha512` file, which should be side by side with nupkg. Here is the profiler output after the change:
publish_perf_more


moozzyk:
Looks good to me :ship::it:

ChengTian:
We decide to make this beta7

PR #2495, opened 2015-08-16
Stop sending ErrorMessage for each processing loop.

troydai:
/cc @davidfowl


davidfowl:
:shipit:

troydai:
Will merge after travis pass.

Issue #1894, opened 2015-05-19
project.json targeting net20 and net35


JamesNK:
I am attempting to use project.json to compile Json.NET and I discovered the framework targets net20 and net35 currently don't work. The assemblies referenced for those targets are always .NET 4.0.

project.json:
C:\Windows\Microsoft.NET\assembly\GAC_MSIL\System\v4.0_4.0.0.0__b77a5c561934e089\System.dll

Class library targeting .NET 2.0:
C:\Windows\Microsoft.NET\Framework\v2.0.50727\System.dll

Also there are automatic references to assemblies that don't exist in .NET 2.0 like System.Core.dll

This is blocking me using project.json to compile all of Json.NET

muratg:
@davidfowl

JamesNK:
What's the plan here?

I'd like to switch over. It would simplify my build scripts, reduce the likely-hood of errors such as https://github.com/JamesNK/Newtonsoft.Json/issues/618 and I could improve the features offered by Json.NET on vnext by using IL gen instead of expression trees for reflection.

anurse:
I think we can do this. It's just a matter of expanding our set of known frameworks and reference assembly folders. I'll leave it to @muratg to schedule, but we can try to do it in beta7. It may not make it, but I think it'll be fairly straightforward so I'm OK trying to take it on for b7.

JamesNK:
@muratg Will this make beta 7? :smile:

anurse:
Ping @Eilon (as @muratg is on vacation)

Eilon:
@anurse let's chat tomorrow on what this would take. We have a design mtg at noon. Let @glennc know we ought to discuss this.

anurse:
We should be able to do this for beta8, it should just be a matter of searching the correct GAC paths. If it turns out to be more work, I'll circle back to this thread for an update.

JamesNK:
Can you point me towards instructions for getting this before beta 8?

anurse:
It should be on the `aspnetvnext` feed now, just `dnvm upgrade -u` (make sure your NuGet.config also points to `https://www.myget.org/F/aspnetvnext/api/v2` and you have the latest beta8 builds of any DNX packages, if you reference them)

JamesNK:
Bug:

System.Data.Entity resolves in .NET 2.0 when it should fail.

Name: fx/System.Data.Entity
[...]

PR #2508, opened 2015-08-18
Add support for net20+ as build targets

anurse:
Lowered the minimum supported .NET build target to 2.0! Want to build some old stuff with DNX? You can now!

/cc @davidfowl

@jamesnk If I give you a private nupkg you can install with `dnvm install` are you up for trying this out in your scenario from #1894 ?


JamesNK:
Sure. Not sure how busy I'm going to be this week but I could look at it in the weekend.

anurse:
It'll probably be in `aspnetvnext` by then :). No problem, I have done tests myself and there's still plenty of time to react to bugs if you encounter problems with it.

src/Microsoft.Dnx.Runtime/DependencyManagement/GacDependencyResolver.cs
```diff
@@ -102,18 +102,27 @@ private bool TryResolvePartialName(string name, SemanticVersion version, out str
...
 
-        private static IEnumerable<string> GetGacSearchPaths()
+        private static IEnumerable<string> GetGacSearchPaths(FrameworkName targetFramework)
         {
             var gacFolders = new[] { "GAC_32", "GAC_64", "GAC_MSIL" };
 
             string windowsFolder = Environment.GetEnvironmentVariable("WINDIR");
 
+            string gacRoot;
+            if(targetFramework.Version.Major < 4)
```
davidfowl: formatting

src/Microsoft.Dnx.Runtime/DependencyManagement/ProjectReferenceDependencyProvider.cs
```diff
@@ -12,6 +13,8 @@ namespace Microsoft.Dnx.Runtime
     public class ProjectReferenceDependencyProvider : IDependencyProvider
     {
         private readonly IProjectResolver _projectResolver;
+        private static readonly Version Version35 = new Version(3, 5);
```
davidfowl: This class doesn't have access to the Constants?

src/Microsoft.Dnx.Runtime/FrameworkReferenceResolver.cs
```diff
@@ -8,6 +8,7 @@
 using System.Runtime.Versioning;
 using System.Xml.Linq;
 using NuGet;
+using System.Linq;
```
davidfowl: sort usings

src/Microsoft.Dnx.Runtime/FrameworkReferenceResolver.cs
```diff
@@ -214,6 +214,12 @@ private static FrameworkInformation GetFrameworkInformation(FrameworkName target
 
         private static FrameworkInformation GetFrameworkInformation(FrameworkName targetFramework, string referenceAssembliesPath)
         {
+            // Check for legacy frameworks
+            if (targetFramework.Identifier.Equals(".NETFramework", StringComparison.Ordinal) && targetFramework.Version <= Constants.Version35)
```
davidfowl: VersionUtility.IsDesktop?

src/Microsoft.Dnx.Runtime/FrameworkReferenceResolver.cs
```diff
@@ -277,7 +342,7 @@ private static FrameworkInformation GetFrameworkInformation(DirectoryInfo direct
 
                             var entry = new AssemblyEntry();
                             entry.Version = version != null ? Version.Parse(version) : null;
-                            frameworkInfo.Assemblies.Add(assemblyName, entry);
+                            frameworkInfo.Assemblies[assemblyName] = entry;
```
davidfowl: Do you have tests that verify this scenario?

test/Microsoft.Dnx.Runtime.Tests/DependencyManagement/FrameworkReferenceResolverFacts.cs
```diff
@@ -0,0 +1,69 @@
...
+        [InlineData("net451", "Microsoft.CSharp", @"REFASMSROOT\.NETFramework\v4.5.1\Microsoft.CSharp.dll", "4.0.0.0")]
+        [InlineData("net452", "Microsoft.CSharp", @"REFASMSROOT\.NETFramework\v4.5.2\Microsoft.CSharp.dll", "4.0.0.0")]
+        [InlineData("net46", "Microsoft.CSharp", @"REFASMSROOT\.NETFramework\v4.6\Microsoft.CSharp.dll", "4.0.0.0")]
+
+        public void FrameworkResolverResolvesCorrectPaths(string shortFrameworkName, string assemblyName, string expectedPath, string expectedVersion)
+        {
+            var progFiles = string.IsNullOrEmpty(Environment.GetEnvironmentVariable("PROGRAMFILES(X86)")) ?
+                Environment.GetEnvironmentVariable("PROGRAMFILES") :
+                Environment.GetEnvironmentVariable("PROGRAMFILES(X86)");
+            var refAsmsRoot = Path.Combine(progFiles, "Reference Assemblies", "Microsoft", "Framework");
```
davidfowl: Use this method https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Runtime/FrameworkReferenceResolver.cs#L141

src/Microsoft.Dnx.Runtime/FrameworkDefinitions.cs
```diff
@@ -265,6 +270,11 @@ private static FrameworkInformation PopulateNet452(string referenceAssembliesPat
             frameworkInfo.Path = Path.Combine(referenceAssembliesPath, ".NETFramework", "v4.5.2");
             frameworkInfo.RedistListPath = Path.Combine(referenceAssembliesPath, ".NETFramework", "v4.5.2", "RedistList", "FrameworkList.xml");
             frameworkInfo.Name = ".NET Framework 4.5.2";
+            frameworkInfo.SearchPaths = new[]
```
davidfowl: What's the reason these are templated? It's pretty confusing

src/Microsoft.Dnx.Runtime/FrameworkReferenceResolver.cs
davidfowl: Or VersionUtility.NetPlatformFrameworkIdentifier

src/Microsoft.Dnx.Runtime/DependencyManagement/ProjectReferenceDependencyProvider.cs
anurse: It's where the constants originated, just forgot to move them out :)

src/Microsoft.Dnx.Runtime/FrameworkDefinitions.cs
anurse: Missed it in a refactoring. I need to add some tests on attempted paths

src/Microsoft.Dnx.Runtime/FrameworkReferenceResolver.cs
anurse: prefer the latter as there is no such thing as `dnx20` :)

test/Microsoft.Dnx.Runtime.Tests/DependencyManagement/FrameworkReferenceResolverFacts.cs
anurse: done

src/Microsoft.Dnx.Runtime/FrameworkReferenceResolver.cs
anurse: added one. Microsoft.Build.Engine is present in both the 2.0 and 3.5 folders (thus the 3.5 one should override it in our logic, I'm confirming that this is the behavior for MSBuild)

anurse: Looks like MSBuild lets you select both. In theory we could do that by allowing a version in the `frameworkAssemblies` entry but that's a larger change. I suggest that if we care about supporting that, we make a new bug for it. Until then, we default to the highest version.

davidfowl: We actually do allow that already. We should flow all the information to make sure it's possible. The only change here would be to support multiple versions of assemblies in the framework information assemblies right?

src/Microsoft.Dnx.Runtime/FrameworkDefinitions.cs
anurse: fixed, but the diff of this line hasn't changed so github isn't hiding it :)

anurse:
:watch: @davidfowl

anurse:
+ @halter73 for review

src/Microsoft.Dnx.Runtime/FrameworkReferenceResolver.cs
```diff
@@ -40,6 +41,15 @@ public class FrameworkReferenceResolver : IFrameworkReferenceResolver
             }
         };
 
+        private readonly bool _useCachedFrameworkDefinitions;
```
davidfowl: nuke this

davidfowl:
:shipit:

onovotny:
Does this PR support building Xamarin projects too? That's essential to cover "the world." It needs to support MonoTouch, MonoAndroid and Xamarin.iOS if building on Windows (not sure if you can build MonoMac from Windows).

anurse:
This PR does not affect Xamarin frameworks. That would be a separate issue and a separate discussion if we want to support it. This is just focused on building for older .NET frameworks.

onovotny:
I'm not sure it should be seen as a separate issue. It's really about DNX being able to build "the world" and Xamarin is part of that now.

It appears to be minor changes to this PR to add that support -- at least for Library projects that don't create IPA/IPK files.

anurse:
It's definitely a different issue IMO, if it's a minor change to add it to this PR, it shouldn't be any less minor to add it once this PR is merged, so there's no need to block this on that discussion. Feel free to open a new issue to discuss that and we can look at adding it in.

onovotny:
I agree that there's no need to block this, I'll open a new issue.

anurse:
Sounds good! Being able to build more targets is definitely something we want to work towards

test/Microsoft.Dnx.Runtime.Tests/DependencyManagement/FrameworkReferenceResolverFacts.cs
```diff
@@ -0,0 +1,128 @@
...
+        public void FrameworkResolverResolvesCorrectPaths(string shortFrameworkName, string assemblyName, string expectedPath, string expectedVersion)
+        {
+            var resolver = new FrameworkReferenceResolver();
+
+            string actualPath;
+            Version actualVersion;
+            Assert.True(resolver.TryGetAssembly(assemblyName, VersionUtility.ParseFrameworkName(shortFrameworkName), out actualPath, out actualVersion));
+            Assert.Equal(Environment.ExpandEnvironmentVariables(expectedPath).Replace("REFASMSROOT", FrameworkReferenceResolver.GetReferenceAssembliesPath()), actualPath);
+
+            // Having this be Version->Version equality caused some problems...
```
halter73: Really. Do you know why? Is there a bug in `Version`?

src/Microsoft.Dnx.Runtime/FrameworkReferenceResolver.cs
```diff
@@ -172,7 +172,7 @@ public static string GetReferenceAssembliesPath()
                     "Reference Assemblies", "Microsoft", "Framework");
         }
 
-        private static FrameworkInformation GetFrameworkInformation(FrameworkName targetFramework)
```
halter73: Is there a reason this is no longer static?

src/Microsoft.Dnx.Runtime/FrameworkReferenceResolver.cs
```diff
@@ -232,12 +238,105 @@ private static FrameworkInformation GetFrameworkInformation(FrameworkName target
...
+            var frameworkInfo = new FrameworkInformation();
+            if (!RuntimeEnvironmentHelper.IsMono)
+            {
+                // Always grab .NET 2.0 data
+                var searchPaths = new List<string>();
+                var net20Dir = Path.Combine(Environment.GetEnvironmentVariable("WINDIR"), "Microsoft.NET", "Framework", "v2.0.50727");
+
+                if (!Directory.Exists(net20Dir))
+                {
+                    return null;
```
halter73: What's the difference between returning null and returning a `FrameworkInformation` object with `Exists` false?

src/Microsoft.Dnx.Runtime/FrameworkReferenceResolver.cs
```diff
@@ -122,8 +123,7 @@ public string GetFrameworkPath(FrameworkName targetFramework)
             {
                 return null;
             }
-
-            return information.Path;
+            return information.SearchPaths.Select(s => Path.Combine(s, "{name}.dll"));
```
halter73: Are facades no more?

src/Microsoft.Dnx.Runtime/FrameworkReferenceResolver.cs
anurse: Not anymore! Good catch :)

test/Microsoft.Dnx.Runtime.Tests/DependencyManagement/FrameworkReferenceResolverFacts.cs
anurse: It actually seemed to be an error with rendering the Expected/Actual values when the assert failed. I'll take a look at nailing down a repro separately

src/Microsoft.Dnx.Runtime/FrameworkReferenceResolver.cs
anurse: Pretty much nothing :). A distinction is kinda needed because there's a bunch of pre-computed framework data but we still need to know if the framework is actually installed (via a Directory.Exists call).

src/Microsoft.Dnx.Runtime.Sources/Impl/Constants.cs
```diff
@@ -25,5 +25,8 @@ internal static class Constants
 
         public static readonly TimeSpan RegexMatchTimeout = TimeSpan.FromSeconds(10);
         public static readonly string AppConfigurationFileName = "app.config";
+
+        public static readonly Version Version35 = new Version(3, 5);
```
victorhurdugaci: This class should be called `Readonlys` not `Constants` :trollface:

src/Microsoft.Dnx.Runtime/FrameworkReferenceResolver.cs
```diff
@@ -232,12 +238,105 @@ private static FrameworkInformation GetFrameworkInformation(FrameworkName target
...
+        private static string SynthesizeFrameworkFriendlyName(FrameworkName targetFramework)
+        {
+            // Names are not present in the RedistList.xml file for older frameworks or on Mono
+            // We do some custom version string rendering to match how net40 is rendered (.NET Framework 4)
+            if (targetFramework.Identifier.Equals(VersionUtility.NetFrameworkIdentifier))
+            {
+                string versionString = targetFramework.Version.Minor == 0 ?
+                    targetFramework.Version.Major.ToString() :
+                    targetFramework.Version.ToString();
+                return ".NET Framework " + versionString;
```
victorhurdugaci: Const?

test/Microsoft.Dnx.Runtime.Tests/DependencyManagement/FrameworkReferenceResolverFacts.cs
```diff
@@ -0,0 +1,131 @@
...
+            Assert.Equal(Version.Parse(expectedVersion).ToString(), actualVersion.ToString());
+        }
+
+        [ConditionalTheory]
+        [FrameworkSkipCondition(RuntimeFrameworks.CLR | RuntimeFrameworks.CoreCLR)]
+        [OSSkipCondition(OperatingSystems.Win2008R2 | OperatingSystems.Win7 | OperatingSystems.Win7And2008R2)]
```
victorhurdugaci: `OperatingSystems.Win7And2008R2` already includes the other two. The OSSkip condition is superfluous because of the previous one

test/Microsoft.Dnx.Runtime.Tests/DependencyManagement/FrameworkReferenceResolverFacts.cs
```diff
@@ -0,0 +1,131 @@
...
+        [InlineData("net35", "Microsoft.Build.Engine", @"REFASMSROOT\v3.5\Microsoft.Build.Engine.dll", "3.5.0.0")]
+
+        public void FrameworkResolverResolvesCorrectPaths(string shortFrameworkName, string assemblyName, string expectedPath, string expectedVersion)
+        {
+            var resolver = new FrameworkReferenceResolver();
+
+            string actualPath;
+            Version actualVersion;
+            Assert.True(resolver.TryGetAssembly(assemblyName, VersionUtility.ParseFrameworkName(shortFrameworkName), out actualPath, out actualVersion));
+            Assert.Equal(Environment.ExpandEnvironmentVariables(expectedPath).Replace("REFASMSROOT", FrameworkReferenceResolver.GetReferenceAssembliesPath()), actualPath);
```
victorhurdugaci: Move this to a helper function? You do this replacement in a lot of places

jbogard:
Heyoo I'm going to give this one a shot, people keep asking for .NET 3.5 support for AutoMapper

project.json all the things!

anurse:
@onovotny When you add an issue for more general support, please @mention me. One thing I want to consider (though perhaps not for 1.0 RTM) is adding support for an arbitrary framework target (by providing paths to framework assemblies if absolutely necessary) so that even the ones we don't have direct support for can be used.

PR #2509, opened 2015-08-19
Use the correct target framework when resolving project references

davidfowl:
- Don't store the target framework when creating the library exporter
- Store the target framework for the LibraryManager so that it's easy to
see the context for the library manager in the debugger.


davidfowl:
Merged to unblock the ci.

Issue #2352, opened 2015-07-27
"dnu publish" should restore for wrapper projects before building them


ChengTian:
@davidfowl 

Today when doing `dnu publish`, we implicitly "no-source" wrapper projects. That is, for a `project.json` wrapping a csproj, we always build nupkg from it, no matter `--no-source` option was given or not.

However, there is no `dnu restore` being done before we call `dnu publish`. So when you `dnu publish` a project referencing a wrapped csproj, building the wrapper project will be failed.

![capture](https://cloud.githubusercontent.com/assets/1383883/8898723/ccf6f2a0-33de-11e5-962a-f537617d3533.PNG)

Or we want to call `dnu restore` during `dnu wrap`?

ChengTian:
I suggest that we restore during `dnu publish` instead of `dnu wrap` because only the publishing scenario needs a lock file.

sayedihashimi:
Looking forward to the fix in beta8. Please let me know if this moves out as its impacting VS scenarios.

PR #2512, opened 2015-08-20
dnu publish restores for wrapper projects

ChengTian:
parent https://github.com/aspnet/dnx/issues/2352

@davidfowl @BrennanConroy @troydai


misc/DnuPublishTestApps/ConsoleAppReferencingWrappedProject/src/Net45Library/Net45Library.csproj
```diff
@@ -0,0 +1,61 @@
...
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="System"/>
+    
+    <Reference Include="System.Core"/>
+    <Reference Include="System.Xml.Linq"/>
+    <Reference Include="System.Data.DataSetExtensions"/>
+    
+
```
BrennanConroy: empty lines

ChengTian: It was generated by VS and I didn't touch it. I guess it's Ok to keep it like this.

troydai: I'm with @ChengTian. It doesn't worth to fight the tool for project file's format.

test/Microsoft.Dnx.Tooling.FunctionalTests/project.json
```diff
@@ -1,5 +1,6 @@
 {
     "dependencies": {
+        "Bootstrapper.FunctionalTests": "1.0.0-*",
```
davidfowl: Nooo, we need to move those helpers into a shared project.

src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
```diff
@@ -107,6 +107,17 @@ private bool EmitNupkg(PublishRoot root)
                 root.Operations.Delete(TargetPath);
             }
 
+            // If this is a wrapper project, we need to generate a lock file before building it
+            if (IsWrappingAssembly())
+            {
+                var success = Restore(root, publishProject: this, restoreDirectory: project.ProjectDirectory)
```
davidfowl: Do we want to suppress the output here? Can this project potentially reference things that aren't in the local set of packages copied?

ChengTian: The output is already suppressed in the `Restore()` helper method:
https://github.com/aspnet/dnx/pull/2512/files#diff-e073db6edc86fb957608d9a10b4e521dR436

All NuGet package references of this project should be installed when you restore for the main project (i.e. the project referencing this project). So the answer is "no".

BTW, the `Restore()` helper method has
```c#
restoreCommand.IgnoreMissingDependencies = true; 
```
so missing package dependencies shouldn't be problems.

test/Microsoft.Dnx.Tooling.FunctionalTests/project.json
ChengTian: Fixing

ChengTian:
Thanks for the feedback. Please check latest commits.

src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
davidfowl: I'm worried about functional failures. What happens if I never restore my project references to csproj? It actually seems like this can happen for other `project.json` projects as well.

ChengTian: If you never restore your wrapper project, then obviously you cannot build the wrapper project and `dnu publish` will fail.

I kind of understand your concern because previously, when you publish an app without any package installed, the publish can still be successful with warnings say that some package dependencies cannot be found. However, if we check this in, you cannot publish without restoring packages first (when you have a wrapper project reference).

ChengTian: However, do we really care the "publish without packages installed" scenario? If you see a publish failure, you can restore and retry.

davidfowl:
:shipit:

Issue #2501, opened 2015-08-17
Cross-process synchronization in DNX (named mutex, semaphore)


kouvel:
There are a couple of named sync objects created in DNX. It looks like they are used to prevent concurrent writes to files:
- https://github.com/aspnet/dnx/blob/219871c6063d00f8297eeafe93266f1048f59a45/src/Microsoft.Dnx.Runtime/Internal/ConcurrencyUtilities.cs#L38
- https://github.com/aspnet/dnx/blob/219871c6063d00f8297eeafe93266f1048f59a45/src/Microsoft.Dnx.Tooling/NuGet/Core/Configuration/Settings.cs#L777

I understand that DNX is currently using Mono. Some tests I had done showed that while Mono allows creating and using named sync objects, the sync objects are actually process-local, so two process cannot synchronize on them. So this may not be working as expected currently.

Also, it has recently been decided that CoreCLR/CoreFX won't be supporting named sync objects at this time. Instead, in an interest to reduce confusion and the impact of breaking changes in the future if we do add that feature, CoreCLR/CoreFX will currently throw PlatformNotSupportedException when attempting to create a named sync object, see:
- https://github.com/dotnet/coreclr/pull/1387
- https://github.com/dotnet/corefx/pull/2796

So if DNX were to switch to CoreCLR/CoreFX, the code above would break.

A couple of potential solutions involving creating a file in some predetermined location with the name currently being used for the sync object:
- If it is not expected for these scenarios to have high contention, a simple, portable solution may be to use a FileStream to create/open a file using FileShare to get exclusive access to the lock file. Another thread/process trying to open the file with the same FileShare value would get an exception, and it can poll for access in a loop with short pauses.
- Another solution is on Unix to use flock() directly on a lock file. flock() can exclusively lock a file and another attempt to exclusively lock the same file would block. The lock is automatically released on process termination (in case of crash). But the lock file cannot be deleted safely since there won't be any ref counting.

CC @ChengTian, @loudej, @davidfowl

ChengTian:
Here are some statistics that may help us make decision on which alternative we should use.

To roughly quantify the degree of contention, I introduce a global counter `ContentionCount` in `dnu restore`.

**Definition: each time a thread/process tries acquiring a named semaphore but has to wait, we do `ContentionCount++`.**

Here are values of `ContentionCount` in different scenarios:

**1 . `dnu restore --parallel` in root dir of Mvc repo:**

Average `ContentionCount`: 4390

[...]

PR #2515, opened 2015-08-20
Cross-process sync utils fall back to process-local on cross-plat CoreCLR

ChengTian:
parent: https://github.com/aspnet/dnx/issues/2501

@davidfowl @anurse @danroth27 

Worked with @BrennanConroy on this and verified the change on cross-plat CoreCLR. This should unblock restore scenarios as long as you don't run multiple instances of `dnu restore` simultaneously.

However, we should explore the flock() option because that's a more robust solution, which provides real interprocess synchronization. We will do some experiments with flock() tomorrow and see whether we can get it in before beta7 sign-off.

### DIFF
--- a/src/Microsoft.Dnx.Runtime/Internal/ConcurrencyUtilities.cs
+++ b/src/Microsoft.Dnx.Runtime/Internal/ConcurrencyUtilities.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Dnx.Runtime/Internal/ConcurrencyUtilities.cs
```diff
@@ -10,6 +11,11 @@ namespace Microsoft.Dnx.Runtime.Internal
 {
     public static class ConcurrencyUtilities
     {
+#if DNXCORE50
+        private static ConcurrentDictionary<string, System.Threading.Semaphore> _nameSemaphore =
```
davidfowl: Move this inside of the semaphore class.

davidfowl: Also, how big can this get? This for every file path touched during restore right? Have we tested MVC after deleting `.dnx\packages`? We should remove the entry in dispose right?

src/Microsoft.Dnx.Runtime/Internal/ConcurrencyUtilities.cs
```diff
@@ -65,5 +71,43 @@ public static void ExecuteWithFileLocked(string filePath, Action<bool> action)
             // should never get here
             throw new TaskCanceledException($"Failed to acquire semaphore for file: {filePath}");
         }
+
+#if DNXCORE50
+        private class Semaphore
+        {
+            private readonly System.Threading.Semaphore _semaphore;
```
moozzyk: nit: do you need fully qualified type names?

src/Microsoft.Dnx.Runtime/Internal/ConcurrencyUtilities.cs
```diff
@@ -65,5 +71,43 @@ public static void ExecuteWithFileLocked(string filePath, Action<bool> action)
             // should never get here
             throw new TaskCanceledException($"Failed to acquire semaphore for file: {filePath}");
         }
+
+#if DNXCORE50
+        private class Semaphore
```
moozzyk: [fxCop] Shouldn't this class implement `IDisposable`?

src/Microsoft.Dnx.Runtime/Internal/ConcurrencyUtilities.cs
ChengTian: We tested restoring for Mvc repo and the peak memory usage was about 500MB. Yes, we should remove entries. Working on it.

src/Microsoft.Dnx.Runtime/Internal/ConcurrencyUtilities.cs
ChengTian: Yes, will do

src/Microsoft.Dnx.Runtime/Internal/ConcurrencyUtilities.cs
ChengTian: Yes, this class itself has the name `Semaphore` but we want to use the one from standard library.

moozzyk: Got it.

src/Microsoft.Dnx.Tooling/NuGet/Core/Configuration/Settings.cs
```diff
@@ -774,7 +775,23 @@ private void ExecuteSynchronized(Action ioOperation)
...
+#if DNXCORE50
+            if (RuntimeEnvironmentHelper.IsWindows)
+            {
+                mutex = new Mutex(false, mutexName);
+            }
+            else
+            {
+                // However, named mutex is not supported on cross-plat CoreCLR
+                // so we fall back to process-local synchronization
+                mutex = new Mutex(false);
```
kouvel: Don't you need a lookup mechanism by name here too, like for the Semaphore case? Otherwise it looks like it would create a new Mutex every time and no two calls would share a mutex to synchronize file access.

ChengTian: yes, good point

anurse:
Looks good to me once @kouvel 's comment is resolved.

ChengTian:
Fix the memory consumption issue. Here are latest statistics:

`dnu restore --no-cache` for MusicStore with an empty packages folder:

On Mono:
Time: 300 sec
Peak Memory Usage: 116MB

On Cross-plat CoreCLR:
Time: 150 sec
Peak Memory Usage: 137MB

Note that restoring on cross-plat CoreCLR is **twice as fast as** restoring on Mono.

I added additional log to check the global shared dictionary and confirmed that all entries get deleted after restoring for a project. So the small difference of memory usage is not introduced by this change.

src/Microsoft.Dnx.Tooling/NuGet/Core/Configuration/Settings.cs
```diff
@@ -766,15 +767,53 @@ private static Settings ReadSettings(IFileSystem fileSystem, string settingsPath
             }
         }
 
+#if DNXCORE50
+        private static Mutex _globalMutex = new Mutex(initiallyOwned: false);
```
ChengTian: This code is from NuGet codebase and DNU actually never runs this code path in multiple threads. To simplify the fix, I just use a global mutex on cross-plat CoreCLR to do the synchronization

davidfowl:
:shipit:

davidfowl:
@ChengTian open a bug on nuget to port this code.

ChengTian:
@davidfowl , NuGet issue created here:
https://github.com/NuGet/Home/issues/1230

ChengTian:
Checked this change into release branch and merged back to dev: https://github.com/aspnet/dnx/commit/e336e171735cffb1def89f6458922a86729510af

Issue #2513, opened 2015-08-20
DTH spent about 300M memory on Microsoft.Dnx.Runtime.Servicing.Breadcrumbs when open MVC solution


troydai:
Open MVC solution in VS with `dnx-clr-win-x86.1.0.0-beta8-15109`.

Noticing that about 48% of the managed memory is of type `String[]` and referred by `Microsoft.Dnx.Runtime.Servicing.Breadcrumbs`

Snapshot:
![image](https://cloud.githubusercontent.com/assets/1329240/9374253/d70386a0-46a8-11e5-8d2c-9717d7afdc94.png)

```
Name                                                                                   	Inc %	           Inc	    Inc Ct	Exc %	         Exc	    Exc Ct	Fold	Fold Ct
 mscorlib!String                                                                       	 52.2	   315,924,416	3,508,660.250	 52.2	 315,924,416	3,508,660.250	   0	      0
+ mscorlib!String[] (Bytes > 1M,Ptrs,ElemSize=4)                                       	 48.8	   295,422,304	3,272,285.500	  0.0	           0	         0	   0	      0
|+ mscorlib!List<String>                                                               	 48.8	   295,422,304	3,272,285.500	  0.0	           0	         0	   0	      0
| + Microsoft.Dnx.Runtime!Microsoft.Dnx.Runtime.Servicing.Breadcrumbs                  	 48.8	   295,422,304	3,272,285.500	  0.0	           0	         0	   0	      0
|  + [static var Microsoft.Dnx.Runtime.Servicing.Breadcrumbs.<Instance>k__BackingField]	 48.8	   295,422,304	3,272,285.500	  0.0	           0	         0	   0	      0
|   + [static vars]                                                                    	 48.8	   295,422,304	3,272,285.500	  0.0	           0	         0	   0	      0
|    + [.NET Roots]                                                                    	 48.8	   295,422,304	3,272,285.500	  0.0	           0	         0	   0	      0
|     + ROOT                                                                           	 48.8	   295,422,304	3,272,285.500	  0.0	           0	         0	   0	      0
```

davidfowl:
Hah! Nice catch! This is an easy one to fix

troydai:
__Observation Note__

1. `PackageDependencyProvider` keeps adding breadcrumbs throughout the entire life cycle. [see here](https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Runtime/DependencyManagement/PackageDependencyProvider.cs#L118)
[...]

PR #2519, opened 2015-08-20
Move breadcrumbing to DefaultHost

davidfowl:
- This ensures bread crumbing only happens at runtime. It removes the side effects when
we use this API to create a library manager.

#2513

/cc @troydai @victorhurdugaci


troydai:
:shipit: let's save MVC team 400 MB memory in their day-to-day life.

PR #2521, opened 2015-08-20
Update `.travis.yml` to run on OSX.

NTaylorMullen:
- Added `MONO_THREADS_PER_CPU=2000` environment variable to ensure `dnu restore` completes successfully.

**Note:** It looks like OSX is currently failing so that will need to be investigated.

### DIFF
--- a/test/Microsoft.Dnx.CommonTestUtils/TestUtils.cs
+++ b/test/Microsoft.Dnx.CommonTestUtils/TestUtils.cs
@@ -30,6 +30,16 @@ namespace Microsoft.Dnx.CommonTestUtils
         {
             var tempDirPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
             Directory.CreateDirectory(tempDirPath);
+
+            if (string.Equals(RuntimeEnvironmentHelper.RuntimeEnvironment.OperatingSystem, "Darwin"))
+            {
+                // Resolves issues on Mac where GetTempPath gives /var and GetCurrentDirectory gives /private/var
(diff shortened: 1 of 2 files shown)


BrennanConroy:
So all the failures are because the file path we are looking for doesn't have /private/ in the front. Not sure why yet.

davidfowl:
We should fix the failures in @NTaylorMullen 's branch before merging this in.

BrennanConroy:
Idea:
In Microsoft.Dnx.CommonTestUtils the CreateTempDir function we save the current directory, then set the current directory to the created temp directory, then set the temp directory string to the value from get current directory, and reset the current directory. Tested it locally on a Mac for the Microsoft.Dnx.Tooling.FunctionTests and it worked

NTaylorMullen:
:+1:

test/Microsoft.Dnx.CommonTestUtils/TestUtils.cs
```diff
@@ -30,6 +30,13 @@ public static DisposableDir CreateTempDir()
         {
             var tempDirPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
             Directory.CreateDirectory(tempDirPath);
+
+            // Resolves issues on Mac where GetTempPath gives /var and GetCurrentDirectory gives /private/var
+            var currentDirectory = Directory.GetCurrentDirectory();
```
davidfowl: Make sure this only runs on mac

davidfowl:
:shipit:

BrennanConroy:
fb88afaeed702b1af730f49b49f696b3f8e8b68f

PR #2530, opened 2015-08-20
fix #2528 by reenabling net20 tests

### LINKED ISSUE #2528: Re-enable net20 targetting tests with guards
I disabled some of the net20 targetting tests because the CI didn't have the framework reference assemblies. We should a) consider installing them, but b) not require that the CI have them by having the test detect their presence and either mark as skipped or pass the test.

anurse:
Fixes #2528 

/cc @davidfowl


davidfowl:
:shipit:

PR #2533, opened 2015-08-21
[Prototype] SemanticVersion performance optimzation

troydai:
Reuse the same copy of the SemanticVersion when the given verison is same.

Example: open MVC solution and build
Before:
image

After:
image


troydai:
Downside: these `SemanticVersion` instances will be forever referred in the static dictionary.

davidfowl:
Changes to make
- Just cache the the original `string` to `SemanticVersion` don't use the tuple. After normalization just store a duplicate entry, that's not a big deal.
- Don't change any of the original parsing logic
- Add a factory method that can be delegated to,`Func<string, SemanticVersion> SemanticVersion.Factory`, set it in DTH only.

davidfowl:
Closing this out

PR #2549, opened 2015-08-24
Performance improvements to make startup time better

davidfowl:
A large amount of changes to help improve the startup time of DNX for precompiled applications. Generally this change gets rid of the dependency walker and the graph walk for runtime. The lock file is now being used for project and package resolution at runtime and thus the `project.lock.json` version was bumped to 2. Split the ApplicationHostContext logic into 2 code paths, InitializeForRutime and Initialize (normal). One is optimized for runtime startup and does less work. As a side effect, some of the error handling about missing dependencies has been removed from runtime. It also means that restore needs to be run when P2P references change (but this was always the case anyways).

Other changes include:
- Moved IProjectResolver and friends to Microsoft.Dnx.Tooling

Performance changes:
- Made ProjectFilesCollection lazy initialize. We never need source files until we perform a compilation
- Got rid of the dependency type parsing and just used a flags enum
- Avoid capturing the execution context when writing breadcrumbs
- Avoid checking to see if package directories exist on runtime startup path (assembly load errors are fine :smile: )
- Made ILibraryExporter lazy in DefaultHost 

There's still a bunch of work to do for compile time performance (we still overly walk the dependency graph but that'll come in another PR).


src/Microsoft.Dnx.Compilation/LibraryExporter.cs
```diff
@@ -146,7 +146,7 @@ public LibraryExport GetNonProjectExports(string name)
                 {
                     var childNode = new Node
                     {
-                        Library = LibraryManager.GetLibraryDescription(dependency.Name),
+                        Library =  dependency.Library,
```
anurse: spacing

src/Microsoft.Dnx.Compilation/CompilationEngine.cs
```diff
@@ -51,7 +56,16 @@ public LibraryExporter CreateProjectExporter(Project project, FrameworkName targ
 
             // This library manager represents the graph that will be used to *load* the compiler and other
             // build time related dependencies
-            var runtimeLibraryManager = _context.ProjectGraphProvider.GetProjectGraph(project, _context.ApplicationEnvironment.RuntimeFramework);
+
```
moozzyk: Is the comment above still applicable?

moozzyk: I think it should be moved to line 68

anurse:
:shipit:. I think the layering still needs some consideration and work, but that can come later

PR #2551, opened 2015-08-24
fix #2390 by using normalized version numbers during dnu publish

### LINKED ISSUE #2390: Setting the version to 1.0.0.0 makes dnu throw a FileNotFoundException
I am getting this when setting the version to 1.0.0.0 in project.json:
...
Microsoft .NET Development Utility CLR-x86-1.0.0-beta6-12256
Copying to output path C:\Users\Dima\AppData\Local\Temp\PublishTemp
System.IO.FileNotFoundException: Could not find file 'D:\Projects\My\Test\src\TestProject\bin\Release\TestProject.1.0.0.nupkg'.

It seems dnu not picking up the last zero from the version number and trying to find *.1.0.0.nupkg instead of *.1.0.0.0.nupkg. Or... the nuget package should be created as *.1.0.0.nupkg instead

JunTaoLuo:
Updated the GetPackagePath helper function in BuildManager to use normalized version string instead for dnu publish. Added a functional test to test for this scenario.


JunTaoLuo:
cc @davidfowl @anurse

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPublishTests.cs
```diff
@@ -453,6 +453,67 @@ public void DnuPublishConsoleApp(string flavor, string os, string architecture)
...
+                    subcommand: "restore",
+                    arguments: "",
+                    workingDir: testEnv.ProjectPath);
+                Assert.Equal(0, exitCode);
+
+                exitCode = DnuTestUtils.ExecDnu(
+                    runtimeHomeDir,
+                    subcommand: "publish",
+                    arguments: string.Format("--no-source --out {0}",
+                        testEnv.PublishOutputDirPath),
```
ChengTian: Use string interpolation

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPublishTests.cs
```diff
@@ -453,6 +453,67 @@ public void DnuPublishConsoleApp(string flavor, string os, string architecture)
...
+  ""frameworks"": {
+    ""dnx451"": {}
+  }
+}")
+                    .WriteTo(testEnv.ProjectPath);
+
+                var exitCode = DnuTestUtils.ExecDnu(
+                    runtimeHomeDir,
+                    subcommand: "restore",
+                    arguments: "",
```
ChengTian: nit:`string.Empty`

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPublishTests.cs
```diff
@@ -453,6 +453,67 @@ public void DnuPublishConsoleApp(string flavor, string os, string architecture)
...
+            var expectedOutputStructure = @"{
+  'approot': {
+    'global.json': '',
+    'packages': {
+      'PROJECT_NAME': {
+        '1.0.0': {
+          '.': ['PROJECT_NAME.1.0.0.nupkg', 'PROJECT_NAME.1.0.0.nupkg.sha512', 'PROJECT_NAME.nuspec'],
+          'root': {
+            '.': ['Config.json', 'project.json', 'project.lock.json']
+          },
```
ChengTian: nit: can be simplified as
```json
'root': ['Config.json', 'project.json', 'project.lock.json']
```
if you just want a folder containing a list of empty files

JunTaoLuo:
The current approach to resolve version normalization is to force the toString method of SemanticVersion to normalize the version number. This will force normalization whenever serialization is required. 

The functionality of determining whether normalization has occurred is now removed since normalization occurs in every case. A DNU pack test for this scenario is removed.

DNU pack and publish tests have been added to test version normalization.

JunTaoLuo:
cc @ChengTian

src/Microsoft.Dnx.Runtime/NuGet/SemanticVersion.cs
```diff
@@ -319,7 +272,11 @@ public int CompareTo(SemanticVersion other)
 
         public override string ToString()
         {
-            return OriginalString;
+            var revision = Version.Revision > 0 ? ("." + Version.Revision.ToString(CultureInfo.InvariantCulture)) : string.Empty;
+            var specialVer = !string.IsNullOrEmpty(SpecialVersion) ? ("-" + SpecialVersion) : string.Empty;
+
+            // SemanticVersion normalizes the missing components to 0.
+            return $"{Version.Major}.{Version.Minor}.{Version.Build}{revision}{specialVer}";
```
ChengTian: 1. Calling `ToString()` on one version object multiple times is common. We shouldn't repeat the work each time when `ToString()` is called.
2. Also note that sometimes we create version objects just to do some comparison and we never call ToString() on them. We should avoid allocating strings in this case.

You may want to use lazy initialization here.

src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
```diff
@@ -413,7 +413,7 @@ private static string Normalize(string projectDir)
 
         private static string GetPackagePath(Runtime.Project project, string outputPath, bool symbols = false)
         {
-            string fileName = project.Name + "." + project.Version + (symbols ? ".symbols" : "") + ".nupkg";
+            string fileName = project.Name + "." + project.Version.ToString() + (symbols ? ".symbols" : "") + ".nupkg";
```
ChengTian: I guess it's Ok to make the code nicer here BTW in this PR:
1. Use string interpolation
2. Replace hard-coded `".nupkg"` with `NuGet.Constants.PackageExtension`

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
```diff
@@ -202,6 +202,40 @@ public void DnuPack_OutPathSpecified(string flavor, string os, string architectu
...
+  }
+}");
+
+                exitCode = DnuTestUtils.ExecDnu(runtimeHomeDir, "restore", "", out stdOut, out stdErr, workingDir: testEnv.RootDir);
+                Assert.Equal(0, exitCode);
+
+                exitCode = DnuTestUtils.ExecDnu(runtimeHomeDir, "pack", "", out stdOut, out stdErr, workingDir: testEnv.RootDir);
+                Assert.Empty(stdErr);
+                Assert.Contains(string.Format(expectedNupkg, Path.GetFileName(testEnv.RootDir), testEnv.RootDir), stdOut);
+                Assert.Contains(string.Format(expectedSymbol, Path.GetFileName(testEnv.RootDir), testEnv.RootDir), stdOut);
```
ChengTian: Instead of checking the contents of log, we should check the actual output files on disk. That's what we care the most in this scenario.

src/Microsoft.Dnx.Runtime/NuGet/SemanticVersion.cs
pranavkm: See - https://github.com/NuGet/NuGet2/blob/2.8.6/src/Core/SemanticVersion.cs#L295-L330

test/Microsoft.Dnx.Runtime.Tests/NuGet/SemanticVersionTest.cs
```diff
@@ -31,7 +31,7 @@ public static IEnumerable<object[]> NormalizedVersionStringTestData
         [MemberData(nameof(NormalizedVersionStringTestData))]
         public void TestGetNormalizeVersionString(SemanticVersion version, string expectation)
         {
-            Assert.Equal(expectation, version.GetNormalizedVersionString());
+            Assert.Equal(expectation, version.ToString());
```
pranavkm: Could you also add a test case for `1.0-beta` -> `1.0.0-beta` - doesn't seem to be covered?

src/Microsoft.Dnx.Runtime/NuGet/SemanticVersion.cs
```diff
@@ -16,9 +16,6 @@ public sealed class SemanticVersion : IComparable, IComparable<SemanticVersion>,
         public SemanticVersion(string version)
             : this(Parse(version))
         {
-            // The constructor normalizes the version string so that it we do not need to normalize it every time we need to operate on it. 
-            // The original string represents the original form in which the version is represented to be used when printing.
-            OriginalString = version;
```
davidfowl: Why remove this?

src/Microsoft.Dnx.Runtime/NuGet/SemanticVersion.cs
```diff
@@ -207,7 +160,7 @@ private static bool TryParseInternal(string version, bool strict, out SemanticVe
                 }
             }
 
-            semVer = new SemanticVersion(NormalizeVersionValue(versionValue), specialVersion, version.Replace(" ", ""));
```
davidfowl: ```version.Replace(" ", "")``` that line isn't relevant?

src/Microsoft.Dnx.Runtime/NuGet/SemanticVersion.cs
JunTaoLuo: I believe the OriginalString is only used for storing the unnormalized version number and was only used for determining whether normalization occurred. From our previous conversation, this functionality will no longer be needed so isn't this field now redundant?

src/Microsoft.Dnx.Runtime/NuGet/SemanticVersion.cs
JunTaoLuo: version.Replace(" ", "") was originally passed in to be stored as OriginalString, which I think is redundant? I removed the constructor that required the OriginalString and removed this argument here. version is not used in the function after this point so it should be safe.

src/Microsoft.Dnx.Runtime/NuGet/SemanticVersion.cs
pranavkm: It was sort of a UI feature. It was apparently confusing when `Install-Package Foo 1.0` would install `Foo (1.0.0)`. We probably shouldn't care about it keeping it around.

JunTaoLuo:
The Bootstrapper test should not be caused by the changes in this PR.

src/Microsoft.Dnx.Runtime/NuGet/SemanticVersion.cs
```diff
@@ -319,7 +275,33 @@ public int CompareTo(SemanticVersion other)
...
+                if (!string.IsNullOrEmpty(SpecialVersion))
+                {
+                    builder
+                        .Append('-')
+                        .Append(SpecialVersion);
+                }
+                
+                _normalizedVersionString = builder.ToString();
+            }
+            return _normalizedVersionString;
```
BrennanConroy: nit: newline between } and return

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
```diff
@@ -202,6 +202,87 @@ public void DnuPack_OutPathSpecified(string flavor, string os, string architectu
...
+                    runtimeHomeDir, 
+                    subcommand: "restore",
+                    arguments: string.Empty,
+                    workingDir: testEnv.ProjectPath);
+                Assert.Equal(0, exitCode);
+
+                exitCode = DnuTestUtils.ExecDnu(
+                    runtimeHomeDir,
+                    subcommand: "pack",
+                    arguments: "",
```
BrennanConroy: string.Empty

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
```diff
@@ -202,6 +202,87 @@ public void DnuPack_OutPathSpecified(string flavor, string os, string architectu
...
+                    runtimeHomeDir,
+                    subcommand: "restore",
+                    arguments: string.Empty,
+                    workingDir: testEnv.ProjectPath);
+                Assert.Equal(0, exitCode);
+
+                exitCode = DnuTestUtils.ExecDnu(
+                    runtimeHomeDir,
+                    subcommand: "pack",
+                    arguments: "",
```
BrennanConroy: string.Empty

BrennanConroy:
LGTM

src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
```diff
@@ -413,7 +413,7 @@ private static string Normalize(string projectDir)
 
         private static string GetPackagePath(Runtime.Project project, string outputPath, bool symbols = false)
         {
-            string fileName = project.Name + "." + project.Version + (symbols ? ".symbols" : "") + ".nupkg";
+            string fileName = project.Name + "." + project.Version.ToString() + (symbols ? ".symbols" : "") + NuGet.Constants.PackageExtension;
```
davidfowl: Use string interpolation here

davidfowl:
As soon as the ci passes :shipit:

Issue #2538, opened 2015-08-22
Adopt local-win policy and make packages folder location in global.json has highest priority


ChengTian:
The current order to resolve packages folder location is:
1. EnvironmentNames.Packages environment variable
2. global.json { "packages": "..." }

It means that, if you `dnu publish` an app, which has its local packages folder specified in global.json, it will still use global packages folder specified in `DNX_PACKAGES/NUGET_PACKAGES` environment variable.

This can lead to some really confusing case, where you correctly publish an app but cannot run it, just because it is using wrong packages folder.

The current global-win policy is unnatural. We should adopt local-win policy.

/cc @davidfowl

PR #2555, opened 2015-08-25
Use local-win policy when resolving packages folder

ChengTian:
parent https://github.com/aspnet/dnx/issues/2538

Also updated wiki page at: https://github.com/aspnet/Home/wiki/Package-Locations (/cc @glennc )

@davidfowl


victorhurdugaci:
The change looks pretty straight forward. I don't expect it to break anything even on the CI since none of our repos set the packages folder in `global.json` (or at least I hope no one does it),

:shipit:

davidfowl:
:shipit:

PR #2557, opened 2015-08-25
fix potential nullref in ProjectAssemblyLoader when using DTH plugins

anurse:
Without this fix, Razor Tag Helpers within the current project are not properly recognized.

@ToddGrun tried out a private build of the fix and it did indeed fix the issue. I believe this meets the beta7 bar.

/review @davidfowl @NTaylorMullen
/fyi @danroth27 @DamianEdwards @Eilon @muratg @ToddGrun


NTaylorMullen:
:shipit:

PR #2559, opened 2015-08-25
copy things to x64 when dnu building

anurse:
I got tired of having half my tests pass when doing incremental builds using `dnu build`

@moozzyk I thought about doing similar for the native code but was scared away by angle-brackets and pointers :stuck_out_tongue_winking_eye:

/cc @victorhurdugaci @halter73 @davidfowl @BrennanConroy @ChengTian


davidfowl:
Seems like you have some invalid JSON, at least that's what travis thinks

moozzyk:
@anurse - Pointers? In MSBuild? :trollface:

anurse:
Yep, invalid JSON fixed. Measuring impact now, might have a potential adjustment to make it faster

anurse:
It's only slow because each invocation of batchcopy starts a new powershell instance. I rewrote it in CMD and now it's 3s **faster** than before for Microsoft.Dnx.Runtime **including** copying x64 bits.

anurse:
Timing for dnu build of Microsoft.Dnx.Runtime

```
Before: 10s
After (-PS1, +x64): 7s
```

davidfowl:
Great just make sure you dont break mono in the process

anurse:
That's what travis is for :). But regardless, I don't see how I could since it uses the `sh` variant

anurse:
Post-build steps have already started running successfully in Travis. I'll wait for it to fully succeed before merge but any further comments on the PR @davidfowl ?

anurse:
I have a branch I want to use this change in ASAP :P so a squirrel soon would rock :)

src/Microsoft.Dnx.Host/project.json
```diff
@@ -34,7 +34,9 @@
     "scripts": {
         "postbuild": [
             "%project:Directory%/../../build/batchcopy \"%project:BuildOutputDir%/Debug/dnx451/*.*\" %project:Directory%/../../artifacts/build/dnx-clr-win-x86/bin",
-            "%project:Directory%/../../build/batchcopy \"%project:BuildOutputDir%/Debug/dnxcore50/*.*\" %project:Directory%/../../artifacts/build/dnx-coreclr-win-x86/bin"
+            "%project:Directory%/../../build/batchcopy \"%project:BuildOutputDir%/Debug/dnx451/*.*\" %project:Directory%/../../artifacts/build/dnx-clr-win-x64/bin",
+            "%project:Directory%/../../build/batchcopy \"%project:BuildOutputDir%/Debug/dnxcore50/*.*\" %project:Directory%/../../artifacts/build/dnx-coreclr-win-x86/bin",
+            "%project:Directory%/../../build/batchcopy \"%project:BuildOutputDir%/Debug/dnxcore50/*.*\" %project:Directory%/../../artifacts/build/dnx-coreclr-win-x64/bin"
```
moozzyk: Mono?

src/Microsoft.Dnx.Project/project.json
```diff
@@ -36,5 +36,14 @@
                 "System.Threading.Tasks": "4.0.11-beta-*"
             }
         }
+    },
+    "scripts": {
+        "postbuild": [
+            "%project:Directory%/../../build/batchcopy \"%project:BuildOutputDir%/Debug/dnx451/*.*\" %project:Directory%/../../artifacts/build/dnx-clr-win-x86/bin/lib/Microsoft.Dnx.Project",
+            "%project:Directory%/../../build/batchcopy \"%project:BuildOutputDir%/Debug/dnxcore50/*.*\" %project:Directory%/../../artifacts/build/dnx-coreclr-win-x86/bin/lib/Microsoft.Dnx.Project",
```
moozzyk: shouldn't this go just to bin? (and not bin/lib/Microsoft.Dnx.Project)?

src/Microsoft.Dnx.Tooling/project.json
```diff
@@ -55,8 +55,10 @@
     "scripts": {
         "postbuild": [
             "%project:Directory%/../../build/batchcopy \"%project:BuildOutputDir%/Debug/dnx451/*.*\" %project:Directory%/../../artifacts/build/dnx-clr-win-x86/bin/lib/Microsoft.Dnx.Tooling",
-            "%project:Directory%/../../build/batchcopy \"%project:BuildOutputDir%/Debug/dnx451/*.*\" %project:Directory%/../../artifacts/build/dnx-mono/bin/lib/Microsoft.Dnx.Tooling",
-            "%project:Directory%/../../build/batchcopy \"%project:BuildOutputDir%/Debug/dnxcore50/*.*\" %project:Directory%/../../artifacts/build/dnx-coreclr-win-x86/bin/lib/Microsoft.Dnx.Tooling"
+            "%project:Directory%/../../build/batchcopy \"%project:BuildOutputDir%/Debug/dnxcore50/*.*\" %project:Directory%/../../artifacts/build/dnx-coreclr-win-x86/bin/lib/Microsoft.Dnx.Tooling",
```
moozzyk: Same as above - shouldn't the target folder be just bin?

src/Microsoft.Dnx.Host/project.json
anurse: Doesn't use this assembly AFAICT. It didn't have Mono before so I didn't want to add it

src/Microsoft.Dnx.Project/project.json
anurse: Nope, this is a tool, so it lives in bin/lib/[toolname] like DNU

src/Microsoft.Dnx.Tooling/project.json
anurse: Nope, see the unchanged lines. Tools like dnu, dth, and Microsoft.Dnx.Project, live in their own worlds under lib

moozzyk: Sounds good. Didn't know about that.

moozzyk:
:ship::it:

src/Microsoft.Dnx.Project/project.json
```diff
@@ -36,5 +36,14 @@
                 "System.Threading.Tasks": "4.0.11-beta-*"
             }
         }
+    },
+    "scripts": {
+        "postbuild": [
```
halter73: Why didn't this have a postbuild step previously?

build/batchcopy.cmd
```diff
@@ -1 +1,14 @@
-@powershell -NoProfile -ExecutionPolicy unrestricted -File "%~dp0\batchcopy.ps1" %*
+@echo off
+
+set SRC=%1
+set DST=%2
+
+set SRC=%SRC:/=\%
+set DST=%DST:/=\%
+
+if not exist %SRC% goto end
```
halter73: I notice the sh script creates the directory if it doesn't exists and carries on. Is there a reason for having different behavior between the two batchcopy scripts?

src/Microsoft.Dnx.Project/project.json
anurse: It looks like it was just missed, mostly because no one works with this project :)

build/batchcopy.cmd
anurse: Not sure, just ported it from PowerShell. I don't see a reason why we couldn't create the directory here, but in general it's not expected that this command would ever be run without having done a build previously.

halter73:
:shipit:

PR #2561, opened 2015-08-25
allow additional files to be included in the nupkg

### LINKED ISSUE #848: Allow additional files in the nupkg
Similar to a nuspec file, project.json should allow you to specify additional files to include in the nupkg. I imagine a format like the following.
```JSON
{
    "files": {
        "tools/EntityFramework.psd1": "tools/",
        "tools/EntityFramework.psm1": "tools/",
        "tools/init.ps1": "tools/"
    }
}
```
We currently do this post-build on the following packages.
* EntityFramework.Commands
* Microsoft.Data.SQLite

anurse:
fixes #848 

blocked on https://github.com/aspnet/FileSystem/pull/120 ... still waiting for a squirrel there ;). Anyone on this review should feel free to go review that and give me a squirrel there :P

With this change, the project.json can have a new `packageFiles` section, for example:

```json
{
    "packageFiles": {
        "destination1/": "source1/**",
        "destination2/": "source2/**",
        "destination2/some_additional_file.txt": "source2a/somefile.txt",
        "destination3/": ["source3/a.txt", "source3/b.txt", "source4/c/**/*.ps1"]
    }
}
```

The value is a JSON object. Keys in this object (`destination1`, etc. in the above example) are destinations relative to the package root. P. If they end with a `/`, they are treated as directories, otherwise they are files. The associated value is one or more globbing patterns. If the globs end up pulling in multiple files, the destination on the left must be a directory, otherwise an error is reported (see tests for example). If the destination is a directory, the file name for each file is calculated using the stem of the globbing match (see aspnet/FileSystem#120).

Open Questions:

1. Name of `PackageFilesSource` class is weird, but I want to keep the Line/Column info around for error reporting...
2. Handling the "arity" of the source/destination pairs is still a little icky. Requiring that directories end with "/" seemed like the easiest because it allows us to have a clear error. Thoughts? That part can be tweaked both now, and after receiving user feedback in the remaining beta.

/cc @davidfowl @Tratcher @BrennanConroy @halter73 @lodejard

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -64,6 +64,7 @@ var PROGRAM_FILES_X86 = '${Environment.GetFolderPath(Environment.SpecialFolder.P
 var MSBUILD = '${Path.Combine(PROGRAM_FILES_X86, "MSBuild", "14.0", "Bin", "MSBuild.exe")}'
 var VS_REDIST_ROOT = '${Path.Combine(PROGRAM_FILES_X86, @"Microsoft Visual Studio 14.0\VC\redist")}'
 var WIN10_SDK_LIB = '${Path.Combine(PROGRAM_FILES_X86, @"Windows Kits\10\Lib")}'
+var VS_ONECORE_ROOT = '${Path.Combine(PROGRAM_FILES_X86, @"Microsoft Visual Studio 14.0\VC\lib\onecore")}'
 var CLANG = '${SearchForClang()}'
 
 var MANAGED_PROJECTS = '${FindAllProjects(
@@ -93,7 +94,7 @@ var WIN_MANAGED_PROJECTS = '${FindAllProjects(
 
 var RUNTIME_TARGETS='${new List<RuntimeTarget>()}'
 
-var CAN_BUILD_ONECORE = '${Directory.Exists(WIN10_SDK_LIB) && Directory.GetFiles(WIN10_SDK_LIB, "onecore.lib", SearchOption.AllDirectories).Any()}'
+var CAN_BUILD_ONECORE = '${Directory.Exists(VS_ONECORE_ROOT) && Directory.GetFiles(VS_ONECORE_ROOT, "vcruntime.lib", SearchOption.AllDirectories).Any() && Directory.Exists(WIN10_SDK_LIB) && Directory.GetFiles(WIN10_SDK_LIB, "onecore.lib", SearchOption.AllDirectories).Any()}'
 
 @{
     // Always build mono
--- a/src/Microsoft.Dnx.Runtime.Json.Sources/JsonArray.cs
+++ b/src/Microsoft.Dnx.Runtime.Json.Sources/JsonArray.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 
 namespace Microsoft.Dnx.Runtime.Json
 {
@@ -20,14 +21,8 @@ namespace Microsoft.Dnx.Runtime.Json
             _array = array;
         }
 
-        public int Length
-        {
-            get { return _array.Length; }
-        }
-
-        public JsonValue this[int index]
-        {
-            get { return _array[index]; }
-        }
+        public int Length => _array.Length;
+        public IEnumerable<JsonValue> Values => _array;
+        public JsonValue this[int index] => _array[index];
     }
 }
--- a/src/Microsoft.Dnx.Runtime/ProjectFilesCollection.cs
+++ b/src/Microsoft.Dnx.Runtime/ProjectFilesCollection.cs
@@ -1,8 +1,10 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using Microsoft.Dnx.Runtime.Json;
 
 namespace Microsoft.Dnx.Runtime
@@ -18,6 +20,8 @@ namespace Microsoft.Dnx.Runtime
 
         public static readonly string[] DefaultBuiltInExcludePatterns = new[] { "bin/**", "obj/**", "**/*.xproj" };
 
+        public static readonly string PackIncludePropertyName = "packInclude";
+
         private PatternGroup _sharedPatternsGroup;
         private PatternGroup _resourcePatternsGroup;
         private PatternGroup _preprocessPatternsGroup;
@@ -25,6 +29,7 @@ namespace Microsoft.Dnx.Runtime
         private PatternGroup _contentPatternsGroup;
         private IDictionary<string, string> _namedResources;
         private IEnumerable<string> _publishExcludePatterns;
+        private IEnumerable<PackIncludeEntry> _packInclude;
 
         private readonly string _projectDirectory;
         private readonly string _projectFilePath;
@@ -76,8 +81,31 @@ namespace Microsoft.Dnx.Runtime
 
             _namedResources = NamedResourceReader.ReadNamedResources(_rawProject, _projectFilePath);
 
+            // Files to be packed along with the project
+            var packIncludeJson = _rawProject.ValueAsJsonObject(PackIncludePropertyName);
+            if (packIncludeJson != null)
+            {
+                _packInclude = packIncludeJson
+                    .Keys
+                    .Select(k => new PackIncludeEntry(k, packIncludeJson.Value(k)))
+                    .ToList();
+            }
+            else
+            {
+                _packInclude = new List<PackIncludeEntry>();
+            }
+
             _initialized = true;
             _rawProject = null;
+        }
+
+        public IEnumerable<PackIncludeEntry> PackInclude
+        {
+            get
+            {
+                EnsureInitialized();
+                return _packInclude;
+            }
         }
 
         public IEnumerable<string> SourceFiles
--- a/src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
+++ b/src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
@@ -241,6 +241,7 @@ namespace Microsoft.Dnx.Tooling
                             AddPackageFiles(_currentProject.ProjectDirectory, _currentProject.Files.PackInclude, packageBuilder, packDiagnostics);
                         }
                         success &= !packDiagnostics.HasErrors();
+                        allDiagnostics.AddRange(packDiagnostics);
 
                         foreach (var path in _currentProject.Files.SourceFiles)
                         {
--- a/src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
+++ b/src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
@@ -51,7 +51,7 @@ namespace Microsoft.Dnx.Tooling
 
             // Resolve all the project names
             var projectFilesToBuild = new List<string>();
-            foreach(var pattern in _buildOptions.ProjectPatterns)
+            foreach (var pattern in _buildOptions.ProjectPatterns)
             {
                 if (pattern.Contains("*"))
                 {
@@ -67,7 +67,7 @@ namespace Microsoft.Dnx.Tooling
             var rootDirectory = Directory.GetCurrentDirectory();
             var patternSearchFolder = new DirectoryInfoWrapper(new DirectoryInfo(rootDirectory));
             var globbingProjects = projectFilesFinder.Execute(patternSearchFolder).Files.Select(file =>
-                Path.Combine(rootDirectory, file));
+                Path.Combine(rootDirectory, file.Path));
             projectFilesToBuild.AddRange(globbingProjects);
 
             var sw = Stopwatch.StartNew();
@@ -225,6 +225,7 @@ namespace Microsoft.Dnx.Tooling
 
                     if (configurationSuccess)
                     {
+                        var packDiagnostics = new List<DiagnosticMessage>();
                         foreach (var sharedFile in _currentProject.Files.SharedFiles)
                         {
                             var file = new PhysicalPackageFile();
@@ -235,6 +236,12 @@ namespace Microsoft.Dnx.Tooling
 
                         var root = _currentProject.ProjectDirectory;
 
+                        if (_currentProject.Files.PackInclude != null && _currentProject.Files.PackInclude.Any())
+                        {
+                            AddPackageFiles(_currentProject.ProjectDirectory, _currentProject.Files.PackInclude, packageBuilder, packDiagnostics);
+                        }
+                        success &= !packDiagnostics.HasErrors();
+
                         foreach (var path in _currentProject.Files.SourceFiles)
                         {
                             var srcFile = new PhysicalPackageFile();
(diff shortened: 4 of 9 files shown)


src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
```diff
@@ -296,6 +310,68 @@ private bool BuildInternal(string projectPath)
...
+                if(pair.Key.EndsWith("/"))
+                {
+                    var dir = pair.Key.Substring(0, pair.Key.Length - 1).Replace('/', Path.DirectorySeparatorChar);
+
+                    foreach(var file in files)
+                    {
+                        yield return new PhysicalPackageFile()
+                        {
+                            SourcePath = file.Path.Replace('/', Path.DirectorySeparatorChar),
+                            TargetPath = Path.Combine(dir, file.Stem.Replace('/', Path.DirectorySeparatorChar))
```
moozzyk: I think it might break in case of absolute, incorrect paths - two cases that comes to my mind are:

`/Windows/` -->You will probably end up witch `\Windows\file.txt` path
// --> you will end up with `\`

src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
```diff
@@ -235,6 +236,11 @@ private bool BuildInternal(string projectPath)
 
                         var root = _currentProject.ProjectDirectory;
 
+                        if(_currentProject.PackageFiles != null && _currentProject.PackageFiles.Any())
```
moozzyk: add spaces after `if` and `foreach` instructions before you get too many comments on that (you have quite a few occurrences) also use `var`

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.IO.Compression;
```
moozzyk: needed?

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
```diff
@@ -71,7 +72,7 @@ public class TestClass : BaseClass {
                 Assert.Equal(0, exitCode);
             }
         }
-
+
```
moozzyk: remove

davidfowl:
My eyes aren't used to destinations being on the left, it feels still feels odd.

src/Microsoft.Dnx.Runtime/PackageFilesSource.cs
```diff
@@ -0,0 +1,42 @@
...
+        public int Column { get; }
+
+        internal PackageFilesSource(JsonValue json)
+            : this(ExtractValues(json), json.Line, json.Column)
+        {
+        }
+
+        public PackageFilesSource(IEnumerable<string> values, int line, int column)
+        {
+            Values = values.ToList().AsReadOnly();
```
davidfowl: Do all callers call an IList<string> directly? If so change the input to list or array or something specific. This isn't a generic type that we want to expose to people is it?

src/Microsoft.Dnx.Runtime/Project.cs
```diff
@@ -93,6 +94,8 @@ public string ProjectDirectory
 
         public ProjectFilesCollection Files { get; private set; }
 
+        public IReadOnlyDictionary<string, PackageFilesSource> PackageFiles { get; private set; }
```
davidfowl: Is there any reason to no merge this in with the existing project files collection?

anurse:
I suppose we could do destination on the right. It's a little weird to have globbing patterns as keys in a dictionary, but it's not awful.

src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
```diff
@@ -243,18 +249,26 @@ private bool BuildInternal(string projectPath)
                             symbolPackageBuilder.Files.Add(srcFile);
                         }
 
-                        using (var fs = File.Create(nupkg))
-                        {
-                            packageBuilder.Save(fs);
-                            _buildOptions.Reports.Quiet.WriteLine("{0} -> {1}", _currentProject.Name, Path.GetFullPath(nupkg));
-                        }
+                        bool packageSuccess = packDiagnostics.All(d => d.Severity != DiagnosticMessageSeverity.Error);
```
ChengTian: packDiagnostics.HasErrors()

ChengTian:
With the following configuration:
```json
{
    "packageFiles": {
        "tools/": "A.txt",
        "./tools/": "B.txt",
        "arbitrary/../tools": "C.txt"
    }
}
```
The current implementation will copy all three .txt files into one tools folder, which looks fine.

However, with
```json
{
    "packageFiles": {
        "tools/final.txt": "A.txt",
        "./tools/final.txt": "B.txt",
        "arbitrary/../tools/final.txt": "C.txt"
    }
}
```
The behavior is undefined because we are using Dictionary, which is unsorted. So we don't know which .txt file gets copied as the last one and overwrites all previous ones.

moozzyk:
Should we even allow overwriting files by default? feels scary (--force option would be useful)

src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
anurse: The source paths come from globbing which has already processed them and made them absolute. And the target paths are used relative to the ZIP file being generated so the worst they can do is mess up a user's package. In either case, if the user types an incorrect path, they will get an incorrect result, which doesn't seem bad to me :)

src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
anurse: I wonder if the format-on-save extension has been updated for VS2015 now .... ;)

src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
moozzyk: I think file.Stem is always relative so you are right - we should be always relative to the ZIP file and then it is on the user to make sure that targets are correct.

src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
moozzyk: Ctrl + K, D

anurse:
I'd be ok with blocking files with duplicate final target paths.

anurse:
I'd also be OK with banning path-traversal characters in the target paths since they are unnecessary :)

src/Microsoft.Dnx.Runtime/Project.cs
```diff
@@ -233,12 +236,25 @@ internal static Project GetProjectFromStream(Stream stream, string projectName,
 
             project.Dependencies = new List<LibraryDependency>();
 
+            // Files to be packed along with the project
+            var packInclude = rawProject.ValueAsJsonObject(PackIncludePropertyName);
```
davidfowl: Parse all of this lazily

anurse: I'm doing so, but it means more of the JSON is retained in memory without being parsed down to smaller data structures. I'm not sure that's actually useful. Perhaps with the ProjectReader change we could make the amount of data loaded configurable and then only have dnu pack load this much but that seems like a larger change.

Anyway, I'll lazy it for now and we can discuss a better option later.

davidfowl: > Perhaps with the ProjectReader change we could make the amount of data loaded configurable and then only have dnu pack load this much but that seems like a larger change.

I'll look into that. I was thinking of a runtime project but this might be a better option.

anurse: So a separate project object? That's not a bad idea either. Also, we should extract ProjectReader out into a package at some point (possibly just a shared source package). We'd still embed it in the runtime, but as @rynowak said it's useful in other situations.

anurse:
:watch: @davidfowl @moozzyk

anurse:
Travis failures are unrelated

src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
```diff
@@ -243,18 +249,26 @@ private bool BuildInternal(string projectPath)
...
-                            using (var fs = File.Create(symbolsNupkg))
+                            using (var fs = File.Create(nupkg))
                             {
-                                symbolPackageBuilder.Save(fs);
-                                _buildOptions.Reports.Quiet.WriteLine("{0} -> {1}", _currentProject.Name, Path.GetFullPath(symbolsNupkg));
+                                packageBuilder.Save(fs);
+                                _buildOptions.Reports.Quiet.WriteLine("{0} -> {1}", _currentProject.Name, Path.GetFullPath(nupkg));
+                            }
+
+                            if (symbolPackageBuilder.Files.Any())
```
davidfowl: This is a new change

anurse: What? No it isn't, it was in my original PR.

anurse: The inline diff looks weird, split diff is clearer. Regardless, this was in the PR from early on.

src/Microsoft.Dnx.Runtime/Project.cs
```diff
@@ -17,6 +17,8 @@ public class Project
         public static readonly TypeInformation DefaultRuntimeCompiler = new TypeInformation("Microsoft.Dnx.Compilation.CSharp", "Microsoft.Dnx.Compilation.CSharp.RoslynProjectCompiler");
         public static readonly TypeInformation DefaultDesignTimeCompiler = new TypeInformation("Microsoft.Dnx.Compilation.DesignTime", "Microsoft.Dnx.Compilation.DesignTime.DesignTimeHostProjectCompiler");
 
+        public static readonly string PackIncludePropertyName = "packInclude";
```
davidfowl: Move this to ProjectFilesCollection

src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
```diff
@@ -225,6 +225,7 @@ private bool BuildInternal(string projectPath)
 
                     if (configurationSuccess)
                     {
+                        var packDiagnostics = new List<DiagnosticMessage>();
```
davidfowl: Why have separate diagnostics? Why not add it to all diagnostics?

src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
```diff
@@ -296,6 +310,97 @@ private bool BuildInternal(string projectPath)
...
+            foreach (var entry in projectFileGlobs)
+            {
+                // Evaluate the globs on the right
+                var matcher = new Matcher();
+                matcher.AddIncludePatterns(entry.SourceGlobs);
+                var results = matcher.Execute(rootDirectory);
+                var files = results.Files.ToList();
+
+                // Check for illegal characters
+                if(string.IsNullOrEmpty(entry.Target))
```
davidfowl: spacing between ifs

src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
```diff
@@ -296,6 +310,97 @@ private bool BuildInternal(string projectPath)
...
+                // Check the arity of the left
+                if (entry.Target.EndsWith("/"))
+                {
+                    var dir = entry.Target.Substring(0, entry.Target.Length - 1).Replace('/', Path.DirectorySeparatorChar);
+
+                    foreach (var file in files)
+                    {
+                        yield return new PhysicalPackageFile()
+                        {
+                            SourcePath = file.Path.Replace('/', Path.DirectorySeparatorChar),
```
davidfowl: There's a helper method in  PathUtility for this operation

src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
anurse: Because if the packaging fails, the WriteDiagnostics(allDiagnostics) call is never hit. I thought about refactoring this to be clearer but wanted to limit the change.

anurse: See https://github.com/aspnet/dnx/pull/2561/files#diff-3ed8b70df3021a1c57d3a4aa849e8d1aR278

anurse: Think we can just rewrite that if though, hang on.

src/Microsoft.Dnx.Runtime.Json.Sources/JsonArray.cs
```diff
@@ -9,6 +10,8 @@ internal class JsonArray : JsonValue
     {
         private readonly JsonValue[] _array;
 
+        public IEnumerable<JsonValue> Values { get { return _array; } }
```
pranavkm: `public IEnumerable<JsonValue> Values => _array;`

src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
```diff
@@ -296,6 +310,97 @@ private bool BuildInternal(string projectPath)
...
+                        "targets must either be a file name or a directory suffixed with '/'. " +
+                        "The root directory of the package can be specified by using a single '/' character.",
+                        projectFilePath,
+                        DiagnosticMessageSeverity.Error,
+                        entry.Line,
+                        entry.Column));
+                    continue;
+                }
+
+                if (entry.Target.Split('/').Any(s => s.Equals(".") || s.Equals("..")))
```
pranavkm: If we're doing path sanitization, would rooted paths also not be supported?

anurse: "rooted" paths in the target refer to the root of the package, so no. See the error message above

src/Microsoft.Dnx.Runtime.Json.Sources/JsonArray.cs
anurse: done

src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
anurse: done

src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
anurse: done

src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
anurse: Looks like we're good. Removed the extra list

src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
```diff
@@ -296,6 +310,97 @@ private bool BuildInternal(string projectPath)
...
+                            projectFilePath,
+                            DiagnosticMessageSeverity.Error,
+                            entry.Line,
+                            entry.Column));
+                    }
+                    else
+                    {
+                        yield return new PhysicalPackageFile()
+                        {
+                            SourcePath = files.Single().Path,
```
pranavkm: `files[0]`

src/Microsoft.Dnx.Runtime/ProjectFilesCollection.cs
```diff
@@ -76,10 +79,33 @@ internal void EnsureInitialized()
...
+            else
+            {
+                _packInclude = new List<PackIncludeEntry>();
+            }
+
             _initialized = true;
             _rawProject = null;
         }
 
+        public IList<PackIncludeEntry> PackInclude
```
pranavkm: `IEnumerable` \ `IReadOnlyList`?

anurse: Had that originally, @davidfowl wanted to be consistent with everything else.

anurse: though it can be IEnumerable now.

anurse:
:watch: @pranavkm @davidfowl

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackAdditionalFilesTests.cs
```diff
@@ -0,0 +1,430 @@
...
+                {
+                    Assert.True(File.Exists(packageOutputPath));
+
+                    string[] entries;
+                    using (var archive = ZipFile.OpenRead(packageOutputPath))
+                    {
+                        entries = archive.Entries.Select(e => e.FullName).Where(IsNotOpcMetadata).ToArray();
+                    }
+
+                    Assert.Equal(0, exitCode);
```
pranavkm: Could you print the `stdError` if the exitCode is non-zero?

pranavkm:
:shipit:

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackAdditionalFilesTests.cs
anurse: It's already printed by the test.

src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
anurse: Nope, we aren't good. The only remaining call is to WriteSummary(allDiagnostics) which is a summary view. We still need to write the individual diagnostic messages from the packaging process. Returning this code.

makefile.shade
```diff
@@ -64,6 +64,7 @@ var PROGRAM_FILES_X86 = '${Environment.GetFolderPath(Environment.SpecialFolder.P
 var MSBUILD = '${Path.Combine(PROGRAM_FILES_X86, "MSBuild", "14.0", "Bin", "MSBuild.exe")}'
 var VS_REDIST_ROOT = '${Path.Combine(PROGRAM_FILES_X86, @"Microsoft Visual Studio 14.0\VC\redist")}'
 var WIN10_SDK_LIB = '${Path.Combine(PROGRAM_FILES_X86, @"Windows Kits\10\Lib")}'
+var VS_ONECORE_ROOT = '${Path.Combine(PROGRAM_FILES_X86, @"Microsoft Visual Studio 14.0\VC\lib\onecore")}'
```
moozzyk: @anurse - why was this needed?

anurse: I was in a situation where I did have the Windows SDK installed but still couldn't build for OneCore because I was missing files in this folder. I'll sync up with you in person.

andreasandreou:
any idea when this will be available?

anurse:
It should be available already, though the name was changed to `packInclude`.

andreasandreou:
@anurse thank you, indeed you can include content using packInclude but there is still no way to get to import the content from package into the project which makes this pretty much useless.

anurse:
Correct, that's something tracked over in the dotnet/cli repo, since DNX is being retired: https://github.com/dotnet/cli/issues/66

andreasandreou:
thank you very much for the follow up @anurse, this is very good news, will track referenced issue.

PR #2579, opened 2015-08-28
Split the project parsing and the project into separate classes

davidfowl:
- Added ProjectReader

/cc @anurse conflicts a' coming

### DIFF
--- a/src/Microsoft.Dnx.Runtime/Project.cs
+++ b/src/Microsoft.Dnx.Runtime/Project.cs
@@ -3,13 +3,9 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Runtime.Versioning;
-using System.Text;
-using Microsoft.Dnx.Runtime.Helpers;
-using Microsoft.Dnx.Runtime.Json;
 using NuGet;
 
 namespace Microsoft.Dnx.Runtime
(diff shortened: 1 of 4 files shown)


rynowak:
It would be really handy if this was available for `TestHost` to use

anurse:
Travis? TRAVIS?! TRAVIS?!!

anurse:
Someone told me not to review code unless it passes Travis sooooooo,... :trollface:

davidfowl:
I agree with that actually the red is bugging me so much :cry: . But mac osx is failing because of relative path and tmp crap. I'll fix that as soon as I get this in so we can be green again

anurse:
Suuuuuure ;)

davidfowl:
@anurse promise :smile:

src/Microsoft.Dnx.Runtime/Project.cs
```diff
@@ -23,19 +19,19 @@ public class Project
 
         public static TypeInformation DefaultCompiler = DefaultRuntimeCompiler;
 
-        private readonly Dictionary<FrameworkName, TargetFrameworkInformation> _targetFrameworks = new Dictionary<FrameworkName, TargetFrameworkInformation>();
-        private readonly Dictionary<FrameworkName, CompilerOptions> _compilationOptions = new Dictionary<FrameworkName, CompilerOptions>();
-        private readonly Dictionary<string, CompilerOptions> _configurations = new Dictionary<string, CompilerOptions>(StringComparer.OrdinalIgnoreCase);
+        // REVIEW: It's kinda hacky making these internal but the reader needs to set them
```
anurse: Could you put them in a constructor? If you did that I think it'd be OK for the constructor to be public and the grossness would go away.

davidfowl: The dictionaries yes but not the default compiler or target framework.

anurse: Why not? If they come from the JSON then they should be fine there. If they don't come from the JSON they don't need to be in ProjectReader :)

davidfowl: I'd have to create the project after parsing everything. Which might be possible but I'd have to look.

anurse: Fair enough for now.

PR #2592, opened 2015-08-31
Avoid closure allocation in ApplicationContext

Maxwe11:


src/Microsoft.Dnx.DesignTimeHost/ApplicationContext.cs
```diff
@@ -95,7 +95,7 @@ public void OnReceive(Message message)
                 _inbox.Enqueue(message);
             }
 
-            ThreadPool.QueueUserWorkItem(ProcessLoop);
+            ThreadPool.QueueUserWorkItem(state => ((ApplicationContext)state).ProcessLoop(null), this);
```
davidfowl: Remove the state parameter from the method. That'll make it clearer.

Maxwe11: done

PR #2593, opened 2015-08-31
Avoid `async` and Unwrap overhead in Bootstrapper

Maxwe11:


davidfowl:
The change looks fine but is there any real overhead?

Maxwe11:
@davidfowl 
- `.Unwrap()` allocates TaskCompletionSource
- compilers generate a lot of code for `async` methods which is unnecessary in this case because we awaiting on completed task.

davidfowl:
@Maxwe11 sure but have you measured anything?

davidfowl:
Can you make the change here too https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.ApplicationHost/Program.cs#L89

Maxwe11:
>sure but have you measured anything?

It's hard to measure it but in general we avoid TaskCompletionSource allocation :smile:  (which if I understand the code correctly will be in memory while user's Main executed)

>Can you make the change here too

done

davidfowl:
@Maxwe11 in general, changes that claim to improve performance should have performance tests/metrics associated with change. 

Still I like this change because it simplifies the logic (assuming nothing regressed like exception behavior) but all of the tests passed :smile:

PR #2595, opened 2015-08-31
Commonizing ConvertMetdataReference

pranavkm:
cc @anurse  \ @davidfowl

### DIFF
--- a/src/Microsoft.Dnx.Compilation.CSharp.Common/MetadataReferenceExtensions.cs
+++ b/src/Microsoft.Dnx.Compilation.CSharp.Common/MetadataReferenceExtensions.cs
@@ -1,0 +1,76 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.Reflection.PortableExecutable;
+using Microsoft.CodeAnalysis;
+
+namespace Microsoft.Dnx.Compilation.CSharp
+{
+    /// <summary>
+    /// Extension methods for <see cref="IMetadataReference"/>
+    /// </summary>
+    public static class MetadataReferenceExtensions
+    {
+        /// <summary>
+        /// Converts an <see cref="IMetadataReference"/> to a <see cref="MetadataReference"/> instance.
+        /// </summary>
+        /// <param name="metadataReference">The <see cref="IMetadataReference"/> to convert.</param>
+        /// <param name="assemblyMetadataFactory">Factory invoked to get instances of <see cref="AssemblyMetadata"/>.</param>
+        /// <returns>The converted <see cref="MetadataReference"/>.</returns>
+        public static MetadataReference ConvertMetadataReference(
+            this IMetadataReference metadataReference,
+            Func<IMetadataFileReference, AssemblyMetadata> assemblyMetadataFactory)
+        {
+            var roslynReference = metadataReference as IRoslynMetadataReference;
+
+            if (roslynReference != null)
+            {
+                return roslynReference.MetadataReference;
+            }
+
+            var embeddedReference = metadataReference as IMetadataEmbeddedReference;
+
+            if (embeddedReference != null)
+            {
+                return MetadataReference.CreateFromImage(embeddedReference.Contents);
+            }
+
+            var fileMetadataReference = metadataReference as IMetadataFileReference;
+
+            if (fileMetadataReference != null)
+            {
+                var metadata = assemblyMetadataFactory(fileMetadataReference);
+                return metadata.GetReference(filePath: fileMetadataReference.Path);
+            }
+
+            var projectReference = metadataReference as IMetadataProjectReference;
+            if (projectReference != null)
+            {
+                using (var ms = new MemoryStream())
+                {
+                    projectReference.EmitReferenceAssembly(ms);
+
+                    return MetadataReference.CreateFromStream(ms, filePath: projectReference.ProjectPath);
+                }
+            }
+
+            throw new NotSupportedException($"Unsupported type '{metadataReference.GetType()}'.");
+        }
+
+        /// <summary>
+        /// Creates a <see cref="AssemblyMetadata"/> for the assembly specified by <paramref name="fileReference"/>.
+        /// </summary>
+        /// <param name="fileReference">The <see cref="IMetadataFileReference"/>.</param>
+        /// <returns>An <see cref="AssemblyMetadata"/>.</returns>
+        public static AssemblyMetadata CreateAssemblyMetadata(this IMetadataFileReference fileReference)
+        {
+            using (var stream = File.OpenRead(fileReference.Path))
+            {
+                var moduleMetadata = ModuleMetadata.CreateFromStream(stream, PEStreamOptions.PrefetchMetadata);
+                return AssemblyMetadata.Create(moduleMetadata);
+            }
+        }
+    }
+}
--- a/src/Microsoft.Dnx.Compilation.CSharp/RoslynCompiler.cs
+++ b/src/Microsoft.Dnx.Compilation.CSharp/RoslynCompiler.cs
@@ -8,7 +8,6 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection;
-using System.Reflection.PortableExecutable;
 using System.Text;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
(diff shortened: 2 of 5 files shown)


src/Microsoft.Dnx.Compilation.CSharp.Common/MetadataReferenceExtensions.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// </summary>
+        /// <param name="metadataReference">The <see cref="IMetadataReference"/> to convert.</param>
+        /// <param name="cache">The <see cref="ICache"/>.</param>
+        /// <param name="onAssemblyMetadataCacheMiss">Invoked when a cache miss occurs when attempting to look 
+        /// up cached <see cref="AssemblyMetadata"/>.</param>
+        /// <returns>The converted <see cref="MetadataReference"/>.</returns>
+        public static MetadataReference ConvertMetadataReference(
+            this IMetadataReference metadataReference,
+            ICache cache,
+            Action<CacheContext, string> onAssemblyMetadataCacheMiss)
```
pranavkm: Thoughts?

pranavkm: Alternatively we can move `FileWriteTimeCacheDependency` to common code and remove the need to do this. That said, in anything outside of DNX, adding a cache dependency would be pretty pointless.

src/Microsoft.Dnx.Compilation.CSharp.Common/MetadataReferenceExtensions.cs
```diff
@@ -0,0 +1,77 @@
...
+            }
+
+            var projectReference = metadataReference as IMetadataProjectReference;
+            if (projectReference != null)
+            {
+                using (var ms = new MemoryStream())
+                {
+                    projectReference.EmitReferenceAssembly(ms);
+
+                    return MetadataReference.CreateFromImage(ms.ToArray());
```
pranavkm: @davidfowl, could we use `CreateFromStream` here?

src/Microsoft.Dnx.Compilation.CSharp.Common/MetadataReferenceExtensions.cs
anurse: It's not the prettiest, but given that this is purely an internal helper that we expect few people to use, I'm not that afraid of it. It's cleaner than bringing FileWriteTimeCacheDependency along when we don't need it.

davidfowl: Uhhhhhh why is the cache in here at all?

src/Microsoft.Dnx.Compilation.CSharp.Common/MetadataReferenceExtensions.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// Converts an <see cref="IMetadataReference"/> to a <see cref="MetadataReference"/> instance.
+        /// </summary>
+        /// <param name="metadataReference">The <see cref="IMetadataReference"/> to convert.</param>
+        /// <param name="cache">The <see cref="ICache"/>.</param>
+        /// <param name="onAssemblyMetadataCacheMiss">Invoked when a cache miss occurs when attempting to look 
+        /// up cached <see cref="AssemblyMetadata"/>.</param>
+        /// <returns>The converted <see cref="MetadataReference"/>.</returns>
+        public static MetadataReference ConvertMetadataReference(
+            this IMetadataReference metadataReference,
+            ICache cache,
```
davidfowl: I don't think this should be exposed and razor shouldn't be using it.

pranavkm: The `ICache` or the method in general?

davidfowl: All of the caching things. Why isn't this helper simple? Converts the metadata reference to a roslyn implementation and nothing more. That's the logic you are trying to commonize right?

src/Microsoft.Dnx.Compilation.CSharp.Common/MetadataReferenceExtensions.cs
anurse: IF you're going to commonize this, then https://github.com/aspnet/dnx/blob/prkrishn/metadataref/src/Microsoft.Dnx.Compilation.CSharp.Common/MetadataReferenceExtensions.cs#L49 requires it. But I agree that commonizing should be discussed a bit further.

src/Microsoft.Dnx.Compilation.CSharp.Common/MetadataReferenceExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// <summary>
+        /// Converts an <see cref="IMetadataReference"/> to a <see cref="MetadataReference"/> instance.
+        /// </summary>
+        /// <param name="metadataReference">The <see cref="IMetadataReference"/> to convert.</param>
+        /// <param name="assemblyMetadataFactory">Factory invoked to get or set cached instances of
+        /// <see cref="AssemblyMetadata"/>.</param>
+        /// <returns>The converted <see cref="MetadataReference"/>.</returns>
+        public static MetadataReference ConvertMetadataReference(
+            this IMetadataReference metadataReference,
+            Func<string, Func<AssemblyMetadata>, AssemblyMetadata> assemblyMetadataFactory)
```
pranavkm: @davidfowl thoughts?

src/Microsoft.Dnx.Compilation.CSharp.Common/MetadataReferenceExtensions.cs
pranavkm: Also, my guess is we could set the `filePath` here.

src/Microsoft.Dnx.Compilation.CSharp.Common/MetadataReferenceExtensions.cs
davidfowl: `Func<string, AssemblyMetadata>` should be enough

src/Microsoft.Dnx.Compilation.CSharp/RoslynCompiler.cs
```diff
@@ -110,8 +119,12 @@ public class RoslynCompiler
                 parseOptions,
                 isMainAspect);
 
-            var embeddedReferences = incomingReferences.OfType<IMetadataEmbeddedReference>()
-                                                       .ToDictionary(a => a.Name, ConvertMetadataReference);
+            var embeddedReferences = incomingReferences
```
davidfowl: Since you're changing this code, remove this.

src/Microsoft.Dnx.Compilation.CSharp.Common/MetadataReferenceExtensions.cs
davidfowl: Who owns the stream?

pranavkm: We create it - https://github.com/aspnet/dnx/pull/2595/files#diff-6d73c10696b867e7a9c5b6a4243462d9R53

src/Microsoft.Dnx.Compilation.CSharp.Common/MetadataReferenceExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+            }
+
+            var projectReference = metadataReference as IMetadataProjectReference;
+            if (projectReference != null)
+            {
+                using (var ms = new MemoryStream())
+                {
+                    projectReference.EmitReferenceAssembly(ms);
+
+                    return MetadataReference.CreateFromStream(ms, filePath: projectReference.ProjectPath);
```
pranavkm: This line changed

src/Microsoft.Dnx.Compilation.CSharp.Common/MetadataReferenceExtensions.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+        /// <summary>
+        /// Converts an <see cref="IMetadataReference"/> to a <see cref="MetadataReference"/> instance.
+        /// </summary>
+        /// <param name="metadataReference">The <see cref="IMetadataReference"/> to convert.</param>
+        /// <param name="assemblyMetadataFactory">Factory invoked to get instances of <see cref="AssemblyMetadata"/>.</param>
+        /// <returns>The converted <see cref="MetadataReference"/>.</returns>
+        public static MetadataReference ConvertMetadataReference(
+            this IMetadataReference metadataReference,
+            Func<IMetadataFileReference, AssemblyMetadata> assemblyMetadataFactory)
```
pranavkm: Round 3

davidfowl:
:shipit: once the ci passes

Issue #2387, opened 2015-07-31
DNX does not select a runtime or resolve runtime assets


ericstj:
Packages can have runtime specific assets.  Currently our .NET packages define a set of runtime IDs with a heirarchy that are used by our packages.  This works fine for UWP, PCL, and Desktop where the MSBuild task uses an active RID when resolving runtime assets, but not on DNX.

This is going to become even more important as we use the RID more in the upcoming release to distinguish between linux and windows assets.

anurse:
Going to capture some of the results of our previous design discussions on this:

1. DNX will have a built-in "runtime.json" that will define DNX-specific runtime IDs. For example `dnx-win8-x64`, `dnx-linux-x64`, etc. These will inherit from base RIDs to join in the general RID ecosystem

2. At runtime, it's pretty clear what RID we are using. OS Version detection might get a bit tricky, but at the very least we can know the OS type (win/linux/darwin) and bitness

3. Restore time is the tricky part. In NuGet3 we require that project.json contains a `runtimes` section, and we will likely support that in DNX as well. However, we're not sure we want to require it. The question then becomes, what do we do when there is no `runtimes` section? Do we assume it contains all the `dnx-*` RIDs? Maybe we do have to require it, but we still need to think on that a bit.

4. Much like we do for CLR quirking versions, we should probably provide ways for `dnu` and `dnx` to take in a runtime to force DNX to select that runtime (if it is compatible with the project).

danroth27:
@Eilon @lodejard We need to figure out what to do here for beta7

anurse:
After chatting with @ericstj, @davidfowl and @Eilon we decided we can move this to beta8

anurse:
Ok, the final design I'm going with is as follows:

1. Every DNX knows it's RID components: (Operating System Version, Architecture).
2. At runtime, the DNX looks up the section in the `targets` of the lock file for it's specific RID. There must be an **exact** match, meaning the project must be restored for the correct operating system(s)
3. Each DNU has a set of RIDs it will restore for by default. These consist of the combination of the currently-running OS version and all known architectures. For example, on Windows 10: `win10-x86`, `win10-x64`, `win10-arm`. On Windows 7, `win7-x86`, `win7-x64`. On Linux: `linux-x86`, `linux-x64`, `linux-arm`, etc.
4. If a project contains a `runtimes` section, we will always use that to produce a set of RIDs to restore for.
[...]

Issue #2063, opened 2015-06-13
architecture is "x86' on arm


moozzyk:
To determine OS architecture we just check the size of the pointer (https://github.com/aspnet/dnx/blob/32358ec9b8970070ff77cbc5beef28c13e3ab09b/src/dnx.host/RuntimeEnvironment.cs#L21). As a result Win10 IoT one core on ARM is retported to be x86.

BrennanConroy:
Here too if we run tests on arm: https://github.com/aspnet/dnx/blob/32358ec9b8970070ff77cbc5beef28c13e3ab09b/test/dnx.host.Tests/RuntimeEnvironmentTests.cs#L42

moozzyk:
And in kestrel where we use it to load libuv.dll.

anurse:
Shouldn't the native bootstrapper just pass this data in and the managed code just read it rather than trying to redetect it? The bootstrapper is unique per architecture and OS (and there's a bootstrapper for Mono too).

moozzyk:
Possibly. Might part of #1970

muratg:
@DamianEdwards is ARM in scope for V1?  

cc @davidfowl @Eilon @danroth27 

Also see https://github.com/aspnet/KestrelHttpServer/issues/106

moozzyk:
Actually this should resolve itself semi-automatically when dnx starts loading native dlls automatically. - then we will be able to remove the code where we have the check.

moozzyk:
We seem to be doing this only in one place: https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Loader/LoadContext.cs#L71

davidfowl:
Do we still need this?

moozzyk:
Possibly - depending on what happens with `LoadContext` and if ARM is going to be officially supported.

PR #2597, opened 2015-08-31
add code to determine the runtime ID for a DNX

anurse:
**Part** of #2387, but not the whole runtime.json feature.

I wanted to get this in to dev relatively quickly because a) it's benign, nothing will use it yet and b) it will affect @moozzyk 's work on #1970.

In theory, we will need to have the runtime ID itself passed in by the native code, so I wanted to put a place for that to go. For now, it is synthesized by the managed code, but it should live in the native code since that is where we will be pivoting the runtime.

After this, `dnx --version` outputs the runtime ID. For example:

image

I had to add an app manifest that enables Windows 10 compatibility in order for [`GetVersionEx`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms724451(v=vs.85).aspx) (and thus `Environment.OperatingSystem.Version`) to return higher than `6.2` (Windows 8).

From here it's relatively straight forward to provide the necessary data to restore and during loading.

/cc @ericstj @davidfowl @moozzyk @lodejard

### DIFF
--- a/src/Microsoft.Dnx.Runtime.Internals/Constants.cs
+++ b/src/Microsoft.Dnx.Runtime.Internals/Constants.cs
@@ -1,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Dnx.Runtime
+{
+    internal static class RuntimeTypes
+    {
+        public static readonly string CoreCLR = nameof(CoreCLR);
+        public static readonly string CLR = nameof(CLR);
+        public static readonly string Mono = nameof(Mono);
+    }
+
+    internal static class RuntimeArchitectures
+    {
+        public static readonly string X86 = "x86";
+        public static readonly string X64 = "x64";
+    }
+
+    internal static class RuntimeOperatingSystems
+    {
+        public static readonly string Windows = nameof(Windows);
+    }
+}
--- a/src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
+++ b/src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Text;
 
 namespace Microsoft.Dnx.Runtime
@@ -20,12 +21,49 @@ namespace Microsoft.Dnx.Runtime
                 str.AppendLine($" OS Version:   {env.OperatingSystemVersion}");
             }
 
+            str.AppendLine($" Runtime Id:   {env.GetRuntimeIdentifier()}");
+
             return str.ToString();
         }
 
         public static string GetShortVersion(this IRuntimeEnvironment env)
         {
             return $"{env.RuntimeType}-{env.RuntimeArchitecture}-{env.RuntimeVersion}";
+        }
+
+        public static string GetRuntimeIdentifier(this IRuntimeEnvironment env)
+        {
+            string os = env.OperatingSystem;
+            string ver = env.OperatingSystemVersion;
+            if (string.Equals(os, RuntimeOperatingSystems.Windows, StringComparison.Ordinal))
+            {
+                os = "win";
+
+                // Convert 6.x to the correct branding version for the RID
+                var parsedVersion = Version.Parse(ver);
+                if(parsedVersion.Major == 6 && parsedVersion.Minor == 1)
+                {
+                    ver = "7";
+                }
+                else if(parsedVersion.Major == 6 && parsedVersion.Minor == 2)
+                {
+                    ver = "8";
+                }
+                else if(parsedVersion.Major == 6 && parsedVersion.Minor == 3)
+                {
+                    ver = "81";
+                }
+                else if(parsedVersion.Major == 10 && parsedVersion.Minor == 0)
+                {
+                    ver = "10";
+                }
+            }
+            else
+            {
+                os = os.ToLower(); // Just use the lower-case full name of the OS as the RID OS
+            }
+
+            return $"{os}{ver}-{env.RuntimeArchitecture.ToLower()}";
         }
     }
 }
--- a/test/Microsoft.Dnx.Host.Tests/RuntimeEnvironmentTests.cs
+++ b/test/Microsoft.Dnx.Host.Tests/RuntimeEnvironmentTests.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Linq;
+using Microsoft.AspNet.Testing.xunit;
 using Microsoft.Dnx.Runtime;
 using Xunit;
 
@@ -53,6 +55,76 @@ namespace dnx.hostTests
             var runtime = Type.GetType("Mono.Runtime") == null ? "CLR" : "Mono";
             Assert.Equal(runtime, runtimeEnv.RuntimeType);
 #endif
+        }
+
+        // Test RID generation
+        [Theory]
+        [InlineData("Windows", "6.1.1234", "x86", "win7-x86")] // 1234 => We only care about major and minor
+        [InlineData("Windows", "6.1.1234", "x64", "win7-x64")]
+        [InlineData("Windows", "6.2.1234", "x86", "win8-x86")]
+        [InlineData("Windows", "6.2.1234", "x64", "win8-x64")]
+        [InlineData("Windows", "6.3.1234", "x86", "win81-x86")]
+        [InlineData("Windows", "6.3.1234", "x64", "win81-x64")]
+        [InlineData("Windows", "10.0.1234", "x86", "win10-x86")]
+        [InlineData("Windows", "10.0.1234", "x64", "win10-x64")]
+        [InlineData("Windows", "10.0.1234", "arm", "win10-arm")]
+        [InlineData("Linux", "", "x86", "linux-x86")]
+        [InlineData("Linux", "", "x64", "linux-x64")]
+        [InlineData("Linux", "", "arm", "linux-arm")]
+        [InlineData("Darwin", "", "x86", "darwin-x86")]
+        [InlineData("Darwin", "", "x64", "darwin-x64")]
+        [InlineData("Darwin", "", "arm", "darwin-arm")]
+        public void RuntimeIdIsGeneratedCorrectly(string osName, string version, string architecture, string expectedRid)
+        {
+            var runtimeEnv = new DummyRuntimeEnvironment()
+            {
+                OperatingSystem = osName,
+                OperatingSystemVersion = version,
+                RuntimeArchitecture = architecture
+            };
+            Assert.Equal(expectedRid, runtimeEnv.GetRuntimeIdentifier());
+        }
+
+        // Test live runtime ids
+        [ConditionalFact]
+        [OSSkipCondition(OperatingSystems.Linux | OperatingSystems.MacOSX)]
+        public void WindowsRuntimeIdIsCorrect()
+        {
+            // Verifying OS version is difficult in a test
+
+            var expectedArch = RuntimeEnvironmentHelper.RuntimeEnvironment.RuntimeArchitecture.ToLowerInvariant();
+            var rid = RuntimeEnvironmentHelper.RuntimeEnvironment.GetRuntimeIdentifier();
+            var osName = new string(rid.TakeWhile(c => !char.IsDigit(c)).ToArray());
+            var arch = rid.Split('-')[1];
+            Assert.Equal("win", osName);
+            Assert.Equal(expectedArch, arch);
+        }
+
+        [ConditionalFact]
+        [OSSkipCondition(OperatingSystems.Linux)]
+        [FrameworkSkipCondition(RuntimeFrameworks.CLR | RuntimeFrameworks.CoreCLR)] // We don't have an OS skip condition for all Windows yet
+        public void MacRuntimeIdIsCorrect()
+        {
+            var arch = RuntimeEnvironmentHelper.RuntimeEnvironment.RuntimeArchitecture.ToLowerInvariant();
+            Assert.Equal($"darwin-{arch}", RuntimeEnvironmentHelper.RuntimeEnvironment.GetRuntimeIdentifier());
+        }
+
+        [ConditionalFact]
+        [OSSkipCondition(OperatingSystems.MacOSX)]
+        [FrameworkSkipCondition(RuntimeFrameworks.CLR | RuntimeFrameworks.CoreCLR)] // We don't have an OS skip condition for all Windows yet
+        public void LinuxRuntimeIdIsCorrect()
+        {
+            var arch = RuntimeEnvironmentHelper.RuntimeEnvironment.RuntimeArchitecture.ToLowerInvariant();
+            Assert.Equal($"linux-{arch}", RuntimeEnvironmentHelper.RuntimeEnvironment.GetRuntimeIdentifier());
+        }
+
+        private class DummyRuntimeEnvironment : IRuntimeEnvironment
+        {
+            public string OperatingSystem { get; set; }
+            public string OperatingSystemVersion { get; set; }
+            public string RuntimeArchitecture { get; set; }
+            public string RuntimeType { get; set; }
+            public string RuntimeVersion { get; set; }
         }
     }
 }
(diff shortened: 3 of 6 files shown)


davidfowl:
Travis is failing

src/Microsoft.Dnx.Runtime.Abstractions/Constants.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Dnx.Runtime
```
davidfowl: We shouldn't expose anything like this publicly. It's an implementation detail that shouldn't be exposed. Especially until we're sure of these values.

davidfowl: Move this to where the other sets of internal constants are.

src/Microsoft.Dnx.Runtime.Abstractions/IRuntimeEnvironment.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.Dnx.Runtime
     public interface IRuntimeEnvironment
```
davidfowl: You can unto the changes here :smile:

anurse:
yep, on it

src/dnx.windows/dnx.windows.vcxproj
```diff
@@ -30,6 +30,9 @@
   <PropertyGroup>
     <IncludePath Condition="'$(RuntimeTargetOS)' == 'WIN7_PLUS_CORESYSTEM'">$(ProgramFiles)\Windows Kits\Win8CoreSystem\inc\crt;$(IncludePath)</IncludePath>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
```
davidfowl: Only `debug|win32`?

src/Microsoft.Dnx.Runtime.Abstractions/Constants.cs
anurse: Ok, but we do already expose the values. These aren't RIDs, they are the values we expose publicly on `IRuntimeEnvironment`

src/dnx.windows/dnx.exe.manifest
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8" standalone="yes"?>
```
moozzyk: Would we need the manifest if we used `IsWindowsXOrGreater` API family?

src/Microsoft.Dnx.Runtime.Abstractions/Constants.cs
```diff
@@ -0,0 +1,29 @@
...
+        public static readonly string CLR = nameof(CLR);
+        public static readonly string Mono = nameof(Mono);
+    }
+
+    public static class RuntimeArchitectures
+    {
+        public static readonly string X86 = nameof(X86);
+        public static readonly string X64 = nameof(X64);
+        public static readonly string Arm = nameof(Arm);
+        public static readonly string Arm64 = nameof(Arm64);
```
moozzyk: At the moment we don't distinguish between arm and arm64 - I am not sure if we will in the future...

src/Microsoft.Dnx.Runtime.Abstractions/Constants.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
```
moozzyk: These usings are not needed. 
Also, waiting for @troydai to point out that there is no license header.

src/dnx.windows/dnx.windows.vcxproj
anurse: That's what I get for editing in VS :). Will fix

anurse: This section isn't even needed, it's just because I toggled something that defaulted to true

moozzyk: These vcxproj files are "carefully handcrafted artifacts" in order to keep the same settings for all native projects (and to remove duplication included by VS) - so you typically want to fix what VS adds.

src/dnx.windows/dnx.exe.manifest
anurse: I believe so, but I'm not sure. Also those are Macros, so they'd have to be added in the native code. I think that would be a viable option once we get closer to having native code pass data to managed code.

src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
```diff
@@ -27,5 +30,45 @@ public static string GetShortVersion(this IRuntimeEnvironment env)
...
+
+            string os = env.OperatingSystem;
+            string ver = env.OperatingSystemVersion;
+            if (string.Equals(os, RuntimeOperatingSystems.Windows, StringComparison.Ordinal))
+            {
+                os = "win";
+
+                // Convert 6.x to the correct branding version for the RID
+                var parsedVersion = Version.Parse(ver);
+                if(parsedVersion.Major == 6 && parsedVersion.Minor == 1)
```
moozzyk: I actually thought would come from native code where I would use IsVersionXOrHiger API to get the version instead of passing what GetVersionEx returns...

src/Microsoft.Dnx.Runtime.Internals/Constants.cs
```diff
@@ -0,0 +1,27 @@
...
+        public static readonly string X64 = "x64";
+        public static readonly string Arm = "arm";
+        public static readonly string Arm64 = "arm64";
+    }
+
+    internal static class RuntimeOperatingSystems
+    {
+        public static readonly string Windows = nameof(Windows);
+        public static readonly string Linux = nameof(Linux);
+        public static readonly string Darwin = nameof(Darwin);
```
moozzyk: What are you thoughts about other OSes not on this list. I actually thought of using uname on non-windows and pass it here otherwise we will get Linux for FreeBSD...

test/Microsoft.Dnx.Host.Tests/RuntimeEnvironmentTests.cs
```diff
@@ -54,5 +56,59 @@ public void RuntimeEnvironment_RuntimeType()
             Assert.Equal(runtime, runtimeEnv.RuntimeType);
 #endif
         }
+
+        // Test RID generation
+        [Theory]
+        [InlineData()]
```
davidfowl: ?

src/dnx.windows/dnx.windows.vcxproj
anurse: Small-batch locally-sourced vcxproj files?

src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
```diff
@@ -27,5 +30,45 @@ public static string GetShortVersion(this IRuntimeEnvironment env)
...
+                    ver = "81";
+                }
+                else if(parsedVersion.Major == 10 && parsedVersion.Minor == 0)
+                {
+                    ver = "10";
+                }
+            }
+            else
+            {
+                os = os.ToLower(); // Just use the lower-case full name of the OS as the RID OS
```
davidfowl: This will end up being wrong on certain distros right? I guess we're assuming that when @moozzyk makes the changes to flow the relevant information,  this will fix itself right?

src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
anurse: I'm totally OK with adjusting it that way when you do the native->managed stuff. I didn't want to dive too far into native code right now and the manifest seemed like a quicker fix. I could remove the manifest and we just wouldn't be able to actually see `win81` and `win10` RIDs until your changes :). The manifest seems like a Good Thing generally though (however, I'm no expert there)

src/Microsoft.Dnx.Runtime.Internals/Constants.cs
anurse: The non-Windows constants aren't actually used. They're left over from earlier in the PR when this was public. I'm trimming them now. We do use the output of `uname` on non-Windows

test/Microsoft.Dnx.Host.Tests/RuntimeEnvironmentTests.cs
anurse: fixing

src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
anurse: yes, right now we only ship one Linux DNX. That will probably change for beta8 and when that changes we'll have to solve this part as well.

test/Microsoft.Dnx.Host.Tests/RuntimeEnvironmentTests.cs
```diff
@@ -54,5 +56,72 @@ public void RuntimeEnvironment_RuntimeType()
...
+        [InlineData("Windows", "6.2.1234", "x86", "win8-x86")]
+        [InlineData("Windows", "6.2.1234", "x64", "win8-x64")]
+        [InlineData("Windows", "6.3.1234", "x86", "win81-x86")]
+        [InlineData("Windows", "6.3.1234", "x64", "win81-x64")]
+        [InlineData("Windows", "10.0.1234", "x86", "win10-x86")]
+        [InlineData("Windows", "10.0.1234", "x64", "win10-x64")]
+        [InlineData("Linux", "", "x86", "linux-x86")]
+        [InlineData("Linux", "", "x64", "linux-x64")]
+        [InlineData("Darwin", "", "x86", "darwin-x86")]
+        [InlineData("Darwin", "", "x64", "darwin-x64")]
```
moozzyk: arm :)

src/Microsoft.Dnx.Runtime.Internals/Constants.cs
```diff
@@ -0,0 +1,23 @@
...
+        public static readonly string CoreCLR = nameof(CoreCLR);
+        public static readonly string CLR = nameof(CLR);
+        public static readonly string Mono = nameof(Mono);
+    }
+
+    internal static class RuntimeArchitectures
+    {
+        public static readonly string X86 = "x86";
+        public static readonly string X64 = "x64";
+    }
```
moozzyk: + arm

src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
moozzyk: I am planning to start working on #1970 this week.

test/Microsoft.Dnx.Host.Tests/RuntimeEnvironmentTests.cs
```diff
@@ -54,5 +56,59 @@ public void RuntimeEnvironment_RuntimeType()
...
+                OperatingSystem = osName,
+                OperatingSystemVersion = version,
+                RuntimeArchitecture = architecture
+            };
+            Assert.Equal(expectedRid, runtimeEnv.GetRuntimeIdentifier());
+        }
+
+        // Test live runtime ids
+        [ConditionalFact]
+        [OSSkipCondition(OperatingSystems.Linux | OperatingSystems.MacOSX | OperatingSystems.Win2008R2)]
```
victorhurdugaci: Why skip on Win2008 ?

test/Microsoft.Dnx.Host.Tests/RuntimeEnvironmentTests.cs
anurse: We don't need it here yet, the managed code doesn't support ARM yet (see https://github.com/aspnet/dnx/issues/2063)

src/Microsoft.Dnx.Runtime.Internals/Constants.cs
anurse: We don't need it here yet, the managed code doesn't support ARM yet (see https://github.com/aspnet/dnx/issues/2063)

src/dnx.windows/dnx.exe.manifest
```diff
@@ -0,0 +1,19 @@
...
+<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1" xmlns:asmv3="urn:schemas-microsoft-com:asm.v3">
+    <!-- Compatibility manifests so we get the right behavior out of APIs -->
+    <!-- See: https://msdn.microsoft.com/en-us/library/windows/desktop/dn481241(v=vs.85).aspx -->
+    <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1"> 
+        <application> 
+            <!-- Windows 10 -->
```
moozzyk: Adding the manifest will change how we treat dnx on windows 10 - before it was a win32 app and now it is onecore because of [this check](https://github.com/aspnet/dnx/blob/dev/src/dnx.windows/dnx.cpp#L16). I think it should be fine. One caveat - people who don't have Windows10 SDK installed won't be able to run dnx they built on their boxes if they are on Windows 10 since they wont have dnx.onecore.dll.

src/Microsoft.Dnx.Runtime.Internals/Constants.cs
moozzyk: Fair enough...

test/Microsoft.Dnx.Host.Tests/RuntimeEnvironmentTests.cs
```diff
@@ -54,5 +56,75 @@ public void RuntimeEnvironment_RuntimeType()
...
+        [InlineData("Windows", "6.3.1234", "x64", "win81-x64")]
+        [InlineData("Windows", "10.0.1234", "x86", "win10-x86")]
+        [InlineData("Windows", "10.0.1234", "x64", "win10-x64")]
+        [InlineData("Windows", "10.0.1234", "arm", "win10-arm")]
+        [InlineData("Linux", "", "x86", "linux-x86")]
+        [InlineData("Linux", "", "x64", "linux-x64")]
+        [InlineData("Linux", "", "arm", "linux-arm")]
+        [InlineData("Darwin", "", "x86", "darwin-x86")]
+        [InlineData("Darwin", "", "x64", "darwin-x64")]
+        [InlineData("Darwin", "", "arm", "darwin-arm")]
```
moozzyk: spiteful...

src/dnx.windows/dnx.exe.manifest
moozzyk: Looks like the manifest will always be needed.

moozzyk:
:ship::it:

test/Microsoft.Dnx.Host.Tests/RuntimeEnvironmentTests.cs
anurse: I suppose we don't support iOS do we :). I'll remove that one.

moozzyk: Linux and Darwin are only x64 now. If/when we have Core CLR for Linux on arm we can support it too (I was able to compile our bootstrapper on the original Raspberry Pi but did not have a CoreClr I could use).

anurse: Right, but this code is only testing that given an OS, Version and Architecture, we can synthesize a runtime ID correctly. So it doesn't have to be limited to what we support now.

src/dnx.windows/dnx.exe.manifest
anurse: Hmm... that's not great. Is that really the only way we have to determine if we're running on OneCore? If so, I'd be open to removing the Windows 10 entry from here for the short-term while we work out the full plan.

src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
anurse: :+1:

Removing the manifest still allows me to proceed with the rest of the changes (since I can still use win7/win8/linux/darwin as pivots, just not win81 and win10), so I think I'll do that just to ensure we don't mess things up too much. We just need to ensure #1970 handles it :)

anurse:
I'm going to remove the manifest after seeing @moozzyk 's comments. We will still need to be able to detect Windows 8.1 and 10 before we can call #2387 done, but we can proceed with the rest without the manifest.

davidfowl:
:shipit: After making the changes @moozzyk suggests

src/dnx.windows/dnx.exe.manifest
moozzyk: This was the simplest way I found back then and I don't know if there is a different way. On the other hand Windows 10 is onecore too. The code between onecore and win32 is the same. The only difference is what we link against.

Issue #2214, opened 2015-07-08
Fix exception messages related to project.json not being found


kichalla:
The behavior varies differently for different commands...we shouldn't be showing exception details if project.json is not found and also notice a Null ref

![image](https://cloud.githubusercontent.com/assets/6559784/8583483/d3d6e20c-2582-11e5-8768-3df5deebd325.png)

PR #2598, opened 2015-08-31
Don't throw from TryGetProject when it cannot get the project

victorhurdugaci:
Fixes https://github.com/aspnet/dnx/issues/2214

`TryGetProject` was throwing when the project file was not found. Probably a recent regression.

Please review @BrennanConroy @anurse or @davidfowl


davidfowl:
Why is there no test here? I'm not sure exactly what you're fixing or if it's even fixed by this change or what other things it affects (DTH, build etc)

victorhurdugaci:
Updated

davidfowl:
:shipit:

PR #2608, opened 2015-09-01
Stopping using DNX_DEFAULT_LIB to pass the path to the runtime folder

moozzyk:
Note: DNX_DEFAULT_LIB is also being used to resolve search paths so it cannot be completely removed until we have an alternate way to pass the value.

#1970


davidfowl:
Does this break helios?

moozzyk:
It shouldn't - from what I checked Helios did not use DNX_DEFAULT_LIB. (The next couple of changes will most likely will though)

davidfowl:
What about mono?

davidfowl:
:shipit: but as you said it's still incomplete.

moozzyk:
Mono is not affected by this change. The second part of this is about changing how we pass things to the bootstrapper in general and this will be where Mono will be affected.

Issue #2542, opened 2015-08-22
Add portable PDB support to dnu pack/publish (and runtime)


davidfowl:
This is to enable xplat debugging. For now it will be emitted when an environment variable is set. We need to figure out if we ever use this format by default.

PR #2609, opened 2015-09-02
Add support for generating portable pdbs

davidfowl:
- Added support for emitting portable PDBs when the pdb generation isn't available
- Added support for env variable DNX_BUILD_PORTABLE_PDB to forcefully turn on ppdb generation

#2542

/cc @tmat 

@pranavkm we need to do something similar for Razor compilation.

### DIFF
--- a/src/Microsoft.Dnx.Compilation.CSharp/RoslynProjectReference.cs
+++ b/src/Microsoft.Dnx.Compilation.CSharp/RoslynProjectReference.cs
@@ -98,15 +98,11 @@ namespace Microsoft.Dnx.Compilation.CSharp
 
                     var sw = Stopwatch.StartNew();
 
-                    if (_supportsPdbGeneration.Value)
-                    {
-                        emitResult = CompilationContext.Compilation.Emit(assemblyStream, pdbStream: pdbStream, manifestResources: resources);
-                    }
-                    else
-                    {
-                        Logger.TraceWarning("PDB generation is not supported on this platform");
-                        emitResult = CompilationContext.Compilation.Emit(assemblyStream, manifestResources: resources);
-                    }
+                    var emitOptions = GetEmitOptions();
+                    emitResult = CompilationContext.Compilation.Emit(assemblyStream,
+                                                                     pdbStream: pdbStream,
+                                                                     manifestResources: resources,
+                                                                     options: emitOptions);
 
                     sw.Stop();
 
@@ -201,28 +197,14 @@ namespace Microsoft.Dnx.Compilation.CSharp
 
                 var sw = Stopwatch.StartNew();
 
-                EmitResult emitResult = null;
-
-                if (_supportsPdbGeneration.Value)
-                {
-                    var options = new EmitOptions(pdbFilePath: pdbPath);
-                    emitResult = CompilationContext.Compilation.Emit(
-                        assemblyStream,
-                        pdbStream: pdbStream,
-                        xmlDocumentationStream: xmlDocStream,
-                        win32Resources: win32resStream,
-                        manifestResources: resources,
-                        options: options);
-                }
-                else
-                {
-                    Logger.TraceWarning("PDB generation is not supported on this platform");
-                    emitResult = CompilationContext.Compilation.Emit(
-                        assemblyStream,
-                        xmlDocumentationStream: xmlDocStream,
-                        manifestResources: resources,
-                        win32Resources: win32resStream);
-                }
+                var emitOptions = GetEmitOptions().WithPdbFilePath(pdbPath);
+                var emitResult = CompilationContext.Compilation.Emit(
+                    assemblyStream,
+                    pdbStream: pdbStream,
+                    xmlDocumentationStream: xmlDocStream,
+                    win32Resources: win32resStream,
+                    manifestResources: resources,
+                    options: emitOptions);
 
                 sw.Stop();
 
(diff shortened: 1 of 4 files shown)


tmat:
:)

src/Microsoft.Dnx.Compilation.CSharp/RoslynProjectReference.cs
```diff
@@ -98,15 +98,11 @@ public Assembly Load(AssemblyName assemblyName, IAssemblyLoadContext loadContext
 
                     var sw = Stopwatch.StartNew();
 
-                    if (_supportsPdbGeneration.Value)
```
victorhurdugaci: I think you can remove `_supportsPdbGeneration` completely

victorhurdugaci:
Does this work everywhere including OneCore (Nano)?

src/Microsoft.Dnx.Compilation.CSharp/RoslynProjectReference.cs
victorhurdugaci: Nevermind. Keep it there, I just noticed it is still used

src/Microsoft.Dnx.Compilation.CSharp/RoslynProjectReference.cs
```diff
@@ -293,6 +271,26 @@ public DiagnosticResult EmitAssembly(string outputPath)
...
+            var usePortablePdbString = Environment.GetEnvironmentVariable(EnvironmentNames.PortablePdb);
+
+            // Use portable pdbs if explicitly specified or the platform doesn't support pdb genertion
+            var usePortablePdb = !_supportsPdbGeneration.Value ||
+                                 string.Equals(usePortablePdbString, "true", StringComparison.OrdinalIgnoreCase) ||
+                                 string.Equals(usePortablePdbString, "1", StringComparison.OrdinalIgnoreCase);
+
+            if (usePortablePdb)
+            {
+                Logger.TraceInformation("Using portable pdb format");
```
victorhurdugaci: Verbose maybe?

src/Microsoft.Dnx.Compilation.CSharp/RoslynProjectReference.cs
```diff
@@ -293,6 +271,26 @@ public DiagnosticResult EmitAssembly(string outputPath)
...
+        private EmitOptions GetEmitOptions()
+        {
+            var emitOptions = new EmitOptions();
+
+            var usePortablePdbString = Environment.GetEnvironmentVariable(EnvironmentNames.PortablePdb);
+
+            // Use portable pdbs if explicitly specified or the platform doesn't support pdb genertion
+            var usePortablePdb = !_supportsPdbGeneration.Value ||
+                                 string.Equals(usePortablePdbString, "true", StringComparison.OrdinalIgnoreCase) ||
+                                 string.Equals(usePortablePdbString, "1", StringComparison.OrdinalIgnoreCase);
```
victorhurdugaci: I'm sure someone could use an upper case "1" :trollface:

victorhurdugaci: Serious now: we should have a helper for these flag environment variables rather that doing `string.Equals(... true, 1)` etc

victorhurdugaci:
Other than the minor comments, looks good :shipit:

victorhurdugaci:
Oh, one more thing: how about a test for the non Windows platforms where pdb generation is not supported to check that we emit pdbs now?

src/Microsoft.Dnx.Compilation.CSharp/RoslynProjectReference.cs
```diff
@@ -293,6 +271,26 @@ public DiagnosticResult EmitAssembly(string outputPath)
             }
         }
 
+        private EmitOptions GetEmitOptions()
+        {
+            var emitOptions = new EmitOptions();
+
+            var usePortablePdbString = Environment.GetEnvironmentVariable(EnvironmentNames.PortablePdb);
+
+            // Use portable pdbs if explicitly specified or the platform doesn't support pdb genertion
```
BrennanConroy: generation

BrennanConroy:
Looks good :shipit:

src/Microsoft.Dnx.Compilation.CSharp/RoslynProjectReference.cs
davidfowl: File a bug for that

davidfowl:
> Does this work everywhere including OneCore (Nano)?

What does "work" mean? Some version of CoreCLR will support it natively but I don't know which version :smile: and it will emit on nano since there's no pdb support there.

> Oh, one more thing: how about a test for the non Windows platforms where pdb generation is not supported to check that we emit pdbs now?

Will do.

src/Microsoft.Dnx.Compilation.CSharp/RoslynProjectReference.cs
davidfowl: Nah it's used elsewhere.

src/Microsoft.Dnx.Compilation.CSharp/RoslynProjectReference.cs
davidfowl: We don't have verbose

davidfowl:
I cant reproduce the travis failure on my linux vm. It's also just timing out so not sure whats up.

muratg:
@davidfowl Do not push  before Travis passes.

davidfowl:
@muratg yes I know how to ci. Mono has been trolling me for hours now.

src/Microsoft.Dnx.Compilation.CSharp/RoslynProjectReference.cs
```diff
@@ -293,6 +271,26 @@ public DiagnosticResult EmitAssembly(string outputPath)
...
+        private EmitOptions GetEmitOptions()
+        {
+            var emitOptions = new EmitOptions();
+
+            var usePortablePdbString = Environment.GetEnvironmentVariable(EnvironmentNames.PortablePdb);
+
+            // Use portable pdbs if explicitly specified or the platform doesn't support pdb generation
+            var usePortablePdb = !_supportsPdbGeneration.Value ||
+                                 string.Equals(usePortablePdbString, "true", StringComparison.OrdinalIgnoreCase) ||
+                                 string.Equals(usePortablePdbString, "1", StringComparison.OrdinalIgnoreCase);
```
anurse: Do we do this double-check thing anywhere else? Everywhere else I've seen only accepts "1" as a value. I'd rather just be clear on what the true value is than try to be unnecessarily flexible :)

anurse:
Once travis is passing, :shipit:

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
```diff
@@ -164,6 +165,51 @@ public void DnuPack_FrameworkSpecified(string flavor, string os, string architec
 
         [Theory]
         [MemberData(nameof(RuntimeComponents))]
+        public void DnuPack_PdbsGeneratedOnAllPlatforms(string flavor, string os, string architecture)
```
halter73: Is there a way to test that setting `DNX_BUILD_PORTABLE_PDB` creates a portable pdb when `SupportsPdbGeneration` returns true.

halter73:
Why create non-portable PDBs? Just curious.

davidfowl:
Closing this out for now until I get more time to investigate why mono is hanging.

PR #2620, opened 2015-09-02
Add MONO_THREADS_PER_CPU to dnu.sh to workaround restore timeouts on Mono

akoeplinger:
This is troubling a lot of people and I honestly don't expect it to be fixed for Mono 4.2.
Adding the workaround to dnu.sh seems like a better solution for now instead of telling everyone to add it themselves.

### DIFF
--- a/scripts/dnu.sh
+++ b/scripts/dnu.sh
@@ -8,5 +8,7 @@ while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symli
 done
 DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
 
+# work around restore timeouts on Mono
+[ -z "$MONO_THREADS_PER_CPU" ] && export MONO_THREADS_PER_CPU=50
 
 exec "$DIR/dnx" "$DIR/lib/Microsoft.Dnx.Tooling/Microsoft.Dnx.Tooling.dll" "$@"


scripts/dnu.sh
```diff
@@ -8,5 +8,7 @@ while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symli
 done
 DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
 
+# work around restore timeouts on Mono
+if [[ "$@" == "restore"* ]]; then export MONO_THREADS_PER_CPU=50; fi
```
davidfowl: I'd rather always do it

akoeplinger: @davidfowl yeah, I did it this way to avoid impacting other commands like build, but I guess it doesn't really matter. Changed it.

scripts/dnu.sh
```diff
@@ -8,5 +8,7 @@ while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symli
 done
 DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
 
+# work around restore timeouts on Mono
+export MONO_THREADS_PER_CPU=50
```
davidfowl: Is 50 enough?

davidfowl: Also one more thing, we need to not override this if people have already set it.

Issue #2607, opened 2015-09-01
[osx] missing native libraries in latest build


natemcmaster:
`dnu restore` with 1.0.0-beta8-15548 throws an error on OS X

>    System.TypeInitializationException: The type initializer for 'System.Net.Http.CurlHandler' threw an exception. ---> System.TypeInitializationException: The type initializer for 'libcurl' threw an exception. ---> System.TypeInitializationException: The type initializer for 'HttpInitializer' threw an exception. ---> System.TypeInitializationException: The type initializer for 'CryptoInitializer' threw an exception. ---> System.DllNotFoundException: Unable to load DLL 'System.Security.Cryptography.Native': The specified module could not be found.

Missing ~~libcurl~~ and System.Security.Cryptography.Native

Looks like this error started in 15545
https://github.com/aspnet/dnx/issues/1793#issuecomment-136835107

davidfowl:
Did you install lib curl?

BrennanConroy:
Looks like we need to copy System.Net.Http.Native.so and System.Security.Cryptography.Native.so to the Microsoft.Dnx.Tooling folder in the dnx

natemcmaster:
@davidfowl Update: libcurl was already installed. It's because DNX is missing System.Security.Cryptography.Native.dylib

BrennanConroy:
fc105f626e1a04511918143644ade299d7b60b38

xasx:
I am seeing similar exceptions on Linux when trying out the HelloWorld sample.

```
$ dnu build --framework dnxcore50
Microsoft .NET Development Utility CoreCLR-x64-1.0.0-beta8-15585


Building helloworld-dnxcore for DNXCore,Version=v5.0
System.TypeInitializationException: The type initializer for 'libcrypto' threw an exception. ---> System.TypeInitializationException: The type initializer for 'CryptoInitializer' threw an exception. ---> System.DllNotFoundException: Unable to load DLL 'System.Security.Cryptography.Native': The specified module could not be found.
 (Exception from HRESULT: 0x8007007E)
   at Interop.CryptoInitializer.EnsureOpenSslInitialized()
   at Interop.CryptoInitializer..cctor()
   --- End of inner exception stack trace ---
   at Interop.CryptoInitializer.Initialize()
   at Interop.libcrypto..cctor()
   --- End of inner exception stack trace ---
   at Interop.libcrypto.EVP_sha1()
[...]

PR #2621, opened 2015-09-02
Move some native assemblies to Tooling for System.Net.Http

BrennanConroy:
@moozzyk @davidfowl

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -516,6 +516,13 @@ var CAN_BUILD_ONECORE = '${Directory.Exists(VS_ONECORE_ROOT) && Directory.GetFil
             "Microsoft.Dnx.Project",
             "Microsoft.Dnx.Tooling"
         };
+        
+        // native assemblies loaded by System.Net.Http.dll at runtime
+        var nativeLibraries = new []
+        {
+            "System.Net.Http.Native",
+            "System.Security.Cryptography.Native"
+        };
 
         foreach (var target in RUNTIME_TARGETS)
         {
@@ -613,6 +620,20 @@ var CAN_BUILD_ONECORE = '${Directory.Exists(VS_ONECORE_ROOT) && Directory.GetFil
                     var targetFilePath = Path.Combine(target.TargetFolder, Path.GetFileName(file));
                     Log.Info("Copying " + file + " to " + targetFilePath);
                     File.Copy(file, targetFilePath, true);
+                }
+                
+                if (target.OS != "win")
+                {
+                    foreach (var native in nativeLibraries)
+                    {
+                        var targetFilePath = Path.Combine(target.TargetFolder, "lib", "Microsoft.Dnx.Tooling", native + nativeExtension);
+                        var sourceFile = Path.Combine(target.TargetFolder, native + nativeExtension);
+                        if (File.Exists(sourceFile))
+                        {
+                            Log.Info("Moving " + native + " to " + targetFilePath);
+                            File.Move(Path.Combine(target.TargetFolder, native + nativeExtension), targetFilePath);
+                        }
+                    }
                 }
 
                 // Copy the managed resources the runtime needs for Linux


makefile.shade
```diff
@@ -516,6 +516,13 @@ var CAN_BUILD_ONECORE = '${Directory.Exists(VS_ONECORE_ROOT) && Directory.GetFil
             "Microsoft.Dnx.Project",
             "Microsoft.Dnx.Tooling"
         };
+        
+        // native assemblies loaded by System.Net.Http.dll at runtime
+        var httpNativeAssemblies = new []
```
davidfowl: Call it native libraries

makefile.shade
```diff
@@ -611,6 +618,18 @@ var CAN_BUILD_ONECORE = '${Directory.Exists(VS_ONECORE_ROOT) && Directory.GetFil
                          || (Path.GetFileName(f).StartsWith("System") && Path.GetFileName(f).EndsWith(nativeExtension))))
                 {
                     var targetFilePath = Path.Combine(target.TargetFolder, Path.GetFileName(file));
+                    if (file.EndsWith(nativeExtension))
+                    {
+                        foreach (var native in httpNativeAssemblies)
```
moozzyk: Why not just blindly copy the files from httpNativeAssemblies instead of checking if the names match for each so/dylib?

davidfowl:
:shipit: When the build passes.

makefile.shade
```diff
@@ -614,6 +621,13 @@ var CAN_BUILD_ONECORE = '${Directory.Exists(VS_ONECORE_ROOT) && Directory.GetFil
                     Log.Info("Copying " + file + " to " + targetFilePath);
                     File.Copy(file, targetFilePath, true);
                 }
+                
+                foreach (var native in nativeLibraries)
+                {
+                    var targetFilePath = Path.Combine(target.TargetFolder, "lib", "Microsoft.Dnx.Tooling", native + nativeExtension);
+                    Log.Info("Moving " + native + " to " + targetFilePath);
+                    File.Move(Path.Combine(target.TargetFolder, native + nativeExtension), targetFilePath);
```
davidfowl: Make sure this only happens for core clr only.

makefile.shade
```diff
@@ -614,6 +621,20 @@ var CAN_BUILD_ONECORE = '${Directory.Exists(VS_ONECORE_ROOT) && Directory.GetFil
...
+                if (target.OS != "win")
+                {
+                    foreach (var native in nativeLibraries)
+                    {
+                        var targetFilePath = Path.Combine(target.TargetFolder, "lib", "Microsoft.Dnx.Tooling", native + nativeExtension);
+                        var sourceFile = Path.Combine(target.TargetFolder, native + nativeExtension);
+                        if (File.Exists(sourceFile))
+                        {
+                            Log.Info("Moving " + native + " to " + targetFilePath);
+                            File.Move(Path.Combine(target.TargetFolder, native + nativeExtension), targetFilePath);
```
moozzyk: @BrennanConroy - did you want to use sourceFile in the File.Move? - from what I can see the first arg has the same value...

BrennanConroy: Ya :/

davidfowl: Did you test this @BrennanConroy ?

BrennanConroy: Yep, there isn't a problem, just can replace some of the stuff with the variable to make it look nicer.

davidfowl:
@BrennanConroy close your bugs https://github.com/aspnet/dnx/issues/2607 :smile:

BrennanConroy:
Troll...

makefile.shade
moozzyk: One more (after)thought - shouldn't we fail (or at least warn) if the expected *Native.{so|dylib} - does not exist? If it does not exist it means that either CoreClr will be busted or it is no longer needed and the script should be updated accordingly.

Issue #2622, opened 2015-09-03
Add target framework to project.lock.json for P2P references


davidfowl:
This will allow us to get rid of that logic at runtime.

PR #2623, opened 2015-09-03
Added the relevant target framework for project references in the lockfile

davidfowl:
- Only a restore change for now so we can have project files with
the framework be generated before taking advantage of it in the runtime.

#2622 

/cc @anurse @BillHiebert @jasonmalinowski @yishaigalatzer

### DIFF
--- a/src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
+++ b/src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
@@ -780,7 +780,6 @@ namespace Microsoft.Dnx.Tooling
                                    IEnumerable<TargetContext> contexts)
         {
             var resolver = new DefaultPackagePathResolver(repository.RepositoryRoot.Root);
-            var previousProjectLibraries = previousLockFile?.ProjectLibraries.ToDictionary(l => Tuple.Create(l.Name, l.Version));
             var previousPackageLibraries = previousLockFile?.PackageLibraries.ToDictionary(l => Tuple.Create(l.Name, l.Version));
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
```diff
@@ -862,14 +856,11 @@ private Task<LockFile> ReadLockFile(string projectLockFilePath)
                     var packageInfo = repository.FindPackagesById(library.Name)
                                                 .FirstOrDefault(p => p.Version == library.Version);
 
-                    var projectInfo = projectResolver.FindProject(library.Name);
-
-                    if (projectInfo != null)
+                    var projectDependency = projectResolver.FindProject(library.Name);
```
BrennanConroy: Super nit: add newline here or remove it from same code above (L813)

BrennanConroy:
:shipit:

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPublishTests.cs
```diff
@@ -671,7 +671,8 @@ public void PublishMultipleProjectsWithDifferentTargetFrameworks(string flavor,
   ""targets"": {
     ""DNX,Version=v4.6"": {
       ""Lib/1.0.0"": {
-        ""type"": ""project""
+        ""type"": ""project"",
```
moozzyk: any tests for reading the lock file?

anurse:
:shipit:

davidfowl:
@mozzyz updated the lock file reading test

Issue #2617, opened 2015-09-02
Adding new testing infrastructure


JunTaoLuo:
Building on the work Wei started in https://github.com/davidfowl/DnxTestingPrototype/tree/wei/add-more-representative-cases

PR #2627, opened 2015-09-03
Adding Wei's work on the new testing infrastructure

JunTaoLuo:
#2617 
Adding Wei's work on the new testing infrastructure to dnx:

New testing utilities are located in test/Microsoft.Dnx.Testing.CommonUtils
Sample tests using the new infrastructure are located in test/Microsoft.Dnx.Testing.SampleTests
Test files (dummy solutions and projects) are located in misc/
Tests will leave the solutions/projects used during testing under TestOutput/

Some of the features of the new infrastructure
- Creating projects and solutions is changed. See the sample test in DnuPublishTests.cs for an example.
- Tests can grab preconstructed projects and solutions. See the sample test in DnuRestoreTests.cs for an example.
- dnu calls are now made through DnxSdk.Dnu.<command> where the DnxSdk object specifies the runtime and is passed in to every test.

Possible future work
- Options for saving or deleting the solutions/projects of the tests. Currently all output is saved and this can lead to clutter in the TestOutput/ directory. Ideally, the output should be saved for failing tests only.
- Relocate some of the hardcoded constants to a centralized file in the style of Microsoft.Dnx.Runtime.EnvironmentNames
- Time limits on tests. Currently all calls to Exec.Run will always wait for the command to finish. To prevent a test from hanging forever maybe we should automatically kill off a test after a certain duration.
- Display difference between files that differ from the expected result in DirDiff.cs


JunTaoLuo:
cc @anurse @davidfowl @Tratcher @victorhurdugaci @BrennanConroy @halter73 @moozzyk @troydai

DNX.sln
```diff
@@ -63,13 +63,13 @@ Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.Dnx.Compilation.C
 EndProject
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.Dnx.Runtime.FunctionalTests", "test\Microsoft.Dnx.Runtime.FunctionalTests\Microsoft.Dnx.Runtime.FunctionalTests.xproj", "{5DCA7A59-3B5E-4751-A563-B78201F97AE3}"
 EndProject
-Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "dnx", "src\dnx\dnx.vcxproj", "{D0E2FB09-0FEA-478A-9068-D6AA420C6DED}"
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "dnx.win32", "src\dnx\dnx.vcxproj", "{D0E2FB09-0FEA-478A-9068-D6AA420C6DED}"
```
victorhurdugaci: What's this?

DNX.sln
```diff
@@ -117,6 +117,10 @@ Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.Dnx.DesignTimeHos
 EndProject
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.Dnx.Runtime.Json.Sources", "src\Microsoft.Dnx.Runtime.Json.Sources\Microsoft.Dnx.Runtime.Json.Sources.xproj", "{F88EB033-3BCA-411A-987F-B705569F9317}"
 EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.Dnx.Testing.CommonUtils", "test\Microsoft.Dnx.Testing.CommonUtils\Microsoft.Dnx.Testing.CommonUtils.xproj", "{00E320EE-F8E7-40C2-9334-A02561EDAEAA}"
+EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.Dnx.Testing.SampleTests", "test\Microsoft.Dnx.Testing.SampleTests\Microsoft.Dnx.Testing.SampleTests.xproj", "{88FF8911-526C-4FBE-B4E2-707CDAF1DEC0}"
```
victorhurdugaci: Samples should go under the \samples folder not \test

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
```diff
@@ -0,0 +1,152 @@
+﻿using System;
```
victorhurdugaci: Missing copyright. Please check the other files too

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
```diff
@@ -0,0 +1,152 @@
...
+        {
+            ReadFile = fileInfo => File.ReadAllText(fileInfo.FullName);
+        }
+
+        public Dir(Dir parent)
+        {
+            ReadFile = parent.ReadFile;
+        }
+
+        public Func<FileInfo, object> ReadFile;
```
victorhurdugaci: Ugh, public field...

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
```diff
@@ -0,0 +1,152 @@
...
+            var directory = new DirectoryInfo(path);
+
+            foreach (var subDirectoryInfo in directory.EnumerateDirectories())
+            {
+                Read(this, subDirectoryInfo);
+            }
+
+            foreach (var file in directory.EnumerateFiles())
+            {
+                this[file.Name] = ReadFile(file);
```
victorhurdugaci: I don't like this. We load the file content in memory for **every** file in that directory. That might take quite a lot of space. How about load it on demand?

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
```diff
@@ -0,0 +1,152 @@
...
+            else
+            {
+                Directory.CreateDirectory(Path.GetDirectoryName(path));
+                File.WriteAllText(path, value?.ToString() ?? string.Empty);
+            }
+        }
+
+        private static void Read(Dir parent, DirectoryInfo directoryInfo)
+        {
+            var directory = new Dir(parent);
```
victorhurdugaci: There's a lot of duplicated code here and in `Load`. Can we just use `Read` directly in `Load`?

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
```diff
@@ -0,0 +1,152 @@
...
+
+            foreach (var subDirectoryInfo in directoryInfo.EnumerateDirectories())
+            {
+                Read(directory, subDirectoryInfo);
+            }
+
+            foreach (var file in directoryInfo.EnumerateFiles())
+            {
+                directory[file.Name] = directory.ReadFile(file);
```
victorhurdugaci: Lazy read this. We don't want all files in memory if they're not needed

test/Microsoft.Dnx.Testing.CommonUtils/Directory/DirDiff.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Microsoft.Dnx.Testing
+{
+    public class DirDiff
+    {
+        public IEnumerable<string> MissingEntries { get; set; } = Enumerable.Empty<string>();
+        public IEnumerable<string> ExtraEntries { get; set; } = Enumerable.Empty<string>();
+        public IEnumerable<string> DifferentEntries { get; set; } = Enumerable.Empty<string>();
+
+        public bool IsEmpty
```
victorhurdugaci: I would call this method `Any` or `NoDiff`

test/Microsoft.Dnx.Testing.CommonUtils/DnxFunctionalTestBase.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Linq;
+
+namespace Microsoft.Dnx.Testing
+{
+    public class DnxSdkFunctionalTestBase
+    {
+        public const string SdkVersionForTestingEnvName = "DNX_SDK_VERSION_FOR_TESTING";
+
+        static DnxSdkFunctionalTestBase()
+        {
```
victorhurdugaci: Oh, don't do this. Static constructors are usually a bad idea, you have no idea when this will be invoked

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/Dnu.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        public Dnu(string sdkPath)
+        {
+            _sdkPath = sdkPath;
+        }
+
+        public ExecResult Publish(
+            string projectPath,
+            string outputPath,
+            bool noSource = false,
```
victorhurdugaci: Why is `noSource` special?

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/Dnu.cs
```diff
@@ -0,0 +1,104 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+
+namespace Microsoft.Dnx.Testing
+{
+    public class Dnu
+    {
+        private readonly string _sdkPath;
```
victorhurdugaci: I think all the command methods in this class need extra arguments just like `Publish`

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/Dnu.cs
```diff
@@ -0,0 +1,104 @@
...
+            return Execute(sb.ToString(), envSetup);
+        }
+
+        public ExecResult Restore(
+            string restoreDir,
+            string packagesDir = null,
+            IEnumerable<string> feeds = null)
+        {
+            var sb = new StringBuilder();
+            sb.Append("restore");
```
victorhurdugaci: You can do some of these appends in one line. I think it more clear: `sb.Append($"restore \"{restoreDir}\")`

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/DnxSdk.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        public FrameworkName TargetFramework { get; set; }
+
+        public Dnu Dnu => new Dnu(Path);
+
+        public Dnx Dnx => new Dnx(Path);
+
+        public static string GetRuntimeHome()
+        {
+            var homePath = Environment.ExpandEnvironmentVariables(Environment.GetEnvironmentVariable("DNX_HOME"));
```
victorhurdugaci: I think there are at least 5 places in our dnx code base where we do this. Please try to find at least one of them and reuse it here

test/Microsoft.Dnx.Testing.CommonUtils/Exec.cs
```diff
@@ -0,0 +1,96 @@
...
+
+    public class Exec
+    {
+        public static ExecResult Run(
+            string program,
+            string commandLine,
+            Action<Dictionary<string, string>> envSetup = null,
+            string workingDir = null)
+        {
+            Console.WriteLine($"Running: {program} {commandLine}");
```
victorhurdugaci: Yet another process run helper. Please reuse one of the existing ones

test/Microsoft.Dnx.Testing.CommonUtils/Exec.cs
```diff
@@ -0,0 +1,96 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Text;
+
+namespace Microsoft.Dnx.Testing
+{
+    public class ExecResult
```
victorhurdugaci: 1 class/file

test/Microsoft.Dnx.Testing.CommonUtils/Exec.cs
```diff
@@ -0,0 +1,96 @@
...
+                {
+                    Console.WriteLine(args.Data);
+                    stderrBuilder.AppendLine(args.Data);
+                }
+            };
+
+            process.BeginOutputReadLine();
+            process.BeginErrorReadLine();
+
+            process.WaitForExit();
```
victorhurdugaci: Possible test hang if the process hangs. Always use a timeout and make it configurable.

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
```diff
@@ -0,0 +1,152 @@
...
+
+        public IEnumerable<KeyValuePair<string, object>> Nodes
+        {
+            get
+            {
+                return _nodes;
+            }
+        }
+
+        public object this[params string[] keys]
```
victorhurdugaci: What's the scenario in which you want to set the same content for multiple files? Why not use a loop when you need that?

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
```diff
@@ -0,0 +1,152 @@
...
+            return new DirDiff
+            {
+                ExtraEntries = nodes1.Keys.Except(nodes2.Keys),
+                MissingEntries = nodes2.Keys.Except(nodes1.Keys),
+                DifferentEntries = nodes1.Keys.Intersect(nodes2.Keys)
+                    .Where(entry => !string.Equals(nodes1[entry].ToString(), nodes2[entry].ToString(), StringComparison.Ordinal))
+            };
+        }
+
+        public IEnumerable<KeyValuePair<string, object>> Flatten()
```
victorhurdugaci: Why not return `IDictionary< >` from there? That way you don't need the call to `ToDictionary` above

test/Microsoft.Dnx.Testing.CommonUtils/JsonUtils.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.IO;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Dnx.Testing
+{
+    public class JsonUtils
+    {
+        public static string NormalizeJson(string json)
+        {
+            return JObject.Parse(json).ToString();
```
victorhurdugaci: what's this?

test/Microsoft.Dnx.Testing.CommonUtils/Solution.cs
```diff
@@ -0,0 +1,108 @@
...
+using Microsoft.Dnx.Runtime;
+using ProjectResolver = Microsoft.Dnx.Tooling.ProjectResolver;
+
+namespace Microsoft.Dnx.Testing
+{
+    public class Solution
+    {
+        public Solution(string rootPath)
+        {
+            RootPath = rootPath;
```
victorhurdugaci: do you need to do some sort path existence check or something?

test/Microsoft.Dnx.Testing.CommonUtils/Solution.cs
```diff
@@ -0,0 +1,108 @@
...
+            get
+            {
+                return Path.Combine(RootPath, "wrap");
+            }
+        }
+
+        public Runtime.Project GetProject(string name)
+        {
+            Runtime.Project project;
+            var resolver = new ProjectResolver(RootPath);
```
victorhurdugaci: Cache the resolver?

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/DnxSdk.cs
davidfowl: Not for this PR

DNX.sln
JunTaoLuo: Not sure why but whenever I open VS, it automatically makes this change. I will revert this by hand.

davidfowl: @moozzyk any clues?

DNX.sln
JunTaoLuo: @davidfowl prefers keeping it under \test.

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
JunTaoLuo: Right... oops

test/Microsoft.Dnx.Testing.CommonUtils/Directory/DirDiff.cs
JunTaoLuo: Good idea. Done.

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/Dnu.cs
JunTaoLuo: Yup, that's better.

test/Microsoft.Dnx.Testing.CommonUtils/Exec.cs
JunTaoLuo: Which ones are you referring to? The only similar helper I found was in the current CommonTestUtils project but I believe the idea is to replace the current test infrastructure so we can't use that one.

test/Microsoft.Dnx.Testing.CommonUtils/Exec.cs
JunTaoLuo: Yup.

test/Microsoft.Dnx.Testing.CommonUtils/Exec.cs
JunTaoLuo: Right. I was thinking of adding this in another PR. I was going to take a look at whether to implement this using environment variables, method arguments or some other method. Also what would be a good default value?

davidfowl: @victorhurdugaci do we do this today?

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
JunTaoLuo: My guess is that Wei had this to allow for flexibility in terms of file read behaviour. However, I haven't seen any scenarios where this is required yet. For now I'll remove it and maybe if there is a need for it in the future I'll add it back in.

davidfowl: By remove it you mean private right?

test/Microsoft.Dnx.Testing.CommonUtils/DnxFunctionalTestBase.cs
JunTaoLuo: I'm not sure why Wei implemented it this way but if I were to guess, I would say that this just ensures that the constructor is only called once. An alternative would be to use a private static flag and checking it inside a non-static constructor. Is that better?

DNX.sln
victorhurdugaci: Then why not make it a test?

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
victorhurdugaci: Either make it private or make it a property

test/Microsoft.Dnx.Testing.CommonUtils/DnxFunctionalTestBase.cs
victorhurdugaci: You can use `IUseFixture<T>` http://stackoverflow.com/questions/12976319/xunit-global-setup-teardown

test/Microsoft.Dnx.Testing.CommonUtils/Exec.cs
victorhurdugaci: I don't think we do but we should start doing it. Just pick a big timeout (3-5 minutes) and that should be enough. We just want to make sure that nothing hangs. Kill the process if the timeout expires

victorhurdugaci: You might consider two ways to configure the timeout. There is a default one that might be configurable globally (env var?) so you can tweak it on slow machines and then this method should also take a timeout in case we know that something is taking a long time

davidfowl: I don think we should do this at all. It's just a bad idea. Avoid it unless things actually start hanging.

DNX.sln
JunTaoLuo: I don't think these tests should be official for now. Instead these should serve as a template for porting over the existing set of tests. These should be removed once the old tests are ported over.

test/Microsoft.Dnx.Testing.CommonUtils/DnxFunctionalTestBase.cs
```diff
@@ -0,0 +1,59 @@
+﻿using System;
```
halter73: The filename should be DnxSdkFunctionalTestBase

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/Dnu.cs
JunTaoLuo: Are there any important ones that I should look into right now specifically? I'm not too familiar with all the options and arguments for every command so I was thinking that I will make changes whenever the need arises when porting new tests?

test/Microsoft.Dnx.Testing.CommonUtils/DnxFunctionalTestBase.cs
JunTaoLuo: Yup I fixed this in an upcoming commit.

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/DnxSdk.cs
JunTaoLuo: I didn't find a suitable helper in the current code base. There is something in DnuEnvironment.cs but that seems to be ifdef'ed out for only DNXCORE50 so I'm not sure if that's a good idea.

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
JunTaoLuo: I think it's cleaner to do `["Scripts/bootstrap.js", "Scripts/jquery.js"] = Dir.EmptyFile` for example to create multiple empty files.

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
JunTaoLuo: It makes sense especially since nodes are represented as Dictionaries<string, object> anyways. Not sure why the conversion to IEnumerable<> would ever make sense.

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/Dnu.cs
JunTaoLuo: It's not, especially since noSource is defaulted to false. I can imagine requiring this default argument for options we want to default true but this is not the case here.

test/Microsoft.Dnx.Testing.CommonUtils/Solution.cs
JunTaoLuo: I certainly can. In fact I think the check would also be good to have in Dir. Note that usually a test should not need to create a Solution object itself and should use GetSolution() to grab a pre-constructed solution instead. Though I suppose a test could theoretically construct a solution using Dir (like in DnxPublishTests) and then create a Solution object to represent it, it is sort of redundant in that scenario.

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
JunTaoLuo: Got it.

test/Microsoft.Dnx.Testing.CommonUtils/Exec.cs
JunTaoLuo: Ok

test/Microsoft.Dnx.Testing.CommonUtils/Solution.cs
JunTaoLuo: I'm going to add this as a lazy type.

test/Microsoft.Dnx.Testing.CommonUtils/DnxFunctionalTestBase.cs
JunTaoLuo: Done using ICollectionFixture<T>

global.json
```diff
@@ -3,4 +3,4 @@
         "src",
         "submodules/FileSystem/src"
     ]
-}
\ No newline at end of file
+}
```
davidfowl: Mistake?

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
davidfowl: yep

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/Dnu.cs
davidfowl: We can expand this as we add tests

test/Microsoft.Dnx.Testing.CommonUtils/Exec.cs
davidfowl: yes leave it, this wil be the only one once we transiton

test/Microsoft.Dnx.Testing.CommonUtils/TestUtils.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+            var rootPath = ProjectResolver.ResolveRootDirectory(Directory.GetCurrentDirectory());
+            var originalSolutionPath = Path.Combine(rootPath, "misc", solutionName);            
+            var tempSolutionPath = GetTestFolder<TTest>(sdk, testName);
+            CopyFolder(originalSolutionPath, tempSolutionPath);
+            return new Solution(tempSolutionPath);
+        }
+        public static string GetTempTestFolder<T>(DnxSdk sdk, [CallerMemberName]string testName = null)
+        {
+
```
davidfowl: clean up extra space

test/Microsoft.Dnx.Testing.CommonUtils/TestUtils.cs
```diff
@@ -0,0 +1,96 @@
...
+        public static Solution GetSolution<TTest>(DnxSdk sdk, string solutionName, [CallerMemberName]string testName = null)
+        {
+            var rootPath = ProjectResolver.ResolveRootDirectory(Directory.GetCurrentDirectory());
+            var originalSolutionPath = Path.Combine(rootPath, "misc", solutionName);            
+            var tempSolutionPath = GetTestFolder<TTest>(sdk, testName);
+            CopyFolder(originalSolutionPath, tempSolutionPath);
+            return new Solution(tempSolutionPath);
+        }
+        public static string GetTempTestFolder<T>(DnxSdk sdk, [CallerMemberName]string testName = null)
```
davidfowl: add space between methods

test/Microsoft.Dnx.Testing.SampleTests/DnuRestoreTests.cs
```diff
@@ -0,0 +1,38 @@
...
+        [MemberData(nameof(DnxSdks))]
+        public void DnuRestoreInstallsIndirectDependency(DnxSdk sdk)
+        {
+            // SimpleChain -> DependencyA -> DependencyB
+            const string appName = "SimpleChain";
+            const string solutionName = "DependencyGraphs";
+
+            var solution = TestUtils.GetSolution<DnuRestoreTests>(sdk, solutionName);
+            var project = solution.GetProject(appName);
+            var tempProjectDir = TestUtils.GetTempTestFolder<DnuRestoreTests>(sdk);
```
davidfowl: What is this for?

global.json
JunTaoLuo: I think my vim auto added a new line at the end of the file which was missing. What's the best practice here, new line or no? I will scan the other files for this later.

test/Microsoft.Dnx.Testing.CommonUtils/TestUtils.cs
JunTaoLuo: VS CodeLens is really throwing off my spacing...

test/Microsoft.Dnx.Testing.CommonUtils/TestUtils.cs
JunTaoLuo: Yup, missed this one when cleaning up.

test/Microsoft.Dnx.Testing.SampleTests/DnuRestoreTests.cs
JunTaoLuo: The current implementation requires two copies of the solution. One is used to build the packages and created a feed. The other copy then uses this feed when restoring the SimpleChain project. It might be possible to simplify this though, I'll look into it.

JunTaoLuo:
Note that with the new infrastructure, keep the test names short or else tests may fail on windows due to long path names. Also, before running tests, delete the TestOutput/ folder. In the future, I may change it so that all tests clear their old outputs before running.

Please review.
@anurse @davidfowl @Tratcher @victorhurdugaci @BrennanConroy @halter73 @moozzyk @troydai

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/Dnu.cs
```diff
@@ -0,0 +1,106 @@
...
+                StandardError = result.StandardError,
+                StandardOutput = result.StandardOutput
+            };
+        }
+
+        public ExecResult Execute(
+            string commandLine, 
+            Action<Dictionary<string, string>> envSetup = null)
+        {
+            var dnxPath = Path.Combine(_sdkPath, "bin", "dnu.cmd");
```
BrennanConroy: We need to keep the same behavior as the old Execute for support on non-windows OS's
https://github.com/aspnet/dnx/blob/dev/test/Microsoft.Dnx.CommonTestUtils/DnuTestUtils.cs#L37

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdkFunctionalTestBase.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        public static IEnumerable<object[]> DnxSdks
+        {
+            get
+            {
+                return ClrDnxSdks.Concat(CoreClrDnxSdks);
+            }
+        }
+
+        public static IEnumerable<object[]> ClrDnxSdks
```
BrennanConroy: Where is the mono runtime?

test/Microsoft.Dnx.Testing.CommonUtils/TestUtils.cs
```diff
@@ -0,0 +1,100 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.Linq;
+using System.Runtime.Versioning;
+using Microsoft.Dnx.Runtime;
+using Microsoft.Dnx.Runtime.Helpers;
+using System.Runtime.CompilerServices;
```
BrennanConroy: Sort

test/Microsoft.Dnx.Testing.SampleTests/DnuPublishTests.cs
```diff
@@ -0,0 +1,162 @@
...
+using System.Runtime.Versioning;
+using Microsoft.Dnx.Runtime;
+using Microsoft.Dnx.Runtime.Helpers;
+using Microsoft.Dnx.Tooling.Publish;
+using Newtonsoft.Json.Linq;
+using Xunit;
+
+namespace Microsoft.Dnx.Testing.SampleTests
+{
+    [Collection("SampleTestCollection")]
```
BrennanConroy: nameof

test/Microsoft.Dnx.Testing.SampleTests/DnuPublishTests.cs
```diff
@@ -0,0 +1,162 @@
...
+                    ["DNX,Version=v4.5.1"] = new JArray(),
+                    ["DNXCore,Version=v5.0"] = new JArray()
+                }
+            };
+
+            var expectedOutputWebConfig = $@"<?xml version=""1.0"" encoding=""utf-8""?>
+<configuration>
+  <appSettings>
+    <add key=""{Constants.WebConfigBootstrapperVersion}"" value="""" />
+    <add key=""{Constants.WebConfigRuntimePath}"" value=""..\approot\runtimes"" />
```
BrennanConroy: Use Path.DirectorySeparatorChar for file paths

test/Microsoft.Dnx.Testing.SampleTests/DnuPublishTests.cs
```diff
@@ -0,0 +1,162 @@
+﻿using System.IO;
```
BrennanConroy: Copyright

test/Microsoft.Dnx.Testing.SampleTests/DnuRestoreTests.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System.IO;
```
BrennanConroy: Copyright

test/Microsoft.Dnx.Testing.CommonUtils/TestUtils.cs
JunTaoLuo: Just to check. I see other files have all Microsoft.Dnx.* after System.* is this the convention?

test/Microsoft.Dnx.Testing.SampleTests/DnuPublishTests.cs
JunTaoLuo: oops, forgot to add these in after relocating tests from the samples directory

test/Microsoft.Dnx.Testing.SampleTests/DnuPublishTests.cs
JunTaoLuo: yup

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
```diff
@@ -0,0 +1,148 @@
...
+namespace Microsoft.Dnx.Testing
+{
+    public class Dir
+    {
+        public const string EmptyFile = "";
+
+        private readonly Dictionary<string, object> _nodes = new Dictionary<string, object>();
+        private Func<FileInfo, object> _readFile;
+
+        public Dir(string rootPath): this()
```
victorhurdugaci: nit: usually we add `: this()` on a new line

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
```diff
@@ -0,0 +1,148 @@
...
+                throw new DirectoryNotFoundException();
+            }
+
+            LoadPath = rootPath;
+            Load(rootPath);
+        }
+
+        public Dir()
+        {
+            _readFile = fileInfo => new Lazy<object>(() => File.ReadAllText(fileInfo.FullName));
```
victorhurdugaci: The lazy is an implementation  detail. Don't return it to the caller, just do the lazy part inside this object

test/Microsoft.Dnx.Testing.SampleTests/DnuPublishTests.cs
JunTaoLuo: done

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdkFunctionalTestBase.cs
JunTaoLuo: added

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/Dnu.cs
JunTaoLuo: For now, just add the argument for additionalArguments for each method so the caller can specify additional arguments. I think this is what Victor meant.

test/Microsoft.Dnx.Testing.CommonUtils/TestUtils.cs
```diff
@@ -0,0 +1,99 @@
...
+            foreach (var folderPath in Directory.EnumerateDirectories(sourceFolder))
+            {
+                var folderName = new DirectoryInfo(folderPath).Name;
+                CopyFolder(folderPath, Path.Combine(targetFolder, folderName));
+            }
+        }
+
+        public static string CreateLocalFeed<TTest>(DnxSdk sdk, Solution solution, [CallerMemberName]string testName = null)
+        {
+            var feed = GetTestFolder<TTest>(sdk, $"{testName}.{Path.GetRandomFileName()}");
```
BrennanConroy: Be careful with this, we were having issues with Mac Travis tests and this function. See DisposableDir

test/Microsoft.Dnx.Testing.CommonUtils/Directory/DirDiff.cs
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Dnx.Testing
+{
+    public class DirDiff
+    {
+        public IEnumerable<string> MissingEntries { get; set; } = Enumerable.Empty<string>();
```
victorhurdugaci: Should the setters be private?

test/Microsoft.Dnx.Testing.CommonUtils/Directory/DirMismatchException.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Microsoft.Dnx.Testing
+{
+    public class DirMismatchException : XunitException
+    {
+        public DirMismatchException(string expectedDirPath, string actualDirPath, DirDiff diff)
+        {
+            Message = $@"The actual directory structure '{actualDirPath}' doesn't match expected structure '{expectedDirPath}'.
+Difference information:
+Extra: {string.Join(", ", diff.ExtraEntries)}
```
victorhurdugaci: Probably you should join them with a new line no just comma. Maybe new line and a space so it has a little indentation?

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/Dnu.cs
```diff
@@ -0,0 +1,117 @@
...
+            return Execute(sb.ToString());
+        }
+
+        public ExecResult PackagesAdd(
+            string packagePath,
+            string packagesDir,
+            string additionalArguments = null,
+            Action<Dictionary<string, string>> envSetup = null)
+        {
+            return Execute($"packages add {packagePath} {packagesDir} {additionalArguments}", envSetup);
```
victorhurdugaci: You need quotes around the arguments representing paths for the case where paths have spaces. See that you handle this in other cases too

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/Dnx.cs
```diff
@@ -0,0 +1,27 @@
...
+        public Dnx(string sdkPath)
+        {
+            _sdkPath = sdkPath;
+        }
+
+        public ExecResult Execute(string commandLine, bool dnxTraceOn = true)
+        {
+            var dnxPath = Path.Combine(_sdkPath, "bin", "dnx");
+            return Exec.Run(
+                dnxPath,
```
victorhurdugaci: You need to do the same trick as in the `dnu` case with `cmd /C` on Windows

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/DnxSdk.cs
victorhurdugaci: FYI, the helper is here: https://github.com/aspnet/dnx/blob/219871c6063d00f8297eeafe93266f1048f59a45/src/Microsoft.Dnx.Tooling/Utils/PathUtilities.cs#L35

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdkFunctionalTestBase.cs
```diff
@@ -0,0 +1,61 @@
...
+            }
+        }
+
+        public static IEnumerable<object[]> CoreClrDnxSdks
+        {
+            get
+            {
+                if (RuntimeEnvironmentHelper.IsWindows)
+                {
+                    yield return new[] { DnxSdk.GetRuntime(SdkVersionForTesting, "coreclr", "win", "x86") };
```
victorhurdugaci: Do we need a linux coreclr case? @BrennanConroy  ?

test/Microsoft.Dnx.Testing.CommonUtils/Exec.cs
```diff
@@ -0,0 +1,107 @@
...
+                RedirectStandardOutput = true,
+                RedirectStandardError = true
+            };
+
+            foreach (var pair in env)
+            {
+#if DNX451
+                processStartInfo.EnvironmentVariables[pair.Key] = pair.Value;
+#else
+                processStartInfo.Environment.Add(pair);
```
victorhurdugaci: Will this fail if I already have set one of the environment variables? For example I set `DNX_TRACE` in my console and then I run a test that sets `DNX_TRACE` again. Will this conflict with the inherited variables?

test/Microsoft.Dnx.Testing.CommonUtils/Exec.cs
```diff
@@ -0,0 +1,107 @@
...
+            var stdoutBuilder = new StringBuilder();
+            var stderrBuilder = new StringBuilder();
+
+            process.OutputDataReceived += (sender, args) =>
+            {
+                // If it is not EOF, we always write out a line
+                // This should preserve blank lines
+                if (args.Data != null)
+                {
+                    Console.WriteLine(args.Data);
```
victorhurdugaci: There should be a way to disable console writing. It creates a lot of noise

test/Microsoft.Dnx.Testing.CommonUtils/TestUtils.cs
```diff
@@ -0,0 +1,99 @@
...
+                return FrameworkNameHelper.ParseFrameworkName("dnxcore50");
+            }
+
+            throw new InvalidOperationException($"Unknown runtime flavor '{flavor}'");
+        }
+
+        public static Solution GetSolution<TTest>(
+            DnxSdk sdk, 
+            string solutionName, 
+            [CallerMemberName]string testName = null,
```
victorhurdugaci: You are making an assumption that might not hold. The caller might not be the test method but another method that the test method calls.

test/Microsoft.Dnx.Testing.SampleTests/SampleTestFixture.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Microsoft.Dnx.Testing.SampleTests
+{
+    public class SampleTestFixture : IDisposable
+    {
+        public SampleTestFixture()
+        {
+            Console.WriteLine($@"
+Environment information:
+  DNX_HOME: {Environment.GetEnvironmentVariable("DNX_HOME")}
```
victorhurdugaci: const for the environment variables

test/Microsoft.Dnx.Testing.CommonUtils/TestUtils.cs
davidfowl: That's ok, we can make sure it holds in code reviews. This is the best way we know the get the test name.

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/DnxSdk.cs
```diff
@@ -0,0 +1,88 @@
...
+    {
+        public string Version { get; set; }
+
+        public string Flavor { get; set; }
+
+        public string Architecture { get; set; }
+
+        public string OperationSystem { get; set; }
+
+        public string Path { get; set; }
```
davidfowl: Change this to Location

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdkFunctionalTestBase.cs
```diff
@@ -0,0 +1,61 @@
...
+            {
+                return ClrDnxSdks.Concat(CoreClrDnxSdks);
+            }
+        }
+
+        public static IEnumerable<object[]> ClrDnxSdks
+        {
+            get
+            {
+                if (RuntimeEnvironmentHelper.IsWindows)
```
davidfowl: We should us IsMono for this one.

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdkFunctionalTestBase.cs
davidfowl: After it works :smile:

test/Microsoft.Dnx.Testing.CommonUtils/Exec.cs
davidfowl: Yah, why are we calling Add

test/Microsoft.Dnx.Testing.CommonUtils/Exec.cs
davidfowl: Yes please, add an optional bool

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/DnxSdk.cs
JunTaoLuo: The util is marked as internal so I don't think tests have access to them? Seems like tests are treated as "different" programs.

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
JunTaoLuo: Got it

test/Microsoft.Dnx.Testing.CommonUtils/Directory/DirMismatchException.cs
JunTaoLuo: Good idea

test/Microsoft.Dnx.Testing.CommonUtils/Exec.cs
JunTaoLuo: I'll open another issue for this to deal with the options required. It will include this, how files are saved and possibly more.

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/Dnu.cs
JunTaoLuo: Yup

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/Dnx.cs
JunTaoLuo: Done.

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
JunTaoLuo: Yup

test/Microsoft.Dnx.Testing.SampleTests/SampleTestFixture.cs
JunTaoLuo: This will be done when I deal with all the environment variables in another Issue

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdkFunctionalTestBase.cs
JunTaoLuo: Do you mean in the else block?

davidfowl: Instead of IsWindows

JunTaoLuo: I'm a little confused. Is the expected behavior 
1 use mono runtime for mono environment
2 use clr x86/x64 for all other environments
?

test/Microsoft.Dnx.Testing.CommonUtils/Exec.cs
JunTaoLuo: There shouldn't be an issue since env is a dictionary itself and cannot contain duplicate keys. But to make it more consistent would "processStartInfo.Environment[pair.Key] = pair.Value;" be better?

test/Microsoft.Dnx.Testing.CommonUtils/TestUtils.cs
JunTaoLuo: These temporary folders are created under the dnx root so they should not get into trouble with var - private/var on Mac

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/Dnx.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            string command;
+            if (RuntimeEnvironmentHelper.IsWindows)
+            {
+                command = "cmd";
+                commandLine = $"/C {Path.Combine(_sdkPath, "bin", "dnx.exe")} {commandLine}";
+            }
+            else
+            {
+                command = "bash";
```
davidfowl: I don't think we do this elsewhere. We just run the script because it has the executable bit. We shouldn't be running bash.

davidfowl: This will also break on CoreCLR on linux/osx where we actually have a native executable.

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
```diff
@@ -0,0 +1,173 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
```
troydai: Just a reminder please sort and remove unused `using` statements. (Because these four `using` statements are always automatically added. So I'm just wondering.)

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
```diff
@@ -0,0 +1,173 @@
...
+
+        private readonly Dictionary<string, object> _nodes = new Dictionary<string, object>();
+        private Func<FileInfo, object> _readFile;
+
+        public Dir(string rootPath)
+            : this()
+        {
+            if (!Directory.Exists(rootPath))
+            {
+                throw new DirectoryNotFoundException();
```
troydai: It would be more proper to throw `new ArgumentException("Directory was not found", nameof(rootPath))`

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
```diff
@@ -0,0 +1,173 @@
...
+        {
+            _readFile = fileInfo => new Lazy<object>(() => File.ReadAllText(fileInfo.FullName));
+        }
+
+        public Dir(Dir parent)
+        {
+            _readFile = parent._readFile;
+        }
+
+        public string LoadPath { get; private set; } = "In Memory";
```
troydai: Since it is set in constructor only, you can drop the `private set;`. New C# 6 feature.

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
JunTaoLuo: Yup good idea. I just checked all the files again and all usings are sorted and none of them are redundant.

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
JunTaoLuo: Will do. I'll also check for other occurances of exceptions.

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
JunTaoLuo: Cool

test/Microsoft.Dnx.Testing.CommonUtils/Directory/Dir.cs
troydai: There are plugs for VS can help organizing the using statement for all files.

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/Dnx.cs
JunTaoLuo: I see. Just to make sure, dnx is an executable whereas dnu is a script so this fix should only be made here. Also, should IsMono be used instead of IsWindows as was the case for grabbing runtime sdks? Isn't mono unsupported on Windows so why did we bother with that?

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/Dnx.cs
```diff
@@ -0,0 +1,36 @@
...
+            _sdkPath = sdkPath;
+        }
+
+        public ExecResult Execute(string commandLine, bool dnxTraceOn = true)
+        {
+            string command;
+            if (RuntimeEnvironmentHelper.IsWindows)
+            {
+                command = "cmd";
+                commandLine = $"/C {Path.Combine(_sdkPath, "bin", "dnx.exe")} {commandLine}";
```
davidfowl: Do you need to quote the path incase _sdkPath has spaces in it?

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/DnuPackOutput.cs
```diff
@@ -0,0 +1,43 @@
...
+{
+    public class DnuPackOutput : ExecResult
+    {
+        public DnuPackOutput(string outputPath, string packageName, string configuration)
+        {
+            RootPath = outputPath;
+            PackageName = packageName;
+            Configuration = configuration;
+            var basePath = Path.Combine(RootPath, Configuration);
+            PackagePath = Directory.EnumerateFiles(basePath, $"*{Constants.PackageExtension}")
```
davidfowl: You don't need to enumerate the files. I'm not sure why we did that. The output is predictable (unless we change it..)

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/Dnu.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        public ExecResult Execute(
+            string commandLine, 
+            Action<Dictionary<string, string>> envSetup = null)
+        {
+            string command;
+            if (RuntimeEnvironmentHelper.IsWindows)
+            {
+                command = "cmd";
+                commandLine = $"/C {Path.Combine(_sdkPath, "bin", "dnu.cmd")} {commandLine}";
```
davidfowl: Same comment here about quoting arguments

davidfowl:
Added 2 minor comments then :shipit:

PR #2635, opened 2015-09-04
Use the project.lock.json to determine the compatible framework for projects

davidfowl:
This removes the last set of nuget code that the runtime uses to determine compatibility. All compatibility moves to restore and the runtime just reads the lock file to build a model. This is the final part of #2622

/cc @anurse

### DIFF
--- a/src/Microsoft.Dnx.Runtime/Project.cs
+++ b/src/Microsoft.Dnx.Runtime/Project.cs
@@ -169,17 +169,10 @@ namespace Microsoft.Dnx.Runtime
 
         public TargetFrameworkInformation GetTargetFramework(FrameworkName targetFramework)
         {
-            TargetFrameworkInformation targetFrameworkInfo;
-            if (_targetFrameworks.TryGetValue(targetFramework, out targetFrameworkInfo))
+            TargetFrameworkInformation targetFrameworkInfo = null;
+            if (targetFramework != null && _targetFrameworks.TryGetValue(targetFramework, out targetFrameworkInfo))
             {
                 return targetFrameworkInfo;
-            }
-
-            IEnumerable<TargetFrameworkInformation> compatibleConfigurations;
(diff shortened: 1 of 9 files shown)


src/Microsoft.Dnx.Tooling/ProjectExtensions.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
```
BrennanConroy: copyright

davidfowl: Do we have a script for this?

src/Microsoft.Dnx.Runtime/Project.cs
```diff
@@ -169,19 +169,12 @@ public static bool TryGetProject(string path, out Project project, ICollection<D
...
             }
 
-            IEnumerable<TargetFrameworkInformation> compatibleConfigurations;
-            if (VersionUtility.GetNearest(targetFramework, GetTargetFrameworks(), out compatibleConfigurations) &&
-                compatibleConfigurations.Any())
-            {
-                targetFrameworkInfo = compatibleConfigurations.FirstOrDefault();
-            }
-
             return targetFrameworkInfo ?? _defaultTargetFrameworkConfiguration;
```
davidfowl: Why not?

BrennanConroy: Nevermind

src/Microsoft.Dnx.Tooling/ProjectExtensions.cs
BrennanConroy: Not that I know of

BrennanConroy:
Squash, copyright and :shipit:

davidfowl:
This is merged. Feel free to keep reviewing.

PR #2654, opened 2015-09-08
removed some leftover debug spew

### LINKED ISSUE #2649: Suggest suppressing the `*** READING ...` messages
`dnu restore --quiet` now outputs numerous lines like
```
*** READING C:\Users\dougbu\.dnx\packages\System.Linq.Expressions\4.0.11-beta-23302\runtime.json
```

What do these messages mean? Since there does not appear to be an action users can take in response, suggest removing the lines from the `dnu restore` output (not just from the `--quiet` output).

These messages do not just appear for `System.Linq.Expressions` though that is by far the most commonly-mentioned package, at least when in the Mvc repo.

More context...
```
Microsoft .NET Development Utility CLR-x86-1.0.0-beta8-15585

Restoring packages for C:\dd\dnx\MVC\samples\MvcMinimalSample.Web\project.json
*** READING C:\Users\dougbu\.dnx\packages\System.Linq.Expressions\4.0.11-beta-23302\runtime.json
... [repeat ad nauseam]
```

anurse:
Looks like this has been here since the original runtime.json work and just never hit in DNX because we didn't provide a runtime ID.

Fixes #2649 

/cc @davidfowl @rynowak

### DIFF
--- a/src/Microsoft.Dnx.Tooling/Restore/LocalWalkProvider.cs
+++ b/src/Microsoft.Dnx.Tooling/Restore/LocalWalkProvider.cs
@@ -56,10 +56,8 @@ namespace Microsoft.Dnx.Tooling
                     .Replace("{name}", match.Library.Name)
                     .Replace("{version}", match.Library.Version.ToString());
 
-                // Console.WriteLine("*** {0}", runtimeJsonPath);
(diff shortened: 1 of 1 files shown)


BrennanConroy:
:shipit: please!

Issue #2440, opened 2015-08-07
Add diagnostic monikers for DNX messages


sayedihashimi:
It would be helpful if dnx errors/warnings each had their own unique error codes. This will be helpful in case Visual Studio needs to de-duplicate errors and it also enables a help link for each error message.

cc @glennc @BillHiebert @davidfowl @DamianEdwards

Eilon:
What kind of DNX errors/warnings is this specifically in reference to?

davidfowl:
Similar to how CSC has CS1034 etc

Eilon:
I got that; I'm wondering for which cases it's interesting.

davidfowl:
Ah! We have a few: 

- Unresolved dependency
- Incompatible dependency
- Unexpected version resolved

victorhurdugaci:
- Moniker: NUAAAA (A = digit)
- Start at 1000. 
- Consecutive but not categorized (could allocate blocks if needed)
- NU1000 - Unexpected error (unhandled)
- Use DNX is the error is DNX specific

PR #2655, opened 2015-09-08
Add error monikers to the messageas generated by DNX/DNU

victorhurdugaci:
Fixes https://github.com/aspnet/dnx/issues/2440

Please review @davidfowl @lodejard @troydai @anurse

### DIFF
--- a/src/Microsoft.Dnx.Runtime.Abstractions/DiagnosticMessage.cs
+++ b/src/Microsoft.Dnx.Runtime.Abstractions/DiagnosticMessage.cs
@@ -8,14 +8,15 @@ namespace Microsoft.Dnx.Runtime
     /// </summary>
     public class DiagnosticMessage
     {
-        public DiagnosticMessage(string message, string filePath, DiagnosticMessageSeverity severity)
-            : this(message, filePath, severity, startLine: 1, startColumn: 0)
+        public DiagnosticMessage(string errorCode, string message, string filePath, DiagnosticMessageSeverity severity)
+            : this(errorCode, message, filePath, severity, startLine: 1, startColumn: 0)
         { }
 
-        public DiagnosticMessage(string message, string filePath, DiagnosticMessageSeverity severity, int startLine, int startColumn)
+        public DiagnosticMessage(string errorCode, string message, string filePath, DiagnosticMessageSeverity severity, int startLine, int startColumn)
                 : this(
+                    errorCode,
                     message,
-                    $"{filePath}({startLine},{startColumn}): {severity.ToString().ToLowerInvariant()}: {message}",
+                    $"{filePath}({startLine},{startColumn}): {severity.ToString().ToLowerInvariant()} {errorCode}: {message}",
                     filePath,
                     severity,
                     startLine,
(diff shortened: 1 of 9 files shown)


anurse:
Can we centralize the error codes and message templates in one place and have a class with a bunch of static methods that create fully-formed Diagnostic Messages?

moozzyk:
:+1: Other than that it looks good to me.

src/Microsoft.Dnx.Runtime.Internals/DiagnosticMonikers.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Dnx.Runtime.Internals
+{
+    internal class DiagnosticMonikers
+    {
+        public static readonly string NU1001 = nameof(NU1001); // The dependency A could not be resolved.
+        public static readonly string NU1002 = nameof(NU1002); // The dependency A in project B does not support framework C."
+        public static readonly string NU1003 = nameof(NU1003); // Invalid A section. The target B is invalid, targets must either be a file name or a directory suffixed with '/'. The root directory of the package can be specified by using a single '/' character.
```
troydai: Can we put the comments above the error code so as to shorten the line?

src/Microsoft.Dnx.Runtime.Abstractions/DiagnosticMessage.cs
```diff
@@ -25,15 +26,17 @@ public DiagnosticMessage(string message, string filePath, DiagnosticMessageSever
...
+            string message,
+            string formattedMessage,
+            string filePath,
+            DiagnosticMessageSeverity severity,
+            int startLine,
+            int startColumn,
+            int endLine,
             int endColumn)
         {
+            ErrorCode = errorCode;
```
troydai: Should we throw if the `errorCode` is `null`?

moozzyk:
Actually are you planning to add any tests for this?

src/Microsoft.Dnx.Runtime.Abstractions/DiagnosticMessage.cs
victorhurdugaci: No. VS works well without it

victorhurdugaci:
Will fix the test failures once we figure out what is the error format that VS understands

victorhurdugaci:
@moozzyk  Testing this is pretty complicated if you want to go the DTH route. Maybe @troydai can help here.

I've updated the tests that cover error message so we get some coverage at least for the out of proc scenario.

davidfowl:
@victorhurdugaci Did you figure out the parsing issue?

victorhurdugaci:
@davidfowl check your e-mail :)

src/Microsoft.Dnx.Runtime.Abstractions/DiagnosticMessage.cs
troydai: All right.

davidfowl:
:shipit: Once you test it with @BillHiebert 's new bits.

PR #2643, opened 2015-09-05
Prune packages with --framework

victorhurdugaci:
Fixes https://github.com/aspnet/dnx/issues/2640
Fixes https://github.com/aspnet/dnx/issues/2428

- Dnu publish should have an option to prune packages when the runtime flag is not passed.
- Dnu publish --no-source doesn't remove tfms for the projects being compiled 

Similar to `--runtime` but allows publishing just a set of target frameworks without runtime.

Also fixed a publish bug when you run `--no-source` and you get too many frameworks.

Please review: @BrennanConroy @davidfowl @anurse

PR #2659, opened 2015-09-09
Enabling running scripts per configuration per framework

### LINKED ISSUE #887: Expose project build configuration in scripts section of project.json
There are a few variables surfaced in the [scripts section of project.json](https://github.com/aspnet/Home/wiki/Project.json-file#scripts). However, the current build configuration is not among them. 

The variable is useful in such a scenario: copying the output to a folder other than default output
``` json
{
  "scripts":{
    "postbuild": [
      "xcopy /F /Y /I bin\\%project:BuildConfiguration%\\* C:\\Playground"
    ]
  }
}
```

It worth to mention that `kpm build` support multiple configurations in one build run. If this variable is exposed, the scripts need to be executed multiple times to match each configuration.

moozzyk:
Required to fix #887

### DIFF
--- a/src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
+++ b/src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
@@ -6,11 +6,10 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Runtime.Versioning;
 using Microsoft.Dnx.Compilation;
 using Microsoft.Dnx.Compilation.Caching;
-using Microsoft.Dnx.Compilation.FileSystem;
 using Microsoft.Dnx.Runtime;
-using Microsoft.Dnx.Runtime.Compilation;
 using Microsoft.Dnx.Tooling.SourceControl;
 using Microsoft.Dnx.Tooling.Utils;
 using Microsoft.Framework.FileSystemGlobbing;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
```diff
@@ -127,192 +125,208 @@ private bool BuildInternal(string projectPath)
...
 
-            var allDiagnostics = new List<DiagnosticMessage>();
+            allDiagnostics.AddRange(projectDiagnostics);
+            WriteSummary(allDiagnostics);
 
+            _buildOptions.Reports.Information.WriteLine("Time elapsed {0}", sw.Elapsed);
+            return success;
+        }
 
+        private bool BuildConfiguration(string baseOutputPath, IEnumerable<System.Runtime.Versioning.FrameworkName> frameworks, List<DiagnosticMessage> allDiagnostics, string configuration)
```
davidfowl: Fix `System.Runtime.Versioning.FrameworkName` that. Fix the namespace.

davidfowl:
This diff is crazy hard to return

davidfowl:
We need to flow the context to the command being run. You need to add some variables to `GetScriptVariable` so that scripts can use `project:Configuration` and `project:Framework` when executing scripts to make decisions.

moozzyk:
@davidfowl d@sdfjeuio

moozzyk:
@davidfowl as I said it is required to fix #887 and not fixes #887. The fix prototype for bug-887 is here https://github.com/aspnet/dnx/commit/69b90084c6703c03ca7edd96d784e86316f901e1. I decided to send this separately since I am waiting for #2643. Without #2643 I don't have framework moniker - only the full name of framework. Also in the prototype I used different names - let me know which ones you want.

moozzyk:
The method was already like 300 lines long - I extracted a couple methods and simplified some `if`s.  It's hard to review but it is now easier to read.

davidfowl:
I think we should just pass the variables in the same time. How does this affect the performance of building a single project in our repository (since it now copies multiple times instead of once)?. I think the `build:TargetFramework` and `build:Configuration` sound better.

moozzyk:
I will wait for @victorhurdugaci and update the PR after he commits #2643. 
I have not checked the performance but from what I understand currently we build only one configuration and then copy explicitly for dnx451 and dnxcore50 (btw. this is also conceptually wrong since it assumes we always build both dnx451 and dnxcore50 which does not have to be the case). After the change the scripts need to/will be changed to copy only once but to %build.Configuration%/%build.TargetFramework% folder. So instead of doing all the copying blindly without really knowing what was built each build will take care of stuff it built.

src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
moozzyk: = don't trust "Extract Method" by VS. Will do.

moozzyk:
Replaced by #2679

PR #2665, opened 2015-09-10
Pawelka/coreclr rewrite

moozzyk:
@janvorli - can you take a look at this? This (at least "Rewriting CoreClr bootstrapper to use new runtime API" part) is related to the [change](https://github.com/dotnet/coreclr/commit/441a97709757ec1b6edcaadb733941e64af8fb60) you made

/cc @davidfowl @BrennanConroy

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -345,12 +345,14 @@ var CAN_BUILD_ONECORE = '${Directory.Exists(VS_ONECORE_ROOT) && Directory.GetFil
         var sourceFiles = new string[] 
         {
             Path.Combine("src", BOOTSTRAPPER_COMMON_FOLDER_NAME, "tpa.cpp"),
-            Path.Combine("src", BOOTSTRAPPER_CORECLR_NAME + ".unix", BOOTSTRAPPER_CORECLR_NAME + ".cpp")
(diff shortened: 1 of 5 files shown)


BrennanConroy:
Tested on Mac and this gets rid of the Trace/BPT error we were seeing.

moozzyk:
Yay!

davidfowl:
Feels like we should have a constant for this `"Microsoft.Dnx.Host.CoreClr"` as it's used in several places in the file. Or at least something like:

```
bootstrapperAssemblyName = "Microsoft.Dnx.Host.CoreClr";
bootstrapperAssemblyFileName = "Microsoft.Dnx.Host.CoreClr.dll";
bootstrapperAssemblyFullName = "Microsoft.Dnx.Host.CoreClr, 1.0.0.0, ...."
```

janvorli:
LGTM
@moozzyk Please ignore the comment on the ExecuteAssembly, I have missed the fact that there were two commits.

moozzyk:
@janvorli I think loading libcoreclrpal.so was added by @ellismg sometime in spring. It has not failed since so no one was aware it is not needed. I will remove loading this lib.

moozzyk:
@davidfowl - created a #define for Microsoft.Dnx.Host.CoreClr but it does not look prettier...

PR #2669, opened 2015-09-10
Removing attempts to load libcoreclrpal.{so|dylib}

moozzyk:
libcoreclrpal no longer exists so the code is not needed.


davidfowl:
:shipit:

Issue #2197, opened 2015-07-07
Implicit DNX min version check


muratg:
If any of the packages in project.json depend on a particular DNX and this is not less than the DNX version  we're currently using, we can detect and throw on the app start.  This would prevent some of the issues described in this bug: https://github.com/aspnet/Home/issues/312#issuecomment-119013660

davidfowl:
@muratg I think we need to do this for beta6

muratg:
@victorhurdugaci has found a potential issue that we'll need to discuss. Tentatively putting in beta 6.

davidfowl:
Reopening this /cc @muratg @glennc

avanderhoorn:
@anurse thanks, better than nothing!

victorhurdugaci:
- Check for the version of Abstractions
- If the app tries to load a newer one, stop it

anurse:
technically we keep that label but put design:done on it. /cc @glennc we still using your app?

victorhurdugaci:
Note: this will be interesting to see in the wild because our dev bits have a higher version than release ones (1.0.0-beta7-123 > 1.0.0-beta7)

PR #2675, opened 2015-09-10
Check for min runtime version at application start

victorhurdugaci:
Fixes: https://github.com/aspnet/dnx/issues/2197

- The error shows up only on run
- The error doesn't show up on build so you can build with an older runtime
- No tests because this is super hard to test
- Most of the change is passing IRuntimeEnvironment up the stack

please review @davidfowl @anurse

### DIFF
--- a/src/Microsoft.Dnx.ApplicationHost/DefaultHost.cs
+++ b/src/Microsoft.Dnx.ApplicationHost/DefaultHost.cs
@@ -22,6 +22,9 @@ namespace Microsoft.Dnx.ApplicationHost
 {
     public class DefaultHost
     {
+        private static readonly string RuntimeAbstractionsPackageName =
+            typeof(IRuntimeEnvironment).GetTypeInfo().Assembly.GetName().Name;
+
         private ApplicationHostContext _applicationHostContext;
 
         private readonly string _projectDirectory;
@@ -124,6 +127,8 @@ Please make sure the runtime matches a framework specified in {Project.ProjectFi
 
             _project = applicationHostContext.Project;
 
+            ValidateMinRuntimeVersion(libraries);
+
             if (options.WatchFiles)
             {
                 fileWatcher.OnChanged += _ =>
@@ -140,11 +145,11 @@ Please make sure the runtime matches a framework specified in {Project.ProjectFi
             var applicationEnvironment = new ApplicationEnvironment(Project, _targetFramework, options.Configuration, hostEnvironment);
 
             var compilationContext = new CompilationEngineContext(
-                applicationEnvironment, 
-                _runtimeEnvironment, 
-                loadContextAccessor.Default, 
-                new CompilationCache(), 
-                fileWatcher, 
+                applicationEnvironment,
+                _runtimeEnvironment,
+                loadContextAccessor.Default,
+                new CompilationCache(),
+                fileWatcher,
                 new ProjectGraphProvider());
 
             // Compilation services available only for runtime compilation
@@ -181,6 +186,28 @@ Please make sure the runtime matches a framework specified in {Project.ProjectFi
             _serviceProvider.Add(typeof(ICompilerOptionsProvider), new CompilerOptionsProvider(projects));
 
             AddBreadcrumbs(libraries);
+        }
+
+        private void ValidateMinRuntimeVersion(IEnumerable<LibraryDescription> libraries)
+        {
+            if (string.Equals(Environment.GetEnvironmentVariable(EnvironmentNames.DnxDisableMinVersionCheck), "1", StringComparison.OrdinalIgnoreCase))
+            {
+                // Min version is disabled
+                return;
+            }
+
+            foreach (var lib in libraries)
+            {
+                // We only need to validate if runtime abstractions is precompiled
+                // because when building from sources the versions are not ordered
+                // correctly
+                if (lib.Type == LibraryTypes.Package &&
+                    string.Equals(lib.Identity.Name, RuntimeAbstractionsPackageName, StringComparison.Ordinal) &&
+                    lib.Identity.Version > new SemanticVersion(_runtimeEnvironment.RuntimeVersion))
+                {
+                    throw new InvalidOperationException($"This application requires DNX version {lib.Identity.Version} or newer to run.");
+                }
+            }
         }
 
         private void AddBreadcrumbs(IEnumerable<LibraryDescription> libraries)
--- a/src/Microsoft.Dnx.Runtime.Sources/Impl/EnvironmentNames.cs
+++ b/src/Microsoft.Dnx.Runtime.Sources/Impl/EnvironmentNames.cs
@@ -23,5 +23,6 @@ namespace Microsoft.Dnx.Runtime
         public const string PortablePdb = "DNX_BUILD_PORTABLE_PDB";
         public const string DnxIsWindows = "DNX_IS_WINDOWS";
         public const string AspNetLoaderPath = "DNX_ASPNET_LOADER_PATH";
+        public const string DnxDisableMinVersionCheck = "DNX_NO_MIN_VERSION_CHECK";
     }
 }


src/Microsoft.Dnx.Runtime/ApplicationHostContext.cs
```diff
@@ -30,18 +30,20 @@ public class ApplicationHostContext
 
         public bool SkipLockfileValidation { get; set; }
 
+        public bool SkipRuntimeMinVersionCheck { get; set; }
+
         public FrameworkReferenceResolver FrameworkResolver { get; set; }
```
davidfowl: Add the `RuntimeVersion` to the ApplicationHostContext

src/Microsoft.Dnx.Runtime/ApplicationHostContext.cs
```diff
@@ -214,7 +221,19 @@ private static Result InitializeCore(ApplicationHostContext context)
             var lockFilePresent = context.LockFile != null;
             context.LockFile = null; // LockFile is no longer needed
 
-            return new Result(libraries, lockFilePresent, validLockFile, lockFileValidationMessage);
+            string minRequiredRuntime = null;
+            if (!context.SkipRuntimeMinVersionCheck)
+            {
+                var abstractions = libraries.FirstOrDefault(lib => lib.Identity.Name.Equals("Microsoft.Dnx.Runtime.Abstractions", StringComparison.Ordinal));
```
anurse: @davidfowl is going to tell you to remove the Linq. No Linq allowed in the runtime core path :).

moozzyk:
Is it possible to add tests?

victorhurdugaci:
Updated with tests

src/Microsoft.Dnx.Runtime/ApplicationHostContext.cs
```diff
@@ -215,7 +224,30 @@ private static Result InitializeCore(ApplicationHostContext context)
...
+            string minRequiredRuntime = null;
+            if (!context.SkipRuntimeMinVersionCheck)
+            {
+                LibraryDescription abstractions = null;
+                foreach (var lib in libraries)
+                {
+                    // Ignore projects references because in that case we are compiling from sources 
+                    // and versions don't match
+                    if (lib.Type == LibraryTypes.Package &&
+                        lib.Identity.Name.Equals("Microsoft.Dnx.Runtime.Abstractions", StringComparison.Ordinal))
```
troydai: Can we extract the string `Microsoft.Dnx.Runtime.Abstractions` in way rather than hard code the string. For example reflecting a type name?

anurse: I'd avoid reflection since this is in the critical runtime path. But putting it in a constant is a good idea :)

src/Microsoft.Dnx.Compilation/CompilationEngine.cs
```diff
@@ -59,7 +59,9 @@ public LibraryExporter CreateProjectExporter(Project project, FrameworkName targ
             {
                 Project = project,
                 RuntimeIdentifiers = _context.RuntimeEnvironment.GetAllRuntimeIdentifiers(),
-                TargetFramework = _context.ApplicationEnvironment.RuntimeFramework
+                TargetFramework = _context.ApplicationEnvironment.RuntimeFramework,
+                SkipRuntimeMinVersionCheck = true,
+                RuntimeEnvironment = _context.RuntimeEnvironment
```
troydai: If runtime version check is skipped here, the `RuntimeEnvironment` instance is not required. Do we have to set it?

src/Microsoft.Dnx.Runtime/ApplicationHostContext.cs
```diff
@@ -215,7 +224,30 @@ private static Result InitializeCore(ApplicationHostContext context)
...
-            return new Result(libraries, lockFilePresent, validLockFile, lockFileValidationMessage);
+            string minRequiredRuntime = null;
+            if (!context.SkipRuntimeMinVersionCheck)
+            {
+                LibraryDescription abstractions = null;
+                foreach (var lib in libraries)
+                {
+                    // Ignore projects references because in that case we are compiling from sources 
+                    // and versions don't match
+                    if (lib.Type == LibraryTypes.Package &&
```
anurse: `Type` is a string I believe. We should use `string.Equals` to compare it in order to be clear about case-sensitivity (it **is** case-sensitive :))

src/Microsoft.Dnx.Tooling/List/DependencyListOperation.cs
```diff
@@ -70,7 +70,8 @@ private LibraryManager CreateLibraryManager()
             var hostContext = new ApplicationHostContext
             {
                 Project = _options.Project,
-                TargetFramework = _framework
+                TargetFramework = _framework,
+                RuntimeEnvironment = RuntimeEnvironmentHelper.RuntimeEnvironment
```
troydai: Do we need to check min runtime version just for `dnu list`?

src/Microsoft.Dnx.Tooling/Publish/DependencyContext.cs
```diff
@@ -18,7 +18,8 @@ public DependencyContext(Runtime.Project project, FrameworkName targetFramework)
             var applicationHostContext = new ApplicationHostContext
             {
                 Project = project,
-                TargetFramework = targetFramework
+                TargetFramework = targetFramework,
+                RuntimeEnvironment = RuntimeEnvironmentHelper.RuntimeEnvironment
```
troydai: Same question as above.

src/Microsoft.Dnx.Runtime/ApplicationHostContext.cs
```diff
@@ -52,6 +56,11 @@ public static IList<LibraryDescription> GetRuntimeLibraries(ApplicationHostConte
                 }
             }
 
+            if (!string.IsNullOrEmpty(result.MinRequiredRuntimeVersion))
```
anurse: This confused me at first because I thought it was always being set to the minimum version (so I was like "but what if it's below the current runtime version?"). Can you change the below code to just extract the version and put the comparison here instead?

test/Microsoft.Dnx.Runtime.Tests/Host/MinRuntimeVersionFacts.cs
```diff
@@ -0,0 +1,125 @@
...
+namespace Microsoft.Dnx.Runtime.Tests.Host
+{
+    public class MinRuntimeVersionFacts
+    {
+        private const string TooHighVersion = "100.0.0";
+
+        private static readonly FrameworkName Dnx451 = new FrameworkName(VersionUtility.DnxFrameworkIdentifier, new Version(4, 5, 1));
+
+        [Fact]
+        public void LoadWithOldRuntime_Throws()
```
anurse: Woo! More AHC Unit Tests!

anurse:
Looking good. A few comments inline.

src/Microsoft.Dnx.Runtime/ApplicationHostContext.cs
victorhurdugaci: I'm with @anurse on this one

src/Microsoft.Dnx.Runtime/ApplicationHostContext.cs
victorhurdugaci: Yes but it is used throughout the code as an enum... (weird, I know). We never check it with a specific case

src/Microsoft.Dnx.Tooling/List/DependencyListOperation.cs
victorhurdugaci: Fixed

src/Microsoft.Dnx.Runtime/ApplicationHostContext.cs
troydai: +1 this one got me, too.

victorhurdugaci: Sure

src/Microsoft.Dnx.Compilation/CompilationEngine.cs
victorhurdugaci: For consistency

src/Microsoft.Dnx.Runtime/ApplicationHostContext.cs
```diff
@@ -215,7 +225,29 @@ private static Result InitializeCore(ApplicationHostContext context)
...
+                        lib.Identity.Name.Equals("Microsoft.Dnx.Runtime.Abstractions", StringComparison.Ordinal))
+                    {
+                        abstractions = lib;
+                        break;
+                    }
+                }
+
+                if (abstractions != null)
+                {
+                    minRequiredRuntime = abstractions.Identity.Version.ToString();
```
davidfowl: Why are you turning this into a string and then back into a version?

src/Microsoft.Dnx.Runtime/ApplicationHostContext.cs
```diff
@@ -31,10 +31,14 @@ public class ApplicationHostContext
 
         public bool SkipLockfileValidation { get; set; }
 
+        public bool SkipRuntimeMinVersionCheck { get; set; }
```
davidfowl: Make this true by default.

davidfowl:
After giving this some thought, I think this change should be here:

https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.ApplicationHost/DefaultHost.cs#L172

It's only relevant at runtime, we're trying to protect people from themselves and since we only throw at runtime, why put it in ApplicationHostContext.

src/Microsoft.Dnx.Runtime/ApplicationHostContext.cs
muratg: Agreed, though if we make other name changes we'll have to catch this case as well.

src/Microsoft.Dnx.ApplicationHost/Microsoft.Dnx.ApplicationHost.xproj
```diff
@@ -13,5 +13,8 @@
   <PropertyGroup>
     <SchemaVersion>2.0</SchemaVersion>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <ProduceOutputsOnBuild>True</ProduceOutputsOnBuild>
```
victorhurdugaci: whoops, forgot these. Will revert

victorhurdugaci:
updated

src/Microsoft.Dnx.ApplicationHost/DefaultHost.cs
```diff
@@ -183,6 +185,22 @@ private void Initialize(RuntimeOptions options, IServiceProvider hostServices, I
...
 
+        private void ValidateMinRuntimeVersion(IEnumerable<LibraryDescription> libraries)
+        {
+            foreach (var lib in libraries)
+            {
+                // We only need to validate if runtime abstractions is precompiled
+                // because when building from sources the versions are not ordered
+                // correctly
+                if (lib.Type == LibraryTypes.Package &&
+                    string.Equals(lib.Identity.Name, "Microsoft.Dnx.Runtime.Abstractions", StringComparison.Ordinal) &&
```
anurse: Constantify, as discussed earlier ;)

anurse:
LGTM

src/Microsoft.Dnx.ApplicationHost/DefaultHost.cs
```diff
@@ -183,6 +185,22 @@ private void Initialize(RuntimeOptions options, IServiceProvider hostServices, I
             AddBreadcrumbs(libraries);
         }
 
+        private void ValidateMinRuntimeVersion(IEnumerable<LibraryDescription> libraries)
+        {
+            foreach (var lib in libraries)
```
davidfowl: We can avoid the 2 for loops and do the servicing and this in a single pass but that's not important right now.

src/Microsoft.Dnx.ApplicationHost/DefaultHost.cs
davidfowl: maybe `typeof(SomethingInThatAssembly).GetTypeInfo().Assembly.GetName().Name`

davidfowl:
:shipit:

src/Microsoft.Dnx.ApplicationHost/DefaultHost.cs
victorhurdugaci: will change now

PR #2679, opened 2015-09-11
Fix for #887

moozzyk:


davidfowl:
I think you should change the scripts after this change goes in. Otherwise we won't be able to do Dev until coherence passes

BrennanConroy:
Can we add linux and darwin to the postbuild scripts?

davidfowl:
@BrennanConroy Sure but not as part of this fix.

BrennanConroy:
File #2684

moozzyk:
@davidfowl this was one of the reasons why scripts are updated as a separate commit - so that they can be committed separately

davidfowl:
@moozzyk The PR shows both. Hard to know what you'll commit

moozzyk:
@davidfowl - I will commit scripts separately after coherence with the first two commits passes - and it can only happen if I get a ship it :)

davidfowl:
:shipit:

davidfowl:
This can be closed right? @moozzyk

moozzyk:
Yes. I will close this. The scripts won't go in in the current state. They either need to be redone a little bit or not touched. At the moment - because we have a hardcoded Debug folder in the source path nothing is being copied on the CI which is a good thing.

davidfowl:
@moozzyk Can you make sure we have a task tracking that. We should be using this change right now to make sure it works well.

moozzyk:
OK. Will create one and assign to myself.

PR #2688, opened 2015-09-11
CORECLR_DIR is gone

moozzyk:


davidfowl:
:shipit: Helios will snap harder 😀

PR #2691, opened 2015-09-11
Changing return code from -1 to 1

moozzyk:
On non-windows the range of the exit codes is [0, 255]. -1 is turned
to 255 which means "Exit code out of range". 1 stands for "Catchall
for general errors" which is our case. We print more detailed error
information to the stderr stream.

### DIFF
--- a/src/dnx.coreclr.unix/dnx.coreclr.cpp
+++ b/src/dnx.coreclr.unix/dnx.coreclr.cpp
@@ -122,7 +122,7 @@ int LoadCoreClr(const char* runtime_directory)
         fprintf(stderr, "failed to locate libcoreclr with error %s\n", error);
 
         FreeCoreClr();
-        return -1;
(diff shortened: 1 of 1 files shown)


davidfowl:
:shipit:

davidfowl:
When travis passes

muratg:
:shipit:

moozzyk:
@davidfowl - you are talking about console commands right? The bootstrapper already returns 1 in case of a failure: https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Host/RuntimeBootstrapper.cs#L33 (as opposed to console commands where some of them return -1)

davidfowl:
What about application host

moozzyk:
Good point - I quick searched only for `return -1` but the application host is async so it returns `Task.FromResult(-1)`

Issue #2405, opened 2015-08-04
When csproj project reference can not be resolved, it still should be sent as project reference


abpiskunov:
When Project A references project B - a wrapped csproj which can not  be resolved by some reason, project B is not sent as project reference to project A, however exist as unresolved dependency. This make its is hard for tooling to show it correctly in Add Reference dialog.

However if you look at dependency info, it can be determined that Project B is actually a project, since it has a path to wrapped project.json, thus DTH could notice that and send it also as project reference in references message.

![image](https://cloud.githubusercontent.com/assets/1592091/9070665/e80794a0-3abe-11e5-8671-22285df42151.png)

abpiskunov:
It still should be unresolved in Dependencies message though.

abpiskunov:
It would be nice if it could be beta7 issue.

abpiskunov:
Please let me know when fix is checked in and I will verify that tooling works with it as expected.

davidfowl:
This is actually a bigger change than it looks like. We'd need to add a new field that determines if the dependency is resolved or not so we can preserve the original type.

davidfowl:
See https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.DesignTimeHost/Models/OutgoingMessages/DependencyDescription.cs

abpiskunov:
I see, sounds good.

troydai:
__Proposal__

_Solution 1_
Adding a new boolean property `Resolved` to the `DependencyDescription` model to indicate if the dependency resolved. And the `Type` property accept the `LibraryDescription`'s `Type` value.

_Solution 2_
Reuse the `Type` property but append the `LibraryDescription.Type` with a string indicate if it is resolved or not. So it will look like:
- `Project`
- `Package`
- `Project:Unresolved`
- `Package:Unresolved`

abpiskunov:
Both of the would require tooling changes in parsing messages we receive from DTH, but Solution1 looks more neat.

troydai:
Either way it's a breaking change requires tooling coordination. Data in solution 1 will be easier to be consumed. ping @davidfowl

davidfowl:
Solution 1 is better 😁

Issue #2502, opened 2015-08-17
Adding DTH integration test


troydai:
Port from https://github.com/troydai/Sparrow

- [x] Protocol Negotiation
- [x] Perform Compilation
- [x] Refresh Dependencies
- [ ] Wrap project
- [ ] Enumerate project (__in progress__: https://github.com/aspnet/dnx/pull/2922)
- [ ] Get compiled assembly
- [ ] File format exception
- [x] Compile error / warning
- [x] Change configuration

anurse:
just attaching myself to the thread so I get notifications under the participating section ;)

troydai:
The first test case has been checked in d4465d83c176bab267bca61b205d7ac7a13b625f.

troydai:
Adding following tests now:
[] Protocol Negotiation
[] Compilation messages
[] Dependencies messages
[] Compilation diagnostics
[] Dependencies diagnostics

troydai:
More work planned.

muratg:
Remaining work can be done in RC1.

troydai:
Not validate any more. Close.

PR #2701, opened 2015-09-14
Add property 'Resolved' to DependencyDescription

troydai:
Addressing: #2405 

/cc @abpiskunov: this will break current DTH protocol need VS side change.

Add a new boolean property `Resolved` to the dependency description in DTH response. Alter the `Type` property to represent the type of the unresolved dependency. For __project__ dependency the type will become `"Project"`. However for __package__ it is still `"Unresolved"` since the `UnresolvedDependencyResolver` can't determine the actual type.

### DIFF
--- a/src/Microsoft.Dnx.DesignTimeHost/Models/OutgoingMessages/DependencyDescription.cs
+++ b/src/Microsoft.Dnx.DesignTimeHost/Models/OutgoingMessages/DependencyDescription.cs
@@ -9,7 +9,7 @@ namespace Microsoft.Dnx.DesignTimeHost.Models.OutgoingMessages
     public class DependencyDescription
     {
         public string Name { get; set; }
-        
+
         public string DisplayName { get; set; }
 
         public string Version { get; set; }
@@ -18,6 +18,8 @@ namespace Microsoft.Dnx.DesignTimeHost.Models.OutgoingMessages
 
         public string Type { get; set; }
 
+        public bool Resolved { get; set; }
+
         public IEnumerable<DependencyItem> Dependencies { get; set; }
 
         public override bool Equals(object obj)
@@ -25,6 +27,7 @@ namespace Microsoft.Dnx.DesignTimeHost.Models.OutgoingMessages
             var other = obj as DependencyDescription;
 
             return other != null &&
+                   Resolved == other.Resolved &&
                    string.Equals(Name, other.Name) &&
                    object.Equals(Version, other.Version) &&
                    string.Equals(Path, other.Path) &&
--- a/test/Microsoft.Dnx.DesignTimeHost.FunctionalTests/DthStartupTests.cs
+++ b/test/Microsoft.Dnx.DesignTimeHost.FunctionalTests/DthStartupTests.cs
@@ -49,14 +49,16 @@ namespace Microsoft.Dnx.DesignTimeHost.FunctionalTests
             {
                 foreach (var combination in RuntimeComponents)
                 {
-                    // The current max protocol version is 2
+                    // The current max protocol version is 3
 
                     // request 1, respond 1
                     yield return combination.Concat(new object[] { 1, 1 }).ToArray();
                     // request 2, respond 2
                     yield return combination.Concat(new object[] { 2, 2 }).ToArray();
                     // request 3, respond 2
-                    yield return combination.Concat(new object[] { 3, 2 }).ToArray();
+                    yield return combination.Concat(new object[] { 3, 3 }).ToArray();
+                    // request 4, respond 3
+                    yield return combination.Concat(new object[] { 4, 3 }).ToArray();
                 }
             }
         }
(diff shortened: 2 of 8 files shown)


troydai:
/review: @davidfowl 
/cc: @muratg @BrennanConroy @victorhurdugaci @anurse 

As expected it should break VS in unresolved dependency situation. I'll testing in VS right now.

victorhurdugaci:
Looks good

muratg:
:shipit:

troydai:
@abpiskunov please comment. This change needs VS side's update as well.

abpiskunov:
Yes, I will add corresponding changes in WTE. Where could I get your privates to tests ? Or it is already in dev builds?

davidfowl:
This needs to be a protocol version change

src/Microsoft.Dnx.DesignTimeHost/Models/OutgoingMessages/DependencyDescription.cs
```diff
@@ -18,6 +18,8 @@ public class DependencyDescription
 
         public string Type { get; set; }
 
+        public bool Resolved { get; set; }
```
davidfowl: This needs to be part of Equals

troydai:
@davidfowl do you think it is good to go?

davidfowl:
Can you add a test for a package that exists but is incompatible (aka unresolved)? Like any package on the internet that doesn't support dnxcore50.

troydai:
If we want the test to be reliable we have to create the package on the fly.

davidfowl:
We have tons of tests that rely on nuget right now (heck even our build relies on nuget). Older version of a package are pretty much immutable so I'd be ok picking something like mvc 3 or something else that we shipped in the past

test/Microsoft.Dnx.DesignTimeHost.FunctionalTests/DthStartupTests.cs
```diff
@@ -202,5 +203,68 @@ public void DthCompilation_RestoreComplete_OnEmptyConsoleApp(string flavor, stri
...
+                    yield return combination.Concat(new object[] { 1, "UnresolvedPackageSample", "NoSuchPackage", "Unresolved" }).ToArray();
+
+                    yield return combination.Concat(new object[] { 2, "UnresolvedProjectSample", "EmptyLibrary", "Unresolved" }).ToArray();
+
+                    yield return combination.Concat(new object[] { 2, "UnresolvedPackageSample", "NoSuchPackage", "Unresolved" }).ToArray();
+
+                    yield return combination.Concat(new object[] { 3, "UnresolvedProjectSample", "EmptyLibrary", "Project" }).ToArray();
+
+                    // Unresolved package dependency's type is still Unresolved
+                    yield return combination.Concat(new object[] { 3, "UnresolvedPackageSample", "NoSuchPackage", "Unresolved" }).ToArray();
```
abpiskunov: Any scenario for wrapped project?

test/Microsoft.Dnx.DesignTimeHost.FunctionalTests/DthStartupTests.cs
```diff
@@ -202,5 +203,68 @@ public void DthCompilation_RestoreComplete_OnEmptyConsoleApp(string flavor, stri
...
+
+            using (var server = DthTestServer.Create(runtimeHomePath, testProject))
+            using (var client = new DthTestClient(server, 0))
+            {
+                client.Initialize(testProject, protocolVersion);
+
+                var message = client.DrainTillFirst("Dependencies");
+                message.EnsureSource(server, client);
+
+                var dependencies = message.Payload["Dependencies"];
```
abpiskunov: What about References message? Could you add a test for that too?

troydai: I'll add tests for reference later in the future.

abpiskunov: Do you have an issue to track it? :) Its easier to do it now, but its up to you of course (btw it would caught another issue with wrapped project path that I sent you earlier)

davidfowl: I agree with Anton. We need to have tests for everything we find. You already did all the had work getting the infrastructure setup to test dth .

davidfowl:
:watch:

troydai:
So. @abpiskunov I'll add the wrap test later in the future. The work is being tracked here https://github.com/aspnet/dnx/issues/2502

I'll add test for incompatible package and asserts references.

troydai:
:watch: @davidfowl

troydai:
Examining the Travis failure now.

troydai:
Travis failure due to myget timeout. I rebase and triggered another run.

davidfowl:
When travis passes :shipit:

troydai:
Travis passed. I'll merge after squash the commits.

PR #2705, opened 2015-09-15
Passing context from native bootstrapper coreclr windows

moozzyk:
@davidfowl @anurse @victorhurdugaci @DamianEdwards 

I am sending this to get early feedback - Travis won't pass. 

If you think I missed some of the variables (as @victorhurdugaci found out yesterday) let me know ASAP.

### DIFF
--- a/src/Microsoft.Dnx.Host.Clr/RuntimeBootstrapper.cs
+++ b/src/Microsoft.Dnx.Host.Clr/RuntimeBootstrapper.cs
@@ -9,23 +9,35 @@ namespace Microsoft.Dnx.Host.Clr
     {
         public static int Execute(string[] argv, FrameworkName targetFramework)
         {
-            var executeMethodInfo = GetBootstrapperType()
-                .GetMethod("Execute", BindingFlags.Static | BindingFlags.Public, null, new[] { typeof(string[]), typeof(FrameworkName) }, null);
-            return (int)executeMethodInfo.Invoke(null, new object[] { argv, targetFramework });
+            var bootstrapperContext = GetBootstrapperContext();
+
+            var executeMethodInfo = GetBootstrapperType(bootstrapperContext.GetType().Assembly)
+                .GetMethod("Execute", BindingFlags.Static | BindingFlags.Public, null,
+                    new[] { typeof(string[]), typeof(FrameworkName), bootstrapperContext.GetType() }, null);
+            return (int)executeMethodInfo.Invoke(null, new object[] { argv, targetFramework, null });
         }
 
         // This method is only called by Helios
         public static Task<int> ExecuteAsync(string[] argv, FrameworkName targetFramework)
         {
-            var executeMethodInfo = GetBootstrapperType()
-                .GetMethod("ExecuteAsync", BindingFlags.Static | BindingFlags.Public, null, new[] { typeof(string[]), typeof(FrameworkName) }, null);
-            return (Task<int>)executeMethodInfo.Invoke(null, new object[] { argv, targetFramework });
+            var bootstrapperContext = GetBootstrapperContext();
+
+            var executeMethodInfo = GetBootstrapperType(bootstrapperContext.GetType().Assembly)
+                .GetMethod("ExecuteAsync", BindingFlags.Static | BindingFlags.Public, null,
+                    new[] { typeof(string[]), typeof(FrameworkName), bootstrapperContext.GetType() }, null);
+            return (Task<int>)executeMethodInfo.Invoke(null, new object[] { argv, targetFramework, null });
         }
 
-        private static Type GetBootstrapperType()
+        private static Type GetBootstrapperType(Assembly dnxHost)
+        {
+            return dnxHost.GetType("Microsoft.Dnx.Host.RuntimeBootstrapper");
+        }
+
+        private static object GetBootstrapperContext()
         {
             var dnxHost = Assembly.Load("Microsoft.Dnx.Host");
-            return dnxHost.GetType("Microsoft.Dnx.Host.RuntimeBootstrapper");
+            var bootstrapperContext = Activator.CreateInstance(dnxHost.GetType("Microsoft.Dnx.Host.BootstrapperContext"));
+            return bootstrapperContext;
         }
     }
 }
--- a/src/Microsoft.Dnx.Host.CoreClr/DomainManager.cs
+++ b/src/Microsoft.Dnx.Host.CoreClr/DomainManager.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
 using System.Security;
 using Microsoft.Dnx.Host;
@@ -11,8 +12,21 @@ using Microsoft.Dnx.Runtime.Common.Impl;
 [SecurityCritical]
 sealed class DomainManager
 {
+    // this structure is used to pass data from native code
+    // and needs to be in sync with its native counterpart
+    [StructLayout(LayoutKind.Sequential)]
+    private unsafe struct NativeBootstrapperContext
+    {
+        public char* OperatingSystem;
+        public char* OsVersion;
+        public char* Architecture;
+        public char* RuntimeDirectory;
+        public char* ApplicationBase;
+        public bool HandleExceptions;
+    }
+
     [SecurityCritical]
-    unsafe static int Execute(int argc, char** argv)
+    unsafe static int Execute(int argc, char** argv, NativeBootstrapperContext* context)
     {
         Logger.TraceInformation($"[{nameof(DomainManager)}] Using CoreCLR");
 
@@ -25,9 +39,20 @@ sealed class DomainManager
 
         try
         {
+            var bootstrapperContext = new BootstrapperContext();
+
+            bootstrapperContext.OperatingSystem = new string(context->OperatingSystem);
+            bootstrapperContext.OsVersion = new string(context->OsVersion);
+            bootstrapperContext.Architecture = new string(context->Architecture);
+            bootstrapperContext.RuntimeDirectory = new string(context->RuntimeDirectory);
+            bootstrapperContext.ApplicationBase = new string(context->ApplicationBase);
+            bootstrapperContext.TargetFramework = new FrameworkName(FrameworkNames.LongNames.DnxCore, new Version(5, 0));
+            bootstrapperContext.HandleExceptions = context->HandleExceptions;
+
             return RuntimeBootstrapper.Execute(
                 arguments,
-                new FrameworkName(FrameworkNames.LongNames.DnxCore, new Version(5, 0)));
+                bootstrapperContext.TargetFramework,
+                bootstrapperContext);
         }
         catch (Exception ex)
         {
--- a/src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
+++ b/src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
@@ -57,7 +57,7 @@ public class EntryPoint
         // Set the default lib path to be next to the entry point location
         Environment.SetEnvironmentVariable(EnvironmentNames.DefaultLib, Path.GetDirectoryName(typeof(EntryPoint).Assembly.Location));
         Environment.SetEnvironmentVariable(EnvironmentNames.ConsoleHost, "1");
-        
+
         var p = Environment.OSVersion.Platform;
         if (p != PlatformID.MacOSX && p != PlatformID.Unix)
         {
@@ -76,7 +76,8 @@ public class EntryPoint
 
         return RuntimeBootstrapper.Execute(arguments,
             // NOTE(anurse): Mono is always "dnx451" (for now).
-            new FrameworkName("DNX", new Version(4, 5, 1)));
+            new FrameworkName("DNX", new Version(4, 5, 1)),
+            null);
     }
 
     private static string[] ExpandCommandLineArguments(string[] arguments)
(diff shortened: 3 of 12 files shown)


src/Microsoft.Dnx.Host/BootstrapperContext.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        public string Architecture { get; set; }
+
+        public uint NumProcessors { get; set; }
+
+        public string RuntimeDirectory { get; set; }
+
+        public string ApplicationBase { get; set; }
+
+        // TargetFramework?
```
moozzyk: @anurse: I am thinking about passing TargetFramework in this struct instead of as a separate parameter - sounds good?

src/Microsoft.Dnx.Host/BootstrapperContext.cs
```diff
@@ -0,0 +1,23 @@
...
+        // IsConsole?
```
moozzyk: @davidfowl: Do we need this flag anymore - I think it was mostly for Helios?

src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
```diff
@@ -76,7 +76,8 @@ public static int Main(string[] arguments)
 
         return RuntimeBootstrapper.Execute(arguments,
             // NOTE(anurse): Mono is always "dnx451" (for now).
-            new FrameworkName("DNX", new Version(4, 5, 1)));
+            new FrameworkName("DNX", new Version(4, 5, 1)),
+            null);
```
moozzyk: This just makes Mono pass.

src/Microsoft.Dnx.Host/BootstrapperContext.cs
davidfowl: We need it but you can rename it if you want something like throw exceptions.

src/Microsoft.Dnx.Host/BootstrapperContext.cs
davidfowl: Yes please

src/Microsoft.Dnx.Host/BootstrapperContext.cs
moozzyk: OK.

src/Microsoft.Dnx.Host.CoreClr/DomainManager.cs
```diff
@@ -11,8 +12,24 @@
...
+    [StructLayout(LayoutKind.Sequential)]
+    private unsafe struct NativeBootstrapperContext
+    {
+        public char* OperatingSystem;
+
+        public char* OsVersion;
+
+        public char* Architecture;
+
+        public uint NumProcessors;
```
davidfowl: Why do we need this?

moozzyk: @DamianEdwards asked for that - they wanted to use this in the perf tests I believe.

DamianEdwards: I specifically asked for number of physical processors, which you can't get from .NET (`Environment`). We'd use this by default in Kestrel to set the number of threads to use to match the number of physical cores in the machine.

src/Microsoft.Dnx.Host/BootstrapperContext.cs
moozzyk: OK. But it will be separate - it would pollute things too much at this stage.

src/Microsoft.Dnx.Host.Clr/RuntimeBootstrapper.cs
```diff
@@ -9,23 +9,35 @@ internal class RuntimeBootstrapper
...
+        private static Type GetBootstrapperType(Assembly dnxHost)
         {
-            var dnxHost = Assembly.Load("Microsoft.Dnx.Host");
             return dnxHost.GetType("Microsoft.Dnx.Host.RuntimeBootstrapper");
         }
+
+        private static object GetBootstrapperContext()
+        {
+            var dnxHost = Assembly.Load("Microsoft.Dnx.Host");
+            dynamic bootstrapperContext = Activator.CreateInstance(dnxHost.GetType("Microsoft.Dnx.Host.BootstrapperContext"));
```
anurse: `dynamic` shouldn't be needed here, right? You aren't actually calling anything.

src/Microsoft.Dnx.Host.CoreClr/DomainManager.cs
anurse: If that's the case, can we clarify the property name? `PhysicalProcessorCount`? (I'd also like to avoid using prefixes like `Num` that we don't use elsewhere, even though this is private ;))

src/dnx.common/include/app_main.h
```diff
@@ -21,3 +21,13 @@ typedef struct CALL_APPLICATION_MAIN_DATA
 #endif
     *FnCallApplicationMain)(
     PCALL_APPLICATION_MAIN_DATA pCallApplicationMainData);
+
+struct bootstrapper_context
```
anurse: So maybe I'm being paranoid, but I'd really like to try and keep these in sync. Could we put this and the matching C# file together somehow? Or at the very least put a big giant comment indicating that these two MUST be kept in sync.

If I had my way and the necessary tools existed, I'd have one of the two code-genned from the other ;). Maybe Roslyn could be used to generate the unmanaged struct? Anyway, that can wait.

src/Microsoft.Dnx.Host.CoreClr/DomainManager.cs
```diff
@@ -11,8 +12,24 @@
 [SecurityCritical]
 sealed class DomainManager
 {
+    [StructLayout(LayoutKind.Sequential)]
+    private unsafe struct NativeBootstrapperContext
+    {
+        public char* OperatingSystem;
```
anurse: As useful as this data is, and we should keep it, I think I'd prefer to have Runtime ID be a first-class member of this struct. Rather than just providing the components (OS Name, Ver, Arch, etc.), I'd like a field for the `RuntimeIdentifier` itself.

src/dnx.common/utils.cpp
```diff
@@ -114,5 +114,44 @@ namespace dnx
...
+            if (version_info.dwMajorVersion == 10)
+            {
+                return L"10";
+            }
+
+            if (version_info.dwMajorVersion == 6)
+            {
+                if (version_info.dwMinorVersion == 3)
+                {
+                    return L"81";
```
anurse: Can we reformat this into a full version number? "8.1", "10.0", etc. for example. I realize not all OSes will fit this criteria but I'd generally like the version numbers to be parsable by System.Version, etc.

If we need to shrink it back down to "81", we can do it after having parsed it.

src/dnx.coreclr/dnx.coreclr.cpp
```diff
@@ -242,8 +243,30 @@ HRESULT ExecuteMain(ICLRRuntimeHost2* pCLRRuntimeHost, PCALL_APPLICATION_MAIN_DA
         return hr;
     }
 
+    auto windows_version = dnx::utils::get_windows_version();
+    if (!windows_version)
+    {
+        trace_writer.write(L"Unexpected windows version", false);
```
anurse: Could we proceed in this case? We could simply use "win-x86" (etc.) as the RID...

src/Microsoft.Dnx.Host.Clr/RuntimeBootstrapper.cs
moozzyk: Not yet but I will need to populate the instance and I only can use reflection so dynamic will make it easier and nicer.

davidfowl: Let's not use dynamic

src/Microsoft.Dnx.Host.CoreClr/DomainManager.cs
moozzyk: This is actually the number of cores :) I can rename it to `PhysicalProcessorCount`

src/dnx.common/include/app_main.h
moozzyk: won't fix :) - this is in the .h so that it can be used across the entire C++ codebase. We already have something similar in Helios etc. If both definitions are not in sync you will learn it quick (and hard). I will add a comment.

src/Microsoft.Dnx.Host.CoreClr/DomainManager.cs
moozzyk: No. This is up to the managed code to decide what to do with this. Native code is supposed to bootstrap the runtime and give all the required data to the managed code without processing it. In other words it should not know concepts (like Runtime ID which are managed stack concepts)

src/dnx.common/utils.cpp
moozzyk: I am all for it. This would be inline with my other comment that in native code we don't want to the logic that belongs to managed code. 
Is `8` just `8` or `8.0`?

src/dnx.coreclr/dnx.coreclr.cpp
moozzyk: Possibly. My understanding this can only happen if someone runs pre-win7. My understanding is that post win10 will return win10 (or win8?) until we update the manifest.

src/Microsoft.Dnx.Host.Clr/RuntimeBootstrapper.cs
moozzyk: OK. I will use reflection to set the properties (for Desktop CLR we will flow the by extending ApplicationMainInfo struct https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Host.Clr/DomainManager.cs#L167

src/dnx.common/include/app_main.h
anurse: A comment is good enough for now.

src/dnx.common/utils.cpp
anurse: `8.0`

src/dnx.coreclr/dnx.coreclr.cpp
anurse: That's a good point. Let's keep it this way.

src/Microsoft.Dnx.Host.CoreClr/DomainManager.cs
anurse: I guess so. Runtime ID is supposed to basically be an identifier of the CLR and Native platform being used so I still think it's part of the native stack, but I can see it both ways.

anurse: Having the version number be formatted like a System.Version mostly removes my concern here though. The version number format "81" was a RID-specific concept so if that was leaking in I figured the whole thing should. Since it's not, I'm OK leaving this as-is

src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
davidfowl: Does it break mono?

src/Microsoft.Dnx.Host/BootstrapperContext.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    public class BootstrapperContext
+    {
+        public string OperatingSystem { get; set; }
+
+        public string OsVersion { get; set; }
+
+        public string Architecture { get; set; }
+
+        public uint PhysicalProcessorCount { get; set; }
```
davidfowl: Why do we need this?

anurse: Are you serious? Didn't we literally just cover this in https://github.com/aspnet/dnx/pull/2705#discussion_r39565970 ?

src/Microsoft.Dnx.Host.CoreClr/DomainManager.cs
davidfowl: Let the CLR team handle this. We shouldn't be exposing this from our APIs

src/Microsoft.Dnx.Host/BootstrapperContext.cs
davidfowl: I didn't see the reply. We should remove it.

src/Microsoft.Dnx.Host.CoreClr/DomainManager.cs
anurse: That makes sense. File/link a corefx bug?

src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
moozzyk: No. At the moment I am working just on passing the struct and not on using what is passed so null is fine here.

src/Microsoft.Dnx.Host/BootstrapperContext.cs
moozzyk: Let's sync up with @DamianEdwards

src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
davidfowl: I believe that when travis passes :stuck_out_tongue_closed_eyes:

moozzyk: :dragon: on linux/mac ATM:

```
Unhandled Exception: System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at System.String.wcslen(Char* ptr)
   at System.String.CtorCharPtr(Char* ptr)
   at DomainManager.Execute(Int32 argc, Char** argv, NativeBootstrapperContext context)
```

davidfowl:
@moozzyk This is merged right?

moozzyk:
@davidfowl  yes (forgot to force push after rebase to close the PR). And the build passed on the CI.

Issue #2671, opened 2015-09-10
Updating search path in global.json won't invalidate lock in case where a package reference can change to a project reference.


PradeepKadubandi:
I showed Troy the issue.

When we send 'RefreshDependencies' message to DTH, we are not getting the dependency message back. This is the log from DTH events:

```
Information: [ProcessingQueue]: OnReceive((19dd91876b82443881b571787736a990, RefreshDependencies, 1) -> null)
Information: [ApplicationContext]: Begin draining inbox.
Information: [ApplicationContext]: Received RefreshDependencies
Information: [ApplicationContext]: Finish draining inbox.
Information: [ApplicationHostContext] Using Lock File Target: DNX,Version=v4.5.1/
Information: [ApplicationHostContext] Using Lock File Target: DNX,Version=v4.5.1/
Information: [ApplicationHostContext] Using Lock File Target: DNX,Version=v4.5.1/win7-x86
Information: [LibraryExporter]: Resolving references for 'WebApplication1' 
Information: [LibraryExporter]: Resolved 137 references for 'WebApplication1' in 1ms
Information: [ApplicationHostContext] Using Lock File Target: DNXCore,Version=v5.0/
Information: [ApplicationHostContext] Using Lock File Target: DNXCore,Version=v5.0/
Information: [ApplicationHostContext] Using Lock File Target: DNX,Version=v4.5.1/win7-x86
Information: [LibraryExporter]: Resolving references for 'WebApplication1' 
Information: [LibraryExporter]: Resolved 182 references for 'WebApplication1' in 2ms
Information: [ApplicationContext]: OnTransmit(ProjectInformation)
Information: [ProcessingQueue]: Send((19dd91876b82443881b571787736a990, ProjectInformation, 1) -> {
  "Name": "WebApplication1",
  "Frameworks": [
    {
      "FrameworkName": "DNX,Version=v4.5.1",
      "FriendlyName": "DNX 4.5.1",
      "ShortName": "dnx451",
      "RedistListPath": "C:\\Program Files (x86)\\Reference Assemblies\\Microsoft\\Framework\\.NETFramework\\v4.5.1\\RedistList\\FrameworkList.xml"
    },
    {
      "FrameworkName": "DNXCore,Version=v5.0",
      "FriendlyName": "DNX Core 5.0",
      "ShortName": "dnxcore50",
      "RedistListPath": null
    }
  ],
  "Configurations": [
    "Debug",
    "Release"
  ],
  "Commands": {
[...]

PR #2709, opened 2015-09-15
[Prototype] Invalidate lock file when search paths in global.json is changed

troydai:
Addressing #2671 

/cc @davidfowl @yishaigalatzer @emgarten


troydai:
I have a follow up change to move the search paths to the end of the lock file.

src/Microsoft.Dnx.Runtime/DependencyManagement/LockFile.cs
```diff
@@ -31,6 +31,35 @@ public bool IsValidForProject(Project project, out string message)
                 return false;
             }
 
+            // Verify the global search paths
+            message = $"The project search paths defined in {GlobalSettings.GlobalFileName} have changed";
+            GlobalSettings globalSettings;
+            if (GlobalSettings.TryGetGlobalSettings(ProjectRootResolver.ResolveRootDirectory(project.ProjectDirectory), out globalSettings))
```
davidfowl: We already resolve the root directory. Why not pass it in? I'd also measure the performance difference this additional file read adds. We already parse it to figure out the nuget path.

src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
```diff
@@ -811,10 +811,17 @@ private Task<LockFile> ReadLockFile(string projectLockFilePath)
             var lockFile = new LockFile();
             lockFile.Islocked = Lock;
 
+            // Set the search paths defined in global settings
+            GlobalSettings globalSettings;
+            if (GlobalSettings.TryGetGlobalSettings(ProjectRootResolver.ResolveRootDirectory(project.ProjectDirectory), out globalSettings))
```
davidfowl: Doesn't this get resolved elsewhere?

troydai:
We have a new solution to the issue.

PR #2710, opened 2015-09-15
Quiet restoration during `--quiet` builds

dougbu:
- use `k-restore` command with new `restoreDir` argument
  - new hook from KoreBuild; see aspnet/Universe@22952e1

(Note this does not mean `build --quiet` commands are actually _quiet_. The functional tests are far too noisy for that :smiling_imp:)


muratg:
Sounds good to me. If package restore fails, we'll still see it thought right? Assuming that's the case, :shipit:

dougbu:
b8fc4b2

PR #2713, opened 2015-09-16
Enabling attaching a debugger to CoreClr dnx apps

### LINKED ISSUE #2565: Consider moving --debug up the stack
Currently we check for the --debug flag as soon as dnx starts. If the flag is specified dnx waits for the debugger to be attached and as soon as it happens it continues execution. 
However there is a limitation on CoreClr where there won't be any debugger events if CoreClr code type is selected until CoreClr is loaded. This results in dnx never waking up despite attaching the debugger to the process if CoreClr code type is selected. The process will wake up when only native code is selected but then it is not possible to debug managed code at all. 

Additional considerations:
- if we move --debug up the stack we may need to add an internal flag that could be used for debugging native code (i.e. bootstrapping) 
- currently --debug flag does not work on non-windows at all  (https://github.com/aspnet/dnx/blob/dev/src/dnx/pal.linux.cpp#L29) but we bootstrap CoreClr differently for non-windows so we need to make sure it works too

moozzyk:
Fixes #2565

### DIFF
--- a/src/Microsoft.Dnx.Host/RuntimeBootstrapper.cs
+++ b/src/Microsoft.Dnx.Host/RuntimeBootstrapper.cs
@@ -81,6 +81,8 @@ namespace Microsoft.Dnx.Host
                 CommandOptionType.MultipleValue);
             var optionDebug = app.Option("--debug", "Waits for the debugger to attach before beginning execution.",
                 CommandOptionType.NoValue);
+            var optionBootstrapperDebug = app.Option("--bootstrapper-debug", "Waits for the debugger to attach before bootstrapping runtime.",
+                CommandOptionType.NoValue);
 #if DNX451
             var optionFramework = app.Option("--framework <FRAMEWORK_ID>", "Set the framework version to use when running (i.e. dnx451, dnx452, dnx46, ...)", CommandOptionType.SingleValue);
 #endif
--- a/src/dnx.coreclr/dnx.coreclr.cpp
+++ b/src/dnx.coreclr/dnx.coreclr.cpp
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 #include "stdafx.h"
-
 #include "dnx.coreclr.h"
 #include "tpa.h"
 #include "utils.h"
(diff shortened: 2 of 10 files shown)


src/dnx.coreclr/dnx.coreclr.cpp
```diff
@@ -162,6 +161,44 @@ HRESULT StopClrHost(ICLRRuntimeHost2* pCLRRuntimeHost)
...
+            {
+                std::wcout << L"Process Id: " << GetCurrentProcessId() << std::endl;
+                std::wcout << L"Waiting for the debugger to attach..." << std::endl;
+
+                while (!IsDebuggerPresent())
+                {
+                    Sleep(250);
+                }
+
+                std::wcout << "Debugger attached." << std::endl;
```
rynowak: Needs more `L`

moozzyk: Yes. Will fix.

davidfowl:
Travis is legit broken on linux

moozzyk:
it's legit fixed now (it was a missing include)

anurse:
Can you confirm my understanding of the switches?

`--debug` - Runs until the earliest point at which managed debugging can be initialized and waits for the debugger (for CoreCLR this is after CLR boot, for Desktop this is right away)

`--bootstrapper-debug` - Pauses immediately after launching, regardless of CLR status.

anurse:
If I'm right, maybe we should just move `--debug` to the first managed code executed. Also, in that case, I'd suggest we do not put `--bootstrapper-debug` in the help text since it is purely for our debugging, not for debugging user apps.

moozzyk:
Yes, since --debug works just fine Desktop CLR and we can leave it as is - especially that it makes bootstrapper debugging easy - you can debug both managed and native code. For Desktop Clr there is no difference between --debug and --bootstrapper-debug. For CoreClr the --bootstrapper-debug option allows debugging the native code. The caveat is that you can only have Native Code selected as code type in VS. After Core Clr is initialized we can attach debugger that has either just Managed/Core Clr or both Native and Managed/CoreClr selected as code types. 
This is also the reason why the code is in the native part. WIth --debug we are now able to debug the transition from native and managed. If we waited for the debugger in the managed code we would end up having a gap where --bootstrapper-debug cannot get you beyond the native code and --debug does not allow you to debug anything before - yes this is the marshaling code you looked at yesterday.
Another caveat is that we don't know debugging is supposed to work on non-windows. 

Currently having internal/undocumented options is not supported. I talked to @davidfowl yesterday and we decided to just add this option as a regular dnx option.

davidfowl:
Any tests for the parsing?

moozzyk:
If you insist I can add some tests for parsing but I don't expect this code to change often if at all. In fact the old code did not work in most cases (when you have an option with an argument that was not --appbase and --debug was after) and no one noticed that. The function that I moved to common is mostly covered by tests added when removing `.` from `dnx . run`

davidfowl:
@moozzyk As long as we have things that exercise this code path (and you tested it locally), I'm fine with it. There is some new logic (it seems like) wrt parsing the args.

moozzyk:
@davidfowl - updated the PR to remove the logic to parse command line. We already had the code that does it so I just moved it to dnx::common and use it. For desktop clr we will now go twice over the parameters but I don't think it is expensive here given that  in 80% case there is less than 5 parameters.

src/Microsoft.Dnx.Host/RuntimeBootstrapper.cs
```diff
@@ -81,6 +81,8 @@ public static Task<int> ExecuteAsync(string[] args, FrameworkName targetFramewor
                 CommandOptionType.MultipleValue);
             var optionDebug = app.Option("--debug", "Waits for the debugger to attach before beginning execution.",
                 CommandOptionType.NoValue);
+            var optionBootstrapperDebug = app.Option("--bootstrapper-debug", "Waits for the debugger to attach before bootstrapping runtime.",
```
troydai: The option is way to long to type easily...

troydai:
Looks good to me.

src/Microsoft.Dnx.Host/RuntimeBootstrapper.cs
moozzyk: it does not matter too much - I am probably the only one who would use it...

PR #2718, opened 2015-09-16
Fix ArgumentException thrown when targeting Client Profile

### LINKED ISSUE #2546: project.json net40 client
Should project.json support .NET 4.0 Client Framework?

With `.NETFramework,Version=v4.0,Profile=Client` I get an error from dnu build:

System.ArgumentException: preprocessorSymbols

Microsoft .NET Development Utility CLR-x86-1.0.0-beta8-15120

anurse:
/cc @davidfowl 

I'm working on porting the test to the new test model but I'm actually seeing if I can just blitz that whole file. If it takes a while, I'll give up.

Fixes #2546


anurse:
Travis failure is because the test is running on Mono but shouldn't. I can't fix it right this second but it shouldn't block review. I'll add a framework skip condition to it.

src/Microsoft.Dnx.Runtime/Internal/FrameworkNameHelper.cs
```diff
@@ -36,15 +37,31 @@ public static FrameworkName ParseFrameworkName(string targetFramework)
 
         public static string MakeDefaultTargetFrameworkDefine(Tuple<string, FrameworkName> frameworkDefinition)
         {
-            var shortName = frameworkDefinition.Item1;
+            var shortName = VersionUtility.GetShortFrameworkName(frameworkDefinition.Item2);
             var targetFramework = frameworkDefinition.Item2;
 
-            if (VersionUtility.IsPortableFramework(targetFramework))
+            if(targetFramework.IsPortableFramework())
```
davidfowl: spaces

test/Microsoft.Dnx.Runtime.Tests/ProjectExtensionsFacts.cs
```diff
@@ -58,6 +58,27 @@ public void GetCompilerOptionsCombinesTargetFrameworkIfNotNull()
             Assert.Equal("x86", options.Platform);
         }
 
+        [Theory]
+        [InlineData("net40", "NET40")]
+        [InlineData(".NETFramework,Version=v4.0", "NET40")]
+        [InlineData(".NETFramework,Version=v4.0,Profile=Client", "NET40_CLIENT")]
+        [InlineData("DOTNET5.1", "DOTNET5_1")]
+        [InlineData(".NETPortable,Version=v4.5,Profile=Profile123", null)]
```
davidfowl: Add portable-x+y+z or whatever it is we parse

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
```diff
@@ -6,575 +6,596 @@
...
-                    Assert.Empty(stdError);
-                    Assert.Equal(0, exitCode);
-                    Assert.True(Directory.Exists(Path.Combine(workingDir, "bin")));
-                    Assert.True(File.Exists(Path.Combine(workingDir, "bin", "Debug", "dnx451", "fr-FR", "ResourcesLibrary.resources.dll")));
-                    Assert.True(File.Exists(Path.Combine(workingDir, "bin", "Debug", "dnxcore50", "fr-FR", "ResourcesLibrary.resources.dll")));
-                }
-            }
-        }
+        //    [Theory]
+        //    [MemberData(nameof(RuntimeComponents))]
```
davidfowl: Whut?

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
```diff
@@ -6,575 +6,596 @@
...
-using Microsoft.Dnx.CommonTestUtils;
+using Microsoft.Dnx.Testing;
+using Newtonsoft.Json.Linq;
 using NuGet;
 using Xunit;
 
 namespace Microsoft.Dnx.Tooling
 {
-    public class DnuPackTests
+    public class DnuPackTests : DnxSdkFunctionalTestBase
```
davidfowl: So I'm thinking we don't generate projects on the fly anymore. We can use the framework to put more projects in misc for each of the scenarios we care about but not during the test itself.

test/Microsoft.Dnx.Runtime.Tests/ProjectExtensionsFacts.cs
anurse: Well that was the gist of my question yesterday. We don't currently support `portable-x+y+z` in project.json.

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
anurse: I'm on board, but `misc` is a terrible name for a folder that contains projects required by tests...

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
anurse: See my comment about porting to the new framework. I'm rolling that back anyway and just porting the one new test. I figured that rewriting each file as we touch it would be a reasonable balance between a big-scale rewrite and only using the framework for new tests, but I'm just not sure I have the time.

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
davidfowl: Who made it in the first place?

anurse: I think it first appeared for `dnu wrap` tests.

davidfowl: Wei!

davidfowl:
> Travis failure is because the test is running on Mono but shouldn't. I can't fix it right this second but it shouldn't block review. I'll add a framework skip condition to it.

Which tests should run on mono?

anurse:
Client Profile doesn't seem to work properly on Mono.

anurse:
Fowler said I could :shipit: when travis passed soooo ;)

PR #2721, opened 2015-09-17
Pawelka/p invoke darwin

moozzyk:


src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -233,6 +237,113 @@ int InvokeDelegate(host_main_fn host_main, int argc, const char** argv)
...
 
+#if defined(PLATFORM_LINUX)
+
+std::string get_os_version()
+{
+    std::vector<std::string> qualifiers { "DISTRIB_ID=", "DISTRIB_RELEASE=" };
+
+    std::ifstream lsb_release;
+    lsb_release.open("/etc/lsb-release", std::ifstream::in);
+    if (lsb_release.is_open())
```
moozzyk: This returns "Ubuntu 15.04". I am not convinced that returning anything more than distro is needed or useful. Someone (@anurse) let me know.

src/Microsoft.Dnx.Host.CoreClr/DomainManager.cs
```diff
@@ -49,6 +49,14 @@ unsafe static int Execute(int argc, char** argv, NativeBootstrapperContext* cont
             bootstrapperContext.TargetFramework = new FrameworkName(FrameworkNames.LongNames.DnxCore, new Version(5, 0));
             bootstrapperContext.HandleExceptions = context->HandleExceptions;
 
+            Logger.TraceInformation($"Operating System: {bootstrapperContext.OperatingSystem}");
```
davidfowl: We shouldn't log this here if application host logs it. We should do it in Dnx.Host so it's consistent.

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -274,9 +387,20 @@ extern "C" int CallApplicationMain(CALL_APPLICATION_MAIN_DATA* data)
...
-    auto result = CallMain(data);
+    MultiByteToWideChar = (MultiByteToWideChar_fn)dlsym(pLibCoreClr, "MultiByteToWideChar");
+    if (!MultiByteToWideChar)
+    {
+        fprintf(stderr, "Could not find MultiByteToWideChar entrypoint in coreclr\n");
+        result = 1;
+    }
+    else
+    {
+        setenv("DNX_FRAMEWORK", "dnxcore50", 1);
```
davidfowl: Why are we still setting this here?

src/Microsoft.Dnx.Host.CoreClr/DomainManager.cs
moozzyk: This is superhelpful to see what actually was passed from native parts without attaching the debugger (and on Mac/Linux you cannnot debug managed (CoreClr)). At the moment this information does not reach the host and when it does it will be broken down to environment and appenvironment etc. I want to leave it here at least for now.

src/dnx.coreclr.unix/dnx.coreclr.cpp
moozzyk: Because  the change is just about passing data from native to managed and not about actually using this data (this is why I could pass null in Mono). We are still using the environment variables across the stack since not all bootstrappers pass the data yet (Clr work is in progress and Mono is left). When all bootstrappers pass the struct I will set environment appenvironment properties accordingly and remove all env variables.

src/dnx.coreclr.unix/dnx.coreclr.cpp
anurse: Yeah, return `DISTRIB_ID` and `DISTRIB_RELEASE` (and OS Name and Version respectively). That should be fine.

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -233,6 +237,113 @@ int InvokeDelegate(host_main_fn host_main, int argc, const char** argv)
...
+{
+    auto dot_position = release.find(".");
+    if (dot_position == std::string::npos)
+    {
+        fprintf(stderr, "Could not determine os version.\n");
+        return "10.1";
+    }
+
+    auto version = stoi(release.substr(0, dot_position));
+    return std::string("10.").append(std::to_string(version - 4));
```
anurse: Can you put a comment here explaining this? I figured it out (10.1 was Darwin 5.0, and so on) but it would be useful to have it recorded here for posterity :)

anurse:
:shipit: after addressing feedback.

src/Microsoft.Dnx.Host.CoreClr/DomainManager.cs
davidfowl: We shouldn't have it. Can you show what a log with DNX_TRACE looks like with ApplicationHost also?

davidfowl: And show mono, and desktop clr as well.

src/dnx.coreclr.unix/dnx.coreclr.cpp
moozzyk: Yes. will do. I took it from here: http://stackoverflow.com/questions/11072804/how-do-i-determine-the-os-version-at-runtime-in-os-x-or-ios-without-using-gesta. This table is especially useful:

15.x.x  OS X 10.11.x El Capitan
14.x.x  OS X 10.10.x Yosemite
13.x.x  OS X 10.9.x Mavericks
12.x.x  OS X 10.8.x Mountain Lion
11.x.x  OS X 10.7.x Lion
10.x.x  OS X 10.6.x Snow Leopard
 9.x.x  OS X 10.5.x Leopard
 8.x.x  OS X 10.4.x Tiger
 7.x.x  OS X 10.3.x Panther
 6.x.x  OS X 10.2.x Jaguar
 5.x    OS X 10.1.x Puma

src/Microsoft.Dnx.Host.CoreClr/DomainManager.cs
moozzyk: Mono and CoreClr are in the works.

src/dnx.coreclr.unix/dnx.coreclr.cpp
anurse: Yeah, I found a similar table on Wikipedia. In theory, we should be using the table but since it's always `10.(Darwin Version - 4)`, this approach seems fine in practice.

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -244,7 +355,7 @@ int CallMain(CALL_APPLICATION_MAIN_DATA* data)
         fprintf(stderr, "Failed to initialize runtime: 0x%08x\n", result);
         return 1;
     }
-
+
```
moozzyk: remove

src/dnx.coreclr.unix/dnx.coreclr.cpp
moozzyk: What do you mean by "OS Name and Version"? OS Name has a separate property but the value returned by this function goes to OsVersion. So in this case you will get:

```
Information: Operating System: Linux
Information: Operating System Version: Ubuntu 15.04
```

src/dnx.coreclr.unix/dnx.coreclr.cpp
anurse: Woah... did you just comment on your own PR?

src/Microsoft.Dnx.Host.CoreClr/DomainManager.cs
moozzyk: This is how the output currently looks like: 
```
PS C:\Temp\TestProject> dnx run
The default servicing root does not exist.
Loaded module: dnx.onecore.coreclr.dll
Found export: CallApplicationMain
Information: [DomainManager] Using CoreCLR
Information: Operating System: Windows
Information: Operating System Version: 10.0
Information: Architecture: x64
Information: Runtime Directory: C:\Users\pawelka\.dnx\runtimes\dnx-coreclr-win-x64.1.0.0-dev\bin\
Information: Application Base: C:\Temp\TestProject
Information: Target Framework: DNXCore,Version=v5.0
Information: Handle Exceptions: True
Information: [LoaderContainer]: Load name=Microsoft.Dnx.ApplicationHost, Culture=neutral, PublicKeyToken=null
Information: [PathBasedAssemblyLoader]: Loaded name=Microsoft.Dnx.ApplicationHost, Culture=neutral, PublicKeyToken=null in 29ms
Information: [Bootstrapper] Runtime Framework: DNXCore,Version=v5.0
Information: [LoaderContainer]: Load name=Microsoft.Dnx.Compilation.Abstractions, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null
...
```

src/dnx.coreclr.unix/dnx.coreclr.cpp
moozzyk: I wanted to earn the _self-critic_ bage...

src/dnx.coreclr.unix/dnx.coreclr.cpp
davidfowl: Ok

davidfowl:
:shipit:

PR #2722, opened 2015-09-17
Remove package references from .Sources projects

kichalla:
@rynowak @davidfowl

### DIFF
--- a/src/Microsoft.Dnx.Compilation.CSharp.Common/project.json
+++ b/src/Microsoft.Dnx.Compilation.CSharp.Common/project.json
@@ -1,11 +1,18 @@
 {
     "version": "1.0.0-*",
     "description": "ASP.NET 5 Roslyn implementation code shared with libraries performing runtime compilation.",
-    "compilationOptions": { "define": [ "TRACE" ], "allowUnsafe": true, "warningsAsErrors": true },
+    "compilationOptions": {
+        "define": [ "TRACE" ],
+        "allowUnsafe": true,
+        "warningsAsErrors": true
+    },
     "dependencies": {
         "Microsoft.CodeAnalysis.CSharp": "1.1.0-beta1-*",
         "Microsoft.Dnx.Compilation.CSharp.Abstractions": "1.0.0-*",
-        "Microsoft.Dnx.Runtime.Sources": { "version": "1.0.0-*", "type": "build" }
+        "Microsoft.Dnx.Runtime.Sources": {
+            "version": "1.0.0-*",
+            "type": "build"
+        }
     },
     "frameworks": {
         "dnx451": {
@@ -16,6 +23,7 @@
         },
         "dnxcore50": {
             "dependencies": {
+                "System.Console": "4.0.0-beta-*",
                 "System.IO": "4.0.11-*"
             }
         }
--- a/src/Microsoft.Dnx.Runtime.Sources/project.json
+++ b/src/Microsoft.Dnx.Runtime.Sources/project.json
@@ -1,6 +1,9 @@
 {
     "version": "1.0.0-*",
-    "compilationOptions": { "define": [ "TRACE" ], "warningsAsErrors": true },
+    "compilationOptions": {
+        "define": [ "TRACE" ],
+        "warningsAsErrors": true
+    },
     "shared": "**/*.cs",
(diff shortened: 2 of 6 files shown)


src/Microsoft.Dnx.Compilation.CSharp.Common/project.json
```diff
@@ -16,6 +23,7 @@
         },
         "dnxcore50": {
             "dependencies": {
+                "System.Console": "4.0.0-beta-*",
```
rynowak: This seems like an odd dep for this package - what's it used for?

src/Microsoft.Framework.CommandLineUtils.Sources/project.json
```diff
@@ -1,12 +1,14 @@
...
     "shared": "**/*.cs",
-    "frameworks" : {
-        "dnx451" : { },
-        "dnxcore50" : { 
+    "frameworks": {
+        "dnx451": { },
+        "dnxcore50": {
             "dependencies": {
-                "System.Collections": "4.0.11-beta-*",
                 "System.Runtime": "4.0.21-beta-*"
```
rynowak: I expected this to just be empty. What happens without `System.Runtime`?

src/Microsoft.Dnx.Runtime.Sources/project.json
```diff
@@ -8,16 +11,7 @@
...
-                "System.ComponentModel": "4.0.1-beta-*",
-                "System.Console": "4.0.0-beta-*",
-                "System.Diagnostics.Debug": "4.0.11-beta-*",
-                "System.Linq": "4.0.1-beta-*",
-                "System.Reflection": "4.0.11-beta-*",
-                "System.Runtime": "4.0.21-beta-*",
-                "System.Runtime.Extensions": "4.0.11-beta-*",
-                "System.Threading": "4.0.11-beta-*",
-                "System.Threading.Tasks": "4.0.11-beta-*"
+                "System.Runtime": "4.0.21-beta-*"
```
rynowak: I expected this to just be empty. What happens without `System.Runtime`?

davidfowl: You need it. These packages still end up turning into dlls (unfortunately)

rynowak: :crying_cat_face:

src/Microsoft.Dnx.Compilation.CSharp.Common/project.json
davidfowl: Logging

src/Microsoft.Dnx.Runtime.Sources/project.json
dougbu: @davidfowl VS also seems to _compile_ `.Sources` projects by default. Does that imply we need to keep some of these dependencies? Or should @kichalla disable building the projects in the `.sln` files (here `DNX.sln`)?

davidfowl: @kichalla shouldn't disable anything. Lets keep this change isolated.

rynowak: He's not talking about the assembly-in-sources-package issue.

VS will show you build errors unless you tell it not to build the project in the solution config.

src/Microsoft.Dnx.Compilation.CSharp.Common/project.json
kichalla: Yes, it for Logger which is part of `Dnx.Runtime.Sources`(which this project references)

src/Microsoft.Framework.CommandLineUtils.Sources/project.json
kichalla: I get a build error saying `error CS8021: No value for RuntimeMetadataVersion found. No assembly containing System.Object was found nor was a value for RuntimeMetadataVersion specified through options.`

kichalla:
@rynowak @davidfowl any more comments?

davidfowl:
:shipit:

kichalla:
1728b3aa4281aac968310e846ca8fe94034869cc

PR #2732, opened 2015-09-17
Fix runtime restore bug (inconsistency with NuGet 3)

anurse:
This should be the issue making us inconsistent with NuGet 3. Not sure if it will fix the clrcompression issue, but if it doesn't we can debug further.

/cc @davidfowl @BrennanConroy @muratg

### DIFF
--- a/src/Microsoft.Dnx.Tooling/Restore/RestoreOperations.cs
+++ b/src/Microsoft.Dnx.Tooling/Restore/RestoreOperations.cs
@@ -26,6 +26,47 @@ namespace Microsoft.Dnx.Tooling
             var sw = new Stopwatch();
             sw.Start();
 
+            var dependencies = new List<LibraryDependency>();
+            if (context.RuntimeSpecs != null && !string.IsNullOrEmpty(context.RuntimeName))
+            {
+                // HACK(davidfowl): This is making runtime.json support package redirects
+
+                // Look up any additional dependencies for this package
+                foreach (var runtimeSpec in context.RuntimeSpecs)
+                {
+                    DependencySpec dependencyMapping;
+                    if (runtimeSpec.Dependencies.TryGetValue(libraryRange.Name, out dependencyMapping))
+                    {
+                        foreach (var runtimeDependency in dependencyMapping.Implementations.Values)
+                        {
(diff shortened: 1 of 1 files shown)


src/Microsoft.Dnx.Tooling/Restore/RestoreOperations.cs
```diff
@@ -26,6 +26,45 @@ public RestoreOperations(IReport report)
             var sw = new Stopwatch();
             sw.Start();
 
+            var dependencies = new List<LibraryDependency>();
+            if (context.RuntimeSpecs != null && !string.IsNullOrEmpty(context.RuntimeName))
+            {
+                // HACK(davidfowl): This is making runtime.json support package redirects
+
+                // Look up any additional dependencies for this package
+                var runtimeDependencies = context.RuntimeSpecs
```
davidfowl: What is this?

anurse: You tell me, it's your code :)

anurse: Ah, the LINQ. It's the equivalent of https://github.com/NuGet/NuGet3/blob/dev/src/NuGet.RuntimeModel/RuntimeGraph.cs#L147 (which is what the NuGet3 code called).

src/Microsoft.Dnx.Tooling/Restore/RestoreOperations.cs
```diff
@@ -46,8 +87,9 @@ public RestoreOperations(IReport report)
                     }
                 }
 
+                dependencies.AddRange(node.Item.Dependencies);
+
                 var tasks = new List<Task<GraphNode>>();
-                var dependencies = node.Item.Dependencies ?? Enumerable.Empty<LibraryDependency>();
```
davidfowl: Not enough old code is removed. It's still evaluating runtime specs for each dependency and that's unnecessary.

anurse: Yeah, I think my update failed to push last night ;)

davidfowl:
:shipit:

PR #2738, opened 2015-09-17
Fix error messages for unresolved/incompatible library

troydai:
The messages were reversed.

/cc @victorhurdugaci @anurse @davidfowl @muratg


src/Microsoft.Dnx.Runtime/DependencyManagement/LibraryManager.cs
```diff
@@ -123,15 +123,15 @@ public IList<DiagnosticMessage> GetAllDiagnostics()
                     string errorCode;
                     if (library.Compatible)
```
davidfowl: This code says: If you're Unresolved but Compatible then show the error:
`The dependency x in project y does not support framework f`

That doesn't make sense...

troydai:
Need to think again.

Issue #2729, opened 2015-09-17
Unresolved project path from lock file cause NullRef in ApplicationHostContext


troydai:
When a relative project path from lock file can't locate the project reference, the `ProjectDependencyProvider` returns a `null`. It then cause a _NullRef_ later in `ApplicaitonHostContext` initialization.

Instead it should be marked as unresolved.

https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Runtime/DependencyManagement/ProjectDependencyProvider.cs#L22

https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Runtime/ApplicationHostContext.cs#L199

https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Runtime/ApplicationHostContext.cs#L63

troydai:
@muratg @glennc for triage
/cc @davidfowl

troydai:
@muratg I'll assign this to me for now.

troydai:
Fixed in 6944973dd0b2b1e2a6ca942b92180a9751d1688e

PR #2739, opened 2015-09-18
Generate an unresolved ProjectDescription for unresolved path

troydai:
/review @anurse @davidfowl 
/cc @rynowak @victorhurdugaci 

Addressing issue #2729 

The change is kind of patchy, but it does fix the problem. The type `ProjectDescription` and `LibraryDescription` can be potentially optimized in the future to simplify the creation. Otherwise for adding a unresolved project I need to add an additional constructor.

The fix solves the issue in both VS and Console.

I'm working on a functional test to cover the case.

### DIFF
--- a/src/Microsoft.Dnx.DesignTimeHost/ApplicationContext.cs
+++ b/src/Microsoft.Dnx.DesignTimeHost/ApplicationContext.cs
@@ -1041,6 +1041,11 @@ namespace Microsoft.Dnx.DesignTimeHost
                 // Add shared files from projects
                 foreach (var reference in dependencyInfo.ProjectReferences)
                 {
+                    if (reference.Project == null)
+                    {
+                        continue;
+                    }
+
                     // Only add direct dependencies as sources
                     if (!project.Dependencies.Any(d => string.Equals(d.Name, reference.Name, StringComparison.OrdinalIgnoreCase)))
                     {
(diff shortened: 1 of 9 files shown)


davidfowl:
I don't trust this change and I'm super surprised nothing failed (but that might be because we guard against stuff in other places). We need to go over all of the uses of ProjectDescription and make sure they all check for null or Unresolved first.

troydai:
It doesn't fail majorly because unresolved dependencies in lock file stop the program from further consuming the project.

davidfowl:
@troydai How much did you test. Did you test all the code paths that look at ProjectReference.Project?

troydai:
Going through all the references now.
- [x] `ProjectDescription.TargetFrameworkInformation`
- [ ] `ProjectDescription.Project`
- [ ] `ProjectReference.Project`

troydai:
I'm still working on integration tests.

troydai:
:watch:

troydai:
:watch: @davidfowl

troydai:
@muratg @victorhurdugaci @BrennanConroy @anurse

src/Microsoft.Dnx.DesignTimeHost/ApplicationContext.cs
```diff
@@ -1153,6 +1158,8 @@ private DependencyInfo ResolveProjectDependencies(Project project, string config
                     var description = CreateDependencyDescription(library, ProtocolVersion);
                     info.Dependencies[description.Name] = description;
 
+                    var isProject = string.Equals(library.Type, LibraryTypes.Project);
```
davidfowl: omg

davidfowl:
:shipit:

Issue #2401, opened 2015-08-03
Add a dnu command to clear the HTTP cache


DamianEdwards:
`dnu` caches results for the various categories of HTTP requests it makes (e.g. feed check, list package, v3 feed check, etc.) for various amounts of time.

While it's possible to run a `dnu restore` at any time with the cache disabled via `--no-cache`, there isn't a way to clear the cache completely, other than just deleting the folder in the user profile.

Propose we add a command to clear the HTTP cache, e.g. `dnu --clear-cache`. Might be nice to support clearing it only for specified feeds too, e.g. `dnu --clear-cache https://www.myget.org`

This hit me today as we found an issue in the NuGet v3 feed root `index.json` file which is now fixed. However, `dnu` caches the request to that file for a week, so customers will need to manually delete the cache folder to see the change in the meantime.

glennc:
Assigning to beta8 for at least the clear-cache command, we will talk about clearing a specific URL but we might punt that part to beta8

MisterJames:
@glennc I could take this one on if you like.

 1. Can I assume that a raw `dnu --clear-cache` would clear out the `DnuFolderPath.HttpCacheDirectory` folder entirely?  Or should it clear the `list_` and `nupkg_` but preserve the dir structure and leave the `index_json.dat`?

 1. When clearing for a specific feed, would `https://www.myget.org` match all entries in the cache with the start of that domain? Or would it be only on direct matches such as `https://www.myget.org/F/aspnetvnext/api/v2/` or `https://www.myget.org/F/roslyn-nightly/`?

glennc:
I had assumed it would nuke the whole folder. But I am not familiar enough with the entire mechanism to say for certain. @davidfowl @anurse @lodejard What do you guys think?

MisterJames:
@glennc Whole folder for the base scenario, yes. But @DamianEdwards had said:

> Might be nice to support clearing it only for specified feeds too, e.g. `dnu --clear-cache https://www.myget.org`

muratg:
Done: https://github.com/aspnet/dnx/commit/3d0edc0914f341794a56fb8c0d927ddbd96dec31

PR #2742, opened 2015-09-18
Adding dnu clearcache command

muratg:
Implements https://github.com/aspnet/dnx/issues/2401

### DIFF
--- a/src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
+++ b/src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
@@ -1,0 +1,73 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+
+namespace Microsoft.Dnx.Tooling
+{
+    internal class ClearCacheCommand
+    {
+        public Reports Reports { get; }
+        public string HttpCacheDirectory { get; }
+        public ClearCacheCommand(Reports reports, string httpCacheDirectory)
+        {
+            Reports = reports;
+            HttpCacheDirectory = httpCacheDirectory;
+        }
+        
+        public int Execute()
+        {
+            if(Directory.Exists(HttpCacheDirectory))
+            {
+                Reports.Information.WriteLine($"Clearing cache directory {HttpCacheDirectory}");
+
+                Exception exception = null;
+
+                for (var i = 0; i < 3; i++)
+                {
+                    try
+                    {
+                        DeleteDirectoryRecursively(HttpCacheDirectory, false);
+                        Reports.Information.WriteLine("Cache cleared.");
+                        return 0;
+                    }
+                    catch (Exception e)
+                    {
+                        exception = e;
+                    }
+                }
+                if (exception != null)
+                {
+                    Reports.Error.WriteLine($"Unable to clear cache directory: {exception.Message}");
+                }
+                return 1;
+            }
+            else
+            {
+                Reports.Error.WriteLine($"Cache directory {HttpCacheDirectory} does not exist.");
+                return 1;
+            }
+        }
+        public void DeleteDirectoryRecursively(string baseDirectory, bool deleteBaseDirectory)
+        {
+            var files = Directory.GetFiles(baseDirectory);
+            var subDirectories = Directory.GetDirectories(baseDirectory);
+
+            foreach (var file in files)
+            {
+                File.Delete(file);
+            }
+
+            foreach(var subDirectory in subDirectories)
+            {
+                DeleteDirectoryRecursively(subDirectory, true);
+            }
+
+            if(deleteBaseDirectory)
+            {
+                Directory.Delete(baseDirectory);
+            }
+        }
+    }
+}
--- a/src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
+++ b/src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
@@ -23,32 +23,20 @@ namespace Microsoft.Dnx.Tooling
             {
                 Reports.Information.WriteLine($"Clearing cache directory {HttpCacheDirectory}");
 
-                Exception exception = null;
+                try
+                {
+                    DeleteDirectoryRecursively(HttpCacheDirectory, deleteBaseDirectory: false);
+                    Reports.Information.WriteLine("Cache cleared.");
+                    return 0;
+                }
+                catch (Exception e)
+                {
+                    Reports.Error.WriteLine($"Unable to clear cache directory: {e.Message}");
+                    return 1;
+                }
+            }
 
-                for (var i = 0; i < 3; i++)
-                {
-                    try
-                    {
-                        DeleteDirectoryRecursively(HttpCacheDirectory, deleteBaseDirectory: false);
-                        Reports.Information.WriteLine("Cache cleared.");
-                        return 0;
-                    }
-                    catch (Exception e)
-                    {
-                        exception = e;
-                    }
-                }
-                if (exception != null)
-                {
-                    Reports.Error.WriteLine($"Unable to clear cache directory: {exception.Message}");
-                }
-                return 1;
-            }
-            else
-            {
-                Reports.Error.WriteLine($"Cache directory {HttpCacheDirectory} does not exist.");
-                return 1;
-            }
+            return 0;
         }
         public void DeleteDirectoryRecursively(string baseDirectory, bool deleteBaseDirectory)
         {
--- a/src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
+++ b/src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
@@ -8,14 +8,14 @@ namespace Microsoft.Dnx.Tooling
 {
     internal class ClearCacheCommand
     {
-        public Reports Reports { get; }
-        public string HttpCacheDirectory { get; }
-
         public ClearCacheCommand(Reports reports, string httpCacheDirectory)
         {
             Reports = reports;
             HttpCacheDirectory = httpCacheDirectory;
         }
+
+        public Reports Reports { get; }
+        public string HttpCacheDirectory { get; }
 
         public int Execute()
         {
@@ -25,9 +25,8 @@ namespace Microsoft.Dnx.Tooling
 
                 try
                 {
-                    DeleteDirectoryRecursively(HttpCacheDirectory, deleteBaseDirectory: false);
+                    FileOperationUtils.DeleteFolder(HttpCacheDirectory);
                     Reports.Information.WriteLine("Cache cleared.");
-                    return 0;
                 }
                 catch (Exception e)
                 {
@@ -37,26 +36,6 @@ namespace Microsoft.Dnx.Tooling
             }
 
             return 0;
-        }
-        public void DeleteDirectoryRecursively(string baseDirectory, bool deleteBaseDirectory)
-        {
-            var files = Directory.GetFiles(baseDirectory);
-            var subDirectories = Directory.GetDirectories(baseDirectory);
-
-            foreach (var file in files)
-            {
-                File.Delete(file);
-            }
-
-            foreach (var subDirectory in subDirectories)
-            {
-                DeleteDirectoryRecursively(subDirectory, deleteBaseDirectory: true);
-            }
-
-            if (deleteBaseDirectory)
-            {
-                Directory.Delete(baseDirectory);
-            }
         }
     }
 }
--- a/src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
+++ b/src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
@@ -10,15 +10,16 @@ namespace Microsoft.Dnx.Tooling
     {
         public Reports Reports { get; }
         public string HttpCacheDirectory { get; }
+
         public ClearCacheCommand(Reports reports, string httpCacheDirectory)
         {
             Reports = reports;
             HttpCacheDirectory = httpCacheDirectory;
         }
-        
+
         public int Execute()
         {
-            if(Directory.Exists(HttpCacheDirectory))
+            if (Directory.Exists(HttpCacheDirectory))
             {
                 Reports.Information.WriteLine($"Clearing cache directory {HttpCacheDirectory}");
 
@@ -28,7 +29,7 @@ namespace Microsoft.Dnx.Tooling
                 {
                     try
                     {
-                        DeleteDirectoryRecursively(HttpCacheDirectory, false);
+                        DeleteDirectoryRecursively(HttpCacheDirectory, deleteBaseDirectory: false);
                         Reports.Information.WriteLine("Cache cleared.");
                         return 0;
                     }
@@ -59,12 +60,12 @@ namespace Microsoft.Dnx.Tooling
                 File.Delete(file);
             }
 
-            foreach(var subDirectory in subDirectories)
+            foreach (var subDirectory in subDirectories)
             {
-                DeleteDirectoryRecursively(subDirectory, true);
+                DeleteDirectoryRecursively(subDirectory, deleteBaseDirectory: true);
             }
 
-            if(deleteBaseDirectory)
+            if (deleteBaseDirectory)
             {
                 Directory.Delete(baseDirectory);
             }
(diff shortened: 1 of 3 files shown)


src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
```diff
@@ -0,0 +1,73 @@
...
+using System;
+using System.IO;
+
+namespace Microsoft.Dnx.Tooling
+{
+    internal class ClearCacheCommand
+    {
+        public Reports Reports { get; }
+        public string HttpCacheDirectory { get; }
+        public ClearCacheCommand(Reports reports, string httpCacheDirectory)
```
Eilon: Fix all the formatting please :smile: Empty line above th ector.

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            Reports = reports;
+            HttpCacheDirectory = httpCacheDirectory;
+        }
+        
+        public int Execute()
+        {
+            if(Directory.Exists(HttpCacheDirectory))
```
Eilon: Format doc!

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
```diff
@@ -0,0 +1,73 @@
...
+                {
+                    try
+                    {
+                        DeleteDirectoryRecursively(HttpCacheDirectory, false);
+                        Reports.Information.WriteLine("Cache cleared.");
+                        return 0;
+                    }
+                    catch (Exception e)
+                    {
+                        exception = e;
```
Eilon: We silently ignore the first failures? Shouldn't this *at least* warn about any errors encountered?

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
```diff
@@ -0,0 +1,73 @@
...
+                }
+                if (exception != null)
+                {
+                    Reports.Error.WriteLine($"Unable to clear cache directory: {exception.Message}");
+                }
+                return 1;
+            }
+            else
+            {
+                Reports.Error.WriteLine($"Cache directory {HttpCacheDirectory} does not exist.");
```
Eilon: Is this really an error? I'm not sure...

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
```diff
@@ -0,0 +1,73 @@
...
+            var subDirectories = Directory.GetDirectories(baseDirectory);
+
+            foreach (var file in files)
+            {
+                File.Delete(file);
+            }
+
+            foreach(var subDirectory in subDirectories)
+            {
+                DeleteDirectoryRecursively(subDirectory, true);
```
Eilon: Named param.

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
```diff
@@ -0,0 +1,73 @@
...
+            {
+                Reports.Information.WriteLine($"Clearing cache directory {HttpCacheDirectory}");
+
+                Exception exception = null;
+
+                for (var i = 0; i < 3; i++)
+                {
+                    try
+                    {
+                        DeleteDirectoryRecursively(HttpCacheDirectory, false);
```
Eilon: Named param.

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
Eilon: I'm getting convinced this isn't an error at all. The base directory is deleted, so presumably running this command twice in a row *guarantees* that the second run will fail. Is that odd? Imagine `git clean -xdf` failing if there was nothing to clear...

muratg: The base Cache directory (`%LocalAppData%\Local\dnu\cache` in Windows) is never deleted, everything _in_ it is deleted.  

Currently `dnu restore` creates the cache folder, so I guess we can not report this at all, and return a 0 instead.

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
muratg: Retry is for transient errors so if we eventually are successful, I think we shouldn't really warn about these.  But perhaps, we want to aggregate the errors and if we don't succeed at the end, display all of the errors (after deduping.)

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
muratg: :+1:

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
muratg: I'll read the doc and fix the formatting :+1:

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
muratg: :+1:

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
muratg: :+1:

src/Microsoft.Dnx.Tooling/ConsoleCommands/ClearCacheConsoleCommand.cs
```diff
@@ -0,0 +1,25 @@
...
+
+using Microsoft.Dnx.Runtime.Common.CommandLine;
+
+namespace Microsoft.Dnx.Tooling
+{
+    internal static class ClearCacheConsoleCommand
+    {
+        public static void Register(CommandLineApplication cmdApp, ReportsFactory reportsFactory)
+        {
+            cmdApp.Command("clearcache", c =>
```
anurse: `clear-cache`

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
Eilon: Do we think there are any common errors that would happen here? And why would retrying resolve those? Why 3 retries? No delay in between retries?

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
Eilon: Ah sorry you're right, base directory isn't deleted. But if it isn't an error to *not* have it to begin with, it should be silently ignored.

src/Microsoft.Dnx.Tooling/ConsoleCommands/ClearCacheConsoleCommand.cs
Eilon: :+1: looks like we hyphenate multi-word commands in dnu in general.

Eilon: Actually :-1/2: because it looks like we randomly do things in random places :smile: 

```
 -s|--source <FEED>
 -f|--fallbacksource <FEED>
 -p|--proxy <ADDRESS>
 --no-cache
 --packages
 --ignore-failed-sources
 --quiet
 --parallel
 -?|-h|--help
```

`--fallbacksource` is one word, but `--ignore-failed-sources` is multi-word.

Do whatever you want :smile:

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
muratg: :+1:

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
muratg: Mainly temporary open handles (antivirus?) could raise `IOException`.  Much more rarely, if two parallel `dnu clearcache` commands are running at the same time, we might get a `DirectoryNotFoundException` or `FileNotFoundException`. These would go away on the retry.

We hit similar issues with TestUtils in the past - file cleanup code there retries 3 times and it seemed to work well. I'm open to other suggestions.

Good catch with the missing delay. Half a second or so sound good?

src/Microsoft.Dnx.Tooling/ConsoleCommands/ClearCacheConsoleCommand.cs
muratg: :+1:

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
Eilon: Never a fan of retries because they can add up to a lot in the event of a flaky system. E.g. running 30 `dnu clearcache` commands (once for each repo let's say), and they might all fail for some reason. With a 0.5sec delay that's 30 repos * 0.5sec * 3 retries = 45 seconds to completely fail.

muratg: OK, I'll remove the retries. We can add them back later if we hear usability issues.

src/Microsoft.Dnx.Tooling/ConsoleCommands/ClearCacheConsoleCommand.cs
anurse: I prefer hyphenating. If we need to change other things to be consistent (eventually) so be it.

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
```diff
@@ -0,0 +1,62 @@
...
+                    Reports.Error.WriteLine($"Unable to clear cache directory: {e.Message}");
+                    return 1;
+                }
+            }
+
+            return 0;
+        }
+        public void DeleteDirectoryRecursively(string baseDirectory, bool deleteBaseDirectory)
+        {
+            var files = Directory.GetFiles(baseDirectory);
```
anurse: How about [`System.IO.Directory.Delete(string path, bool recursive)`](https://msdn.microsoft.com/en-us/library/fxeahc5f(v=vs.110).aspx)? :)

anurse: It should be safe to delete the base directory.

muratg: Yup, `restore` creates it if it doesn't exist. I was thinking to keep it safe if we change the restore behavior in the future though.

But fair enough, I'll make the change to simplify the code.

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
```diff
@@ -0,0 +1,62 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+
+namespace Microsoft.Dnx.Tooling
+{
+    internal class ClearCacheCommand
+    {
+        public Reports Reports { get; }
```
Eilon: Properties after ctors.

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
davidfowl: How about using the helper that already exists :smile:  https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Tooling/Utils/FileOperationUtils.cs#L20

src/Microsoft.Dnx.Tooling/ConsoleCommands/ClearCacheConsoleCommand.cs
```diff
@@ -0,0 +1,26 @@
...
+
+using Microsoft.Dnx.Runtime.Common.CommandLine;
+
+namespace Microsoft.Dnx.Tooling
+{
+    internal static class ClearCacheConsoleCommand
+    {
+        public static void Register(CommandLineApplication cmdApp, ReportsFactory reportsFactory)
+        {
+            cmdApp.Command("clear-cache", c =>
```
davidfowl: clear-http-cache

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            if (Directory.Exists(HttpCacheDirectory))
+            {
+                Reports.Information.WriteLine($"Clearing cache directory {HttpCacheDirectory}");
+
+                try
+                {
+                    DeleteDirectoryRecursively(HttpCacheDirectory, deleteBaseDirectory: false);
+                    Reports.Information.WriteLine("Cache cleared.");
+                    return 0;
```
Eilon: Can remove this `return`. (Hard to read code with lots of different `return`s in lots of different places.

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
anurse: Is there a reason we wrote a helper to do something the BCL already does?

davidfowl: https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Tooling/Utils/FileOperationUtils.cs#L15

There used to be logic that marked readonly files as not readonly I think.

muratg: OK, will switch to the helper. As `dnu restore` creates the cache directory if it doesn't exist, it is safe to do.

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
muratg: Will do, though I've seen both styles across the code. (See Publish and Install command implementations for example.

src/Microsoft.Dnx.Tooling/ConsoleCommands/ClearCacheConsoleCommand.cs
muratg: :+1:

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
muratg: :+1:

Eilon:
:shipit:

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+        public int Execute()
+        {
+            if (Directory.Exists(HttpCacheDirectory))
+            {
+                Reports.Information.WriteLine($"Clearing cache directory {HttpCacheDirectory}");
+
+                try
+                {
+                    FileOperationUtils.DeleteFolder(HttpCacheDirectory);
+                    Reports.Information.WriteLine("Cache cleared.");
```
victorhurdugaci: I think we have some extension methods for `Reports` to write information directly.

src/Microsoft.Dnx.Tooling/ClearCache/ClearCacheCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+    internal class ClearCacheCommand
+    {
+        public ClearCacheCommand(Reports reports, string httpCacheDirectory)
+        {
+            Reports = reports;
+            HttpCacheDirectory = httpCacheDirectory;
+        }
+
+        public Reports Reports { get; }
+        public string HttpCacheDirectory { get; }
```
victorhurdugaci: Do you need this as a property?

victorhurdugaci: Same about `Reports`

muratg:
Fixed: https://github.com/aspnet/dnx/commit/3d0edc0914f341794a56fb8c0d927ddbd96dec31

PR #2746, opened 2015-09-18
Add unix RIDs to DNX

BrennanConroy:
@anurse @davidfowl @muratg

### DIFF
--- a/src/Microsoft.Dnx.Tooling/Restore/ImplicitPackagesWalkProvider.cs
+++ b/src/Microsoft.Dnx.Tooling/Restore/ImplicitPackagesWalkProvider.cs
@@ -32,7 +32,15 @@ namespace Microsoft.Dnx.Tooling
             new RuntimeSpec("win10", "win81"),
             new RuntimeSpec("win10-x86", "win10", "win81-x86"),
             new RuntimeSpec("win10-x64", "win10", "win81-x64"),
-            new RuntimeSpec("win10-arm", "win10", "win81-arm"));
+            new RuntimeSpec("win10-arm", "win10", "win81-arm"),
+            new RuntimeSpec("unix", "any"),
+            new RuntimeSpec("osx.10.10", "unix"),
+            new RuntimeSpec("osx.10.10-x64", "osx.10.10"),
+            new RuntimeSpec("linux", "unix"),
+            new RuntimeSpec("centos.7.1", "linux"),
+            new RuntimeSpec("centos.7.1-x64", "centos.7.1"),
+            new RuntimeSpec("ubuntu.14.04", "linux"),
+            new RuntimeSpec("ubuntu.14.04-x64", "ubuntu.14.04"));
 
         public static readonly SemanticVersion ImplicitRuntimePackageVersion = new SemanticVersion("0.0.0");
         public static readonly string ImplicitRuntimePackageId = "Microsoft.NETCore.Platforms";


muratg:
:shipit: to unblock. @anurse please have a look at this also.

anurse:
:shipit:

davidfowl:
:shipit:

src/Microsoft.Dnx.Tooling/Restore/ImplicitPackagesWalkProvider.cs
```diff
@@ -32,7 +32,15 @@ internal class ImplicitPackagesWalkProvider : IWalkProvider
...
-            new RuntimeSpec("win10-arm", "win10", "win81-arm"));
+            new RuntimeSpec("win10-arm", "win10", "win81-arm"),
+            new RuntimeSpec("unix", "any"),
+            new RuntimeSpec("osx.10.10", "unix"),
+            new RuntimeSpec("osx.10.10-x64", "osx.10.10"),
+            new RuntimeSpec("linux", "unix"),
+            new RuntimeSpec("centos.7.1", "linux"),
+            new RuntimeSpec("centos.7.1-x64", "centos.7.1"),
+            new RuntimeSpec("ubuntu.14.04", "linux"),
+            new RuntimeSpec("ubuntu.14.04-x64", "ubuntu.14.04"));
```
moozzyk: do we really want to/need to distinguish among versions within a linux distribution? (and we are going to have a runtime for each?

src/Microsoft.Dnx.Tooling/Restore/ImplicitPackagesWalkProvider.cs
```diff
@@ -32,7 +32,15 @@ internal class ImplicitPackagesWalkProvider : IWalkProvider
...
             new RuntimeSpec("win10-x86", "win10", "win81-x86"),
             new RuntimeSpec("win10-x64", "win10", "win81-x64"),
-            new RuntimeSpec("win10-arm", "win10", "win81-arm"));
+            new RuntimeSpec("win10-arm", "win10", "win81-arm"),
+            new RuntimeSpec("unix", "any"),
+            new RuntimeSpec("osx.10.10", "unix"),
+            new RuntimeSpec("osx.10.10-x64", "osx.10.10"),
+            new RuntimeSpec("linux", "unix"),
+            new RuntimeSpec("centos.7.1", "linux"),
+            new RuntimeSpec("centos.7.1-x64", "centos.7.1"),
```
moozzyk: are we going to support non x64 versions?

src/Microsoft.Dnx.Tooling/Restore/ImplicitPackagesWalkProvider.cs
```diff
@@ -32,7 +32,15 @@ internal class ImplicitPackagesWalkProvider : IWalkProvider
             new RuntimeSpec("win10", "win81"),
             new RuntimeSpec("win10-x86", "win10", "win81-x86"),
             new RuntimeSpec("win10-x64", "win10", "win81-x64"),
-            new RuntimeSpec("win10-arm", "win10", "win81-arm"));
+            new RuntimeSpec("win10-arm", "win10", "win81-arm"),
+            new RuntimeSpec("unix", "any"),
```
moozzyk: does this mean !windows && !darwin?

BrennanConroy: This is fallback logic, i.e. "linux" falls back to "unix" falls back to "any", or "darwin" falls back to "unix" falls back to "any"

moozzyk: I was asking how we (will) calculate if something is "unix"

src/Microsoft.Dnx.Tooling/Restore/ImplicitPackagesWalkProvider.cs
muratg: I thought for X-Plat it was only X64 too.  Good catch.

src/Microsoft.Dnx.Tooling/Restore/ImplicitPackagesWalkProvider.cs
muratg: I believe this is coming from what @BrennanConroy  discovered in the packages. We would need a way to keep this list up-to-date somehow if we will need to have runtimes per linux distros!

moozzyk: as per the comment in this file: " // Based on values from the real Microsoft.NETCore.Platforms, but without 'aot' since DNX doesn't support .NET Native" - I was just wondering how we are going to handle this and - as you pointed out - how to keep this file in-synch

muratg:
@davidfowl I believe that was the way CoreCLR folks described it on the board a few weeks ago.

src/Microsoft.Dnx.Tooling/Restore/ImplicitPackagesWalkProvider.cs
anurse: This data defines the fallback behavior. We aren't required to identify if an OS is `unix`, but a package that can be implemented the same way on OSX and Linux would mark itself as `unix` and via this fallback logic we would know we can use it.

src/Microsoft.Dnx.Tooling/Restore/ImplicitPackagesWalkProvider.cs
anurse: In theory we should probably consider 32-bit versions. For example, RIDs are used even on Mono, and if  you are using a 32-bit Mono P-Invoking into a 32-bit native library, we should know that :)

src/Microsoft.Dnx.Tooling/Restore/ImplicitPackagesWalkProvider.cs
anurse: We don't have to keep it in perfect sync. This is what we choose to use as a "build-in" version of the package. If the project takes a direct dependency on the real `Microsoft.NETCore.Platforms`, they can get the updated version.

PR #2749, opened 2015-09-19
Use platform specific runtime to insert CoreCLR bits

BrennanConroy:


BrennanConroy:
Travis wont pass until aspnetcidev has new dnx's

makefile.shade
```diff
@@ -537,48 +538,42 @@ var CAN_BUILD_ONECORE = '${Directory.Exists(VS_ONECORE_ROOT) && Directory.GetFil
...
 
             // Parse and copy lib assemblies
             foreach (var project in sdkProjects)
             {
-                var projectFolder = Path.Combine(ROOT, "src", project);
-                var projectAssemblies = ListDependencyAssemblies(projectFolder, target.Framework, coreclrFolder);
-
-                projectAssemblies.ExceptWith(runtimeAssemblies);
+                var projFolder = Path.Combine(ROOT, "src", project);
+                var projAssemblies = ListDependencyAssemblies(projFolder, target, ROOT);
```
muratg: What are projectAssemblies vs projAssemblies?

muratg:
You'll need to check this in to unblock CoreCLR and DNX builds or is it just for Travis?

cc: @anurse

makefile.shade
troydai: Should change the name to `sdkAssemblies`

makefile.shade
```diff
@@ -610,18 +609,15 @@ var CAN_BUILD_ONECORE = '${Directory.Exists(VS_ONECORE_ROOT) && Directory.GetFil
...
                     extraBinaries.Add("System.IO.Pipes.dll");
-                    extraBinaries.Add("System.Net.Security.dll");
+                    extraBinaries.Add("System.Net.Http.dll");
                     extraBinaries.Add("System.Runtime.Extensions.dll");
-                    extraBinaries.Add("System.Security.Cryptography.Encoding.dll");
-                    extraBinaries.Add("System.Security.Cryptography.Primitives.dll");
                     extraBinaries.Add("System.Security.Cryptography.Algorithms.dll");
                     extraBinaries.Add("System.Security.SecureString.dll");
                     extraBinaries.Add("System.Text.Encoding.CodePages.dll");
                     extraBinaries.Add("System.Threading.dll");
```
troydai: What's the rule for us to decide what assemblies are required here? How do we know when to change?

makefile.shade
```diff
@@ -947,7 +943,7 @@ var CI_DARWIN_DROP_PATH = '${Environment.GetEnvironmentVariable("CI_DARWIN_DROP_
                         var projectText = File.ReadAllText(projectFile);
                         var project = (JsonObject)Json.Deserialize(projectText);
 
-                        object configsObject;
+                        //object configsObject;
```
troydai: remove commented code.

makefile.shade
```diff
@@ -1001,27 +997,11 @@ macro name='DoRestore'
...
-                ExecuteAndRedirectOutput("dnu", args, dotnetcoreclrDepFile);
-            }
-            else
-            {
-                var args = string.Format(@"/C dnu list {0} --assemblies --runtime {1}",
-                                        Path.Combine(ROOT, "src", BOOTSTRAPPER_HOST_NAME),
-                                        runtimeCandidates.First(Directory.Exists));
-                ExecuteAndRedirectOutput("cmd", args, dotnetcoreclrDepFile);
-            }
+            var target = RUNTIME_TARGETS.Where(t => t.Flavor == "coreclr").First();
```
troydai: Case sensitive?

makefile.shade
BrennanConroy: The rule is we wont have any of these soon, but these are the ones we currently need because the nupkgs don't differentiate platform assemblies.

BrennanConroy:
This is to unblock CoreCLR

src/dnx.common/tpa.cpp
```diff
@@ -18,7 +18,6 @@ const std::vector<const dnx::char_t*> CreateTpaBase(bool native_images)
             _X("Microsoft.Dnx.Host.CoreClr.ni.dll"),
             _X("Microsoft.Dnx.Loader.ni.dll"),
             _X("Microsoft.Dnx.Runtime.Abstractions.ni.dll"),
-            _X("mscorlib.ni.dll"),
```
moozzyk: why don't we need these anymore?

makefile.shade
```diff
@@ -1348,39 +1328,96 @@ functions @{
...
-            argument += (" --runtime " + coreclrFolder);
+            runtimeTarget = "win7";
         }
-
-        int result;
-        if (!IsMono)
+        else if(target.OS == "linux")
         {
-            result = ExecuteAndRedirectOutput("cmd", string.Format("/C dnu {0}", argument), tmpFile);
+            runtimeTarget = "ubuntu.14.04";
```
moozzyk: This is scary... I am running Ubuntu 15.04....

src/dnx.common/tpa.cpp
davidfowl: This looks pretty broken. How can mscorlib not be on the TPA list.

davidfowl:
The makefile changes don't look as nice as I would have expected. Having lock files makes gathering dependencies alot easier. I get that we're trying to unblock the build but it feels like this is tacked on instead of designed with the lock file in mind.

muratg:
@davidfowl It's been a long time we consumed CoreCLR builds and we're one week away from escrow. If this is not the way to unblock, let's meet early on Monday to come up with a plan and heads down implement it.

cc @anurse @Eilon

makefile.shade
anurse: Yeah, this is very short-term (like, should be fixed early next week) while the CLR team finish Cross-Plat packaging.

makefile.shade
anurse: Then CoreCLR is technically unsupported on your OS :). As far as I know, they are only targeting the current LTS version: 14.04.

It may, of course, technically work, but it's not a production-supported distro as far as I know.

makefile.shade
```diff
@@ -1348,39 +1328,96 @@ functions @{
...
+        var projects = obj.ValueAsJsonObject("libraries");
+        
+        var projectAssemblies = new HashSet<string>();
+        foreach (var package in runtimePackages.Keys)
+        {
+            var lib = runtimePackages.ValueAsJsonObject(package);
+            if(lib.ValueAsString("type") == "project")
+            {
+                var name = package.Substring(0, package.IndexOf('/'));
+                if (name.EndsWith(".Sources") || name.Contains("Microsoft.Dnx.Runtime.Internals"))
```
anurse: We should probably rename `Microsoft.Dnx.Runtime.Internals` to `Microsoft.Dnx.Runtime.Internals.Sources` to be consistent. Not something that should be done in this PR, just noting. /cc @davidfowl

src/dnx.common/tpa.cpp
anurse: From our testing it looks like mscorlib is already treated as a TPA, but perhaps that's wrong. It certainly works fine without it in the TPA.

dougbu:
> It's been a long time we consumed CoreCLR builds

@muratg on a related note, could we pick up a Roslyn build containing the recent fix for aspnet/External#33? That fix isn't included in DNX beta8-15620.

src/dnx.common/tpa.cpp
davidfowl: Ugh. ok

muratg:
@dougbu I'll start a thread. Current POR with the CoreCLR was they push the current Roslyn that we are using.

makefile.shade
```diff
@@ -1348,39 +1327,96 @@ functions @{
...
 
-    HashSet<string> ListDependencyAssemblies(string projectFolder, string framework, string coreclrFolder = null)
+    HashSet<string> ListDependencyAssemblies(string projectFolder, RuntimeTarget target, string ROOT)
     {
-        var tmpFile = Path.GetTempFileName();
-        var argument = string.Format("list {0} --framework {1} --assemblies", projectFolder, framework);
-
-        if (coreclrFolder != null)
+        var runtimeTarget = "";
+        var args = "--runtime {0}";
```
anurse: It seems confusing to have this here. Can you merge the declaration with the full initialization down on line 1348?

BrennanConroy:
Updated

BrennanConroy:
Waiting for a fix from Andrew to add a couple RIDs, Travis wont pass for this PR, because it assumes we are using a newer coreclr than it will grab.

BrennanConroy:
@davidfowl @anurse 

All extraBinaries removed, verified on Windows and Linux, currently building on Mac. Please review

muratg:
:shipit: Unblocking this to get at least one build with the new CoreCLR.

BrennanConroy:
Only System.Net.Http is in extra binaries, should be able to remove it later today

troydai:
Travis is failing. Is it expected?

BrennanConroy:
Yes, read the comments

anurse:
:shipit:

Issue #2754, opened 2015-09-19
Add support for producing ecma compliant "exe"


davidfowl:
This will allow us to use dnx to produce the host as well. 

https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Host.Mono/EntryPoint.cs

akoeplinger:
I presume this fixes https://github.com/aspnet/dnx/issues/1997 as well?

borgdylan:
With this option, a compiler will still have to use the ".dll" extension when ordered to spit out files to disk?

davidfowl:
@akoeplinger Hah, didn't even see that.

davidfowl:
@bojanrajkovic You can do what you want to really. We're not changing the .dll extension.

davidfowl:
Makes me wonder if we should call it emitEntryPoint so we don't confuse people when the output isn't an exe.

davidfowl:
Done, the compiler option is emitEntryPoint

akoeplinger:
Great :clap:

bojanrajkovic:
@davidfowl Heh, I think you meant @borgdylan. Was wondering what I'd gotten myself into this time.

davidfowl:
Oops! 😀

borgdylan:
@davidfowl I thought the answer was not meant for me. Thanks for clearing this. Now I can implement the right behaviour in my provider.

xanather:
Will it ever be possible for DNX to produce a native binary for a target platform (replaces dnx.exe) as a static app-specific bootstrapper? I think that would be pretty cool.

borgdylan:
I found it easy to implement this for my own language support. Would this have any application apart from writing DNX hosts or command line tools to have no dependencies that are external to the BCL?

PR #2761, opened 2015-09-20
Added compilation option for emitting exe

davidfowl:
- Exe in this sense doesn't make the output file
an exe, it's about being able to create things that
have an ECMA compliant entrypoint. This will allow us
to use dnu to build the mono host. See https://github.com/aspnet/dnx/issues/2762

Will add dnu pack test when @anurse commits his changes

#2754


src/Microsoft.Dnx.Runtime/ProjectReader.cs
```diff
@@ -463,7 +463,8 @@ private static CompilerOptions GetCompilationOptions(JsonObject rawObject)
                 Optimize = rawOptions.ValueAsNullableBoolean("optimize"),
                 KeyFile = rawOptions.ValueAsString("keyFile"),
                 DelaySign = rawOptions.ValueAsNullableBoolean("delaySign"),
-                StrongName = rawOptions.ValueAsNullableBoolean("strongName")
+                StrongName = rawOptions.ValueAsNullableBoolean("strongName"),
+                EmitExecutable = rawOptions.ValueAsNullableBoolean("emitExe")
```
Eilon: Come on, man, trailing comma!,

Eilon:
:shipit:

PR #2769, opened 2015-09-21
Passing context from Mono

moozzyk:


src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
```diff
@@ -205,4 +220,83 @@ private static int BootstrapperOptionValueNum(string candidate)
...
+        catch(Exception ex)
+        {
+        }
+
+
+        try
+        {
+            // Red Hat based Linux without lsb_release
+            var redHatRelease = File.ReadAllText("/etc/redhat-release");
+            var m = Regex.Match(redHatRelease, @"^(?<distro>\S+) \S+ \S+ (?<version>\d\.\d)\..+$");
```
victorhurdugaci: What will happen when the version will have more than 1 digit in each group?

moozzyk: Will change to `(?<version>\d+\.\d+)\..+$")

src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
```diff
@@ -205,4 +220,83 @@ private static int BootstrapperOptionValueNum(string candidate)
...
+        {
+            var lsb_release = RunProgram("lsb_release", "-s -i -r");
+            if (!string.IsNullOrEmpty(lsb_release))
+            {
+                osVersion = lsb_release.Replace(Environment.NewLine, " ");
+                return;
+            }
+        }
+        catch(Exception ex)
+        {
```
davidfowl: What happens here?

src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
```diff
@@ -205,4 +220,83 @@ private static int BootstrapperOptionValueNum(string candidate)
...
+            var redHatRelease = File.ReadAllText("/etc/redhat-release");
+            var m = Regex.Match(redHatRelease, @"^(?<distro>\S+) \S+ \S+ (?<version>\d+\.\d+)\..+$");
+            if (m.Success)
+            {
+                osVersion = string.Format("{0} {1}", m.Groups["distro"], m.Groups["version"]);
+                return;
+            }        
+        }
+        catch(Exception ex)
+        {
```
davidfowl: And here?

src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
```diff
@@ -74,10 +75,24 @@ public static int Main(string[] arguments)
...
 
+        string operatingSystem, osVersion, architecture;
+        GetOsDetails(out operatingSystem, out osVersion, out architecture);
+
+        var bootstrapperContext = new BootstrapperContext
+        {
+            OperatingSystem = operatingSystem,
+            OsVersion = osVersion,
+            Architecture = architecture,
+            RuntimeDirectory = Path.GetDirectoryName(typeof(EntryPoint).Assembly.Location),
```
davidfowl: Edge case, this will fail if built using dnx run

src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
moozzyk: See the next line. Running on "I don't know what OS is"

src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
moozzyk: Nothing - on Redhat based Linux distros there might not be `lsb_release` so we ignore the error and try `/etc/redhat-release`

src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
moozzyk: Then it was always broken - because we already use this: https://github.com/aspnet/dnx/pull/2769/files#diff-f79c267a7f222c20dba5339e1498a6d3R59. Where can we get the information from in the case you describe?

src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
moozzyk: Should actually use just `catch`

src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
moozzyk: Should actually use just `catch`

davidfowl:
:shipit:

PR #2772, opened 2015-09-21
Pawelka/remove envvars

moozzyk:
This is part 1. A few things are missing:
- passing runtime (Mono, CoreClr, Clr)
- removing HandleExceptions
- reoving DNX_FRAMEWORK


moozzyk:
Looking at Travis...

src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
```diff
@@ -98,12 +98,14 @@ private static string GetRuntimeOsName(this IRuntimeEnvironment env)
...
                 }
+
+                return os + ver;
             }
             else
             {
                 os = os.ToLowerInvariant(); // Just use the lower-case full name of the OS as the RID OS
             }
-            return os + ver;
+            return os;
```
moozzyk: Changes in this file are workaround for #2773

src/Microsoft.Dnx.Runtime.Sources/Impl/EnvironmentNames.cs
```diff
@@ -16,12 +16,9 @@ internal static class EnvironmentNames
         public const string AppBase = "DNX_APPBASE";
```
moozzyk: This cannot be removed because dnu publish sets this environment variable in the scripts it generates. Anyone know why we are doing this and if it should be removed?

davidfowl: It should only be read for mono

src/Microsoft.Dnx.Host/RuntimeEnvironment.cs
```diff
@@ -14,33 +15,22 @@ public class RuntimeEnvironment : IRuntimeEnvironment
...
-            if (Environment.GetEnvironmentVariable(EnvironmentNames.DnxIsWindows) == "1")
-            {
-                _osName = RuntimeOperatingSystems.Windows;
-            }
+            RuntimeArchitecture = bootstrapperContext.Architecture;
+            _osName = bootstrapperContext.OperatingSystem;
+            _osVersion = bootstrapperContext.OsVersion;
         }
 
         public string OperatingSystem
```
davidfowl: Turn these into get only properties and remove the fields

src/Microsoft.Dnx.Runtime.Sources/Impl/EnvironmentNames.cs
```diff
@@ -16,12 +16,9 @@ internal static class EnvironmentNames
...
         public const string Framework = "DNX_FRAMEWORK";
         public const string Configuration = "DNX_CONFIGURATION";
-        public const string ConsoleHost = "DNX_CONSOLE_HOST";
-        public const string DefaultLib = "DNX_DEFAULT_LIB";
         public const string BuildKeyFile = "DNX_BUILD_KEY_FILE";
         public const string BuildDelaySign = "DNX_BUILD_DELAY_SIGN";
         public const string PortablePdb = "DNX_BUILD_PORTABLE_PDB";
-        public const string DnxIsWindows = "DNX_IS_WINDOWS";
         public const string AspNetLoaderPath = "DNX_ASPNET_LOADER_PATH";
```
davidfowl: Remove?

src/Microsoft.Dnx.Runtime.Sources/Impl/EnvironmentNames.cs
```diff
@@ -16,12 +16,9 @@ internal static class EnvironmentNames
         public const string AppBase = "DNX_APPBASE";
         public const string Framework = "DNX_FRAMEWORK";
         public const string Configuration = "DNX_CONFIGURATION";
```
davidfowl: Remove this?

src/Microsoft.Dnx.Runtime.Sources/Impl/EnvironmentNames.cs
```diff
@@ -16,12 +16,9 @@ internal static class EnvironmentNames
         public const string AppBase = "DNX_APPBASE";
         public const string Framework = "DNX_FRAMEWORK";
```
davidfowl: Remove this?

src/Microsoft.Dnx.Runtime.Sources/Impl/EnvironmentNames.cs
moozzyk: Why? I don't see any code that reads this variable... (besides the one I removed because we pass the appBase in the bootstrapper context) Also, don't we always pass the `--appbase` parameter from the script?

davidfowl: You're right, we parse it on the dnx command line...

src/Microsoft.Dnx.Runtime.Sources/Impl/EnvironmentNames.cs
moozzyk: This is separate (I called it out in the description) - we don't even pass it around too much - it didn't have to be env var at all.

src/Microsoft.Dnx.Runtime.Sources/Impl/EnvironmentNames.cs
moozzyk: Need to find what it is and how this is used. It was not used/set in the native layer.

src/Microsoft.Dnx.Runtime.Sources/Impl/EnvironmentNames.cs
moozzyk: Yes, forgot to call it out. It will be a separate commit since it entails removing the logic that copies the loader.

src/Microsoft.Dnx.Runtime.Sources/Impl/EnvironmentNames.cs
moozzyk: Then I will remove it completely (including the scripts)

davidfowl: Where do we include it?

src/Microsoft.Dnx.Runtime.Sources/Impl/EnvironmentNames.cs
davidfowl: I'm working on that for the httpplatformhandler so I'll remove this.

davidfowl:
@moozzyk Are you done with this change?

src/Microsoft.Dnx.Runtime.Sources/Impl/EnvironmentNames.cs
moozzyk: Sounds good.

src/Microsoft.Dnx.Runtime.Sources/Impl/EnvironmentNames.cs
moozzyk: https://github.com/aspnet/dnx/blob/ba748dbe1abbcd243b775a27eb97e085aeb58b77/src/Microsoft.Dnx.Tooling/Publish/PublishRoot.cs#L174

This is to pass to the script template to fill `{0}` here:

```
export SET {0}=""$DIR/{1}""
exec ""{2}{3}"" --appbase ""${0}"" Microsoft.Dnx.ApplicationHost --configuration {4} {5} ""$@""";
```

moozzyk:
@davidfowl - I need to fix "Turn these into get only properties and remove the fields". `DNX_APPBASE` can be taken care of separately depending on what we agree on. If there are no other comments then that's pretty much it.

moozzyk:
I will merge it as is. I fixed getters in a later review. We need to talk about `DNX_APPBASE`

PR #2780, opened 2015-09-22
Update embedded RID definitions for Ubuntu and OS X

### LINKED ISSUE #2594: DNX on OSX/Linux loads the wrong System.Runtime.InteropServices.RuntimeInformation package
`System.Runtime.InteropServices.RuntimeInformation.IsPlatform(OSPlatform.Windows)` always returns true, even on OSX and Linux.

**Repro**
Run this on different platforms. It will always print "Windows".
```json
{
	"dependencies": {},
	"frameworks": {
		"dnxcore50":{
			"dependencies": {
				"System.Console":"4.0.0-*",
				"System.Runtime.InteropServices.RuntimeInformation": "4.0.0-*"
			}
		}
	}
}
```

```c#
using System;
using System.Runtime.InteropServices;

namespace Test
{
	public class Program
	{
		public void Main(string[] args)
		{
			if(RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
			{
				Console.WriteLine("Windows");
			}
			if(RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
			{
				Console.WriteLine("Darwin");
			}
			if(RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
			{
				Console.WriteLine("Linux");
			}
		}
	}
}
```

**Versions**
```
1.0.0-beta8-15545    coreclr x64  darwin       
and
1.0.0-beta8-15545    coreclr x64  linux      

System.Runtime.InteropServices.RuntimeInformation 4.0.0-beta-23231

Pulling from https://www.myget.org/F/aspnetlitedev/api/v2
```

This is related to #2387. Our myget feed publishes a package of System.Runtime.InteropServices.RuntimeInformation that includes the Windows version of the dll. This gets loaded at runtime, and thus misinforms the `IsPlatform` API about which OS is actually running.

anurse:
Fixes #2594 

/cc @davidfowl @BrennanConroy @ericstj

### DIFF
--- a/src/Microsoft.Dnx.Tooling/Restore/ImplicitPackagesWalkProvider.cs
+++ b/src/Microsoft.Dnx.Tooling/Restore/ImplicitPackagesWalkProvider.cs
@@ -33,14 +33,30 @@ namespace Microsoft.Dnx.Tooling
             new RuntimeSpec("win10-x86", "win10", "win81-x86"),
             new RuntimeSpec("win10-x64", "win10", "win81-x64"),
             new RuntimeSpec("win10-arm", "win10", "win81-arm"),
+
             new RuntimeSpec("unix", "any"),
-            new RuntimeSpec("osx.10.10", "unix"),
-            new RuntimeSpec("osx.10.10-x64", "osx.10.10"),
+            new RuntimeSpec("unix-x64", "unix"),
+
+            new RuntimeSpec("osx", "unix"),
+            new RuntimeSpec("osx-x64", "osx", "unix-x64"),
+
+            new RuntimeSpec("osx.10.10", "osx"),
+            new RuntimeSpec("osx.10.10-x64", "osx-x64", "osx.10.10"),
+
             new RuntimeSpec("linux", "unix"),
-            new RuntimeSpec("centos.7.1", "linux"),
-            new RuntimeSpec("centos.7.1-x64", "centos.7.1"),
-            new RuntimeSpec("ubuntu.14.04", "linux"),
-            new RuntimeSpec("ubuntu.14.04-x64", "ubuntu.14.04"));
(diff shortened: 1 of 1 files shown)


BrennanConroy:
:shipit:

src/Microsoft.Dnx.Tooling/Restore/ImplicitPackagesWalkProvider.cs
```diff
@@ -33,14 +47,30 @@ internal class ImplicitPackagesWalkProvider : IWalkProvider
...
+            new RuntimeSpec("centos-x64", "centos", "linux-x64"),
+
+            new RuntimeSpec("centos.7.1", "centos"),
+            new RuntimeSpec("centos.7.1-x64", "centos.7.1", "centos-x64"),
+
+            new RuntimeSpec("ubuntu", "linux"),
+            new RuntimeSpec("ubuntu-x64", "ubuntu", "linux-x64"),
+
+            new RuntimeSpec("ubuntu.14.04", "ubuntu"),
+            new RuntimeSpec("ubuntu.14.04-x64", "ubuntu.14.04", "ubuntu-x64"));
```
moozzyk: Where do we get this list from?

anurse: Microsoft.NETCore.Platforms package

muratg: Do we need to have a process or automation to keep these two in sync?

anurse: I don't think so. It would take more effort than it's worth since we only need to modify this when DNX adds support for a new platform.

PR #2795, opened 2015-09-23
fix #438 by adding support for dependency targets

### LINKED ISSUE #438: Clean up rules around what is a valid dependency version in the project.json file
The current rules are too fuzzy and need to be more clearly defined. Today we have 3 kinds of references:

- **SDK references** - These are reference assemblies or things in the GAC. Things like System.Data on the full CLR or System.Runtime when targeting a PCL target framework.
- **Project references** - These are references to other projects (self explanatory really)
- **Package references** - References to nuget packages

Today we allow `""` for project dependencies since versions don't matter for projects. We also allow it for SDK references so you don't have to know what version System.Data actually is. This leads to ambiguity so we're going to stop allowing empty versions:

```JSON
{
  "dependencies": {
    "ClassLibrary1": "1.0.0-*",
    "Newtonsoft.Json": "5.0.8"
  },  
  "frameworks": {
    "net45": { 
      "dependencies": {
        "System.Data": "4.0.0.0"       
      }
    }
  }
}
```

**ClassLibrary1** - This is an example of a project reference. The version is ignored completely for source references. Whatever version of the source you have as  project reference will be used.
**System.Data** - This is an SDK reference for .NET 4.5, if there's no version 4.0.0.0 on this platform this will be an error
**Newtonsoft.Json** - This is a package reference

anurse:
Targets can be used to restrict a dependency so that it can only be
satisfied by a particular type of library.

The recent changes to put projects in the lock file made this way simpler than I initially feared :). Since the lock file controls if the runtime will use a project or package to load a dependency (or if it will consider the dependency unresolved), the changes are limited to `dnu restore`

Fixes #438 

/cc @davidfowl @troydai @halter73

### DIFF
--- a/misc/DependencyTargets/src/A/project.json
+++ b/misc/DependencyTargets/src/A/project.json
@@ -1,0 +1,18 @@
+﻿{
+    "version": "1.0.0",
+
+    "configurations": {
+        "package": {
+            "compilationOptions": { "define": [ "PACKAGE" ] }
+        }
+    },
+
+    "frameworks": {
+        "dnx451": { },
+        "dnxcore50": {
+            "dependencies": {
+                "System.Runtime": "4.0.0-beta-*"
+            }
+        }
+    }
+}
--- a/src/Microsoft.Dnx.Runtime.Abstractions/DependencyTargets.cs
+++ b/src/Microsoft.Dnx.Runtime.Abstractions/DependencyTargets.cs
@@ -1,0 +1,42 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Dnx.Runtime
+{
+    public static class DependencyTargets
+    {
+        public static readonly string Project = nameof(Project);
+        public static readonly string Package = nameof(Package);
+
+        private const string NoTargetDisplayName = "Dependency";
+
+        public static string GetDisplayForTarget(string target)
+        {
+            // Normalize the case of the known targets
+            if (string.Equals(target, Project, StringComparison.OrdinalIgnoreCase))
+            {
+                return Project;
+            }
+            else if (string.Equals(target, Package, StringComparison.OrdinalIgnoreCase))
+            {
+                return Package;
+            }
+            else
+            {
+                return NoTargetDisplayName;
+            }
+        }
+
+        public static bool SupportsPackage(string target)
+        {
+            return string.IsNullOrEmpty(target) || string.Equals(target, Package, StringComparison.OrdinalIgnoreCase);
+        }
+
+        public static bool SupportsProject(string target)
+        {
+            return string.IsNullOrEmpty(target) || string.Equals(target, Project, StringComparison.OrdinalIgnoreCase);
+        }
+    }
+}
--- a/src/Microsoft.Dnx.Tooling/Restore/NuGetDependencyResolver.cs
+++ b/src/Microsoft.Dnx.Tooling/Restore/NuGetDependencyResolver.cs
@@ -29,7 +29,7 @@ namespace Microsoft.Dnx.Runtime
 
         public LibraryDescription GetDescription(LibraryRange libraryRange, FrameworkName targetFramework)
         {
-            if (libraryRange.IsGacOrFrameworkReference)
+            if (!DependencyTargets.SupportsPackage(libraryRange.Target))
             {
                 return null;
             }
--- a/src/Microsoft.Dnx.Tooling/Restore/RemoteWalkProvider.cs
+++ b/src/Microsoft.Dnx.Tooling/Restore/RemoteWalkProvider.cs
@@ -28,9 +28,14 @@ namespace Microsoft.Dnx.Tooling
 
         public async Task<WalkProviderMatch> FindLibrary(LibraryRange libraryRange, FrameworkName targetFramework, bool includeUnlisted)
         {
+            if (!DependencyTargets.SupportsPackage(libraryRange.Target))
+            {
+                return null;
+            }
+
             var results = await _source.FindPackagesByIdAsync(libraryRange.Name);
             PackageInfo bestResult = null;
-            if(!includeUnlisted)
+            if (!includeUnlisted)
             {
                 results = results.Where(p => p.Listed);
             }
@@ -54,6 +59,7 @@ namespace Microsoft.Dnx.Tooling
             return new WalkProviderMatch
             {
                 Library = new LibraryIdentity(bestResult.Id, bestResult.Version, isGacOrFrameworkReference: false),
+                LibraryType = LibraryTypes.Package,
                 Path = bestResult.ContentUri,
                 Provider = this,
             };
--- a/src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
+++ b/src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
@@ -463,12 +463,14 @@ namespace Microsoft.Dnx.Tooling
                         if (!IgnoreMissingDependencies)
                         {
                             if (!node.LibraryRange.IsGacOrFrameworkReference &&
-                                 node.LibraryRange.VersionRange != null &&
                                  missingItems.Add(node.LibraryRange))
                             {
-                                var errorMessage = string.Format("Unable to locate {0} {1}",
-                                    node.LibraryRange.Name.Red().Bold(),
-                                    node.LibraryRange.VersionRange);
+                                var versionString = node.LibraryRange.VersionRange == null ?
+                                    string.Empty :
+                                    (" " + node.LibraryRange.VersionRange.ToString());
+                                var errorMessage =
+                                    $"Unable to locate {DependencyTargets.GetDisplayForTarget(node.LibraryRange.Target)} " +
+                                    $"{node.LibraryRange.Name.Red().Bold()}{versionString}";
                                 summary.ErrorMessages.GetOrAdd(projectJsonPath, _ => new List<string>()).Add(errorMessage);
                                 Reports.Error.WriteLine(errorMessage);
                                 success = false;
@@ -505,7 +507,7 @@ namespace Microsoft.Dnx.Tooling
                             graphItems.Add(node.Item);
                         }
 
-                        context.Libraries.Add(node.Item.Match.Library);
+                        context.Matches.Add(node.Item.Match);
                     }
                 });
             }
@@ -829,7 +831,7 @@ namespace Microsoft.Dnx.Tooling
             // Record all libraries used
             foreach (var item in graphItems.OrderBy(x => x.Match.Library, new LibraryComparer()))
             {
-                if (item.Match.Provider is ImplicitPackagesWalkProvider)
+                if (item.Match.LibraryType.Equals(LibraryTypes.Implicit))
                 {
                     continue;
                 }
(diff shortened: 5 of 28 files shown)


src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
```diff
@@ -840,19 +843,23 @@ private Task<LockFile> ReadLockFile(string projectLockFilePath)
...
                 }
-                else if (packageInfo != null)
+                else if (item.Match.LibraryType.Equals(LibraryTypes.Package))
                 {
+                    var packageInfo = repository.FindPackagesById(library.Name)
+                                                .FirstOrDefault(p => p.Version == library.Version);
+
+                    if (packageInfo == null)
+                    {
+                        throw new InvalidOperationException($"Unresolved package: {library.Name}");
```
davidfowl: Does this stop at the first unresolved package?

anurse: Yes, but it did before. This is really an internal error if this happens. It could only happen if a package was explicitly deleted between the time it was downloaded and the time the lock file is written.

src/Microsoft.Dnx.Runtime.Abstractions/DependencyTargets.cs
```diff
@@ -0,0 +1,40 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Dnx.Runtime
+{
+    public static class DependencyTargets
+    {
+        public static readonly string Project = nameof(Project);
+        public static readonly string Package = nameof(Package);
```
troydai: We have the very similar things in `LibraryTypes`. Reusable?

src/Microsoft.Dnx.Runtime.Abstractions/DependencyTargets.cs
```diff
@@ -0,0 +1,40 @@
...
+            {
+                return Project;
+            }
+            else if (string.Equals(target, Package, StringComparison.OrdinalIgnoreCase))
+            {
+                return Package;
+            }
+            else
+            {
+                return "Dependency";
```
troydai: Don't hard code this since neither `Project` and `Package` is hard coded.

src/Microsoft.Dnx.Runtime.Abstractions/DependencyTargets.cs
troydai: Well, on second thought. Never mind.

anurse: Yeah, I considered reusing them but they aren't quite the same.

src/Microsoft.Dnx.Runtime.Abstractions/DependencyTargets.cs
anurse: Well, they are all hard-coded, but I'll put a constant in here

anurse:
Travis should be fixed, other code review feedback addressed. Ping @davidfowl @troydai @halter73

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuRestoreTests2.cs
```diff
@@ -0,0 +1,69 @@
...
+    {
+        [Theory]
+        [MemberData(nameof(ClrDnxSdks))]
+        public void DnuRestore_DoesNotAllowProjectToSatisfyPackageDependency(DnxSdk sdk)
+        {
+            var solution = TestUtils.GetSolution<DnuRestoreTests2>(sdk, "DependencyTargets");
+            var bOutputPath = Path.Combine(solution.ArtifactsPath, "B");
+
+            // Build a package for B
+            sdk.Dnu.Restore(solution.GetProject("B")).EnsureSuccess();
```
JunTaoLuo: This is very similar to what TestUtils.CreateLocalFeed is doing in the new infrastructure. See if you can use it. I believe the only missing capability is specifying a dnu pack configuration.

src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
```diff
@@ -463,12 +463,15 @@ public RestoreCommand(FrameworkName fallbackFramework)
...
                         {
                             if (!node.LibraryRange.IsGacOrFrameworkReference &&
-                                 node.LibraryRange.VersionRange != null &&
                                  missingItems.Add(node.LibraryRange))
                             {
-                                var errorMessage = string.Format("Unable to locate {0} {1}",
+                                var versionString = node.LibraryRange.VersionRange == null ?
+                                    string.Empty :
+                                    (" " + node.LibraryRange.VersionRange.ToString());
+                                var errorMessage = string.Format("Unable to locate {2} {0}{1}",
```
JunTaoLuo: use string interpolation?

anurse:
Fixing travis and @JunTaoLuo 's comment. I'd like to get this in today for beta8 so the sooner I get sign-off the better :)

misc/DependencyTargets/src/A/project.json
```diff
@@ -0,0 +1,23 @@
+﻿{
+    "version": "1.0.0",
+    "description": "A Class Library",
+    "authors": [ "anurse" ],
```
troydai: Microsoft, actually?

src/Microsoft.Dnx.Runtime.Abstractions/DependencyTargets.cs
```diff
@@ -0,0 +1,42 @@
...
+            }
+            else
+            {
+                return NoTargetDisplayName;
+            }
+        }
+
+        public static bool SupportsPackage(string target)
+        {
+            return string.IsNullOrEmpty(target) || string.Equals(target, Package, StringComparison.OrdinalIgnoreCase);
```
troydai: Weo we support this case?
`"target": ""`

misc/DependencyTargets/src/A/project.json
anurse: Hah, I'll just remove the author :)

src/Microsoft.Dnx.Runtime.Abstractions/DependencyTargets.cs
anurse: Yes, it's the same as not specifying `target`

src/Microsoft.Dnx.Tooling/Restore/NuGetDependencyResolver.cs
```diff
@@ -29,7 +29,7 @@ public IEnumerable<string> GetAttemptedPaths(FrameworkName targetFramework)
 
         public LibraryDescription GetDescription(LibraryRange libraryRange, FrameworkName targetFramework)
         {
-            if (libraryRange.IsGacOrFrameworkReference)
+            if(!DependencyTargets.SupportsPackage(libraryRange.Target))
```
troydai: Format

src/Microsoft.Dnx.Tooling/Restore/RemoteWalkProvider.cs
```diff
@@ -28,6 +28,11 @@ public RemoteWalkProvider(IPackageFeed source)
 
         public async Task<WalkProviderMatch> FindLibrary(LibraryRange libraryRange, FrameworkName targetFramework, bool includeUnlisted)
         {
+            if(!DependencyTargets.SupportsPackage(libraryRange.Target))
```
troydai: Format

src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
```diff
@@ -880,31 +887,31 @@ private Task<LockFile> ReadLockFile(string projectLockFilePath)
...
-                    var packageInfo = repository.FindPackagesById(library.Name)
-                                                .FirstOrDefault(p => p.Version == library.Version);
+                    var packageInfo = repository.FindPackagesById(match.Library.Name)
+                                                .FirstOrDefault(p => p.Version == match.Library.Version);
 
-                    var projectDependency = projectResolver.FindProject(library.Name);
-                    if (projectDependency != null)
+                    if (match.LibraryType.Equals(LibraryTypes.Project))
                     {
+                        var projectDependency = projectResolver.FindProject(match.Library.Name);
```
troydai: Still could it be the `projectDependency` is `null`?

src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
```diff
@@ -880,31 +887,31 @@ private Task<LockFile> ReadLockFile(string projectLockFilePath)
...
-                    if (library.Name == project.Name)
+                    if (match.Library.Name == project.Name)
                     {
                         continue;
                     }
 
-                    var packageInfo = repository.FindPackagesById(library.Name)
-                                                .FirstOrDefault(p => p.Version == library.Version);
+                    var packageInfo = repository.FindPackagesById(match.Library.Name)
+                                                .FirstOrDefault(p => p.Version == match.Library.Version);
```
troydai: Move the `packageInfo` in to the `else if` scope where has its only usage.

test/Microsoft.Dnx.Testing.CommonUtils/DnxSdk/Dnx.cs
```diff
@@ -15,6 +15,12 @@ public Dnx(string sdkPath)
             _sdkPath = sdkPath;
         }
 
+        public ExecResult Execute(Project project, string commandLine = null, bool dnxTraceOn = false)
+        {
+            string fullCommandLine = $"-p \"{project.ProjectDirectory}\" {commandLine ?? "run"}";
```
troydai: `var`

troydai: How about just inline?

src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
anurse: It would be a unrecoverable internal error if that happened. A library is only marked with `LibraryTypes.Project` if it came from the `projectResolver`

src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
```diff
@@ -840,19 +842,23 @@ private Task<LockFile> ReadLockFile(string projectLockFilePath)
...
 
-                var packageInfo = repository.FindPackagesById(library.Name)
-                                            .FirstOrDefault(p => p.Version == library.Version);
-
-                var projectDependency = projectResolver.FindProject(library.Name);
-
-                if (projectDependency != null)
+                if (item.Match.LibraryType.Equals(LibraryTypes.Project))
                 {
+                    var projectDependency = projectResolver.FindProject(library.Name);
```
troydai: Could `projectDependency` be `null`?

src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
troydai: I like the current flow but rather than a `NullReferenceException` can we have a meaningful message in case it runs into the bad situation.

troydai: Not blocking.

troydai:
Rest my case. Looks good enough for me :shipit: when Travis pass and comments are addressed.

PR #2808, opened 2015-09-24
fix #2802: Loading multiple satellite assemblies of same project fails in coreclr

### LINKED ISSUE #2802: Loading multiple satellite assemblies of same project fails in coreclr
In the current code, we are not preserving the culture information during satellite assembly creation. So if app tries to load multiple satellite assemblies of same project, first one succeeds but subsequent ones fails with error `Assembly with same name is already loaded`. 

To fix this issue, we need to add `AssemblyCultureAttribute` during satellite assembly compilation

kirthik:
cc @davidfowl

### DIFF
--- a/src/Microsoft.Dnx.Compilation.CSharp/RoslynProjectReference.cs
+++ b/src/Microsoft.Dnx.Compilation.CSharp/RoslynProjectReference.cs
@@ -304,6 +304,8 @@ namespace Microsoft.Dnx.Compilation.CSharp
             CSharpCompilationOptions compilationOptions,
             Stream assemblyStream)
         {
+            const string assemblyCultureName = "System.Reflection.AssemblyCultureAttribute";
+
             var resources = resourceDescriptors
                 .Select(res => new ResourceDescription(res.Name, res.StreamFactory, isPublic: true));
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.Dnx.Compilation.CSharp/RoslynProjectReference.cs
```diff
@@ -312,6 +314,11 @@ private EmitOptions GetEmitOptions(out bool emitPdb)
                 references: CompilationContext.Compilation.References,
                 options: compilationOptions);
 
+            compilation = compilation.AddSyntaxTrees(new[]
```
davidfowl: Culture is *never* null here right?

kirthik: Yes since I am using `assemblyName.CultureName` that should never be null. If no value is specified it defaults to `neutral`. Should I add that check still just in case?

davidfowl: You tell me. If it's never null then maybe add a Debug.Assert

kirthik: I will add a check to AddSyntaxtree only if `cultureName` is not null or empty.

kirthik:
cc @davidfowl - Added null check for culturename

davidfowl:
One more thing, where does `AssemblyCultureAttribute` live on .NET Core? Will it always be available? If you only reference System.Runtime in dnxcore50 what happens?

kirthik:
Yes, it is in `System.Runtime` assembly.

davidfowl:
When travis passes :shipit:

kirthik:
merged 21cd6896974759bf8a1d48790d9515b56d8ef3d0

Issue #864, opened 2014-11-06
Commands should not display stack traces for normal user errors


dougbu:
kpm.cmd and k.cmd often end with stack traces though the errors hit are hardly exceptional.  For example `kpm.cmd unsupported` displays
```
Specify --help for a list of available options and commands.
System.Exception: TODO: Error: unrecognized argument 'unsupported'
   at Microsoft.Framework.Runtime.Common.CommandLine.CommandLineApplication.HandleUnexpecteArg(CommandLineApplication command, String[] args, Int32 index, String argTypeName)
   at Microsoft.Framework.Runtime.Common.CommandLine.CommandLineApplication.Execute(String[] args)
   at Microsoft.Framework.PackageManager.Program.Main(String[] args)
```
instead of simply providing the help information.

Similarly `k.cmd test` displays the following for a sample compilation error.
```
System.IO.FileLoadException: Could not load file or assembly 'Microsoft.AspNet.Mvc.TagHelpers.Test' or one of its depend
encies. General Exception (Exception from HRESULT: 0x80131500)
System.IO.FileLoadException: Could not load file or assembly 'Microsoft.AspNet.Mvc.TagHelpers.Test' or one of its depend
encies. General Exception (Exception from HRESULT: 0x80131500)
File name: 'Microsoft.AspNet.Mvc.TagHelpers.Test' ---> Microsoft.Framework.Runtime.Roslyn.RoslynCompilationException: C:
\dd\k\Mvc.1\test\Microsoft.AspNet.Mvc.TagHelpers.Test\SelectTagHelperTest.cs(310,66): error CS1003: Syntax error, '>' ex
pected
   at Microsoft.Framework.Runtime.Roslyn.RoslynProjectReference.Load(IAssemblyLoadContext loadContext)
   at Microsoft.Framework.Runtime.Loader.ProjectAssemblyLoader.Load(String name, IAssemblyLoadContext loadContext)
   at Microsoft.Framework.Runtime.Loader.ProjectAssemblyLoader.Load(String name)
   at klr.host.LoaderContainer.Load(String name)
   at klr.hosting.RuntimeBootstrapper.<>c__DisplayClass0.<ExecuteAsync>b__8(AssemblyName assemblyName)
   at klr.hosting.RuntimeBootstrapper.<>c__DisplayClass0.<ExecuteAsync>b__11(Object sender, ResolveEventArgs a)
[...]

Issue #2707, opened 2015-09-15
Don't return -1 exit code from managed code


moozzyk:
Follow up to https://github.com/aspnet/dnx/pull/2691 but in managed code - on non-windows valid exit codes are 0-255. However, in several places we return `-1` which is then converted to 255 which is wrong. We should return 1.
(Especially review all `return -1` and `Task.FromResult(-1)` occurrences)

muratg:
@moozzyk If simple to do and not risky, we can pull back to beta 8.

PR #2813, opened 2015-09-24
Cleaning up user errors for dnx and dnu commands

JunTaoLuo:
#864 Hiding stack traces for exceptions thrown in non-user code while running dnx/dnu. 

We should never throw generic exceptions but instead use predefined ones (e.g. ArgumentNullException) or subclassed ones (e.g. CommandParsingException). If needed, these exceptions should be handled explicitly in all command line applications in the general try/catch block.


src/Microsoft.Dnx.ApplicationHost/Program.cs
```diff
@@ -36,49 +36,58 @@ public Task<int> Main(string[] args)
...
+                }
+            }
+            catch (Exception ex)
             {
-                options.ApplicationName = host.Project.EntryPoint ?? host.Project.Name;
+                Console.Error.WriteLine($"Error: {ex.Message}");
+                return Task.FromResult(-1);
             }
-
+
```
JunTaoLuo: I should remove this random spacing change

src/Microsoft.Dnx.Tooling/NuGet/ContentModel/Infrastructure/Parser.cs
```diff
@@ -110,7 +110,7 @@ internal override bool TryMatch(ContentItem item, IDictionary<string, ContentPro
                 ContentPropertyDefinition propertyDefinition;
```
JunTaoLuo: It seems like all the files under Tooling/Nuget/ContentModel don't have copyright. Is this intentional?

src/Microsoft.Dnx.Tooling/Program.cs
```diff
@@ -80,7 +80,28 @@ public int Main(string[] args)
...
-            return app.Execute(args);
+            try
+            {
+                return app.Execute(args);
+            }
+            catch (CommandParsingException ex)
+            {
+                AnsiConsole.GetError(_runtimeEnv.OperatingSystem == "Windows").WriteLine(ex.Message.Red().Bold());
+                ex.Command.ShowHelp();
+                return 1;
```
JunTaoLuo: I've seen some discrepancies with error codes. In the application host I see we return -1 on errors. Elsewhere I've seen 1. Is there a reason to use one or the other? @davidfowl @anurse

moozzyk: #2707

src/Microsoft.Dnx.Tooling/Program.cs
```diff
@@ -80,7 +80,28 @@ public int Main(string[] args)
...
             FeedsConsoleCommand.Register(app, reportsFactory);
 
-            return app.Execute(args);
+            try
+            {
+                return app.Execute(args);
+            }
+            catch (CommandParsingException ex)
+            {
+                AnsiConsole.GetError(_runtimeEnv.OperatingSystem == "Windows").WriteLine(ex.Message.Red().Bold());
```
JunTaoLuo: @anurse This was done to mimic ReportFactory.cs in tooling where ansiconsole was used. I thought that maybe color options were only available for windows environments?

moozzyk: Talk to @BrennanConroy

JunTaoLuo: Talked to @BrennanConroy and this should be fine.

src/Microsoft.Dnx.Tooling/Program.cs
JunTaoLuo: I'm going to leave the -1 return codes for now just for consistency. The -1 should probably be all changed in one batch.

moozzyk: Sounds good.

src/Microsoft.Dnx.Tooling/NuGet/ContentModel/Infrastructure/Parser.cs
JunTaoLuo: tracking in #2822

anurse: It's because they are copied from NuGet. I think it should be OK to add them /cc @Eilon

JunTaoLuo:
Need signoff @davidfowl @anurse @moozzyk Thanks.

src/Microsoft.Dnx.Tooling/Program.cs
anurse: Cool, just wanted to check :). Could you use a named parameter here to clarify the purpose of the parameter? Something like:

```csharp
AnsiConsole.GetError(whatever_the_parameter_name_is: _runtimeEnv.OperatingSystem == "Windows").WriteLine(ex.Message.Red().Bold());
```

src/Microsoft.Dnx.Runtime.CommandParsing.Sources/CommandGrammar.cs
```diff
@@ -48,13 +48,13 @@ private CommandGrammar(Func<string, string> variable, bool preserveSurroundingQu
...
-            var grammer = new CommandGrammar(variables, preserveSurroundingQuotes);
+            var grammar = new CommandGrammar(variables, preserveSurroundingQuotes);
             var cursor = new Cursor(text, 0, text.Length);
 
-            var result = grammer.Parse(cursor);
+            var result = grammar.Parse(cursor);
             if (!result.Remainder.IsEnd)
             {
-                throw new Exception("TODO: malformed command text");
+                throw new ArgumentException($"Malformed command text '{text}'");
```
anurse: Pass the parameter name to the constructor as well (there's an overload that takes it)

src/Microsoft.Dnx.Tooling/NuGet/ContentModel/Infrastructure/Parser.cs
```diff
@@ -110,7 +110,7 @@ internal override bool TryMatch(ContentItem item, IDictionary<string, ContentPro
...
                 if (!propertyDefinitions.TryGetValue(_token, out propertyDefinition))
                 {
-                    throw new Exception(string.Format("Unable to find property definition for {{{0}}}", _token));
+                    throw new ArgumentException($"Unable to find property definition for '{_token}'");
```
anurse: Parameter name -> ctor :)

src/Microsoft.Dnx.Tooling/Restore/LockFileFormat.cs
```diff
@@ -522,7 +522,7 @@ private SemanticVersion ReadSemanticVersion(JToken json, string property)
             var valueToken = json[property];
             if (valueToken == null)
             {
-                throw new Exception(string.Format("TODO: lock file missing required property {0}", property));
+                throw new ArgumentException($"lock file missing required property '{property}'");
```
anurse: param name -> ctor - you get the idea now :P

src/Microsoft.Dnx.Tooling/Program.cs
JunTaoLuo: of course!

src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
```diff
@@ -225,7 +225,7 @@ public RestoreCommand(FrameworkName fallbackFramework)
             var diagnostics = new List<DiagnosticMessage>();
             if (!Runtime.Project.TryGetProject(projectJsonPath, out project, diagnostics))
             {
-                throw new Exception("TODO: project.json parse error");
+                throw new InvalidOperationException("project.json parse error");
```
anurse: We should now have error information in the `diagnostics` value. Could you format that (if it's present) and display it? It's possible this could fail but diagnostics would be empty (in which case just use text like "Invalid project.json").

Something like

```
Errors occurred parsing project.json:
project.json([Line#], [Col#]): [Message]
etc.
```

All of that should be available in `diagnostics` and should be pretty straightforward to add with some `string.Joins` and Linq :). Maybe throw it in a helper if it looks like it's going to be a bit complicated.

src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
```diff
@@ -596,7 +596,7 @@ private IEnumerable<string> GetRestoreRuntimes(IEnumerable<string> projectRuntim
                     {
                         if (imports != null)
                         {
-                            throw new Exception(string.Format("More than one runtime.json file has declared imports for {0}", runtimeName));
+                            throw new ArgumentException($"More than one runtime.json file has declared imports for '{runtimeName}'");
```
anurse: param name -> ctor ;)

src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
```diff
@@ -611,7 +611,7 @@ private IEnumerable<string> GetRestoreRuntimes(IEnumerable<string> projectRuntim
                     {
                         if (imports != null)
                         {
-                            throw new Exception(string.Format("Circular import for {0}", runtimeName));
+                            throw new ArgumentException($"Circular import for '{runtimeName}'");
```
anurse: param name? ctor!

src/Microsoft.Dnx.Tooling/Restore/RestoreOperations.cs
```diff
@@ -112,7 +112,7 @@ public RestoreOperations(IReport report)
             {
                 if (item.Match.Library.Name == name)
                 {
-                    throw new Exception(string.Format("TODO: Circular dependency references not supported. Package '{0}'.", name));
+                    throw new InvalidOperationException($"Circular dependency references not supported. Package '{name}'.");
```
anurse: Rephrase: "A circular dependency reference to '{name}' was detected"

src/Microsoft.Dnx.Tooling/Utils/LockFileUtils.cs
```diff
@@ -747,7 +747,7 @@ public SelectionCriteriaEntryBuilder(SelectionCriteriaBuilder builder, Selection
                         ContentPropertyDefinition propertyDefinition;
                         if (!propertyDefinitions.TryGetValue(key, out propertyDefinition))
                         {
-                            throw new Exception("Undefined property used for criteria");
+                            throw new ArgumentException($"Undefined property '{key}' used for criteria");
```
anurse: param name!

src/Microsoft.Dnx.Tooling/Utils/LockFileUtils.cs
```diff
@@ -762,7 +762,7 @@ public SelectionCriteriaEntryBuilder(SelectionCriteriaBuilder builder, Selection
                             }
                             else
                             {
-                                throw new Exception("Undefined value used for criteria");
+                                throw new ArgumentException($"Undefined value '{value}' used for criteria");
```
anurse: name of parameter

src/Microsoft.Dnx.Tooling/Utils/LockFileUtils.cs
```diff
@@ -775,7 +775,7 @@ public SelectionCriteriaEntryBuilder(SelectionCriteriaBuilder builder, Selection
                         ContentPropertyDefinition propertyDefinition;
                         if (!propertyDefinitions.TryGetValue(key, out propertyDefinition))
                         {
-                            throw new Exception("Undefined property used for criteria");
+                            throw new ArgumentException($"Undefined property '{key}' used for criteria");
```
anurse: nom du paramètre

src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandLineApplication.cs
```diff
@@ -147,7 +147,7 @@ public int Execute(params string[] args)
                             if (!option.TryParse(longOption[1]))
                             {
                                 command.ShowHint();
-                                throw new Exception(string.Format("TODO: Error: unexpected value '{0}' for option '{1}'", longOption[1], option.LongName));
+                                throw new CommandParsingException(command, $"Error: unexpected value '{longOption[1]}' for option '{option.LongName}'");
```
anurse: Remove the "Error:" from the front and capitalize "unexpected". "Error" is a prefix that catcher of the exception should control.

src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandLineApplication.cs
```diff
@@ -192,7 +192,7 @@ public int Execute(params string[] args)
                             if (!option.TryParse(shortOption[1]))
                             {
                                 command.ShowHint();
-                                throw new Exception(string.Format("TODO: Error: unexpected value '{0}' for option '{1}'", shortOption[1], option.LongName));
+                                throw new CommandParsingException(command, $"Error: unexpected value '{shortOption[1]}' for option '{option.LongName}'");
```
anurse: As above, remove Error

src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandLineApplication.cs
```diff
@@ -211,7 +211,7 @@ public int Execute(params string[] args)
                     if (!option.TryParse(arg))
                     {
                         command.ShowHint();
-                        throw new Exception(string.Format("TODO: Error: unexpected value '{0}' for option '{1}'", arg, option.LongName));
+                        throw new CommandParsingException(command, $"Error: unexpected value '{arg}' for option '{option.LongName}'");
```
anurse: And again ;)

src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandLineApplication.cs
```diff
@@ -257,7 +257,7 @@ public int Execute(params string[] args)
             if (option != null)
             {
                 command.ShowHint();
-                throw new Exception(string.Format("TODO: Error: missing value for option"));
+                throw new CommandParsingException(command, $"Error: missing value for option '{option.LongName}'");
```
anurse: <del>Error</del>

src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandLineApplication.cs
```diff
@@ -470,7 +470,7 @@ private void HandleUnexpectedArg(CommandLineApplication command, string[] args,
             if (command._throwOnUnexpectedArg)
             {
                 command.ShowHint();
-                throw new Exception(string.Format("TODO: Error: unrecognized {0} '{1}'", argTypeName, args[index]));
+                throw new CommandParsingException(command, $"Error: unrecognized {argTypeName} '{args[index]}'");
```
anurse: :bomb: Error :boom:

src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandOption.cs
```diff
@@ -41,13 +41,13 @@ public CommandOption(string template, CommandOptionType optionType)
                 }
                 else
                 {
-                    throw new Exception("TODO: invalid template pattern " + template);
+                    throw new ArgumentException($"Invalid template pattern '{template}'");
```
anurse: nombre del parámetro

src/Microsoft.Framework.CommandLineUtils.Sources/CommandLine/CommandOption.cs
```diff
@@ -41,13 +41,13 @@ public CommandOption(string template, CommandOptionType optionType)
...
                 }
             }
 
             if (string.IsNullOrEmpty(LongName) && string.IsNullOrEmpty(ShortName) && string.IsNullOrEmpty(SymbolName))
             {
-                throw new Exception("TODO: invalid template pattern " + template);
+                throw new ArgumentException($"Invalid template pattern '{template}'");
```
anurse: il nome del parametro

src/Microsoft.Dnx.Runtime.CommandParsing.Sources/CommandGrammar.cs
JunTaoLuo: yup

anurse:
Looking good, just a few clean-ups and one slightly larger item to expand on our project.json parsing errors.

I may have been a little bored on the bus and gone a little crazy with the parameter name comments ;P

src/Microsoft.Dnx.Tooling/NuGet/ContentModel/Infrastructure/Parser.cs
JunTaoLuo: Actually I think I should probably not use an argument exception here since _token is technically not passed in to that method but is instead a member variable, would InvalidOperationException be better?

anurse: Yeah, I think that's probably fine.

JunTaoLuo:
@anurse mais oui monsieur

src/Microsoft.Dnx.ApplicationHost/Program.cs
```diff
@@ -242,23 +253,21 @@ private Task<int> ExecuteMain(DefaultHost host, string applicationName, string[]
...
             string applicationName,
             Exception innerException)
         {
             if (host.Project.Commands.Any())
             {
-                // Throw a nicer exception message if the command
-                // can't be found
-                throw new InvalidOperationException(
-                    string.Format("Unable to load application or execute command '{0}'. Available commands: {1}.",
+                Console.Error.WriteLine(string.Format("Unable to load application or execute command '{0}'. Available commands: {1}.",
```
davidfowl: I think this is the wrong place for this code. It needs to go lower in the stack.

anurse: Lower how? Host? Isn't ApplicationHost the lowest layer that knows about Commands?

JunTaoLuo: I can potentially create a new exception type and handle it in the catch of Main instead of ExecuteMain. That's a lot of code to handle one scenario and seems a little heavy handed though.

davidfowl: Here's where we print errors today:

https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Host/RuntimeBootstrapper.cs#L27

I think that we need to keep this as the only place that prints errors to stdout.

anurse: So we want to swallow all user exceptions? There are benefits to allowing actual exceptions from user code to bubble up to Watson, though I suppose those are most relevant when the app has been packaged separately as it's own executable and not `dnx.exe`. The initial thought here was to just capture exceptions caused by our code and allow exceptions caused by user code to bubble.

On a side note, maybe the error handling you referenced above is why I can never get unhandled exceptions to break in the debugger... I don't know if it works for users, but I can't ever get them to work.

JunTaoLuo: I believe the `RuntimeBootstrapper` currently prints out the exceptions AND their stack traces. We want to hide the stack trace for non-user code and therefore will need to differentiate between our exceptions and user exceptions. This will probably require subclasses of `Exception`.

davidfowl: Or some extra data on exceptions that we throw

JunTaoLuo: something like `ex.Data["suppressStacktrace"] = true`?

davidfowl: Yep!

moozzyk: @anurse - "never" should not be the case. This could start happening after my recent changes to dnx. I wanted to file a bug on this on the tooling repo but I could not repro this. This never worked on CoreClr since CoreClr AVs if exception is not handled.

anurse: I'd prefer a (non-abstract) base class, it's less brittle. Something like `Microsoft.Dnx.Runtime.RuntimeException`. Either way we have to go through each exception and figure out what to do, may as well do it "right". When we have other exceptions to throw (like InvalidOperationException) we can wrap them in the InnerException of RuntimeException

src/Microsoft.Dnx.ApplicationHost/Program.cs
```diff
@@ -36,47 +36,56 @@ public Task<int> Main(string[] args)
...
 
-            if (string.IsNullOrEmpty(options.ApplicationName) ||
-                string.Equals(options.ApplicationName, "run", StringComparison.Ordinal))
+                if (string.IsNullOrEmpty(options.ApplicationName) ||
+                    string.Equals(options.ApplicationName, "run", StringComparison.Ordinal))
+                {
+                    options.ApplicationName = host.Project.EntryPoint ?? host.Project.Name;
+                }
+            }
+            catch (Exception ex)
```
troydai: Do we expect anything throw here?

src/Microsoft.Dnx.ApplicationHost/Program.cs
```diff
@@ -36,47 +36,56 @@ public Task<int> Main(string[] args)
...
             {
-                options.ApplicationName = host.Project.EntryPoint ?? host.Project.Name;
+                Console.Error.WriteLine($"Error: {ex.Message}");
+                return Task.FromResult(-1);
```
troydai: Do we have unified plan for exit code?

src/Microsoft.Dnx.Tooling/Program.cs
```diff
@@ -81,7 +81,28 @@ public int Main(string[] args)
...
-            return app.Execute(args);
+            try
+            {
+                return app.Execute(args);
+            }
+            catch (CommandParsingException ex)
+            {
+                AnsiConsole.GetError(useConsoleColor: _runtimeEnv.OperatingSystem == "Windows").WriteLine($"Error: {ex.Message}".Red().Bold());
+                ex.Command.ShowHelp();
+                return 1;
```
troydai: Now here error code is `1`

src/Microsoft.Dnx.ApplicationHost/Program.cs
moozzyk: Yes #2707

src/Microsoft.Dnx.ApplicationHost/Program.cs
davidfowl: Nah, that type then has to be shared throughout the stack. Use the property bag on exception

anurse: Then make a class in Sources that wraps up the exception and sets the property? Otherwise our `throw`s are going to get gross...

```csharp
var ex = new Exception("something");
ex.Data["suppressStackTrace"] = true
throw ex;
```

src/Microsoft.Dnx.Tooling/Program.cs
JunTaoLuo: This is the correct exit code, we'll change all other instances to `1` in #2707.

src/Microsoft.Dnx.ApplicationHost/Program.cs
JunTaoLuo: Going to change this since @davidfowl didn't like this. We'll throw here and catch/handle exception suppression in `RuntimeBootstrapper.cs` instead.

JunTaoLuo:
Anything else before signoff? @anurse @troydai @davidfowl @moozzyk

src/Microsoft.Dnx.Tooling/Program.cs
troydai: good.

troydai:
:shipit: looks good to me.

davidfowl:
Few things:
- Show some examples of what exceptions look like for the different cases (paste pictures)
- don't write tests using the old infrastructure

JunTaoLuo:
travis passed, merged to dev

JunTaoLuo:
@davidfowl The tests were written using the new infrastructure.

JunTaoLuo:
dnx:

non-existent project
image

non-existent command
image

dnu:

image

unrecognized or missing arguments/options display a similar message.

davidfowl:
@JunTaoLuo turn off DNX_TRACE and show the screenshots

davidfowl:
I think we need to hide the stack trace for compilation errors as well:

https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.ApplicationHost/Program.cs#L236

PR #2818, opened 2015-09-24
Writing pid if DNX_DEBUG_PID_PATH specified and correct

moozzyk:


moozzyk:
(travis will most likely first because the build is currently broken)

davidfowl:
:shipit:

anurse:
Actually, making this a real feature (not just for debugging in VS) is extremely useful for *nix daemon management. Many daemon launchers have a command line switch you can use to write a "PID File" which is just a file containing the PID (https://stackoverflow.com/questions/8296170/what-is-a-pid-file-and-what-does-it-contain). For example:

```
dnx --pid-file /var/run/mycoolapp_kestrel.pid kestrel
```

If people don't hate the idea, I'd be up for giving it a try as a separate PR (which could also remove this, I assume, since VS could just specify that command line argument). So no action is needed on this PR, just wanted to see what folks thought.

moozzyk:
The thought was that until we know how to debug CoreClr on non-windows we should not do this. Currently there is no support for this at all https://github.com/aspnet/dnx/blob/dev/src/dnx/pal.linux.cpp#L27-L30

Whenever I needed to debug anything on Linux I just started dnx directly from the debugger. 

Once it is possible to debug CoreClr we should revisit this.

DamianEdwards:
@anurse nice, although we'd still want the environment variable for the VS scenario as the cmd line is actually in the web.config file in that case and we don't want it to change to include a switch to drop the PID.

Issue #2664, opened 2015-09-09
Testing Infrastructure Improvements


JunTaoLuo:
Following up on #2617 

Some of the changes that I plan to add into the testing framework:

1 Organize environment variables for the test environment in Microsoft.Dnx.Testing.TestEnvironmentNames
- This will be similar to Microsoft.Dnx.Runtime.EnvironmentNames
- Prefix test specific names with DNX_TEST_*

2 Trace
- Set DNX_TEST_TRACE=1
- When enabled, test headers ("running test DnuPublishTest.PublishConsoleApp") test commands ("running dnu publish ...") will be displayed otherwise, only test summaries will be displayed

3 File retention behavior
- Set DNX_TEST_SAVE_FILES="None|All|Failed"
- Files saved to TestOutput/ will be deleted if None is set or if the test passed and Failed is set
- File deletion will occur at the end of each test method

4 stdout/stderr log (deferred)
- save the output of each test to files stdout.log/stderr.log inside each test directory in TestOutput/
- parallelization in xUnit makes redirecting output from Console difficult 
- also needs to determine a way to detect test method pass/failure without re-implementing xunit runner
- pending discussion

5 Test method headers
- announce the name of the test method to the output so it's easier to identify which test is running when scanning logs

6 Remove redundant testing (deferred)
- Currently some tests are run for each runtime in every environment and make tests very slow
- this requires changes to makefile and naming convention
- pending further discussion

7 Directory skip comparison
- currently when comparing directories, either the contents of all files are compared or no content is compared
- tag files/dirs. so that some files (especially binaries or directories of no interest) can be checked for existence only

8 Non-blocking execution of commands (deferred)
- This will be useful for cases such as DthTestServer in DTH FunctionalTests
- this is isolated to DTH for now so will not be part of this change

9 Specify a specific runtime via Environment Variable
- instead of all runtimes, specify one via EV so re-running tests during debug will be much faster

10 reverse runtime/testname structure
- this is required for 11

11 ensure restore and cache restored project
- some tests requires only one restore across all runtimes

12 make the default logs less verbose
- do not output execution logs if dnx/dnu commands are successful

13 renaming Microsoft.Dnx.Testing.CommonUtils -> Microsoft. (deferred)
- this is a simple change but I might not have time for this

14 DnxTestContext (deferred)
- see anurse's comment below for details
- I might not have time for this

JunTaoLuo:
[...]

PR #2823, opened 2015-09-24
Johluo/2664 testing infra improvements

JunTaoLuo:
#2664 Adding items 1, 2, 3, 5, 7, 9, 10, and 12.

If these are approved I'll go through the existing tests that uses the framework and make the necessary updates.


test/Microsoft.Dnx.Testing.CommonUtils/TestLogger.cs
```diff
@@ -0,0 +1,42 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Dnx.Testing
+{
+    internal static class TestLogger
```
JunTaoLuo: This is a copy of Logger from Runtime. I'm creating a new one here so that the logging from tests can be controlled by a separate environment variable DNX_TEST_TRACE. This allows for clarity and verbosity control when logging the output for tests.

test/Microsoft.Dnx.Testing.SampleTests/DnuCommonUtilTests.cs
```diff
@@ -0,0 +1,86 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Newtonsoft.Json.Linq;
+using Xunit;
+
+namespace Microsoft.Dnx.Testing.SampleTests
+{
+    [Collection(nameof(SampleTestCollection))]
+    public class DnuCommonUtilTests : DnxSdkFunctionalTestBase
```
JunTaoLuo: I've added these tests for testing what I'm working on in the testing infrastructure. Some of these will eventually be removed or moved to something like Testing.Framework.Tests to prevent regression in the test infrastructure.

test/Microsoft.Dnx.Testing.CommonUtils/Directory/DirItem.cs
```diff
@@ -0,0 +1,28 @@
...
+        }
+
+        public DirItem()
+            : this(Dir.EmptyFile, true)
+        {
+        }
+
+        public bool SkipComparison { get; set; }
+
+        public object Item { get; set; }
```
JunTaoLuo: setter might not be necessary

JunTaoLuo:
Todo: explain some of what I did and the use cases

test/Microsoft.Dnx.Testing.SampleTests/DnuCommonUtilTests.cs
```diff
@@ -0,0 +1,86 @@
...
+    {
+        [Theory, TraceTest]
+        [MemberData(nameof(DnxSdks))]
+        public void FailTest(DnxSdk sdk)
+        {
+            var feedSolution = TestUtils.GetSolution<DnuCommonUtilTests>(sdk, "DependencyGraphsProject");
```
davidfowl: The using pattern is cleaner and makes it so that user code doesn't have to write try {}, finally{}

JunTaoLuo: Do you mean the CleanUpTestDir() function? If so, that was intentional since cleanup should, by default, only occur when the test passed. Failed tests should retain the files used during the test. I know it's kind of hacky but the only way I have right now of telling whether the test passed is if the last statement is reached. Unless I want to  extend the xUnit test runner class that is.

test/Microsoft.Dnx.Testing.CommonUtils/Exec.cs
```diff
@@ -80,7 +79,6 @@ public class Exec
             {
                 if (args.Data != null)
                 {
-                    Console.WriteLine(args.Data);
```
anurse: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1:

anurse:
:shipit: - I had more ideas but decided to leave them for future PRs ;)

JunTaoLuo:
merged to dev

Issue #2681, opened 2015-09-11
OOM when using MVC solution in VS


davidfowl:
Seems we're holding onto a bunch more library descriptions. This can easily be fixed by caching more strings, versions, package ids, paths.

davidfowl:
@rynowak let me know if you see the problem still. The changes drop the memory that is being held onto. There's another issue to just reduce allocations in general which will be submitted later.

rynowak:
This is still happening. There's too much stuff rooted - I get into this state working with either MVC solution for about an hour.

![image](https://cloud.githubusercontent.com/assets/1430011/10056255/2a3dd4f4-61ef-11e5-8d03-4b0c47d934ac.png)

Dump is available on corpnet.

davidfowl:
@rynowak if you see more issues, please reopen again :smile:

PR #2827, opened 2015-09-25
Several DTH and compilation performance improvements

davidfowl:
There were some general regressions from beta7 to beta8 wrt performance of DTH and compilation. This fix intends to rectify thoseand also improve on some of the older logic (<= beta7)

- Don't create the build time load context until required. For it's
not needed unless a compile module is being loaded or a custom compiler
is being loaded.
- The LibraryExporter no longer holds onto the build load context
- Fixed a disposal caching bug
- Nuke IProjectGraphProvider because it provided no real value
- Made ProjectExportContext which acts as a disposable cache entry
for the build time load context.
- Added friendly name to LoadContext for easy debugging
- Removed background compilation from DTH as it was harmful for large
solutions. Things are only compiled on demand now.
- ProjectAssemblyLoader will only take the subset of project that it needs

#2681 

/cc @rynowak @troydai


src/Microsoft.Dnx.Compilation/BuildLoadContext.cs
```diff
@@ -0,0 +1,83 @@
+﻿using System;
```
troydai: License and registration please.

src/Microsoft.Dnx.Loader/LoadContext.cs
```diff
@@ -103,7 +113,11 @@ static LoadContext()
 
         public LoadContext()
         {
-            _contextId = Guid.NewGuid().ToString();
+        }
```
troydai: Seems that `BuildLoadContext` is the only user of this default constructor and it doesn't add the `LoadContext` itself to the `_contexts`. It feels like way too subtle...

src/Microsoft.Dnx.Compilation/BuildLoadContext.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.IO;
+using System.Reflection;
+using Microsoft.Dnx.Runtime;
+using Microsoft.Dnx.Runtime.Loader;
+
+namespace Microsoft.Dnx.Compilation
+{
+    // This load context isn't tracked in the dictionary, it's a shim
+    // so that we don't allocte a real load context until absolutely necessary
```
rynowak: allocate

src/Microsoft.Dnx.Loader/LoadContext.cs
davidfowl: It could be another base class but it has to be implemented on CoreCLR and on CLR and all of the implementation is the same except for the fact that we don't track the LoadContext in the static.

src/Microsoft.Dnx.Compilation/BuildLoadContext.cs
```diff
@@ -0,0 +1,83 @@
...
+            catch (FileNotFoundException)
+            {
+                return LoadContext.LoadWithoutDefault(assemblyName);
+            }
+        }
+
+        public override Assembly LoadAssembly(AssemblyName assemblyName)
+        {
+            throw new NotImplementedException();
+        }
```
rynowak: Shouldn't these two do the same thing?

rynowak:
:shipit: after fixing minor issues

src/Microsoft.Dnx.Compilation/BuildLoadContext.cs
davidfowl: allocte

src/Microsoft.Dnx.Compilation/BuildLoadContext.cs
davidfowl: Nope. This should never be called. The base class only calls this via assembly resolve. This is a pass through implementation.

src/Microsoft.Dnx.Runtime/Loader/RuntimeLoadContext.cs
```diff
@@ -13,14 +13,15 @@ public class RuntimeLoadContext : LoadContext
...
-        public RuntimeLoadContext(IEnumerable<LibraryDescription> libraries,
+        public RuntimeLoadContext(string friendlyName,
+                                  IEnumerable<LibraryDescription> libraries,
                                   ICompilationEngine compilationEngine,
-                                  IAssemblyLoadContext defaultContext)
+                                  IAssemblyLoadContext defaultContext) : base(friendlyName)
         {
             // TODO: Make this all lazy
             // TODO: Unify this logic with default host
             var projects = libraries.Where(p => p.Type == LibraryTypes.Project)
```
troydai: Just use `OfType<ProjectDescription>` could be enough, right?

troydai:
:shipit:

PR #2829, opened 2015-09-25
Don't wait for debugger if no --debug switch...

moozzyk:
...even if DNX_DEBUG_PID_PATH defined


troydai:
:shipit:

anurse:
:shipit:

moozzyk:
Plans have changed - we will not be using `DNX_DEBUG_PID_PATH` at all... so I removed even more code

anurse:
:shipit:

davidfowl:
:shipit:

PR #2835, opened 2015-09-26
Made changes to DTH testing infrastructure and added coverage

davidfowl:
- Added basic coverage for compile modules in DTH
- Fixed project context association
- Removed dependency on old test framework

/cc @troydai


troydai:
:shipit:

Issue #1043, opened 2015-01-07
Polish the signature of CallApplicationMain in klr


troydai:
Here's the function pointer declaration (https://github.com/aspnet/KRuntime/blob/dev/src/klr/klr.h#L11)

``` C++
typedef struct CALL_APPLICATION_MAIN_DATA
{
    LPCWSTR applicationBase; // application base of managed domain
    LPCWSTR klrDirectory; // path to KLR helper directory
    int argc; // Number of args in argv
    LPCWSTR* argv; // Array of arguments
    int exitcode; // Exit code from Managed Application
} *PCALL_APPLICATION_MAIN_DATA;

typedef HRESULT (STDAPICALLTYPE *FnCallApplicationMain)(
    PCALL_APPLICATION_MAIN_DATA pCallApplicationMainData
    );
```

The `pCallApplicationMainData` will contain the exit code, there is no need to return the it from the method again. The return type can be instead change to `bool` so that the caller can be further simplified.

troydai:
@glennc @muratg If we're not going to fix it, should we just close it?

muratg:
@davidfowl @lodejard What are your thoughts on this one?

davidfowl:
Part of any clean up that @moozzyk does

moozzyk:
The `HRESULT` returned by the `FnCallApplicationMain` does not represent the return value of the target function but the result of the operation of calling the target function (e.g. if for whatever reason we cannot start CoreClr the FnCallApplicationMain will return the HRESULT representing the failure and will not set the `exitCode`). 

What we should probably fix is actually to remove this line:
https://github.com/aspnet/dnx/blob/dev/src/dnx.clr/dnx.clr.cpp#L30

moozzyk:
Fixed in #2857

PR #2857, opened 2015-09-28
Cleaning up leftovers in native code

### LINKED ISSUE #1603: Rationalize type and function usage in native parts of DNX
Currently we mix different aliases of types across the code e.g. `WCHAR*`, `LPWSTR`, `LPTSTR` are used all over the place (e.g. bool GetTrustedPlatformAssembliesList(WCHAR* szDirectory, bool bNative, LPWSTR pszTrustedPlatformAssemblies, size_t cchTrustedPlatformAssemblies))
even though they are basically the same type. 

Similarly we mix usage of `_T` and `L` macros and also mix string functions like `_tcs*` and `wcs*`  and 
`_tprintf_s` vs. `wprintf_s`

We should just use one alias or ideally - switch to STL and not use them at all. 

We could consider creating our own alias for `std::string` and `std::wstring` (like dnx::string_t) type that would be used across the codebase regardless of what platform we are using.

moozzyk:
Fixes #1603, #1043

### DIFF
--- a/src/dnx.clr/dnx.clr.cpp
+++ b/src/dnx.clr/dnx.clr.cpp
@@ -7,6 +7,7 @@
 
 #include "ClrBootstrapper.h"
 #include "app_main.h"
+#include  <iostream>
 
 IClrBootstrapperPtr g_clrBootstrapper;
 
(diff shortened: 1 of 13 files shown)


src/dnx.clr/dnx.clr.cpp
```diff
@@ -23,15 +24,14 @@ extern "C" __declspec(dllexport) HRESULT __stdcall CallApplicationMain(PCALL_APP
...
-        dnx::utils::wait_for_debugger(data->argc, data->argv);
+        dnx::utils::wait_for_debugger(data->argc, data->argv, L"--debug");
 
         g_clrBootstrapper = NULL;
         data->exitcode = bootstrapper->CallApplicationMain(data->argc, data->argv);
     }
     else
     {
-        printf_s("Failed to initialize runtime (%x)", hr);
-        data->exitcode = hr;
```
davidfowl: Why don't we set the exit code anymore?

davidfowl:
:shipit:

src/dnx.clr/dnx.clr.cpp
moozzyk: We only return exitcode if hr is S_OK (https://github.com/aspnet/dnx/blob/dev/src/dnx/pal.win32.cpp#L146) so it is not needed here because hr is not S_OK.

PR #2859, opened 2015-09-29
Remove fixture from DTH tests

troydai:
Fully rely on new test infrastructure.

@davidfowl @JunTaoLuo 

Testing time was significantly increased. 
```
Microsoft.Dnx.DesignTimeHost.FunctionalTests  Total: 96, Errors: 0, Failed: 0, Skipped: 0, Time: 1084.000s
```


test/Microsoft.Dnx.DesignTimeHost.FunctionalTests/DthStartupTests.cs
```diff
@@ -531,5 +515,15 @@ public void AddDepsReturnsReferences(DnxSdk sdk)
                 }
             }
         }
+
+        private Project GetTestProject(DnxSdk sdk, string projectName)
```
davidfowl: Get rid of this helper and do it inline

test/Microsoft.Dnx.DesignTimeHost.FunctionalTests/DthStartupTests.cs
```diff
@@ -308,24 +295,24 @@ public void DthNegative_BrokenProjectPathInLockFile_V1(DnxSdk sdk)
...
 
             using (var disposableDir = new DisposableDir())
             using (var server = DthTestServer.Create(sdk))
             using (var client = server.CreateClient())
             {
-                // copy the project to difference location so that the project path in its lock file is invalid
-                var targetPath = Path.Combine(disposableDir, "BrokenProjectPathSample");
-                Testing.TestUtils.CopyFolder(testProject, targetPath);
+                var movedProjectPath = Path.Combine(disposableDir, projectName);
+                TestUtils.CopyFolder(project.ProjectDirectory, movedProjectPath);
```
davidfowl: Why do you need to do this?

test/Microsoft.Dnx.Testing.CommonUtils/ProjectExtensions.cs
```diff
@@ -26,5 +26,22 @@ public static Project UpdateProjectFile(this Project project, Action<JObject> up
             Project.TryGetProject(project.ProjectFilePath, out project);
             return project;
         }
+
+        public static Project Restore(this Project project, DnxSdk sdk, bool ensureSuccess)
```
davidfowl: These are in the wrong place.

test/Microsoft.Dnx.DesignTimeHost.FunctionalTests/DthStartupTests.cs
troydai: Render the project path in lock file invalid.

test/Microsoft.Dnx.Testing.CommonUtils/ProjectExtensions.cs
troydai: This is a extension of the Project not dnu.

davidfowl: Get rid of all helpers. Make all of the tests look the same.

test/Microsoft.Dnx.DesignTimeHost.FunctionalTests/DthStartupTests.cs
davidfowl: Why do you need to copy the project twice?

troydai: First time restore it to generate a correct lock file. Then move it to another place but do not move its dependency so that the relative path in lock file is invalid.

davidfowl: Why can't you change the file in place?

troydai: Because 1. Editing the json file is move complex; 2. Once we update the lock file format the test needs to be rebaseline.

Here actually only two files are copied and it mimic the real situation when the bug was found.

troydai: Another reason. Once we have the replacement of _fixture_ ready, we will still copy the project to a 3rd place and avoid editing the one copy reused across tests.

davidfowl: > Because 1. Editing the json file is move complex; 2. Once we update the lock file format the test needs to be rebaseline.

We have checked in lock files? 
What are you editing?

> Another reason. Once we have the replacement of fixture ready, we will still copy the project to a 3rd place and avoid editing the one copy reused across tests.
 Add a line note

When we have that this test won't use a fixture.

troydai: What? 
We didn't check in lock file.
We don't edit file now.

davidfowl:
:shipit:

PR #2865, opened 2015-09-29
Changing return code from -1 to 1

### LINKED ISSUE #2707: Don't return -1 exit code from managed code

moozzyk:
Fixes #2707

### DIFF
--- a/test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.Old.cs
+++ b/test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.Old.cs
@@ -364,7 +364,7 @@ public class TestClass : BaseClass {
                     out stdOut,
                     out stdErr);
 
-                Assert.NotEqual(0, exitCode);
+                Assert.Equal(1, exitCode);
                 Assert.NotEmpty(stdErr);
                 Assert.DoesNotContain("POST_BUILD_SCRIPT_OUTPUT", stdOut);
                 Assert.DoesNotContain("POST_PACK_SCRIPT_OUTPUT", stdOut);
@@ -475,7 +475,7 @@ public class TestClass : BaseClass {
                     $"{projectJsonPath} -s {emptyLocalFeed}",
                     out stdOut,
                     out stdErr);
-                Assert.NotEqual(0, exitCode);
+                Assert.Equal(1, exitCode);
 
                 exitCode = DnuTestUtils.ExecDnu(
                     runtimeHomeDir,
@@ -484,7 +484,7 @@ public class TestClass : BaseClass {
                     out stdOut,
                     out stdErr);
 
-                Assert.NotEqual(0, exitCode);
+                Assert.Equal(1, exitCode);
                 Assert.NotEmpty(stdErr);
                 var unresolvedDependencyErrorCount = stdErr
                     .Split(new[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries)
--- a/src/Microsoft.Dnx.ApplicationHost/Program.cs
+++ b/src/Microsoft.Dnx.ApplicationHost/Program.cs
@@ -60,7 +60,7 @@ namespace Microsoft.Dnx.ApplicationHost
 
                 if (host.Project == null)
                 {
-                    return Task.FromResult(-1);
+                    return Task.FromResult(1);
                 }
 
                 var lookupCommand = string.IsNullOrEmpty(options.ApplicationName) ? "run" : options.ApplicationName;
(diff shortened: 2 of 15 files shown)


muratg:
:shipit:

troydai:
Any tests?

moozzyk:
@troydai - existing tests are all passing and some of them are checking that the exit code is not 0. I can probably add tests for some conditions (e.g. not sure about this: https://github.com/aspnet/dnx/pull/2865/files#diff-0b1faa794a3c5fb04d329d96e22f8391L55) if you convince me that they add value.

JunTaoLuo:
For the current tests that assert exit code is not 0, can we make it more specific and assert that they are equal to 1 where appropriate? I was trying to do this previously but we had a mix of -1 and 1 then.

troydai:
Test non-zero. That makes sense. I was wondering why there is no test broken. Only thing I'd like to have covered in tests is that instead of testing non-zero how about testing `>0` so as to cover the essence of this change.

moozzyk:
@JunTaoLuo - I can try that.

moozzyk:
Testing `> 0` is not much better than `!= 0` I think. But I will try what @JunTaoLuo proposed and see how far I can get.

JunTaoLuo:
Also, should we just go ahead and make `dnu` consistent as well? 
E.g. https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Tooling/ConsoleCommands/BuildConsoleCommand.cs#L49
Same with `PackConsoleCommand` and `PublishConsoleCommand`

moozzyk:
I checked that - this value is not being used anywhere let alone returned as exit code so I left it as is.

JunTaoLuo:
I agree we currently don't use them anywhere but I think they are being returned as exit codes:

image

moozzyk:
@JunTaoLuo - now I see this. I missed that they are returned from the OnExecute and not from the `Register` function. Will fix that because as you said - they are being returned as exit code.

moozzyk:
@JunTaoLuo @troydai - Updated the PR:
- all the tests that used `NotEqual(0, exitCode)` are checking against the expected exit code (1 or 2)
- all the commands now return 1 in case of errors (good catch @JunTaoLuo - I missed the `}` and therefore thought that we never checked/used the value)

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.Old.cs
```diff
@@ -475,7 +475,7 @@ public void DnuPack_ShowUnresolvedDependencyWhenBuildFails(string flavor, string
                     $"{projectJsonPath} -s {emptyLocalFeed}",
                     out stdOut,
                     out stdErr);
-                Assert.NotEqual(0, exitCode);
```
JunTaoLuo: Missed the `Assert.NotEqual(0, exitCode);` on line 487

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.Old.cs
```diff
@@ -475,7 +475,7 @@ public void DnuPack_ShowUnresolvedDependencyWhenBuildFails(string flavor, string
...
+                Assert.Equal(1, exitCode);
```
JunTaoLuo: Missed the `Assert.NotEqual(0, exitCode);` on line 487

moozzyk: Fixed.

davidfowl:
Travis?

muratg:
Travis failure seems like a timeout, no?

davidfowl:
Then force a re-run

troydai:
I kicked off a re-run. You're welcome.

troydai:
Looks good to me. :shipit: if Travis passes.

muratg:
One of the runs seem passed.  Not sure why it' s not reflected here.

JunTaoLuo:
Not sure what has happened with travis but the Linux run seem to error due to timeout very frequently. It usually takes 2-3 tries to get a success.

davidfowl:
@JunTaoLuo I think we have an actual bug somewhere.

davidfowl:
:shipit:

moozzyk:
@muratg @troydai  - thanks for following up and kicking off the travis build!

Issue #2825, opened 2015-09-25
Testing infrastructure improvements round 2


JunTaoLuo:
Items not completed in #2664 should be addressed here.
- [ ] 1 stdout/stderr log
- save the output of each test to files stdout.log/stderr.log inside each test directory in TestOutput/
- parallelization in xUnit makes redirecting output from Console difficult 
- also needs to determine a way to detect test method pass/failure without re-implementing xunit runner
- [ ] 2 Remove redundant testing
- Currently some tests are run for each runtime in every environment and make tests very slow
- this requires changes to makefile and naming convention
- [ ] 3 Non-blocking execution of commands
- This will be useful for cases such as DthTestServer in DTH FunctionalTests
- [x] 4 renaming Microsoft.Dnx.Testing.CommonUtils -> Microsoft.Dnx.Testing.Framework
- self explanatory
- [ ] 5 DnxTestContext
- see anurse's comment in #2664 
- [ ] 6 Test fixtures
- organize collections of tests
- I see that in xUnit, organizing tests is achieved as nested classes: https://github.com/xunit/xunit/blob/master/test/test.xunit.assert/Asserts/CollectionAssertsTests.cs
- [X] 7 Shared solution among tests
- restore once capabilities can help speed up tests
- Implemented by @troydai in #2877
- [ ] 8 Centralized usings before and after test
- reduce boilerplate code in test methods
- [ ] 9 Long Path !!!
- short term solution: shorten runtime names via #2889, also make sure test names are short
- long term solution: determine if index files are required.
- [ ] 10 Hide the dnx output for expected errors
- Sometimes we give invalid commands to make sure certain error messages are displayed. In these cases it's best to hide the output from failed dnx/dnu calls.
- [ ] 11 File retention 
- Currently this is implemented as a cleanup function in TestUtils and in Dispose of the TestFixtureBase
- Cleanup after each test method needs to be called explicitly at the end of each test and we no longer use TestFixtureBase so this implementation needs to be revisited
- [ ] 12 Test Trace 
- Currently this is implemented as an explicit attribute for each test, this should be avoided
- [ ] 13 Port over old tests 
- do this a portion at a time

JunTaoLuo:
1 Test fixtures

2 Shared Solution dirs

3 centralized usings before and after test

PR #2868, opened 2015-09-29
Renaming test infrastructure and updating tests

JunTaoLuo:
Item 4 of #2825.


JunTaoLuo:
cc @anurse

troydai:
I know it's hard but can we separate rename from updating tests?

test/Microsoft.Dnx.ApplicationHost.FunctionalTests/ApplicationHostTestFixture.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Dnx.Testing.Framework;
+
+namespace Microsoft.Dnx.ApplicationHost.FunctionalTests
+{
+    public class ApplicationHostTestFixture : DnxSdkFunctionalTestFixtureBase
```
troydai: Do we actually need this?

troydai:
So please remove all the fixtures and collection attribute.

JunTaoLuo:
Travis is passing, pinging. @troydai @anurse @moozzyk .

davidfowl:
@JunTaoLuo do we have to add TraceTest on every test?

JunTaoLuo:
Note that this probably conflicts with #2865 since I renamed a few files. I'll fix that in another commit after that PR is merged.

JunTaoLuo:
@davidfowl unfortunately yes. Unless we want to do something during test discovery.

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuRestoreTests.cs
```diff
@@ -1,193 +1,74 @@
...
-  ""dependencies"": {
-    ""alpha"": ""0.1.0""
-  }
-}");
-                DnuTestUtils.ExecDnu(
-                    runtimeHomePath,
-                    subcommand: "restore",
-                    arguments: $"{projectDir} -s {_fixture.PackageSource} --packages {packagesDir}");
+            // Delete the project A
+            CommonTestUtils.TestUtils.DeleteFolder(solution.GetProject("A").ProjectDirectory);
```
davidfowl: Why the 2 part type name?

JunTaoLuo: `DeleteFolder` is not in the new infrastructure, so the one from the old one is used. I'll probably add it in a future change.

test/Microsoft.Dnx.ApplicationHost.FunctionalTests/AppHostTests.Old.cs
```diff
@@ -0,0 +1,309 @@
...
+            var runtimeHomeDir = _fixture.GetRuntimeHomeDir(flavor, os, architecture);
+
+            string stdOut, stdErr;
+            var exitCode = TestUtils.ExecBootstrapper(
+                runtimeHomeDir,
+                arguments: string.Empty,
+                stdOut: out stdOut,
+                stdErr: out stdErr);
+
+            Assert.NotEqual(0, exitCode);
```
moozzyk: I merged #2707 - can you try checking the actual exit code.

moozzyk: (or maybe it will get fixed automatically when you rebase on the latest dev)

JunTaoLuo: I will need to make a new commit to port over the changes from `AppHostTests.cs` to `AppHostTests.Old.cs`

JunTaoLuo:
pinging again @troydai @moozzyk @anurse @davidfowl

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuRestoreTests.Old.cs
```diff
@@ -0,0 +1,192 @@
...
+using Microsoft.Dnx.CommonTestUtils;
+using NuGet;
+using Xunit;
+
+namespace Microsoft.Dnx.Tooling.FunctionalTests.Old
+{
+    [Collection(nameof(PackageManagerFunctionalTestCollection))]
+    public class DnuRestoreTests
+    {
+        private readonly PackageManagerFunctionalTestFixture _fixture;
```
troydai: Fixture still needed?

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
```diff
@@ -73,6 +80,8 @@ public void P2PDifferentFrameworks(DnxSdk sdk)
 
             // Assert
             Assert.Equal(0, result.ExitCode);
+
+            TestUtils.CleanUpTestDir<DnuPackTests>(sdk);
```
troydai: Interesting. Why do we have to cleanup the test dir afterwards? Can the coming up test clean it? If the assertion fails, this line won't run.

troydai: The tear down should happen in test class's dispose method.

test/Microsoft.Dnx.DesignTimeHost.FunctionalTests/DthStartupTests.cs
```diff
@@ -520,6 +536,8 @@ public void AddDepsReturnsReferences(DnxSdk sdk)
                     Assert.Equal("DoesNotExist", error2["Source"].Value<string>("Name"));
                 }
             }
+
+            TestUtils.CleanUpTestDir<DthStartupTests>(sdk);
```
troydai: This won't be run if error happens.

test/Microsoft.Dnx.ApplicationHost.FunctionalTests/ApplicationHostTestCollection.cs
```diff
@@ -2,11 +2,12 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.Dnx.CommonTestUtils;
+using Microsoft.Dnx.Testing.Framework;
 using Xunit;
 
 namespace Microsoft.Dnx.ApplicationHost.FunctionalTests
 {
-    [CollectionDefinition("ApplicationHostTestCollection")]
+    [CollectionDefinition(nameof(ApplicationHostTestCollection))]
```
troydai: Do we still need this fixture?

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuRestoreTests.Old.cs
JunTaoLuo: Do we need to delete it from old tests?

test/Microsoft.Dnx.ApplicationHost.FunctionalTests/AppHostTests.Old.cs
```diff
@@ -0,0 +1,309 @@
...
+using System.IO;
+using System.Runtime.Versioning;
+using Microsoft.Dnx.CommonTestUtils;
+using Microsoft.Dnx.Tooling;
+using Microsoft.Dnx.Runtime;
+using Microsoft.Dnx.Runtime.Common.Impl;
+using NuGet;
+using Xunit;
+
+namespace Microsoft.Dnx.ApplicationHost.Old
```
troydai: I assume this will be eventually retired with the fixture. Am I right?

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuRestoreTests.Old.cs
troydai: We got a timeline for retiring the old tests?

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
JunTaoLuo: Can't use the dispose method since I need to name of the test method in the clean up. This is run at the end of each test to delete any files copied from misc during the test. The coming up test can clean it but this cleans it eagerly.

test/Microsoft.Dnx.DesignTimeHost.FunctionalTests/DthStartupTests.cs
JunTaoLuo: The default behaviour is to keep the files for reference if the test failed.

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuRestoreTests.Old.cs
JunTaoLuo: As soon as possible?

test/Microsoft.Dnx.ApplicationHost.FunctionalTests/ApplicationHostTestCollection.cs
JunTaoLuo: I believe it is required for the DnxRuntimeFixture.

test/Microsoft.Dnx.ApplicationHost.FunctionalTests/AppHostTests.Old.cs
JunTaoLuo: Right

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
troydai: Here's the limitation of binding folder to the test class name, you can't leverage the setup-teardown mechanism of xUnit.

moozzyk: Then, I think there should be a disposable test scope which would do the cleanup. The tests would use the `using` statement to define the scope. Otherwise you may leave leftovers behind. It is also quite easy to miss adding `TestUtils.CleanUpTestDir<DnuPackTests>(sdk);` or even hard to know when it is needed...

JunTaoLuo: The issue with the using pattern is that dispose is called at the end of the test regardless of whether the test passed or failed. In fact, there isn't a good way of detecting this as part of xUnit. Hence the not so ideal way of cleaning up.

troydai: I get the point that it is suppose to clean the folder only if test failed. However it introduces another issue, it the `Theory` case it _may_ clean the test folder anyway because the folder name is decided by __test method name__ + __sdk name__, if there is another factor in the data set (like this case: https://github.com/aspnet/dnx/blob/dev/test/Microsoft.Dnx.DesignTimeHost.FunctionalTests/DthStartupTests.cs#L52) then even failed test's record will be erased. 

The point I want to make here is:
1. Function like retaining test trace should be built into the test infrastructure rather than required the test to explicitly invoke it.
2. It's a unstable and weird implementation.

test/Microsoft.Dnx.ApplicationHost.FunctionalTests/ApplicationHostTestCollection.cs
troydai: Cool.

test/Microsoft.Dnx.ApplicationHost.FunctionalTests/AppHostTests.Old.cs
troydai: Cool.

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuRestoreTests.Old.cs
troydai: Cool.

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
JunTaoLuo: This makes sense, I'll make a point for it in #2825 . We should revisit this implementation of this feature in the future but not in this PR.

troydai: I'm find with that.

troydai:
Please ensure all the issues mentioned in this PR are tracked.

:shipit:

JunTaoLuo:
merged to dev

Issue #2773, opened 2015-09-22
We need logic to translate non-Windows OS, OSName to RID


moozzyk:
Currently we create a valid RuntimeId only for Windows. For other OSes we just concatenate os and os version (https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs#L106). This kind of worked before we started passing actual OS details but now in my branch that provides the details we get:

`darwin10.9-x64`
`linuxUbuntu 12.04-x64`

Also note that we say that we only recognize Ubuntu 14.04 as a supported version but Travis is running on 12.04.

anurse:
Yeah, this is going to get a bit hairy. We can do fairly simple processing of the OS name to get it in to a RID.

The Travis issue is concerning. I'm not sure right now what the best option is... /cc @davidfowl

moozzyk:
I am just hacking it - will show you soon...

moozzyk:
Yes, creating a RID should be simple. For now I am just working around this issue since I don't want to make the PR any bigger.

anurse:
Ouch, this should happen for beta8. /cc @muratg @DamianEdwards @Eilon 

OSX is working, but Ubuntu is reporting a RID of `linux.Ubuntu 14.04-x64` when it should be `ubuntu.14.04-x64`

muratg:
Ugh, yeah I think we should take it.

anurse:
We could consider taking a small fix here as well to allow Linux Mint 17.2 to behave as though it is Ubuntu 14.04 (since they share a package base). Thoughts @DamianEdwards? This would not be a formal declaration of support, just kinda like we do with Mono: "It should work, but no formal guarantees"

dougbu:
FYI on my up-to-date Mint system:
```
> dnu --version
Microsoft .NET Development Utility
 Version:      1.0.0-rc1-15775
 Type:         CoreClr
 Architecture: x64
 OS Name:      Linux
 OS Version:   LinuxMint 17.2
 Runtime Id:   linux.LinuxMint 17.2-x64
> uname -a
Linux GarnetAv6 3.16.0-38-generic #52~14.04.1-Ubuntu SMP Fri May 8 09:43:57 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
>  cat /etc/lsb-release
DISTRIB_ID=LinuxMint
DISTRIB_RELEASE=17.2
[...]

PR #2873, opened 2015-09-30
Fix Linux RID synthesis logic

anurse:
Fixed the way we synthesize RIDs on Linux. The native code passes the values of `DISTRIB_ID` and `DISTRIB_RELEASE` from `/etc/lsb-release` as the Os "Version". For example, if lsb-release has the following:

```
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=14.04
DISTRIB_CODENAME=trusty
DISTRIB_DESCRIPTION="Ubuntu 14.04.3 LTS"
```

The native code sends:

```
OS Name: Linux
OS Version: Ubuntu 14.04
```

Before this change, we'd syntheize a RID of `linux.Ubuntu 14.04-x64` which is totally wrong ;). After this change, the RID will be `ubuntu.14.04-x64`

This change ALSO includes some minor additions to our RID mappings to smooth the experience for people using **unsupported** distros/OS versions which still seem to work well for now.

* `linuxmint.17.2` is defined, which falls back to `ubuntu.14.04`. They should be equivalently compatible, so we want to smooth out that experience. However, this is **not** formal support, nor will it be fully tested. (Sign-off activities should continue to take place primarily on the officially supported distro: Ubuntu 14.04)
* `osx.10.11` is defined, which falls back to `osx.10.10`. Again, this is just for convenience and to be a little optimistic. We hope El Capitan will work, but aren't formally supporting it. We can always remove it later.

/cc @davidfowl @muratg @Eilon @moozzyk @dougbu


anurse:
Waiting for full sign-off on this going in to `release` so review, but be aware that it might change a little (we may remove the additional RID definitions)

src/Microsoft.Dnx.Tooling/Restore/ImplicitPackagesWalkProvider.cs
```diff
@@ -70,7 +75,11 @@ internal class ImplicitPackagesWalkProvider : IWalkProvider
             new RuntimeSpec("ubuntu-x64", "ubuntu", "linux-x64"),
 
             new RuntimeSpec("ubuntu.14.04", "ubuntu"),
-            new RuntimeSpec("ubuntu.14.04-x64", "ubuntu.14.04", "ubuntu-x64"));
+            new RuntimeSpec("ubuntu.14.04-x64", "ubuntu.14.04", "ubuntu-x64"),
+
```
moozzyk: do we also want to add ubuntu 15.04?

anurse:
Got sign-off.

src/Microsoft.Dnx.Tooling/Restore/ImplicitPackagesWalkProvider.cs
anurse: If you confirm for me that `dnu restore` works on CoreCLR on Ubuntu 15.04, I'm willing to add it. It would be similar to Linux Mint, just an unblocker, not formal support.

src/Microsoft.Dnx.Tooling/Restore/ImplicitPackagesWalkProvider.cs
```diff
@@ -70,7 +75,11 @@ internal class ImplicitPackagesWalkProvider : IWalkProvider
...
+            // Linux Mint 17.2 is identical to Ubuntu 14.04 for our purposes (though we are not formally supporting it either)
+            new RuntimeSpec("linuxmint.17.2", "ubuntu.14.04"),
+            new RuntimeSpec("linuxmint.17.2-x64", "ubuntu.14.04-x64"));
```
moozzyk: will it be `linuxmint.17.2-x64` or just `mint.17.2-x64`?

src/Microsoft.Dnx.Tooling/Restore/ImplicitPackagesWalkProvider.cs
moozzyk: Leave it for now...

src/Microsoft.Dnx.Tooling/Restore/ImplicitPackagesWalkProvider.cs
moozzyk: I saw the `lsb-release` output from Mint and it will be `linuxmint.17.2-x64`

moozzyk:
:ship::it:

dougbu:
Is this the fix for #2773? Comments don't make it clear.

anurse:
Fixes #2773

dougbu:
:dancers:

muratg:
@anurse Can you check this in ASAP?

anurse:
Yep, I'll merge it momentarily

PR #2879, opened 2015-10-01
Adding support for CommandLineArguments to IRuntimeEnvironment

grahamehorner:
The idea behind this change is to allow any/all applications hosted by dnx read access to the commandline arguments passed to dnx from an instance of the IRuntimeEnvironment


davidfowl:
We don't want this API. Instead we should just get the API back on .NET Core.

moozzyk:
(It also wouldn't work on CoreClr)

PR #2880, opened 2015-10-01
fix the build by reporting all linuxes as ubuntu.14.04 for now...

anurse:
Also had to rebuild the test package.

It's not pretty... I'm filing an issue to fix this linux RID thing in RC 1 once and for all :)

/cc @BrennanConroy @muratg @davidfowl 

I'd like to see Travis pass before polluting release with changes that won't actually work.


anurse:
Filed https://github.com/aspnet/dnx/issues/2881 to fix this in RC...

BrennanConroy:
As long as the nupkg got built correctly :shipit:

anurse:
I can send you a Hex diff if you want ;P

moozzyk:
:ship: :it:

anurse:
Merged to both `release` and `dev`

Issue #2876, opened 2015-09-30
Race condition in breadcrumbs


Tratcher:
```
[Smoke:WebListener:CoreClr:x86] Unhandled Exception: System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
[Smoke:WebListener:CoreClr:x86]    at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
[Smoke:WebListener:CoreClr:x86]    at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
[Smoke:WebListener:CoreClr:x86]    at System.Collections.Generic.List`1.Enumerator.MoveNext()
[Smoke:WebListener:CoreClr:x86]    at Microsoft.Dnx.Runtime.Servicing.Breadcrumbs.WriteAllBreadcrumbsInternal()
[Smoke:WebListener:CoreClr:x86]    at Microsoft.Dnx.Runtime.Servicing.Breadcrumbs.<>c.<WriteAllBreadcrumbs>b__16_0(Object state)
[Smoke:WebListener:CoreClr:x86]    at System.Threading.QueueUserWorkItemCallback.WaitCallback_Context(Object state)
[Smoke:WebListener:CoreClr:x86]    at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
[Smoke:WebListener:CoreClr:x86]    at System.Threading.QueueUserWorkItemCallback.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem()
[Smoke:WebListener:CoreClr:x86]    at System.Threading.ThreadPoolWorkQueue.Dispatch()
[Smoke:WebListener:CoreClr:x86]    at System.Threading._ThreadPoolWaitCallback.PerformWaitCallback()
```
This happened running the MusicStore functional tests. This is the second time I've seen it happen today.
@victorhurdugaci

Tratcher:
Something has definitely regressed here, this caused Universe Tests to fail four times today in both dev and release. I also saw it once in a local build (ServerTests?).
/cc: @muratg

muratg:
@victorhurdugaci Please investigate ASAP.

PR #2885, opened 2015-10-01
Prevent a concurrent collection edit in breadcrumbs writing

victorhurdugaci:
Fixes https://github.com/aspnet/dnx/issues/2876

Please review @moozzyk @Tratcher

### DIFF
--- a/src/Microsoft.Dnx.Runtime/Servicing/Breadcrumbs.cs
+++ b/src/Microsoft.Dnx.Runtime/Servicing/Breadcrumbs.cs
@@ -105,6 +105,12 @@ namespace Microsoft.Dnx.Runtime.Servicing
             // The lock ensures that no add is happening while or after we set the flag
             lock (_addLock)
             {
+                if (_writeWasCalled)
+                {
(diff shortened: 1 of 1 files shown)


pranavkm:
:shipit:

davidfowl:
How does this happen?

moozzyk:
@davidfowl  `dnx --appbase . Microsoft.Dnx.ApplicationHost Microsoft.Dnx.ApplicationHost web` (transformed from `dnx -p Microsoft.Dnx.ApplicationHost web`) - both hosts were on the stack and since breadcrumb writing is not blocking and happens on a background thread both started writing to a file.

Tratcher:
https://github.com/aspnet/Hosting/commit/818575c20178103a5b61a30867e68f35995bacbb

victorhurdugaci:
@davidfowl that can happen if write is called multiple times. And it happened because of what @moozzyk and @Tratcher said

davidfowl:
Hah, I guess this is just being defensive then.

davidfowl:
:shipit:

PR #2889, opened 2015-10-02
Shortening runtime names

JunTaoLuo:
Reduce the chance of getting into long path issue, number 9 of #2825. This change will shorten the path names by 12-18 characters for non-mono runtimes.


davidfowl:
:shipit:

PR #2893, opened 2015-10-02
correct the fallback order for runtime.json

### LINKED ISSUE #2891: Incorrect runtime.json processing
When restoring a package that has a runtime.json, we end up picking the least specific runtime ID, not the most specific runtime ID. As far as we can tell, this hasn’t produced any known critical issues, but it means we’re picking the wrong dependencies. For example, System.Net.Primitives has the following runtime.json:

```json
{
  "runtimes": {
    "win7": {
      "System.Net.Primitives": {
        "runtime.win7.System.Net.Primitives": "4.0.11-beta-23401"
      }
    },
    "osx.10.10": {
      "System.Net.Primitives": {
        "runtime.osx.10.10.System.Net.Primitives": "4.0.11-beta-23401"
      }
    },
    "unix": {
      "System.Net.Primitives": {
        "runtime.unix.System.Net.Primitives": "4.0.11-beta-23401"
      }
    }
  }
}
```

On `osx.10.10-x64`, we end up picking `runtime.unix.System.Net.Primitives`.

/cc @Eilon @muratg @davidfowl @DamianEdwards @danroth27 @BrennanConroy - May be a candidate for beta8. I'll send an email shortly ;)

anurse:
fixes #2891 

/cc @Eilon @muratg @davidfowl @DamianEdwards @danroth27 @BrennanConroy

### DIFF
--- a/src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
+++ b/src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
@@ -412,8 +412,7 @@ namespace Microsoft.Dnx.Tooling
                                 runtimeName,
                                 runtimeFiles,
                                 runtimeDependencies,
-                                runtimeNames,
-                                _ => false);
+                                runtimeNames);
                             runtimeStopwatch.Stop();
                             Reports.WriteVerbose($" Scanned Runtime graph in {runtimeStopwatch.ElapsedMilliseconds:0.00}ms");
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
```diff
@@ -229,8 +229,8 @@ public RestoreCommand(FrameworkName fallbackFramework)
                     .Where(x => x.Severity == DiagnosticMessageSeverity.Error)
                     .Select(x => x.Message);
 
-                throw new InvalidOperationException(errorMessages.Any() ? 
-                    $"Errors occured when while parsing project.json:{Environment.NewLine}{string.Join(Environment.NewLine, errorMessages)}" : 
+                throw new InvalidOperationException(errorMessages.Any() ?
```
anurse: Just fallout from running Ctrl-K Ctrl-D :)

anurse:
Travis only failed on OSX and it was because of a fix I made in dev that needs to be ported here.

davidfowl:
Add a test. The fact that we have no coverage doesnt inspire confidence :smile:

anurse:
Adding much more than a mocked-up unit test is going to take a significant effort, so it depends on when we need to get this in.

anurse:
I'll build a unit test to start with. We can see what that looks like

anurse:
Added a unit test.

anurse:
ping @davidfowl

davidfowl:
:shipit:

anurse:
Merged

PR #2904, opened 2015-10-05
minor polish to friendly error messages

anurse:
I was helping @kichalla out with something and realized that the friendly error messages needed a small tweak to improve the user story a little. First, we need a way to dump the full stack trace when diagnosing issues, originally I thought a separate environment variable would be good, but I think DNX_TRACE can do the job. Second, we need to handle AggregateException separately and write out the individual exceptions since even single exceptions thrown in async code get wrapped in AEs.

For: RC1

* dump full exception stack trace when DNX_TRACE=1
* dump each exception inside an AggregateException

/cc @JunTaoLuo @troydai @halter73

### DIFF
--- a/src/Microsoft.Dnx.Tooling/Program.cs
+++ b/src/Microsoft.Dnx.Tooling/Program.cs
@@ -98,12 +98,28 @@ namespace Microsoft.Dnx.Tooling
                 throw;
             }
 #else
+            catch (AggregateException aex)
+            {
+                foreach (var exception in aex.InnerExceptions)
+                {
+                    DumpException(exception);
+                }
+                return 1;
+            }
             catch (Exception ex)
             {
-                AnsiConsole.GetError(useConsoleColor: _runtimeEnv.OperatingSystem == "Windows").WriteLine($"Error: {ex.Message}".Red().Bold());
+                DumpException(ex);
                 return 1;
             }
 #endif
+        }
+
+        private void DumpException(Exception ex)
+        {
+            AnsiConsole
+                .GetError(useConsoleColor: _runtimeEnv.OperatingSystem == "Windows")
+                .WriteLine($"Error: {ex.Message}".Red().Bold());
+            Logger.TraceError($"Full Exception: {ex.ToString()}");
         }
     }
 }


src/Microsoft.Dnx.Tooling/Program.cs
```diff
@@ -98,12 +98,26 @@ public int Main(string[] args)
                 throw;
             }
 #else
+            catch(AggregateException aex)
```
davidfowl: formatting!

troydai: formatting!

src/Microsoft.Dnx.Tooling/Program.cs
```diff
@@ -98,12 +98,26 @@ public int Main(string[] args)
...
+            {
+                foreach(var exception in aex.InnerExceptions)
```
troydai: formatting!

src/Microsoft.Dnx.Tooling/Program.cs
```diff
@@ -98,12 +98,26 @@ public int Main(string[] args)
...
-                AnsiConsole.GetError(useConsoleColor: _runtimeEnv.OperatingSystem == "Windows").WriteLine($"Error: {ex.Message}".Red().Bold());
+                DumpException(ex);
                 return 1;
             }
 #endif
         }
+
+        private void DumpException(Exception ex)
+        {
+            AnsiConsole.GetError(useConsoleColor: _runtimeEnv.OperatingSystem == "Windows").WriteLine($"Error: {ex.Message}".Red().Bold());
```
troydai: Any change to wrap this line?

troydai:
:shipit: after Travis passes and format fixes.

src/Microsoft.Dnx.Tooling/Program.cs
```diff
@@ -98,12 +98,28 @@ public int Main(string[] args)
...
                 return 1;
             }
 #endif
         }
+
+        private void DumpException(Exception ex)
+        {
+            AnsiConsole
+                .GetError(useConsoleColor: _runtimeEnv.OperatingSystem == "Windows")
+                .WriteLine($"Error: {ex.Message}".Red().Bold());
```
halter73: Why ex.Message instead of ex.ToString() like you use for TraceError?

anurse: Because we only want to show the short summary message here. `ToString` includes the stack trace

anurse: For example, when `DNX_TRACE` is off, we want to see:

```
Error: Something bad happened
```

But with `DNX_TRACE`:

```
Error: Something bad happened
[Error] Full Exception: System.SomethingBadHappenedException: Something bad happend
 at blah.blah.call.stack
 at blah.blah.call.stack
```

davidfowl: example?

halter73:
:ship:

davidfowl:
:shipit:

PR #2907, opened 2015-10-05
Add build targets to DNX to install dependencies to volatile feeds.

pranavkm:
Used to perform gated builds.

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -667,6 +667,15 @@ var CI_DARWIN_DROP_PATH = '${Environment.GetEnvironmentVariable("CI_DARWIN_DROP_
     }
     nuget-pack packageVersion='${FULL_VERSION}' outputDir='${BUILD_DIR2}' extra='-NoPackageAnalysis -Properties RuntimeNamePrefix=${RUNTIME_NAME_PREFIX}' nugetPath='.nuget/nuget.exe' each='var nuspecFile in Files.Include(Path.Combine(BUILD_DIR2, "*.nuspec"))'
 
+var CI_DEPENDENCIES_PATHS = '${Environment.GetEnvironmentVariable("CI_DEPENDENCIES_PATHS")}'
+
+#nuget-dependencies-install target='install' description='Install dependencies to publish directory' if='!string.IsNullOrEmpty(CI_DEPENDENCIES_PATHS)'
+  - var dependencyPaths = CI_DEPENDENCIES_PATHS.Split(new[] { (char)';' }, StringSplitOptions.RemoveEmptyEntries);
+  - var publishFeed = Environment.GetEnvironmentVariable("NUGET_PUBLISH_FEED");
+  
+  kpm-publish targetPackagesDir='${Environment.GetEnvironmentVariable("PACKAGES_PUBLISH_DIR")}' each='var sourcePackagesDir in dependencyPaths'
+  nuget-resilient-publish nugetFeed='${publishFeed}' each='var sourcePackagesDir in dependencyPaths' if='!string.IsNullOrEmpty(publishFeed)'
+
 - // ===================== TESTING ===================== 
 
 #test-package target='integration-test'


pranavkm:
cc @BrennanConroy

makefile.shade
```diff
@@ -667,6 +667,15 @@ var CI_DARWIN_DROP_PATH = '${Environment.GetEnvironmentVariable("CI_DARWIN_DROP_
     }
     nuget-pack packageVersion='${FULL_VERSION}' outputDir='${BUILD_DIR2}' extra='-NoPackageAnalysis -Properties RuntimeNamePrefix=${RUNTIME_NAME_PREFIX}' nugetPath='.nuget/nuget.exe' each='var nuspecFile in Files.Include(Path.Combine(BUILD_DIR2, "*.nuspec"))'
 
+var CI_DEPENDENCIES_PATHS = '${Environment.GetEnvironmentVariable("CI_DEPENDENCIES_PATHS")}'
+
+#nuget-dependencies-install target='install' description='Install dependencies to publish directory' if='!string.IsNullOrEmpty(CI_DEPENDENCIES_PATHS)'
+  - var dependencyPaths = CI_DEPENDENCIES_PATHS.Split(new[] { (char)';' }, StringSplitOptions.RemoveEmptyEntries);
```
pranavkm: Making it a list of values to maybe allow Roslyn gated builds.

BrennanConroy:
:+1: just need to add the variables to the CI

makefile.shade
moozzyk: wouldn't the path delimiter be `:` on non-windows (and not `;`)? 
Also why you need the cast?

makefile.shade
```diff
@@ -667,6 +667,15 @@ var CI_DARWIN_DROP_PATH = '${Environment.GetEnvironmentVariable("CI_DARWIN_DROP_
     }
     nuget-pack packageVersion='${FULL_VERSION}' outputDir='${BUILD_DIR2}' extra='-NoPackageAnalysis -Properties RuntimeNamePrefix=${RUNTIME_NAME_PREFIX}' nugetPath='.nuget/nuget.exe' each='var nuspecFile in Files.Include(Path.Combine(BUILD_DIR2, "*.nuspec"))'
 
+var CI_DEPENDENCIES_PATHS = '${Environment.GetEnvironmentVariable("CI_DEPENDENCIES_PATHS")}'
```
moozzyk: Can't this go into the target? If it can't, shouldn't it be with other global vars at the top of the file?

makefile.shade
```diff
@@ -667,6 +667,15 @@ var CI_DARWIN_DROP_PATH = '${Environment.GetEnvironmentVariable("CI_DARWIN_DROP_
...
+
+#nuget-dependencies-install target='install' description='Install dependencies to publish directory' if='!string.IsNullOrEmpty(CI_DEPENDENCIES_PATHS)'
+  - var dependencyPaths = CI_DEPENDENCIES_PATHS.Split(new[] { (char)';' }, StringSplitOptions.RemoveEmptyEntries);
+  - var publishFeed = Environment.GetEnvironmentVariable("NUGET_PUBLISH_FEED");
+  
+  kpm-publish targetPackagesDir='${Environment.GetEnvironmentVariable("PACKAGES_PUBLISH_DIR")}' each='var sourcePackagesDir in dependencyPaths'
+  nuget-resilient-publish nugetFeed='${publishFeed}' each='var sourcePackagesDir in dependencyPaths' if='!string.IsNullOrEmpty(publishFeed)'
```
moozzyk: Does this work on non-windows? Should it?

makefile.shade
pranavkm: It could. I was just following the pattern from https://github.com/aspnet/dnx/pull/2907#discussion-diff-41206792R642

makefile.shade
pranavkm: I don't see why it wouldn't. That said, we're running this from a Windows build agent on the CI so it's ok it it isn't cross-plat as yet. In fact, we can remove the `target="install"` since we specifically run this build target as opposed to relying on the `install` target.

makefile.shade
pranavkm: The cast is a Spark \ shade thing.

moozzyk:
LGTM (but check Travis)

makefile.shade
troydai: It's the __sake__ thing ...

troydai:
:shipit:

Issue #2448, opened 2015-08-10
Suggestions to improve publish performance to Kudu


DamianEdwards:
Publish time to Azure using Kudu isn't awesome right now. Here's some suggestions of things we can try to make it better:

- [x] Make 'dnu publish' copy already unzipped package contents rather than copy then unzip at destination. This should help speed up deployments as there will be significantly less IO on the durable (slow) drive.
- [x] `dnu build` (called by `dnu publish`) on project with commands configured results in `project.json` in bin folder, which will cause attempted restore on next `dnu restore` (as it dives into the bin folder).
(done in 0dda8bf86863364cc20421f1af7494f1b2a3256f)
- [ ] Kudu sync seems to copy more than just the expected changed items across on each deployment, perhaps time-stamps aren’t being set properly or something.
- [ ] `dnu restore` could potentially use the package cache (once installed) during restore and create junctions in place, so that restore on a cold site doesn’t have to hit the feed or copy any files for packages in the cache.

troydai:
Currently working __item 2__. I plan to stop dnu restore from diving into subfolder of a folder already containing a `project.json`

troydai:
Update: looks like we already did it however __partially__ (https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs#L107)
If the given `restoreDirectory` doesn't contain a `project.json` it will still dive into deeper level.

muratg:
Yeah, nuget/dotnet tasks to improve restore time are tracked elsewhere. Closing

PR #2910, opened 2015-10-05
Prevent dnu restore logic from diving in nested project folder

troydai:
Addressing the second item in issue: #2448 

/review: @anurse @JunTaoLuo @BrennanConroy @victorhurdugaci @moozzyk 
/cc: @davidfowl @DamianEdwards @muratg

### DIFF
--- a/src/Microsoft.Dnx.Tooling/Restore/RestoreProjectsCollector.cs
+++ b/src/Microsoft.Dnx.Tooling/Restore/RestoreProjectsCollector.cs
@@ -1,0 +1,54 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.Dnx.Tooling
+{
+    internal class RestoreProjectsCollector
+    {
+        public static bool Find(string restoreDirectory, out IEnumerable<string> projectJsonFiles)
+        {
+            if (string.Equals(Runtime.Project.ProjectFileName,
+                              Path.GetFileName(restoreDirectory),
+                              StringComparison.OrdinalIgnoreCase) && File.Exists(restoreDirectory))
+            {
+                // If the path is a project.json file we don't do recursive search in subfolders
+                projectJsonFiles = new List<string> { restoreDirectory };
+
+                return true;
+            }
+
+            if (!Directory.Exists(restoreDirectory))
+            {
+                projectJsonFiles = Enumerable.Empty<string>();
+                return false;
+            }
+
+            var result = new List<string>();
+            CollectProjectFiles(restoreDirectory, result);
+
+            projectJsonFiles = result;
+            return true;
+        }
+
+        private static void CollectProjectFiles(string directory, List<string> results)
+        {
+            var candidate = Path.Combine(directory, Runtime.Project.ProjectFileName);
+            if (File.Exists(candidate))
+            {
+                results.Add(candidate);
+            }
+            else
+            {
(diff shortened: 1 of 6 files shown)


troydai:
Ping

src/Microsoft.Dnx.Tooling/Restore/RestoreProjectsCollector.cs
```diff
@@ -0,0 +1,54 @@
...
+
+namespace Microsoft.Dnx.Tooling
+{
+    internal class RestoreProjectsCollector
+    {
+        public static bool Find(string restoreDirectory, out IEnumerable<string> projectJsonFiles)
+        {
+            if (string.Equals(Runtime.Project.ProjectFileName,
+                              Path.GetFileName(restoreDirectory),
+                              StringComparison.OrdinalIgnoreCase) && File.Exists(restoreDirectory))
```
victorhurdugaci: **File**.Exist(restore**Directory**) ?

src/Microsoft.Dnx.Tooling/Restore/RestoreProjectsCollector.cs
```diff
@@ -0,0 +1,54 @@
...
+            var result = new List<string>();
+            CollectProjectFiles(restoreDirectory, result);
+
+            projectJsonFiles = result;
+            return true;
+        }
+
+        private static void CollectProjectFiles(string directory, List<string> results)
+        {
+            var candidate = Path.Combine(directory, Runtime.Project.ProjectFileName);
```
victorhurdugaci: Add the runtime namespace

src/Microsoft.Dnx.Tooling/Restore/RestoreProjectsCollector.cs
```diff
@@ -0,0 +1,54 @@
...
+            }
+
+            var result = new List<string>();
+            CollectProjectFiles(restoreDirectory, result);
+
+            projectJsonFiles = result;
+            return true;
+        }
+
+        private static void CollectProjectFiles(string directory, List<string> results)
```
victorhurdugaci: Why not return the list instead of using an argument?

src/Microsoft.Dnx.Tooling/Restore/RestoreProjectsCollector.cs
victorhurdugaci: Case insensitive comparison?

troydai: Yes. This could be a file.

src/Microsoft.Dnx.Tooling/Restore/RestoreProjectsCollector.cs
troydai: There is a name conflict here with `Microsoft.Dnx.Project` therefore the `Runtime` prefix is required.

src/Microsoft.Dnx.Tooling/Restore/RestoreProjectsCollector.cs
troydai: Now including globbing, we compare file in case-insensitive way across platforms. It's not ideal but consistent.

muratg: If we change the "case insensitive everywhere" behavior in the future, we should remember to change that here as well.  How do we remember that?  Perhaps a test?

troydai:
@halter73 approved.

victorhurdugaci:
:shipit:

troydai:
yo!

iamjasonp:
@mconnew identified this particular issue I happened to be the one to shout loudly about it at the right time :smile: thanks!

PR #2914, opened 2015-10-06
Added magic environment variable for azure deployments

davidfowl:
- This command will instruct publish to use "%home%\site" as the base path instead of a relative
one for azure deployments in "web.config". It will not be required once azure gets the new
http platform handler module.

/cc @davidebbo @vijayrkn @sayedihashimi

### DIFF
--- a/src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
+++ b/src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
@@ -551,9 +551,12 @@ namespace Microsoft.Dnx.Tooling.Publish
             root.Reports.Information.WriteLine($"Using command '{command}' as the entry point for web.config.");
 
             var azurePublishValue = Environment.GetEnvironmentVariable("DNU_PUBLISH_AZURE");
-            var basePath = string.Equals(azurePublishValue, "true", StringComparison.Ordinal) ||
-                           string.Equals(azurePublishValue, "1", StringComparison.Ordinal)
-                           ? @"%home%\site" : "..";
+            var publishingToAzure = string.Equals(azurePublishValue, "true", StringComparison.Ordinal) ||
+                                    string.Equals(azurePublishValue, "1", StringComparison.Ordinal) ||
+                                    !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("WEBSITE_SITE_NAME"));
+
+            var basePath = publishingToAzure ? @"%home%\site" : "..";
+            var baseLogPath = publishingToAzure ? @"\\?\%home%\LogFiles" : @"..\logs";
 
             var targetDocument = XDocument.Parse($@"<configuration><system.webServer>
   <handlers>
@@ -562,7 +565,7 @@ namespace Microsoft.Dnx.Tooling.Publish
   <httpPlatform processPath=""{basePath}\approot\{command}.cmd""
                 arguments="""" 
                 stdoutLogEnabled=""true"" 
-                stdoutLogFile=""..\logs\stdout.log"">
+                stdoutLogFile=""{baseLogPath}\stdout.log"">
   </httpPlatform>
 </system.webServer></configuration>");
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
```diff
@@ -564,9 +569,7 @@ private bool GenerateWebConfigFileForWwwRootOut(PublishRoot root, Runtime.Projec
             var attributesToOverwrite = new[]
             {
                 "processPath",
-                "arguments",
-                "stdoutLogEnabled",
-                "stdoutLogFile"
```
muratg: Why remove these three?

muratg:
Assuming this is the correct full path, :shipit:

Would it be a good idea for @kichalla to test it on Azure Websites (as opposed to Ebbo's private deployment?)

moozzyk:
Travis failures look legit but I don't see how they could be caused by this change. Otherwise looks good to me.

src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
```diff
@@ -550,11 +550,16 @@ private bool GenerateWebConfigFileForWwwRootOut(PublishRoot root, Runtime.Projec
...
+
             var targetDocument = XDocument.Parse($@"<configuration><system.webServer>
   <handlers>
     <add name=""httpplatformhandler"" path=""*"" verb=""*"" modules=""httpPlatformHandler"" resourceType=""Unspecified"" />
   </handlers>
-  <httpPlatform processPath=""..\approot\{command}.cmd""
+  <httpPlatform processPath=""{basePath}\approot\{command}.cmd""
                 arguments="""" 
                 stdoutLogEnabled=""true"" 
                 stdoutLogFile=""..\logs\stdout.log"">
```
kichalla: change this too?

src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
borgdylan: @muratg Those attributes should never be overwritten. It would break the configurations for people who know how to set them themselves and need automated deployments. The choice of log file path and server arguments should be left up to the user and not dnu since not everyone is using kestrel.

davidfowl: We don't even need to overwrite arguments. I'll fix the tests.

borgdylan: Can the httpPlatform element be copied from web.config present in the source wwwroot folder?

src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
```diff
@@ -550,11 +550,16 @@ private bool GenerateWebConfigFileForWwwRootOut(PublishRoot root, Runtime.Projec
 
             root.Reports.Information.WriteLine($"Using command '{command}' as the entry point for web.config.");
 
+            var azurePublishValue = Environment.GetEnvironmentVariable("DNU_PUBLISH_AZURE");
```
davidebbo: Instead of relying on a new env variable which requires a Kudu change, I suggest using an existing one to detect if you're running in Azure Web App. e.g. just check if `WEBSITE_SITE_NAME` is defined.

src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
davidfowl: Why?

borgdylan: Because we can leave a web.config in that folder so that dnu copies stuff over. Currently it copies some IIS config stuff like error page settings.

davidfowl: Put it in wwwroot

muratg:
A few questions, but assuming you tested this, :shipit:

davidfowl:
@muratg http platform handler shenanigans

PR #2920, opened 2015-10-06
dnx and dnu returns zero if no arguments are passed

### LINKED ISSUE #2912: We should not return 2 if there are no args
Currently if the user calls `dnx` without arguments we will show help and return 2 as exit code. I don't think this is an error to call dnx without arguments (e.g. we don't print any error message in red) so the exit code should be just 0

mholo65:
Fixes #2912 

//cc @moozzyk

### DIFF
--- a/src/Microsoft.Dnx.Host/RuntimeBootstrapper.cs
+++ b/src/Microsoft.Dnx.Host/RuntimeBootstrapper.cs
@@ -113,7 +113,7 @@ namespace Microsoft.Dnx.Host
             if (!app.IsShowingInformation && app.RemainingArguments.Count == 0)
             {
                 app.ShowHelp();
-                return Task.FromResult(2);
+                return Task.FromResult(0);
             }
 
             // Some options should be forwarded to Microsoft.Dnx.ApplicationHost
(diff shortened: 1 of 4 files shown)


moozzyk:
I think commands in the Console.Commands need to be fixed as well. Did you run tests? I would expect changes in tests - e.g. here: https://github.com/aspnet/dnx/blob/4b4bf875b5d9b3d6a3f8e7608992d7d511dc83b3/test/Microsoft.Dnx.ApplicationHost.FunctionalTests/AppHostTests.Old.cs#L47

mholo65:
I ran ```build.sh``` and went for coffee :) thought that it would run tests as well. Now i realized in need to run ```build.sh verify``` to run tests.

Currently running and fixing. Will let You know when I've fixed all failing tests.

moozzyk:
@mholo65 yes you need to run `build verify`. Ideally you would run tests on WIndows as well.

mholo65:
Fixed tests.

I will look into ConsoleCommands and test on Windows tomorrow.

What about [Microsoft.Dnx.ApplicationHost](https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.ApplicationHost/Program.cs#L180)?

davidfowl:
We did this on purpose. Showing the help is an error code. I believe git and other tools do something similar.

moozzyk:
@davidfowl - I filed this bug after talking to you about this yesterday... Anyways - why is it useful to return non-zero exit code if no args were provided?

davidfowl:
@moozzyk 

1. Because other well known programs do it
2. We had a bug once where the arguments changed and help was being output on the CI but nothing was failing

PR #2925, opened 2015-10-07
Pawelka/json parser

moozzyk:
/cc @davidfowl - looks like this can work. 

(do not review it's just a prototype)

### DIFF
--- a/src/dnx.common/utils.cpp
+++ b/src/dnx.common/utils.cpp
@@ -57,6 +57,9 @@ namespace dnx
         bool strings_equal_ignore_case(const dnx::char_t* s1, const dnx::char_t* s2)
         {
 #if defined(_WIN32)
+            auto value = web::json::value::parse(_X("{ \"a\" : 123 }"));
+            std::wcout << value[_X("a")] << std::endl;
+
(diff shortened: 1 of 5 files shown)


src/dnx.common/utils.cpp
```diff
@@ -56,6 +57,9 @@ namespace dnx
         bool strings_equal_ignore_case(const dnx::char_t* s1, const dnx::char_t* s2)
         {
 #if defined(_WIN32)
+            auto value = web::json::value::parse(_X("{ \"a\" : 123 }"));
```
moozzyk: this

davidfowl: oooooo :smile:

moozzyk:
```
PS C:\Source\Universe> dnx --version
123
123
123
123
123
123
123
123
123
123
123
123
123
123
123
123
123
123
123
123
123
123
123
123
123
123
123
123
123
123
123
123
Microsoft .NET Execution environment
 Version:      1.0.0-t151007060838
 Type:         Clr
 Architecture: x86
 OS Name:      Windows
 OS Version:   10.0
 Runtime Id:   win10-x86
```

Issue #2887, opened 2015-10-01
Stack traces for Roslyn compilation errors aren't supressed


shirhatti:
````
C:\Users\vagrant\Documents\Visual Studio 2015\Projects\ConsoleApp1\src\ConsoleApp1> dnx ConsoleApp1 test
Microsoft.Dnx.Compilation.CSharp.RoslynCompilationException: C:\Users\vagrant\Documents\Visual Studio 2015\Projects\ConsoleApp1\src\ConsoleApp1\Program.cs(14,17): DNXCore,Version=v5.0 error CS0234: The type or namespace name 'Console' does not exist in the namespace 'System' (are you missing an assembly reference?)
   at Microsoft.Dnx.Compilation.CSharp.RoslynProjectReference.Load(AssemblyName assemblyName, IAssemblyLoadContext loadContext)
   at Microsoft.Dnx.Compilation.CompilationEngine.LoadProject(Project project, FrameworkName targetFramework, String aspect, IAssemblyLoadContext loadContext, AssemblyName assemblyName)
   at Microsoft.Dnx.Runtime.Loader.ProjectAssemblyLoader.Load(AssemblyName assemblyName, IAssemblyLoadContext loadContext)
   at Microsoft.Dnx.Runtime.Loader.ProjectAssemblyLoader.Load(AssemblyName assemblyName)
   at Microsoft.Dnx.Host.LoaderContainer.Load(AssemblyName assemblyName)
   at Microsoft.Dnx.Host.DefaultLoadContext.LoadAssembly(AssemblyName assemblyName)
   at Microsoft.Dnx.Runtime.Loader.AssemblyLoaderCache.GetOrAdd(AssemblyName name, Func`2 factory)
   at Microsoft.Dnx.Runtime.Loader.LoadContext.Load(AssemblyName assemblyName)
   at System.Runtime.Loader.AssemblyLoadContext.LoadFromAssemblyName(AssemblyName assemblyName)
   at Microsoft.Dnx.Runtime.Loader.LoadContext.Microsoft.Dnx.Runtime.IAssemblyLoadContext.Load(AssemblyName assemblyName)
   at Microsoft.Dnx.ApplicationHost.DefaultHost.GetEntryPoint(String applicationName)
   at Microsoft.Dnx.ApplicationHost.Program.ExecuteMain(DefaultHost host, String applicationName, String[] args)
   at Microsoft.Dnx.ApplicationHost.Program.Main(String[] args)
[...]

PR #2928, opened 2015-10-07
Change Roslyn complation exception search

pakrym:
Make it recursive so as to handle bot first level and wrapped compilation exception.

/cc @troydai @victorhurdugaci @davidfowl 

Address: #2887


misc/ApplicationHostTestProjects/src/SyntaxException/program.cs
```diff
@@ -0,0 +1,7 @@
+﻿class Program
+{
+	public static void Main()
+	{
+		System.Console1.WriteLine("Test");
```
davidfowl: No tabs please.

misc/ApplicationHostTestProjects/src/SyntaxException/project.json
```diff
@@ -0,0 +1,10 @@
+﻿{
+   "frameworks": {
+        "dnx451": { },
+        "dnxcore50": {
+            "dependencies": {
+                "System.Runtime": "4.0.21-beta-*"
+            }
```
davidfowl: 2 spaces not 4

src/Microsoft.Dnx.ApplicationHost/Program.cs
```diff
@@ -229,19 +229,29 @@ private Task<int> ExecuteMain(DefaultHost host, string applicationName, string[]
...
-            // from version to version. It's safest to catch both types of exceptions
-            catch (Exception ex) when (ex is FileLoadException || ex is FileNotFoundException)
+            catch (Exception ex)
             {
-                if (ex.InnerException is ICompilationException)
+                // Try to find compilation exception recursively to supress its stack
+                var innerException = ex;
+                do
                 {
-                    throw SuppressStackTrace(ex.InnerException);
```
davidfowl: When would you need to dive like this?

misc/ApplicationHostTestProjects/src/SyntaxException/program.cs
victorhurdugaci: Using a type name that doesn't exist is not a syntax error. Either rename the project or add an actual syntax error

src/Microsoft.Dnx.ApplicationHost/Program.cs
pakrym: The way dnx loads assembly when running on clr\coreclr differs so we get ICompilationException  either wrapped into FileLoadException or first level. So its either loop trough inner exceptions searching for ICompilationException or two if's like:

    if (ex is ICompilationException)
    {
        throw SuppressStackTrace(ex);
    }

    if (ex.InnerException is ICompilationException)
    {
        throw SuppressStackTrace(ex.InnerException);
    }

test/Microsoft.Dnx.ApplicationHost.FunctionalTests/AppHostTests.cs
```diff
@@ -90,5 +90,27 @@ public void RunP2PDifferentFrameworks(DnxSdk sdk)
 
             TestUtils.CleanUpTestDir<AppHostTests>(sdk);
         }
+        
+        [Theory, TraceTest]
+        [MemberData(nameof(DnxSdks))]
+        public void DxtSupressesRoslynExceptionStackTrace(DnxSdk sdk)
```
victorhurdugaci: Dxt?

test/Microsoft.Dnx.ApplicationHost.FunctionalTests/AppHostTests.cs
```diff
@@ -90,5 +90,27 @@ public void RunP2PDifferentFrameworks(DnxSdk sdk)
...
+            var project = solution.GetProject("CompilationException");
+
+            sdk.Dnu.Restore(solution.RootPath).EnsureSuccess();
+
+            // Act
+            var result = sdk.Dnx.Execute(project);
+
+            // Assert
+            Assert.Equal(1, result.ExitCode);
+            Assert.DoesNotContain("Exception", result.StandardOutput);
```
victorhurdugaci: How about StdErr? Usually error messages end up there

pakrym:
ping @troydai @victorhurdugaci @davidfowl

src/Microsoft.Dnx.ApplicationHost/Program.cs
```diff
@@ -229,19 +229,29 @@ private Task<int> ExecuteMain(DefaultHost host, string applicationName, string[]
...
+                    if (innerException is ICompilationException)
+                    {
+                        throw SuppressStackTrace(innerException);
+                    }
+                    innerException = innerException.InnerException;
```
troydai: Could the `innerException == null` here (therefore cause `NullRef`)? There is nothing guard this from happening. Or is there any implicit rules there that the captured the exception must have an inner exception?

troydai:
:shipit:

pakrym:
@JunTaoLuo  @davidfowl  What if user writes console application using Roslyn and he has roslyn exception inside of it. We would suppress it's stack trace if we put try/catch around ExecuteMain.

davidfowl:
ICompilationException is something that we expose. If you use roslyn directly they won't get exceptions for errors. If you throw an exception that implements ICompilationException, you get our default suppression behavior.

pakrym:
Reopened the issue https://github.com/aspnet/dnx/issues/2887

Issue #2771, opened 2015-09-21
Batch restore on a directory re-locks entire directory after first invalid lock file


ericstj:
When restoring for a single project the following line of code updates the lock state for the entire command.
https://github.com/aspnet/dnx/blob/a0af28b330f858c0ddc2063adfbd0ad18225c4ee/src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs#L256

This causes the remaining lock files all to be rewritten, and their \* dependencies re-evaluated to latest.

/cc @weshaggard

ericstj:
Just to be clear, I'd only expect regeneration of the 'invalid' lock files, not the entire repo.

davidfowl:
@ericstj Yep. That sounds like a bug.

weshaggard:
We keep hitting this issue in our corefx repo which is causing folks some grief. Any thoughts on how to work around this or ETA on a fix?

dagood:
I have a temporary fix where I store `Lock` in a local variable (`relock`) in `RestoreForProject` and use that local variable going on. I then passed `relock` into another method that previously used `Lock`. It worked for my scenario but I don't know if there's a broader impact.

This is part of a larger set of changes right now, but I could cherry-pick and submit a pull request.

davidfowl:
Just FYI, you shouldn't be checkin in lock files, in fact, we're going to be removing the locked property.

ericstj:
Are you dropping * versions too and moving the file to intermediate directory?

ericstj:
Also, I suspect this will make our perf problem worse.  If restore always writes the lock we are back to the perf issue that made us notice this bug in the first place.

davidfowl:
We'll keep supporting *. Moving to an intermediate directory came up but nobody really agrees where it should go and it would require existing tools to update. 

@ericstj What performance issue is that?

weshaggard:
For us it is critical to have the lock file for * versioning scenarios otherwise we cannot use it. We need our builds to be reliable and repeatable. 

As for the performance issue we try and restore all our packages up front and we cannot do that in parallel by launching multiple nuget/dnu restore commands because of all the races. That is why we started using this approach. So the only way for us to avoid that race problem would be to serialize all these restores and that is much slower.

weshaggard:
To add to this we want to switch to nuget.exe v3 at some point which doesn't currently support this restore all under a directory so we need to figure out how to support this scenario there as well.

ericstj:
[...]

PR #2929, opened 2015-10-07
Fix a single invalid lock file causing relock of all following lock files

dagood:
A quick fix for #2771.

I added a local variable `relock` to store whether or not a relock should happen. This is instead of using the `Lock` property. `Lock` seems to be intended to override validity and force relocking, so `relock` is initialized with `Lock`'s value.

I also passed `relock` to `WriteLockFile` to replace a usage of `Lock`.

### DIFF
--- a/src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
+++ b/src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
@@ -220,8 +220,10 @@ namespace Microsoft.Dnx.Tooling
 
             var lockFile = await ReadLockFile(projectLockFilePath);
 
+            bool relock = Lock;
+
             var useLockFile = false;
-            if (Lock == false &&
+            if (relock == false &&
                 Unlock == false &&
                 lockFile != null &&
                 lockFile.Islocked)
@@ -235,7 +237,7 @@ namespace Microsoft.Dnx.Tooling
                 Reports.Information.WriteLine("Updating the invalid lock file with {0}",
                     "dnu restore --lock".Yellow().Bold());
                 useLockFile = false;
-                Lock = true;
+                relock = true;
             }
 
             Func<string, string> getVariable = key =>
@@ -515,7 +517,8 @@ namespace Microsoft.Dnx.Tooling
                               graphItems,
                               repository,
                               projectResolver,
-                              targetContexts);
+                              targetContexts,
+                              relock);
             }
 
             if (!SkipRestoreEvents)
@@ -821,13 +824,14 @@ namespace Microsoft.Dnx.Tooling
                                    List<GraphItem> graphItems,
                                    PackageRepository repository,
                                    IProjectResolver projectResolver,
-                                   IEnumerable<TargetContext> contexts)
+                                   IEnumerable<TargetContext> contexts,
+                                   bool relock)
         {
             var resolver = new DefaultPackagePathResolver(repository.RepositoryRoot.Root);
             var previousPackageLibraries = previousLockFile?.PackageLibraries.ToDictionary(l => Tuple.Create(l.Name, l.Version));
 
             var lockFile = new LockFile();
-            lockFile.Islocked = Lock;
+            lockFile.Islocked = relock;
 
             // Use empty string as the key of dependencies shared by all frameworks
             lockFile.ProjectFileDependencyGroups.Add(new ProjectFileDependencyGroup(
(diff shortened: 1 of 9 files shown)


davidfowl:
@dagood Can you write a test for this?

dagood:
@davidfowl I can try, but this would be the first time I looked at the DNX tests. It seems like the best place for it would be [here](https://github.com/aspnet/dnx/blob/dev/test/Microsoft.Dnx.Tooling.FunctionalTests/DnuRestoreTests.cs) but I don't follow where the test data comes from yet.

davidfowl:
@dagood test data comes from here https://github.com/aspnet/dnx/tree/dev/misc

dagood:
@davidfowl I'm getting ramped up on this, but I think I'm having trouble running the tests on my machine. How is this typically done? I'm trying to copy what Travis is doing (changing to the test directory and running `dnx test -parallel none` with the built `dnx`, so in my case `dnvm run clr-win-x86-dev test -parallel none`) but it takes ~30 minutes to run and of 360 tests, 4 fail and 5 are skipped.

Maybe related to this: is the project supposed to build correctly in VS2015? I haven't been able to get that to work and I've been using build.cmd instead.

dagood:
@davidfowl I added a test and with the current version of code I was able to get it to run on OSX in [build 4963](https://travis-ci.org/aspnet/dnx/builds/86505028). I resubmitted since the Linux error looked transient, but I got some new errors that don't seem to be related to the test. I pulled dev and resubmitted with no apparent change. (I still can't run locally, so I've been relying on CI on this pull request to run this. This also means that it would be very difficult for me to make sure the test fails on the old version of DNX.) Any ideas?

weshaggard:
cc @ericstj

src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
```diff
@@ -234,7 +236,7 @@ public RestoreCommand(FrameworkName fallbackFramework)
                 Reports.Information.WriteLine("Updating the invalid lock file with {0}",
                     "dnu restore --lock".Yellow().Bold());
                 useLockFile = false;
-                Lock = true;
+                relock = true;
```
weshaggard: I didn't look deeply but I suspect this is the key line that fixes the problem? It is expected that this Lock property should only ever be set by the command line and not by other invalid lock files.

dagood: Yes, the rest of the changes are following from this, using the local variable instead of the property and adding tests.

dagood:
@davidfowl I got the test running on my dev machine with some fiddling and it worked. I ran two test validation runs with the current version of the code: the [first run](https://travis-ci.org/aspnet/dnx/builds/86505028) succeeded for OSX and failed for Linux, and the [second run](https://travis-ci.org/aspnet/dnx/builds/87958706) succeeded for Linux and failed for OSX. Between the two runs I merged with `dev` again, but I'm not sure what to make of those errors. Any ideas?

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuRestoreTests.cs
```diff
@@ -98,5 +98,22 @@ public void Restore_NormalizesVersionCasing(DnxSdk sdk)
...
+        {
+            var solution = TestUtils.GetSolution<DnuRestoreTests>(
+                sdk,
+                "ThreeLockedProjects",
+                appendSolutionNameToTestFolder: true);
+
+            var result = sdk.Dnu.Restore(solution.SourcePath).EnsureSuccess();
+
+            var outputLines = result.StandardOutput.Split('\n');
+            Assert.Contains(outputLines, line => line.Contains("Following lock file") && line.Contains("Project1"));
```
davidfowl: Instead, can we parse the lock file and check locked:true?

dagood: I don't understand what that check would validate. As far as I can tell, if an invalid lock file is locked, it's recreated with `--lock`, but if it's unlocked, it's recreated without locking. I did notice that my Project2 lock file wasn't locked after trying that out, though, so I fixed that.

muratg:
As discussed in the [Roadmap](https://github.com/aspnet/Home/wiki/Roadmap) we will be moving to [Dotnet CLI](https://github.com/dotnet/cli) in ASP.NET 5 RC 2 and retire DNX, DNU and DNVM command line tools.  Closing this PR.

Thank you for your contribution!

PR #2932, opened 2015-10-07
Temporary pull request to test Linux Travis timeout issue

pakrym:
Caused by absence of test output.

Try enabling test logging unconditionaly to see if it's environment variable problem or output redirection


troydai:
:shipit: when Travis passes. Thanks for fixing this!

PR #2941, opened 2015-10-09
React to aspnet/Universe#290 fix (2 of 2)

dougbu:
- react to KoreBuild changes in `makefile.shade`
- use `DnxTest()` macro rather than local `KTest()` function
  - `KTest()` broke due to `K()` -> `Dnx()` rename

nits:
- restore `--quiet`'s operation on `dnu restore`
  - lost in commit 015e8a5

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -166,25 +166,25 @@ var CAN_BUILD_ONECORE = '${Directory.Exists(VS_ONECORE_ROOT) && Directory.GetFil
 #package-windows target='package-runtime' if='CanBuildForWindows'
 #package-windows if='!CanBuildForWindows'
 
-- // ===================== BUILD ===================== 
+- // ===================== BUILD =====================
 
-- // ===================== COMMON ===================== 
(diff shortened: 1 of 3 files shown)


makefile.shade
```diff
@@ -769,7 +769,7 @@ var CI_DARWIN_DROP_PATH = '${Environment.GetEnvironmentVariable("CI_DARWIN_DROP_
                         {
                             if(framework == target.Framework)
                             {
-                                KTest(projectFile);
+                                DnxTest(projectFile, false);
```
victorhurdugaci: What's false? Can you use named arguments if sake supports them?

victorhurdugaci:
looks good. :shipit:

makefile.shade
dougbu: This is `false` because it's useless for `DnuTest` to do a single project in parallel. (The parameter is `testParallel`.)

I'll check again on named arguments. Think that failed last I tried though I see a few named parameters elsewhere.

PR #2946, opened 2015-10-09
fix packaging of System.Globalization.Extensions

anurse:
Need approval for beta 8. I'm kinda assuming it's approved, but I'll come bother one of you folks to be sure :).

Added a small amount of tracing that should show up in TeamCity artifacts output to help us diagnose these in the future. Basically when copying files we write a `files.lst` showing what got copied and from where. This file is excluded from packaging so it should not show up there, but I'll double-check that.

/cc @ericstj @muratg @moozzyk @Eilon @danroth27

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -1158,6 +1158,7 @@ functions @{
         
         try
         {
+            var filesList = new List<string>();
             foreach (var file in dependencies)
             {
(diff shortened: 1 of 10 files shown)


anurse:
Verbal approval from @Eilon @muratg . As I expected, the answer to "Without this fix, beta 8 doesn't work. Should we take it?" was, "Um... Yes". :trollface:

moozzyk:
:ship::it:

Issue #2895, opened 2015-10-02
OS version should be the one reported by the OS


victorhurdugaci:
We changed the way `IRuntimeEnvironment` handles OS version. It used to return the kernel version as returned by `GetVersionEx`/`uname`.

Now we do some transformations and return something else like "7.0" for win 7 instead of "6.1.x.x".

The transformation should be an util/extension method because only DNX cares about it and IRuntimeEnvironment should return the string as given by the OS.

cc @moozzyk @davidfowl

moozzyk:
I assume that we would also return unprocessed values for other OSes

anurse:
Agreed, but we should provide the distro id/release in a field as well so we can generate the correct RID.

moozzyk:
:+1: - I never really liked cramming distro and version to one property

muratg:
If not risky, we can give it a shot RC1.  Otherwise, I'll put it in 1.0.0 backlog.

moozzyk:
If we put it in backlog then we will never be able to fix this as it will be a breaking change.

muratg:
We'll be doing other changes in this area, so putting this in RC1 for now.

pakrym:
So what we have right now in the IRuntimeEnvironment:

    OperatingSystem = Windows/Linux/Darvin
    OperatingSystemVersion = 
       On Win: Windows release number 
           7.0 for Windows 7, 8.1 for Windows 8.1 and Windows 2012 Server etc.
       On Linux: Release number (DISTRIB_NUMBER from /etc/lsb_release)  
           14.04 for Ubuntu 14.04, etc.
           Note: right now we always report Ubuntu.14.04 runtime identifier for dnu purposes
       On OSX: We are translating release number returned by uname to OSX version
           15.x.x -> 10.11.x El Capitan
           14.x.x -> 10.10.x Yosemite
           13.x.x -> 10.9.x  Mavericks
           etc.

I propose adding field to contain processed version number (OperatingSystemReleaseVersion?)  from above and keep actual kernel version (GetVersionEx on Windows, uname -r on Linux and OSX) in OperatingSystemVersion.

Also maybe we want to store Linux distributive name separately?

victorhurdugaci:
I think we should have the same behavior as before (before Pawel's change). OsVersion is a string reporting `GetVersionEx` on Windows and `uname -a` on *nix.

[...]

PR #2949, opened 2015-10-09
OS version should be the one reported by the OS

pakrym:
Issue #2895

@moozzyk @victorhurdugaci

### DIFF
--- a/src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
+++ b/src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
@@ -214,6 +214,7 @@ public class EntryPoint
 
         operatingSystem = unameOutput[0];
         architecture = TranslateArchitecture(unameOutput[1]);
+
         if (operatingSystem == "Darwin")
         {
             // sw_vers returns versions in format "10.10.4" so ".4" needs to be removed
@@ -223,36 +224,34 @@ public class EntryPoint
             return;
         }
 
-        try
-        {
-            var lsb_release = RunProgram("lsb_release", "-s -i -r");
-            if (!string.IsNullOrEmpty(lsb_release))
-            {
-                osVersion = lsb_release.Replace(Environment.NewLine, " ");
-                return;
-            }
-        }
-        catch
-        {
-        }
-
-        try
-        {
-            // Red Hat based Linux without lsb_release
-            var redHatRelease = File.ReadAllText("/etc/redhat-release");
-            var m = Regex.Match(redHatRelease, @"^(?<distro>\S+) \S+ \S+ (?<version>\d+\.\d+)\..+$");
-            if (m.Success)
-            {
-                osVersion = string.Format("{0} {1}", m.Groups["distro"], m.Groups["version"]);
-                return;
-            }
-        }
-        catch
-        {
-        }
-
-        Console.WriteLine("Could not determine OS version information. Defaulting to the empty string.");
         osVersion = string.Empty;
+        var qualifiers = new[] { "ID=", "VERSION_ID=" };
+        try
+        {
+            var osRelease = File.ReadAllLines("/etc/os-release");
+            foreach (var qualifier in qualifiers)
+            {
+                foreach (var line in osRelease)
+                {
+                    if (line.StartsWith(qualifier))
+                    {
+                        if (osVersion.Length != 0)
+                        {
+                            osVersion += " ";
+                        }
+                        osVersion += line.Substring(qualifier.Length).Trim('"', '\'');
+                    }
+                }
+            }
+        }
+        catch
+        {
+        }
+
+        if (osVersion.Length == 0)
+        { 
+             Console.WriteLine("Could not determine OS version information. Defaulting to the empty string.");
+        }
     }
 
     private static string TranslateArchitecture(string architecture)
--- a/src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
+++ b/src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
@@ -46,16 +46,16 @@ namespace Microsoft.Dnx.Runtime
             else
             {
                 var arch = env.RuntimeArchitecture.ToLowerInvariant();
-                if (env.OperatingSystemVersion.Equals("7.0", StringComparison.Ordinal))
+                if (env.OperatingSystemVersion.Equals("6.1", StringComparison.Ordinal))
                 {
                     yield return "win7-" + arch;
                 }
-                else if (env.OperatingSystemVersion.Equals("8.0", StringComparison.Ordinal))
+                else if (env.OperatingSystemVersion.Equals("6.2", StringComparison.Ordinal))
                 {
                     yield return "win8-" + arch;
                     yield return "win7-" + arch;
                 }
-                else if (env.OperatingSystemVersion.Equals("8.1", StringComparison.Ordinal))
+                else if (env.OperatingSystemVersion.Equals("6.3", StringComparison.Ordinal))
                 {
                     yield return "win81-" + arch;
                     yield return "win8-" + arch;
@@ -78,18 +78,16 @@ namespace Microsoft.Dnx.Runtime
             if (string.Equals(os, RuntimeOperatingSystems.Windows, StringComparison.Ordinal))
             {
                 os = "win";
-
-                // Convert 6.x to the correct branding version for the RID
-                var parsedVersion = Version.Parse(ver);
-                if (env.OperatingSystemVersion.Equals("7.0", StringComparison.Ordinal))
+                
+                if (env.OperatingSystemVersion.Equals("6.1", StringComparison.Ordinal))
                 {
                     ver = "7";
                 }
-                else if (env.OperatingSystemVersion.Equals("8.0", StringComparison.Ordinal))
+                else if (env.OperatingSystemVersion.Equals("6.2", StringComparison.Ordinal))
                 {
                     ver = "8";
                 }
-                else if (env.OperatingSystemVersion.Equals("8.1", StringComparison.Ordinal))
+                else if (env.OperatingSystemVersion.Equals("6.3", StringComparison.Ordinal))
                 {
                     ver = "81";
                 }
--- a/src/dnx.common/include/utils.h
+++ b/src/dnx.common/include/utils.h
@@ -23,7 +23,7 @@ namespace dnx
         dnx::xstring_t remove_file_from_path(const dnx::xstring_t& path);
 
 #if defined(_WIN32)
-        const wchar_t* get_windows_version();
+        std::wstring get_windows_version();
 #endif
         int find_bootstrapper_option_index(int argc, dnx::char_t**argv, const dnx::char_t* optionName);
         int find_first_non_bootstrapper_param_index(int argc, dnx::char_t**argv);
--- a/src/dnx.common/utils.cpp
+++ b/src/dnx.common/utils.cpp
@@ -128,7 +128,7 @@ namespace dnx
         }
 
 #if defined (_WIN32)
-        const wchar_t* get_windows_version()
+        std::wstring get_windows_version()
         {
             OSVERSIONINFO version_info;
             ZeroMemory(&version_info, sizeof(OSVERSIONINFO));
@@ -137,31 +137,9 @@ namespace dnx
 #pragma warning(disable:4996)
             GetVersionEx(&version_info);
 #pragma warning(default:4996)
-
-            if (version_info.dwMajorVersion == 10)
-            {
-                return L"10.0";
-            }
-
-            if (version_info.dwMajorVersion == 6)
-            {
-                if (version_info.dwMinorVersion == 3)
-                {
-                    return L"8.1";
-                }
-
-                if (version_info.dwMinorVersion == 2)
-                {
-                    return L"8.0";
-                }
-
-                if (version_info.dwMinorVersion == 1)
-                {
-                    return L"7.0";
-                }
-            }
-
-            return nullptr;
+            return std::to_wstring(version_info.dwMajorVersion)
+               .append(L".")
+               .append(std::to_wstring(version_info.dwMinorVersion));
         }
 
 #endif
--- a/src/dnx.coreclr.unix/dnx.coreclr.cpp
+++ b/src/dnx.coreclr.unix/dnx.coreclr.cpp
@@ -244,10 +244,10 @@ int InvokeDelegate(host_main_fn host_main, int argc, const char** argv, const bo
 
 std::string get_os_version()
 {
-    std::vector<std::string> qualifiers { "DISTRIB_ID=", "DISTRIB_RELEASE=" };
+    std::vector<std::string> qualifiers { "ID=", "VERSION_ID=" };
 
     std::ifstream lsb_release;
-    lsb_release.open("/etc/lsb-release", std::ifstream::in);
+    lsb_release.open("/etc/os-release", std::ifstream::in);
     if (lsb_release.is_open())
     {
         std::string os_version;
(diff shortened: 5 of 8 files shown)


src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
```diff
@@ -223,36 +224,34 @@ private static void GetOsDetails(out string operatingSystem, out string osVersio
             return;
         }
 
+        osVersion = string.Empty;
+        var qulifiers = new[] { "ID=", "VERSION_ID=" };
```
moozzyk: typo - should be `qualifiers`

src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
```diff
@@ -3,12 +3,23 @@
...
 using System;
 using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
 using System.Text;
 
 namespace Microsoft.Dnx.Runtime
 {
-    internal static class RuntimeEnvironmentExtensions
+    internal static class RuntimeEnvironmentExtensions
```
moozzyk: nit - remove trailing space

src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
```diff
@@ -3,12 +3,23 @@
...
+using System.Linq;
 using System.Text;
 
 namespace Microsoft.Dnx.Runtime
 {
-    internal static class RuntimeEnvironmentExtensions
+    internal static class RuntimeEnvironmentExtensions 
     {
+
+        private static Dictionary<Version, Version> _windowsVersionMapping = new Dictionary<Version, Version>()
```
moozzyk: nit - do you need `()`?

src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
```diff
@@ -46,27 +57,13 @@ public static IEnumerable<string> GetAllRuntimeIdentifiers(this IRuntimeEnvironm
...
-                    yield return "win81-" + arch;
-                    yield return "win8-" + arch;
-                    yield return "win7-" + arch;
-                }
-                else if (env.OperatingSystemVersion.Equals("10.0", StringComparison.Ordinal))
+                Version parsedVersion;
+                if (!Version.TryParse(env.OperatingSystemVersion, out parsedVersion))
+                    yield break;
+
+                foreach (var version in _windowsVersionMapping.Where(p=>p.Key <= parsedVersion).Select(p=>p.Value))
```
moozzyk: I think the loop could be just: 
`return _windowsVersionMapping.Where(p => p.Key <= parsedVersion).Select(p => $"win{ToShortVersion(version)}-{arch}")`;

However AFAIR there was a concern of using LINQ in the lower layers of the host due to performance hit.. - @anurse

src/dnx.common/utils.cpp
```diff
@@ -128,7 +128,7 @@ namespace dnx
         }
 
 #if defined (_WIN32)
-        const wchar_t* get_windows_version()
+        xstring_t get_windows_version()
```
moozzyk: make it `std::wstring` - the convention I followed was to use native types in code that is platform specific and x-types in code that is platform agnostic.

src/dnx.common/include/utils.h
```diff
@@ -23,7 +23,7 @@ namespace dnx
         dnx::xstring_t remove_file_from_path(const dnx::xstring_t& path);
 
 #if defined(_WIN32)
-        const wchar_t* get_windows_version();
+        dnx::xstring_t get_windows_version();
```
moozzyk: same as below (i.e. `std::wstring`)

src/dnx.common/utils.cpp
```diff
@@ -137,31 +137,10 @@ namespace dnx
...
-                }
-
-                if (version_info.dwMinorVersion == 1)
-                {
-                    return L"7.0";
-                }
-            }
-
-            return nullptr;
+            std::wstring ver = std::to_wstring(version_info.dwMajorVersion)
```
moozzyk: `auto` (or even don't assign to a variable?)

src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
```diff
@@ -223,36 +224,34 @@ private static void GetOsDetails(out string operatingSystem, out string osVersio
...
             }
         }
         catch
         {
         }
 
-        try
-        {
-            // Red Hat based Linux without lsb_release
-            var redHatRelease = File.ReadAllText("/etc/redhat-release");
```
victorhurdugaci: Did we drop this completely? Is there an `/etc/os-release` on redhat?

src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
```diff
@@ -140,5 +124,10 @@ public static string GetRuntimeOsName(this IRuntimeEnvironment env)
             }
             return os;
         }
+
+        private static string ToShortVersion(Version version)
+        {
+            return version.Major + (version.Minor != 0? version.Minor.ToString(CultureInfo.InvariantCulture) : string.Empty);
```
moozzyk: nit: formatting around `?`

moozzyk: nit: formatting around `?`

src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
victorhurdugaci: The value could be a string. You don't need an object

victorhurdugaci: Also, I would avoid this dictionary completely and do the if else that was before because this code is on the startup hot path

src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
```diff
@@ -46,27 +57,13 @@ public static IEnumerable<string> GetAllRuntimeIdentifiers(this IRuntimeEnvironm
...
-                else if (env.OperatingSystemVersion.Equals("8.1", StringComparison.Ordinal))
-                {
-                    yield return "win81-" + arch;
-                    yield return "win8-" + arch;
-                    yield return "win7-" + arch;
-                }
-                else if (env.OperatingSystemVersion.Equals("10.0", StringComparison.Ordinal))
+                Version parsedVersion;
+                if (!Version.TryParse(env.OperatingSystemVersion, out parsedVersion))
+                    yield break;
```
moozzyk: use `{`, `}` even if this is just one line

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -257,11 +257,23 @@ std::string get_os_version()
             {
                 if (line.compare(0, qualifier.length(), qualifier) == 0)
                 {
+                    auto value = line.substr(qualifier.length());
+                    if (value.length() == 0)
+                    {
+                        continue;
+                    }
+                    // trim quotes
+                    if (value[0] == '"' && value[value.length() - 1] == '"')
```
victorhurdugaci: What is `value` has length 1?

test/Microsoft.Dnx.Host.Tests/RuntimeEnvironmentTests.cs
```diff
@@ -35,12 +35,12 @@ public void RuntimeEnvironment_InitializedCorrectly()
 
         // Test RID generation
         [Theory]
-        [InlineData("Windows", "7.0", "x86", "win7-x86")] // 1234 => We only care about major and minor
-        [InlineData("Windows", "7.0", "x64", "win7-x64")]
-        [InlineData("Windows", "8.0", "x86", "win8-x86")]
-        [InlineData("Windows", "8.0", "x64", "win8-x64")]
-        [InlineData("Windows", "8.1", "x86", "win81-x86")]
-        [InlineData("Windows", "8.1", "x64", "win81-x64")]
+        [InlineData("Windows", "6.1", "x86", "win7-x86")] // 1234 => We only care about major and minor
```
victorhurdugaci: Is that 1234 => comment still applicable?

src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
moozzyk: Yes.

pakrym: Do we support Redhat? There is on on CentOS and it's consistent with how we find version in dnx.coreclr.unix

moozzyk: CentOs is based on Redhat - this is why we checked redhat-release before. It seems that os-release is both on CentOs and debian so we can just use os-release and do not have to special case.

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -257,11 +257,23 @@ std::string get_os_version()
             {
                 if (line.compare(0, qualifier.length(), qualifier) == 0)
                 {
+                    auto value = line.substr(qualifier.length());
+                    if (value.length() == 0)
```
moozzyk: probably as per @victorhurdugaci `s comment this should be `< 2` and trim quotes always.

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -257,11 +257,23 @@ std::string get_os_version()
...
+                    {
+                        continue;
+                    }
+                    // trim quotes
```
moozzyk: remove this comment it does not add value...

src/dnx.coreclr.unix/dnx.coreclr.cpp
```diff
@@ -244,10 +244,10 @@ int InvokeDelegate(host_main_fn host_main, int argc, const char** argv, const bo
 
 std::string get_os_version()
 {
-    std::vector<std::string> qualifiers { "DISTRIB_ID=", "DISTRIB_RELEASE=" };
+    std::vector<std::string> qualifiers { "ID=", "VERSION_ID=" };
```
moozzyk: because we always expect `"` you could change the qualifiers to contain the leading `"` and then just copy length-1 chars.

moozzyk:
I think you will need to revert/fix this: https://github.com/aspnet/Testing/commit/1e9cb8ad3a83aa4cd66b579942b420315795f2f7 after you commit this change

src/dnx.coreclr.unix/dnx.coreclr.cpp
victorhurdugaci: Nope. They don't always have quotes. Mine doesn't:

```
Name="Ubuntu"
ID=ubuntu
...
VERSION_ID="14.04"
```

moozzyk: You are right: _Variable assignment values must be enclosed in double or single quotes if they include spaces, semicolons or other special characters outside of A-Z, a-z, 0-9_

moozzyk: Should we also check single quotes then?

test/Microsoft.Dnx.Host.Tests/RuntimeEnvironmentTests.cs
pakrym: Not sure about this, we still care only about major, minor so i kept it

victorhurdugaci: Ok but update the message. what is "1234 =>" ?

moozzyk: Long time ago this test looked like this: 
https://github.com/aspnet/dnx/commit/91bfcb2beae37e9848988d957757f02188d38a33#diff-e55dc1e04a00ac49a7639d3a48ea217dL36

I think we should remove the comment - it is no longer understandable

pakrym:
@moozzyk  @victorhurdugaci I drop dictionary completely and revert to ifs?

victorhurdugaci:
Yes

moozzyk:
:+1:

pakrym:
Added checking for single quotes, reverted to ifs. Anything else?
/cc @moozzyk @victorhurdugaci

src/dnx.coreclr.unix/dnx.coreclr.cpp
victorhurdugaci: You're still not handling `value.length() == 1`. The check should be `<2` instead of `==0`

src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
```diff
@@ -3,6 +3,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
```
victorhurdugaci: Do you still need linq?

src/Microsoft.Dnx.Host.Mono/EntryPoint.cs
```diff
@@ -223,36 +224,34 @@ private static void GetOsDetails(out string operatingSystem, out string osVersio
...
-                return;
+                foreach (var line in osRelease)
+                {
+                    if (line.StartsWith(qualifier))
+                    {
+                        if (osVersion.Length != 0)
+                        {
+                            osVersion += " ";
+                        }
+                        osVersion += line.Substring(qualifier.Length).Trim('"').Trim('\'');
```
victorhurdugaci: nit: Trim accepts a char array as argument

moozzyk: :+1:

src/dnx.coreclr.unix/dnx.coreclr.cpp
pakrym: I do checking for quote removal in next if. <2 wont work for VERSION_ID=5

moozzyk: This is probably theoretical but now the check for empty string should go after trimming quotes...

victorhurdugaci:
Looks good. :shipit: from me but please wait for @moozzyk to sign off too.

Also, don't forget to rebase & squash

moozzyk:
:ship: :it:

src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
```diff
@@ -78,18 +78,16 @@ public static string GetRuntimeOsName(this IRuntimeEnvironment env)
...
-                else if (env.OperatingSystemVersion.Equals("8.0", StringComparison.Ordinal))
+                else if (env.OperatingSystemVersion.Equals("6.2", StringComparison.Ordinal))
                 {
                     ver = "8";
                 }
-                else if (env.OperatingSystemVersion.Equals("8.1", StringComparison.Ordinal))
+                else if (env.OperatingSystemVersion.Equals("6.3", StringComparison.Ordinal))
                 {
                     ver = "81";
                 }
```
davidfowl: What about windows 10?

pakrym: Didn't change, it reports 10

moozzyk: I actually did check this :)

src/dnx.coreclr.unix/dnx.coreclr.cpp
anurse: Better safe than sorry

anurse:
:shipit:

Issue #2850, opened 2015-09-28
Publish over aggressively deletes package contents


davidfowl:
See #2741. Today the publish logic tries to keep only the things it requires by using the lock file input to post process the published files. Instead, we need to only delete the set of things we know of in the lock file that aren't relevant to the current publish. For example, if we're targeting dnx451 instead of dnxcore50, then we can delete all of the known dnxcore50 things.

JunTaoLuo:
merged to dev

Issue #2628, opened 2015-09-03
Rebuild dnu publish off the lock file


anurse:
Right now we do a lot of rewalking in dnu publish. @lodejard, @davidfowl and I discussed this and decided we should change the logic a bit, but want to do so after the changes I and @victorhurdugaci are doing. This bug is to track those changes.

* [ ] Add a static API to AppHostContext that takes a project (and other related data) and creates one AHC for every TFM/RID combo in the lock file
* [ ] Change Publish to use that API to get a set of all the targets to publish for
* [ ] Change `--framework` and `--runtime` to filter this list of AHCs. This means that if you want to publish from windows to linux, you MUST have restored for linux (or the lock file won't have it).
* [ ] Revisit embedding multiple runtimes. Should that work? How?

anurse:
This should probably be done in beta8 and I think I can take it on (after some rebalancing of other issues), but I didn't want it to block #2387

anurse:
This should be scheduled for rc1

/cc @muratg @davidfowl @moozzyk

anurse:
I think @JunTaoLuo is taking this actually. @davidfowl assigned a different version of the same bug to him :).

JunTaoLuo:
@anurse yup.

JunTaoLuo:
This feature will no longer be implemented

PR #2953, opened 2015-10-10
modifying pruning logic to less aggressively purge files after publish

JunTaoLuo:
#2850

### DIFF
--- a/src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
+++ b/src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Runtime.Versioning;
@@ -322,10 +321,7 @@ namespace Microsoft.Dnx.Tooling.Publish
             }
 
             // Prune the packages folder only leaving things that are required
-            if (!PrunePackages(root))
-            {
-                return false;
-            }
+            PrunePackages(root);
 
             // If --wwwroot-out doesn't have a non-empty value, we don't need a public app folder in output
             if (string.IsNullOrEmpty(WwwRootOut))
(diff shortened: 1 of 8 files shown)


JunTaoLuo:
I'm not sure if checking in `*.nupkg*` into `misc` for use in tests is a good idea and if not, I can just create test package on the fly. I didn't since it took extra steps in terms of pack, package add, which are not really relevant to the what the test is for.

src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
```diff
@@ -375,54 +361,85 @@ private bool PrunePackages(PublishRoot root)
...
                 {
                     var packagesDir = resolver.GetInstallPath(library.Name, library.Version);
-                    var manifest = resolver.GetManifestFilePath(library.Name, library.Version);
-
-                    keep.Add(manifest);
 
                     foreach (var path in library.RuntimeAssemblies)
                     {
-                        keep.Add(CombinePath(packagesDir, path));
+                        keep.Add(Path.Combine(packagesDir, Path.GetDirectoryName(path)));
```
JunTaoLuo: Instead of collecting files to keep, this collects `TFM`s. An alternative is to compare the lock files of the original project and the published output. Or some kind of fuzzy best match to figure it out from the `TFM` of the published lockfile.

src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
```diff
@@ -375,54 +361,85 @@ private bool PrunePackages(PublishRoot root)
...
+                }
+            }
 
-                    foreach (var specialFolder in specialFolders)
-                    {
-                        var specialFolderPath = CombinePath(packagesDir, specialFolder);
+            var frameworkSpecificContents = new List<string>
+            {
+                "lib",
+                "ref"
```
JunTaoLuo: May want to consider pruning unused runtimes, especially after #2628.

src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
```diff
@@ -375,54 +361,85 @@ private bool PrunePackages(PublishRoot root)
...
+            };
 
-                        if (!Directory.Exists(specialFolderPath))
-                        {
-                            continue;
-                        }
+            var extensionsToPrune = new List<string>
+            {
+                "xml"
```
JunTaoLuo: maybe `.pdb`s as well?

src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
```diff
@@ -375,54 +361,85 @@ private bool PrunePackages(PublishRoot root)
...
+            };
 
-                        keep.AddRange(Directory.EnumerateFiles(specialFolderPath, "*.*", SearchOption.AllDirectories));
-                    }
-                }
-            }
+            var exemptFrameworks = new List<FrameworkName>
+            {
+                VersionUtility.ParseFrameworkName("native")
```
JunTaoLuo: Currently the only recognized but potentially indirectly referenced `TFM`. Note that currently `xamarinmac20` and `xamarinios10` are common unrecognized `TMF`s and will not be pruned.

JunTaoLuo:
I have the fix in #2932 but the test hanged on linux again :astonished: :trollface: 
@troydai @pakrym

src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
davidfowl: Ugh, not sure about this.

src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
davidfowl: This is pretty weird...

src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
JunTaoLuo: Which part is strange?

davidfowl: That we need it. We should only be deleting things that exist in the lock file. All of this logic should be inverse.

JunTaoLuo: So we should compare the lockfile from the publish output to the lockfile from the original project? And find the difference between the restored packages and packages that are required for publish?

JunTaoLuo: I'm probably missing something here but I couldn't figure a way of using the lock file contents.

1. The lock file in the publish output only lists the files required by the frameworks specified during publish in the `targets` section. There is a `libraries` section which lists all the files in the package. However, there are files there that are used but not listed in `targets` like `runtimes/*` so I can't just prune the difference between the two sections.

2. I can check the difference between the `targets` section of the lockfiles in the project directory and the publish output. This has two problems. One, if you publish for the same frameworks you stored with, which I assume is the most common usage, nothing will be pruned. Two, even if you publish for one of the many frameworks you restored with. Pruning will miss such a large number of files, it'll almost become unnecessary. 

For example, say you restored the `System.Text.Encoding` package for `dnx451` and `dnxcore50` and you publish for `dnx451` only. The difference between the lockfiles' `targets` section will be only `lib/dnxcore50/*.dll`. Not only will this miss tons of unused frameworks like `lib/(MonoAndroid10|xamarinios10|net45|...)/*.dll`, it will also miss files that never appear in the `targets` section like `xml`, `sha512`, `nupkg`.

Hence I tried a compromise where instead of white listing a set of files and delete the rest, we black list a smaller set of files to prune that we are confident that won't be used.

But I'm assuming you still prefer option 2?

src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
```diff
@@ -375,54 +361,85 @@ private bool PrunePackages(PublishRoot root)
...
+                        {
+                            FrameworkName frameworkName;
+
+                            try
+                            {
+                                frameworkName = VersionUtility.ParseFrameworkName(frameworkFolder.Name);
+                            }
+                            catch
+                            {
+                                continue;
```
moozzyk: can this really happen? and if this does can it really be ignored?

troydai:
We shouldn't check in binary files. For testing nupkg we can generate it on the fly.

src/Microsoft.Dnx.Tooling/Publish/PublishRoot.cs
```diff
@@ -33,11 +34,12 @@ public PublishRoot(Runtime.Project project, string outputPath, IServiceProvider
             TargetRuntimesPath = Path.Combine(outputPath, AppRootName, "runtimes");
             Operations = new PublishOperations();
         }
-
+
```
JunTaoLuo: will remove spacing

JunTaoLuo:
Note that if you publish without restoring, no lock file will be generated for the main project and no pruning will occur.

src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
```diff
@@ -352,95 +352,94 @@ public bool PostProcess(PublishRoot root)
...
-                "redist",
-                "runtimes"
-            };
-
-            if (!root.NoSource)
+            if (root.MainProjectLockFile == null)
             {
-                // 'shared' folder is build time dependency, so we only copy it when deploying with source
-                specialFolders.Add("shared");
+                return true;
```
JunTaoLuo: Maybe we should remove the return value for PrunePackages. There are no cases for which we want to report a non-successful pruning that stops further publishing progress. If there are problems, such as no lock file in the main project, just don't prune anything.

src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
```diff
@@ -352,95 +352,94 @@ public bool PostProcess(PublishRoot root)
...
                 {
-                    var packagesDir = resolver.GetInstallPath(library.Name, library.Version);
-                    var manifest = resolver.GetManifestFilePath(library.Name, library.Version);
-
-                    keep.Add(manifest);
+                    var packageDir = resolver.GetInstallPath(library.Name, library.Version);
+                    
+                    if (library.Name != root.MainProjectName && library.Type == "package")
+                    {
+                        filesToRemove.Add(resolver.GetHashPath(library.Name, library.Version));
```
JunTaoLuo: I'm assuming all package dependencies have a hash and package file. If this is not a fair assumption I'll check for existence of these files before adding them to `filesToRemove`

src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
```diff
@@ -352,95 +352,94 @@ public bool PostProcess(PublishRoot root)
...
                 {
-                    var packagesDir = resolver.GetInstallPath(library.Name, library.Version);
-                    var manifest = resolver.GetManifestFilePath(library.Name, library.Version);
-
-                    keep.Add(manifest);
+                    var packageDir = resolver.GetInstallPath(library.Name, library.Version);
+                    
+                    if (library.Name != root.MainProjectName && library.Type == LibraryTypes.Package)
+                    {
+                        filesToRemove.Add(resolver.GetHashPath(library.Name, library.Version));
```
JunTaoLuo: I'm assuming all package dependencies have a hash and package file. If this is not a fair assumption I'll check for existence of these files before adding them to `filesToRemove`

JunTaoLuo:
ping @davidfowl @moozzyk  @troydai

src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
```diff
@@ -352,95 +352,94 @@ public bool PostProcess(PublishRoot root)
...
                     }
+                }
+            }
 
-                    foreach (var specialFolder in specialFolders)
-                    {
-                        var specialFolderPath = CombinePath(packagesDir, specialFolder);
+            foreach (var target in root.MainProjectLockFile.Targets.Where(projectTarget =>
+                !root.PublishedLockFile.Targets.Any(publishTarget =>
+                projectTarget.TargetFramework == publishTarget.TargetFramework)))
```
JunTaoLuo: Only pruning by `TFM`. Maybe we should prune by `RID` if one is specified as well?

src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
davidfowl: Sounds good.

src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
```diff
@@ -352,95 +352,94 @@ public bool PostProcess(PublishRoot root)
...
             {
                 foreach (var library in target.Libraries)
                 {
-                    var packagesDir = resolver.GetInstallPath(library.Name, library.Version);
-                    var manifest = resolver.GetManifestFilePath(library.Name, library.Version);
-
-                    keep.Add(manifest);
+                    var packageDir = resolver.GetInstallPath(library.Name, library.Version);
+                    
+                    if (library.Name != root.MainProjectName && string.Equals(library.Type, LibraryTypes.Package, StringComparison.OrdinalIgnoreCase))
```
davidfowl: Why this filter?

JunTaoLuo: You could have project dependencies which do not have any hash/package files. And the main project when published with --no-source is considered a package and in this case the package files should be preserved no?

src/Microsoft.Dnx.Tooling/Publish/PublishProject.cs
davidfowl: Maybe move this into a method.

JunTaoLuo: Difficult to do lockfile based `RID` filtering for now since the published project is not restored with the runtimes specified by publish. Could potentially revisit this once publish is refactored to be based on lock file.

JunTaoLuo:
ping @moozzyk @anurse @troydai @davidfowl

troydai:
:shipit:

davidfowl:
:shipit:

PR #2962, opened 2015-10-12
Suppress ICompilationException coming from EntryPointExecutor.Execute

pakrym:
Issue #2887
Related to https://github.com/aspnet/dnx/pull/2928

/cc @troydai @victorhurdugaci @davidfowl

### DIFF
--- a/src/Microsoft.Dnx.ApplicationHost/Program.cs
+++ b/src/Microsoft.Dnx.ApplicationHost/Program.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.ExceptionServices;
 using System.Threading.Tasks;
 using Microsoft.Dnx.Compilation;
 using Microsoft.Dnx.Compilation.FileSystem;
@@ -231,18 +232,7 @@ namespace Microsoft.Dnx.ApplicationHost
             }
             catch (Exception ex)
             {
-                // Try to find compilation exception recursively to supress its stack
-                var innerException = ex;
-                do
-                {
-                    if (innerException is ICompilationException)
-                    {
-                        throw SuppressStackTrace(innerException);
-                    }
-                    innerException = innerException.InnerException;
-                }
-                while (innerException != null);
-
+                SuppressCompilationExceptions(ex);
                 if (ex is FileLoadException || ex is FileNotFoundException)
                 {
                     ThrowEntryPointNotfoundException(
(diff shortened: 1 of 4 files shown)


src/Microsoft.Dnx.ApplicationHost/Program.cs
```diff
@@ -221,13 +221,30 @@ private bool ParseArgs(string[] args, out RuntimeOptions defaultHostOptions, out
             return false;
         }
 
-        private Task<int> ExecuteMain(DefaultHost host, string applicationName, string[] args)
+        private async Task<int> ExecuteMain(DefaultHost host, string applicationName, string[] args)
```
moozzyk: don't make it async

src/Microsoft.Dnx.ApplicationHost/Program.cs
```diff
@@ -221,13 +221,30 @@ private bool ParseArgs(string[] args, out RuntimeOptions defaultHostOptions, out
...
         {
             Assembly assembly = null;
 
             try
             {
-                assembly = host.GetEntryPoint(applicationName);
+                try
```
moozzyk: this nested try-catch is gross

davidfowl:
I don't quite understand what this is fixing. Can you show all of the cases that cause compilation errors and how this fixes each of them?

pakrym:
@davidfowl 
There are two source of compilation errors:
1. When we are searching for EntryPoint method and compiling application code on the way (`host.GetEntryPoint`)
2. Some code in application EntryPoint is doing compilation and throws compilation exception (AspNet hosting is searching for Startup type which causes related projects to compile). That was the case that didn't work for you here: https://github.com/aspnet/dnx/pull/2928#commitcomment-13703368

I moved `EntryPointExecutor.Execute` into `try/catch` that handles `ICompilationException` stack suppressing, so exception from entry point execution would be also hidden.

src/Microsoft.Dnx.ApplicationHost/Program.cs
```diff
@@ -250,8 +240,6 @@ private Task<int> ExecuteMain(DefaultHost host, string applicationName, string[]
                         applicationName,
                         ex.InnerException);
                 }
-
-                throw;
```
moozzyk: Isn't this `throw` still needed? I think without this we will suppress all exceptions but FileLoad/FileNotFound and compilation exceptions...

pakrym: Fixed.

src/Microsoft.Dnx.ApplicationHost/Program.cs
```diff
@@ -259,7 +247,33 @@ private Task<int> ExecuteMain(DefaultHost host, string applicationName, string[]
                 return Task.FromResult(1);
             }
 
-            return EntryPointExecutor.Execute(assembly, args, host.ServiceProvider);
+            return EntryPointExecutor.Execute(assembly, args, host.ServiceProvider)
+                        .ContinueWith(t =>
+                        {
+                            t.Exception?.Handle(e =>
```
moozzyk: nice! ...but make sure that it does not convert non-compilation exceptions to `AggregateExceptions`.... (i.e. before if an application threw it would just report the actual exception and not the exception wrapped in `AggregateException`)

moozzyk:
:ship::it: (after you confirm we are not wrapping application exceptions in AggregateExceptions and Travis passes)

src/Microsoft.Dnx.ApplicationHost/Program.cs
pakrym: We lock using GetAwaiter().GetResult() which doesn't wrap exception into AggregateException

moozzyk: Yes, but as per MSDN `Handle()` does:  _if any exceptions went unhandled, all unhandled exceptions will be put into a new AggregateException which will be thrown_ - so just I want to be sure we don't start wrapping exceptions accidently...

pakrym: I checked, exception thrown from user code is output correctly

moozzyk: :clap:

davidfowl: What the hall is Handle () :smile: That's new to me

moozzyk: @davidfowl - Do you know why I asked to check if it does not wrap the user's exception in the AggregateException? Because I had to look `Handle` up on msdn...

Issue #2715, opened 2015-09-16
Flow additional project.json properties into the assembly


phenning:
Currently the following project.json settings only flow into the nupkg:

```
  "description": "ClassLibrary4 Class Library",
  "language": "es",
  "copyright": "my copy",
  "title": "my title",
```

We should flow these into the Assembly attributes as well..

```
[assembly: AssemblyTitle("my title")]
[assembly: AssemblyDescription("ClassLibrary4 Class Library")]
[assembly: AssemblyCopyright("my copy")]
[assembly: AssemblyCulture("es")]
```

rustd:
@glennc @muratg can you look at this?

PR #2966, opened 2015-10-12
Flow project.json properties into the assembly.

cesarblum:
- title -> AssemblyTitle
- description -> AssemblyDescription
- copyright -> AssemblyCopyright

Language has already been handled by #2802.
#2715


cesarbs:
@victorhurdugaci @davidfowl @JunTaoLuo @troydai @phenning @rustd

src/Microsoft.Dnx.Compilation.CSharp/RoslynCompiler.cs
```diff
@@ -252,58 +252,43 @@ private CompilationModules GetCompileModules(CompilationTarget target)
...
-            {
-                versionAttributes.AppendLine($"[assembly:{assemblyVersionName}(\"{RemovePrereleaseTag(project.Version)}\")]");
-            }
+            var attributes = new StringBuilder();
 
-            if (!foundAssemblyInformationalVersion)
+            foreach (var attributeValue in attributeNameToProjectValue)
             {
-                versionAttributes.AppendLine($"[assembly:{assemblyInformationalVersion}(\"{project.Version}\")]");
+                attributes.AppendLine($"[assembly:{attributeValue.Key}(\"{attributeValue.Value}\")]");
```
victorhurdugaci: Do you want to add that attribute even when the value is null?

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
```diff
@@ -83,5 +85,34 @@ public void P2PDifferentFrameworks(DnxSdk sdk)
 
             TestUtils.CleanUpTestDir<DnuPackTests>(sdk);
         }
+
+        [Theory, TraceTest]
```
victorhurdugaci: What's a TraceTest?

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
```diff
@@ -83,5 +85,34 @@ public void P2PDifferentFrameworks(DnxSdk sdk)
...
+            // Act
+            var result = sdk.Dnu.Pack(project.ProjectDirectory);
+
+            // Assert
+            Assert.Equal(0, result.ExitCode);
+
+            var assemblyPath = Path.Combine(result.RootPath, "Debug", "dnx451", "Test.dll");
+            var assembly = Assembly.LoadFrom(assemblyPath);
+            var attributes = assembly.GetCustomAttributes(true);
+            Assert.Equal(project.Title, attributes.OfType<AssemblyTitleAttribute>().First().Title);
```
victorhurdugaci: Should you call Single() instead of First()? That will implicitly assert that you have a single attribute of that type

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
```diff
@@ -83,5 +85,34 @@ public void P2PDifferentFrameworks(DnxSdk sdk)
...
+            Assert.Equal(project.Description, attributes.OfType<AssemblyDescriptionAttribute>().First().Description);
+            Assert.Equal(project.Copyright, attributes.OfType<AssemblyCopyrightAttribute>().First().Copyright);
+            Assert.Equal(project.AssemblyFileVersion.ToString(), attributes.OfType<AssemblyFileVersionAttribute>().First().Version);
+            Assert.Equal(project.Version.ToString(), attributes.OfType<AssemblyInformationalVersionAttribute>().First().InformationalVersion);
+            Assert.Equal(project.Version.Version, assembly.GetName().Version);
+
+            TestUtils.CleanUpTestDir<DnuPackTests>(sdk);
```
victorhurdugaci: This statement will be called only if all the asserts pass (asserts throw). Is that expected?

src/Microsoft.Dnx.Compilation.CSharp/RoslynCompiler.cs
```diff
@@ -252,58 +252,43 @@ private CompilationModules GetCompileModules(CompilationTarget target)
...
-                }
-                else if (string.Equals(attributeName, assemblyVersionName, StringComparison.Ordinal))
-                {
-                    foundAssemblyVersion = true;
-                }
-                else if (string.Equals(attributeName, assemblyInformationalVersion, StringComparison.Ordinal))
+                if (attributeNameToProjectValue.ContainsKey(attributeName))
                 {
-                    foundAssemblyInformationalVersion = true;
+                    attributeNameToProjectValue.Remove(attributeName);
```
victorhurdugaci: Nit: You could build the attributes string right here, no need to iterate over the remaining collection one more time.

victorhurdugaci: Also, it would eliminate the need for doing dictionary operations (remove)

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
```diff
@@ -83,5 +85,34 @@ public void P2PDifferentFrameworks(DnxSdk sdk)
...
+            sdk.Dnu.Restore(solution.RootPath).EnsureSuccess();
+
+            // Act
+            var result = sdk.Dnu.Pack(project.ProjectDirectory);
+
+            // Assert
+            Assert.Equal(0, result.ExitCode);
+
+            var assemblyPath = Path.Combine(result.RootPath, "Debug", "dnx451", "Test.dll");
+            var assembly = Assembly.LoadFrom(assemblyPath);
```
davidfowl: We should use `System.Reflection.Metadata` instead of using `Assembly.Load`. You can't load this assembly with the same name twice.

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
```diff
@@ -83,5 +85,34 @@ public void P2PDifferentFrameworks(DnxSdk sdk)
...
+
+            sdk.Dnu.Restore(solution.RootPath).EnsureSuccess();
+
+            // Act
+            var result = sdk.Dnu.Pack(project.ProjectDirectory);
+
+            // Assert
+            Assert.Equal(0, result.ExitCode);
+
+            var assemblyPath = Path.Combine(result.RootPath, "Debug", "dnx451", "Test.dll");
```
davidfowl: The result of pack has an output directory so there's no need to hard code this much path. This will also fail for other target frameworks.

misc/AssemblyInfo/src/Test/project.json
```diff
@@ -0,0 +1,18 @@
+{
+  "title": "Hello title",
+  "description": "Hello description",
+  "language": "en",
+  "copyright": "Hello copyright",
```
troydai: You may want to add some unicode code point here. For example U+00A9 is commonly used for copy right sign.

src/Microsoft.Dnx.Compilation.Abstractions/CompilationProjectContext.cs
```diff
@@ -24,6 +27,9 @@ public class CompilationProjectContext
             CompilationTarget target,
```
troydai: Personally I find a long constructor very discouraging to use. You need to remember the sequence of the parameter or heavily relies on the IDE to give you a clue. I suggest to make properties settable.

src/Microsoft.Dnx.Compilation.CSharp/RoslynCompiler.cs
```diff
@@ -252,58 +252,43 @@ private CompilationModules GetCompileModules(CompilationTarget target)
...
-        private static CSharpCompilation ApplyVersionInfo(CSharpCompilation compilation, CompilationProjectContext project,
+        private static CSharpCompilation ApplyProjectInfo(CSharpCompilation compilation, CompilationProjectContext project,
             CSharpParseOptions parseOptions)
         {
-            const string assemblyFileVersionName = "System.Reflection.AssemblyFileVersionAttribute";
-            const string assemblyVersionName = "System.Reflection.AssemblyVersionAttribute";
-            const string assemblyInformationalVersion = "System.Reflection.AssemblyInformationalVersionAttribute";
+            var attributeNameToProjectValue = new Dictionary<string, string>(StringComparer.Ordinal)
+            {
+                ["System.Reflection.AssemblyTitleAttribute"] = project.Title,
```
troydai: Can we retrieve the name through reflection instead of hard code the names?

src/Microsoft.Dnx.Compilation.Abstractions/CompilationProjectContext.cs
moozzyk: devil's advocate - when a property has a getter and a setter you never know whether it is just for easy initialization or it actually being reset later. This can lead to wrong assumptions and bugs - where the code originally assumes it is only for initialization but then someone starts resetting the value. Also if you have many properties you probably have to rely on the IDE/Intellisense as well - only you can set them in any order...

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
JunTaoLuo: That's something I added previously. If you enable `DNX_TEST_TRACE` it will print out `Test Information: Begin test/End test: <test>`. This is sort of redundant now given we use the `-verbose` flag for xunit which prints out the beginning of the test

src/Microsoft.Dnx.Compilation.Abstractions/CompilationProjectContext.cs
anurse: I agree. Rather have a large constructor than accidentally-mutable objects...

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
```diff
@@ -108,5 +113,38 @@ public void DnuPack_ResourcesNoArgs(DnxSdk sdk)
...
+
+            // Assert
+            Assert.Equal(0, result.ExitCode);
+
+#if DNX451
+            var assemblyPath = Path.Combine(result.RootPath, result.Configuration, "dnx451", "Test.dll");
+#else
+            var assemblyPath = Path.Combine(result.RootPath, result.Configuration, "dnxcore50", "Test.dll");
+#endif
+            var assembly = Assembly.LoadFrom(assemblyPath);
```
cesarbs: Still pending on using System.Reflection.Metadata.

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
```diff
@@ -2,7 +2,12 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.IO;
+<<<<<<< 87a241624c0efb80b5e34e8fff185ec7d7e3148a
```
cesarbs: Oops.

src/Microsoft.Dnx.Compilation.Abstractions/CompilationProjectContext.cs
khellang: If only C# had [immutable object initializers](https://github.com/dotnet/roslyn/issues/229) :smile:

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
```diff
@@ -108,5 +111,81 @@ public void DnuPack_ResourcesNoArgs(DnxSdk sdk)
...
+
+            sdk.Dnu.Restore(solution.RootPath).EnsureSuccess();
+
+            // Act
+            var result = sdk.Dnu.Pack(project.ProjectDirectory);
+
+            // Assert
+            Assert.Equal(0, result.ExitCode);
+
+#if DNX451
```
davidfowl: This isn't right. Use the result object and the sdk object to get the appropriate output path. The target framework the test is being compiled with is irrelevant.

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
```diff
@@ -108,5 +111,81 @@ public void DnuPack_ResourcesNoArgs(DnxSdk sdk)
 
             TestUtils.CleanUpTestDir<DnuPackTests>(sdk);
          }
+         
+        [Theory, TraceTest]
+        [MemberData(nameof(DnxSdks))]
+        public void AssemblyInfo(DnxSdk sdk)
```
davidfowl: This needs a better test name

test/Microsoft.Dnx.Compilation.CSharp.Tests/RoslynCompilerTest.cs
```diff
@@ -0,0 +1,81 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Runtime.Versioning;
+using Microsoft.CodeAnalysis;
+using Microsoft.Dnx.Compilation.Caching;
+using Microsoft.Dnx.Runtime;
+#if DNX451
```
victorhurdugaci: Why not #if def the entire file?

cesarbs: We might want to add new tests which are not restricted to dnx451. The only reason this is #ifdef'd is that it uses Moq, which doesn't work with CoreCLR.

test/Microsoft.Dnx.Tooling.FunctionalTests/DnuPackTests.cs
cesarbs: Renamed to ProjectPropertiesFlowIntoAssembly.

src/Microsoft.Dnx.Compilation.CSharp/RoslynCompiler.cs
```diff
@@ -355,7 +327,7 @@ private static string RemovePrereleaseTag(string version)
 
             foreach (var d in dirs)
             {
-                ctx.Monitor(new FileWriteTimeCacheDependency(d));
+                ctx?.Monitor(new FileWriteTimeCacheDependency(d));
```
davidfowl: Unit test? make it null

davidfowl:
:shipit:

davidfowl:
very nice

src/Microsoft.Dnx.Compilation.CSharp/RoslynCompiler.cs
cesarbs: It is already :) I made that change because the first time I ran my test, it blew up there with a NullReferenceException.

RehanSaeed:
Was this never re-implemented for .NET Core 1.1.1? Only AssemblyTitle seems to be driven by the title propery of project.json.

davidfowl:
You mean the csproj stuff? I believe it is implemented. If you migrate from `project.json` to `csproj` it'll turn off all of the generation. AFAIK, if you clean up the csproj it does the right thing. What are you seeing and what does your project file look like?

Issue #2970, opened 2015-10-13
KeyNotFoundException when running dnu restore with Contentful.NET package


rytmis:
Running `dnu restore` (1.0.0-beta7, clr, win-x86) on the following project.json file:

```json
{
  "version": "1.0.0-*",
  "description": "RestoreRepro Console Application",
  "authors": ["your name"],
  "tags": [""],
  "projectUrl": "",
  "licenseUrl": "",

  "dependencies": {
		"Contentful.NET": "0.0.5-Alpha"  
  },

  "commands": {
    "ConsoleApplication": "ConsoleApplication"
  },

  "frameworks": {
    "dnx451": {}
  }
}
```

results in a KeyNotFoundException:

```
C:\temp\RestoreRepro
λ dnu restore
Microsoft .NET Development Utility CLR-x86-1.0.0-beta7-15532

  CACHE https://api.nuget.org/v3/index.json
Restoring packages for C:\temp\RestoreRepro\project.json
Writing lock file C:\temp\RestoreRepro\project.lock.json
----------
System.Collections.Generic.KeyNotFoundException: The given key was not present in the dictionary.
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at Microsoft.Dnx.Tooling.RestoreCommand.WriteLockFile(LockFile previousLockFile, String projectLockFilePath, Project project, List`1 graphItems, PackageRepository repository, IProjectResolver projectResolver, IEnumerable`1 contexts)
   at Microsoft.Dnx.Tooling.RestoreCommand.<RestoreForProject>d__61.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.Dnx.Tooling.RestoreCommand.<>c__DisplayClass60_0.<<Execute>b__2>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task)
[...]

PR #2971, opened 2015-10-13
Fixed nuget version casing mismatch

davidfowl:
- Always update the nuspec with the resolved version and id on
install.

#2970

### DIFF
--- a/misc/MismatchedVersionCasing/src/A/project.json
+++ b/misc/MismatchedVersionCasing/src/A/project.json
@@ -1,0 +1,8 @@
+{
+  "frameworks": {
+    "dnx451": { }
+  },
+  "dependencies": {
+    "Contentful.NET": "0.0.5-Alpha"
+  }
+}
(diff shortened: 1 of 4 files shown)


misc/MismatchedVersionCasing/src/A/project.json
```diff
@@ -0,0 +1,8 @@
+{
+  "frameworks": {
+    "dnx451": { }
+  },
+  "dependencies": {
+    "Contentful.NET": "0.0.5-Alpha"
```
moozzyk: Does this test depend on this package being on NuGet?

davidfowl: Yes, like all of our other tests and our source code :smile:

moozzyk: I think other tests mostly depend on packages we kind of control, no?

src/Microsoft.Dnx.Tooling/Utils/NuGetPackageUtils.cs
```diff
@@ -77,23 +77,21 @@ internal static class NuGetPackageUtils
                     // Fixup the casing of the nuspec on disk to match what we expect
                     var nuspecFile = Directory.EnumerateFiles(targetPath, "*" + NuGet.Constants.ManifestExtension).Single();
 
-                    if (!string.Equals(nuspecFile, targetNuspec, StringComparison.Ordinal))
```
moozzyk: Have you considered doing IgnoreCase check *and* printing the warning if the names still don't match? (I don't feel very strongly about about that though...)

davidfowl: Nah, doesn't matter. If we found it then we should use it.

misc/MismatchedVersionCasing/src/A/project.json
davidfowl: Some, some dont

Issue #2683, opened 2015-09-11
React to externalized file watcher changes


davidfowl:
- Delete IApplicationShutdown and merge it with IApplicationLifetime (see https://github.com/aspnet/Hosting/pull/264)
- Remove IFileWatcher interfaces from the runtime

davidfowl:
@victorhurdugaci assigned to you

davidfowl:
Need to do this post beta8

muratg:
@davidfowl still working on this?

davidfowl:
@pakrym You did this right?

muratg:
I think this was done.

victorhurdugaci:
@muratg @davidfowl you were supposed to be at the Christmas party... cc @Eilon fyi

muratg:
@victorhurdugaci I really wanted to go, but decided to take a nap before going (I had like 3 hours sleep last night) and couldn't wake up. :hankey:

pakrym:
@davidfowl yes, there is not IApplicationShutdown anymore

davidfowl:
@pakrym Well close da bug :smile:

PR #2972, opened 2015-10-13
Get rid of IFileWatcher

davidfowl:
Part of #2683

/cc @anurse @victorhurdugaci


victorhurdugaci:
Will this break tooling?

moozzyk:
does this break dnx-watch?

davidfowl:
@victorhurdugaci No, in beta8, @BillHiebert removed the --watch flag AFAIK

@moozzyk No, the watcher has the logic to watch the files :smile:

moozzyk:
:ship::it: (once travis passes)

PR #2977, opened 2015-10-14
Fixing the issue where coreclr tests are using the default runtime

moozzyk:
...which is never the one we just built

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -102,6 +102,10 @@ var RUNTIME_TARGETS='${new List<RuntimeTarget>()}'
 var CAN_BUILD_ONECORE = '${Directory.Exists(VS_ONECORE_ROOT) && Directory.GetFiles(VS_ONECORE_ROOT, "vcruntime.lib", SearchOption.AllDirectories).Any() && Directory.Exists(WIN10_SDK_LIB) && Directory.GetFiles(WIN10_SDK_LIB, "onecore.lib", SearchOption.AllDirectories).Any()}'
 
 @{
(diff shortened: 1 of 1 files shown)


moozzyk:
This PR depends on https://github.com/aspnet/Universe/pull/299

/cc @davidfowl @muratg @pranavkm

makefile.shade
```diff
@@ -102,6 +102,10 @@ var RUNTIME_TARGETS='${new List<RuntimeTarget>()}'
 var CAN_BUILD_ONECORE = '${Directory.Exists(VS_ONECORE_ROOT) && Directory.GetFiles(VS_ONECORE_ROOT, "vcruntime.lib", SearchOption.AllDirectories).Any() && Directory.Exists(WIN10_SDK_LIB) && Directory.GetFiles(WIN10_SDK_LIB, "onecore.lib", SearchOption.AllDirectories).Any()}'
 
 @{
+  E("CORECLR_RUNTIME_VERSION", "1.0.0-dev");
```
dougbu: This will execute unconditionally even on machines where `dev-install` has never run. Should instead change `test-package` target to create an alias for new runtime and set this environment variable there, just before calls to `DnxTest()`.

dougbu:
:watch:

davidfowl:
I think this is breaking me right now...

dougbu:
Suggest closing this and aspnet/Universe#299. We should do #3009 instead.

muratg:
@moozzyk is aware of #3009, I think he'll have a look at that one.

PR #2978, opened 2015-10-14
Using OSS signing on Mono if project json requires real signing

### LINKED ISSUE #2452: Compilation options 'CryptoPublicKey' and 'CryptoKeyFile' can't both be specified at the same time.
I'm attempting to dnu build a project.json that has a strong name. If I have a "keyFile" value then I get this error:

    Compilation options 'CryptoPublicKey' and 'CryptoKeyFile' can't both be specified at the same time.

If I don't have a keyFile value then the build successfully creates a DLL that is strong named using a key that I didn't specify. I have no idea where it came from.

### LINKED ISSUE #2558: Skip strong name signing on Mono until Roslyn supports it
Building projects that specify a `keyFile` in project.json (like [serilog](https://github.com/serilog/serilog/blob/b53cc797bc4a96a49912987b69a49762e12edb92/src/Serilog/project.json#L11)) doesn't work on Mono currently since the following error is generated:

```
error CS7027: Error signing output with public key from file '../../assets/Serilog.snk' -- Assembly signing not supported.
```

Since signing isn't supported by Roslyn on non-Windows platforms yet, signing should just be skipped and a warning printed instead.

moozzyk:
If project json contains the `keyFile` property the assembly should
be signed using the provided snk. However real signing is not
supported on Mono. Currently we just error out in this case, after
this change we will fall back to using OSS signing. (Fixes #2558)

Also adding a warning for a case where if both `strongName` and `keyFile`
settings are provided we would just display errors from Roslyn which might
be hard to understand/troubleshoot because the names used in the errors
are different from the names we use in project.json (Fixes #2452)

### DIFF
--- a/src/Microsoft.Dnx.Compilation.CSharp.Common/CompilerOptionsExtensions.cs
+++ b/src/Microsoft.Dnx.Compilation.CSharp.Common/CompilerOptionsExtensions.cs
@@ -72,7 +72,6 @@ namespace Microsoft.Dnx.Compilation.CSharp
             bool allowUnsafe = compilerOptions.AllowUnsafe ?? false;
             bool optimize = compilerOptions.Optimize ?? false;
             bool warningsAsErrors = compilerOptions.WarningsAsErrors ?? false;
-            bool strongName = compilerOptions.StrongName ?? false;
 
             Platform platform;
             if (!Enum.TryParse(value: platformValue, ignoreCase: true, result: out platform))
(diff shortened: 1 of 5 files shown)


moozzyk:
This requires #2977

davidfowl:
This combo needs to be supported in general. The combos will be:
- strongName: true - is sign with built in key
- ossSign + keyFile - what you have above (the default xplat form as well not just for mono)
- keyFile + delay sign - usual 
- keyFile

moozzyk:
1), 2) and 4) are already supported

3) do we really want to fallback to OSS Signing if the user specified keyFile? My thinking was that we wanted to do this on Mono because we treat Mono as dnx451 so the user has no way of saying "use keyFile on Desktop CLR and use OSS on Mono because keyFile is not supported" . For CoreClr "keyFile" is __never__ valid and the user has the option to specify compilation options for dnxcore50 so falling back is kind of encouraging bad practices.

davidfowl:
Of course it's valid. It's what we'll do on xplat in general by just on mono

moozzyk:
I must be missing something - if real signing (i.e. with snk) is supported/valid on CoreClr why we even need a fallback? Currently we just don't sign the assembly at all and print this warning (warning DNX1001: Strong name generation is not supported on CoreCLR. Skipping strong name generation.).

davidfowl:
Real signing is only supported on Windows period.

davidfowl:
We need to preserve the strong name ( public key) and oss sign

moozzyk:
I guess I don't see how the behavior we currently have and this PR is different from what you are saying...

davidfowl:
2 reasons, a keyFile + plus odd sign is valid be shouldn't warn. The mono/xplat error only applies if the oss sign flag wasn't set

moozzyk:
"keyFile" + "strongName" - we already talked about that. It results in this error from Roslyn: `Compilation options 'CryptoPublicKey' and 'CryptoKeyFile' can't both be specified at the same time.` and we didn't want to decide for the user which method of signing should be used on full Clr. 
Falling back to OSS signing on Mono (and the warning) only happens if strongName is not specified. If both are specified you get 1) because it's incorrect to have both.

davidfowl:
We're talking past each other. That's not what I'm suggesting. There's a missing combination here:

- OSS sign (without our generated public key) + a provided key

I'm suggesting that if oss sign is true and a key file is provided, then use that key file instead of generating one.

src/Microsoft.Dnx.Compilation.CSharp.Common/CompilerOptionsExtensions.cs
```diff
@@ -80,13 +79,46 @@ private static CSharpCompilationOptions GetCompilationOptions(ICompilerOptions c
...
+                {
+                    return options.WithCryptoPublicKey(StrongNameKey);
+                }
+
+                var delaySignString = Environment.GetEnvironmentVariable(EnvironmentNames.BuildDelaySign);
+                var delaySign =
+                    delaySignString == null
+                        ? compilerOptions.DelaySign
+                        : string.Equals(delaySignString, "true", StringComparison.OrdinalIgnoreCase) ||
+                            string.Equals(delaySignString, "1", StringComparison.OrdinalIgnoreCase);
```
troydai: nit: you don't need to ignore case when comparing a number

Issue #2937, opened 2015-10-08
Wrong version of dependency is copied to output directory


pakrym:
1. .\build & .\build dev-install
2. Look into %USERPROFILE%\.dnx\runtimes\dnx-clr-win-x86.1.0.0-rc1-15844\bin\lib\Microsoft.Dnx.Tooling\

Microsoft.Dnx.Project.dll is CoreClr version of this assembly

..\Microsoft.Dnx.Project\Microsoft.Dnx.Project.dll is correct

moozzyk:
Can this be this related to #2702?

pakrym:
Yes, looks similar.

pakrym:
This issue is caused by hardcoding "dnxcore50" in https://github.com/aspnet/dnx/blob/dev/makefile.shade#L1252

PR #2981, opened 2015-10-14
Fix issue when dependency with incorrect framework was copied to output folder

pakrym:
Targets #2937

/cc @moozzyk


moozzyk:
:ship::it:

PR #2998, opened 2015-10-16
Fix preprocessor when emitEntryPoint is set to true

pakrym:
@davidfowl


davidfowl:
:shipit:

pakrym:
re @davidfowl

Issue #2982, opened 2015-10-15
TaskCancelledException discovering tests


rynowak:
DTH is throwing every time while trying to load an assembly for test discovery. I hit this consistently with MVC.NoFun.sln and DNX 1.0.0-rc1-15899

```
Exception: System.AggregateException: One or more errors occurred. ---> System.InvalidOperationException: System.AggregateException: One or more errors occurred. ---> System.Threading.Tasks.TaskCanceledException: A task was canceled.
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at System.Threading.Tasks.Task`1.get_Result()
   at Microsoft.Dnx.Compilation.DesignTime.DesignTimeHostProjectCompiler.CompileProject(CompilationProjectContext projectContext, Func`1 referenceResolver, Func`1 resourcesResolver)
   at Microsoft.Dnx.Compilation.LibraryExporter.<>c__DisplayClass17_0.<ExportProject>b__0(CacheContext ctx)
   at Microsoft.Dnx.Compilation.Caching.CacheExtensions.<>c__DisplayClass0_0`1.<Get>b__0(CacheContext ctx)
   at Microsoft.Dnx.Compilation.Caching.Cache.CreateEntry(Object k, Func`2 acquire)
   at Microsoft.Dnx.Compilation.Caching.Cache.<>c__DisplayClass5_0.<AddEntry>b__0()
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.LazyInitValue()
   at System.Lazy`1.get_Value()
   at Microsoft.Dnx.Compilation.Caching.Cache.Get(Object key, Func`2 factory)
   at Microsoft.Dnx.Compilation.Caching.CacheExtensions.Get[T](ICache cache, Object key, Func`2 factory)
   at Microsoft.Dnx.Compilation.LibraryExporter.ExportProject(ProjectDescription project, String aspect)
   at Microsoft.Dnx.Compilation.LibraryExporter.GetExport(LibraryDescription library, String aspect)
   at Microsoft.Dnx.Compilation.LibraryExporter.GetExport(String name, String aspect)
   at Microsoft.Dnx.Compilation.LibraryExporter.GetExport(String name)
[...]

PR #3008, opened 2015-10-18
Fix NullRef caused by Sources response

troydai:
Addressing: #2982 

In [the removal of `IFileWatcher`](https://github.com/aspnet/dnx/commit/1a222049e0973ae3971b57d8ea8cffbb9ca34bd0), the code of initializing `ProjectSources` property was [removed](https://github.com/aspnet/dnx/commit/1a222049e0973ae3971b57d8ea8cffbb9ca34bd0?diff=unified#diff-177e2f2da3a478ae5072533e84a39961L26). The `null` property cause an `NullRef` exception to be thrown when `DesignTimeHostProjectCompiler` [talks to DTH](https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Compilation.DesignTime/ProcessingQueue.cs#L102). The exception eventually causes the `Close` [event](https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Compilation.DesignTime/ProcessingQueue.cs#L154) and consequently [cancels all the unfinished `Task` waiting for response](https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Compilation.DesignTime/DesignTimeHostCompiler.cs#L54).

There are a few problems make the issue difficult to detect:
1) DesignTime compiler's `ProcessQueue` is different from DTH. It is not well guarded.
2) The `ProcessQueue` [doesn't print the error message](https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Compilation.DesignTime/ProcessingQueue.cs#L154) in this scenario.

### DIFF
--- a/src/Microsoft.Dnx.Compilation.DesignTime/ProcessingQueue.cs
+++ b/src/Microsoft.Dnx.Compilation.DesignTime/ProcessingQueue.cs
@@ -62,36 +62,38 @@ namespace Microsoft.Dnx.Compilation.DesignTime
                             // Embedded Refs (special)
                             // Blob 1
                             // Blob 2
-                            var compileResponse = new CompileResponse();
-                            compileResponse.AssemblyPath = obj.ValueAsString(nameof(CompileResponse.AssemblyPath));
-                            compileResponse.Diagnostics = ValueAsCompilationMessages(obj, (nameof(CompileResponse.Diagnostics)));
-                            int contextId = obj.ValueAsInt("ContextId");
-                            int blobs = obj.ValueAsInt("Blobs");
-
-                            var embeddedReferencesCount = _reader.ReadInt32();
-                            compileResponse.EmbeddedReferences = new Dictionary<string, byte[]>();
-                            for (int i = 0; i < embeddedReferencesCount; i++)
+                            if (ProjectCompiled != null)
                             {
-                                var key = _reader.ReadString();
-                                int valueLength = _reader.ReadInt32();
-                                var value = _reader.ReadBytes(valueLength);
-                                compileResponse.EmbeddedReferences[key] = value;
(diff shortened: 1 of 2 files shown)


troydai:
I'll working on a few tests to cover this scenario. It will be complicated since It involve two processes DTH and TestHost. I tested this fix locally, so let's get this in first to fix user scenarios.

src/Microsoft.Dnx.Compilation.DesignTime/ProcessingQueue.cs
```diff
@@ -98,8 +98,11 @@ private void ReceiveMessages()
                             //    "MessageType": "Sources",
                             //    "Files": [],
                             //}
-                            var files = obj.ValueAsStringArray("Files");
-                            ProjectSources(files);
+                            if (ProjectSources != null)
```
troydai: Either one of the change can fix this issue. But I added both.

muratg:
:shipit:

davidfowl:
Ahhhh my bad 😀

davidfowl:
Go fully defensive admit check for all events

troydai:
Guarded all the event. I'll check in this change soon since it blocks major design time scenarios.

davidfowl:
Add a test

troydai:
Can I add a test in a follow up PR. I'd like to have this check in and unblock people quickly.

troydai:
There: https://github.com/aspnet/dnx/issues/3012

I'll check in the change once my local build verify passed and then begin working on #3012

moozzyk:
+1 for adding the test separately - I think we need this fix ASAP

davidfowl:
:shipit:

troydai:
I'll merge shortly.

Issue #3007, opened 2015-10-18
`dnu` and `dnx` commands fail silently with DNX Core on Travis / Linux builds


dougbu:
This failure blocks fixing aspnet/Universe#290.

See https://travis-ci.org/aspnet/dnx/jobs/85979374 for an example.
```
info: Exec
info:   program: /usr/bin/env
info:   commandline: bash -c "source \"/home/travis/build/aspnet/dnx/packages/KoreBuild/build/dnvm.sh\" && dnvm use default -runtime coreclr && dnu --verbose restore --parallel src test samples"
info:   workingdir: /home/travis/build/aspnet/dnx
Adding /home/travis/.dnx/runtimes/dnx-coreclr-linux-x64.1.0.0-rc1-15910/bin to process PATH
info: Exec
info:   program: git
info:   commandline: submodule update --init
info:   workingdir: 
```

Note the `dnvm use` writes its usual successful output at line 1078. The immediately-following `dnu --verbose restore` command outputs nothing but apparently has exit code `0` (doesn't fail the build).

Ran this test with a branch (`dougbu/dnxcore.restore.290`) of current `dev` to run `dnu restore` with DNX Core and to add debugging output. Any expected failures now that all `release` fixes are in `dev`?

The same branch builds fine on Ubuntu 14.04 and Mint 17.02 (a Ubuntu 14.04 derivative). Can't report on Mac until Travis restarts those servers. Anyone have a Ubuntu 12.04 system to debug further?

One suitable workaround would be to use Ubuntu 14.04 machines on Travis. Is that possible?

dougbu:
This problem is more general. Look for example at the end of recent `dev` build https://travis-ci.org/aspnet/dnx/jobs/85845411. Not sure what `dnx` commands do anything but return a `0` exit code when using DNX Core on Ubuntu 12.04.
```
info: Adding /home/travis/build/aspnet/dnx/artifacts/test/dnx-coreclr-linux-x64/runtimes/dnx-coreclr-linux-x64.1.0.0-t151016235319/bin to the PATH
info: Exec
info:   program: dnx
info:   commandline: test -parallel none
info:   workingdir: test/Microsoft.Dnx.Compilation.CSharp.Tests
info: Exec
info:   program: dnx
info:   commandline: test -parallel none
info:   workingdir: test/Microsoft.Dnx.Compilation.Tests
info: Exec
info:   program: dnx
info:   commandline: test -parallel none
[...]

PR #3009, opened 2015-10-18
Run only the expected tests

dougbu:
- do not use `DnxTest()` macro; it loops through frameworks as `test-package` target does already
  - `DnxTest()` also attempts to use `default` alias when testing `dnxcore50` though it's in `$env:Path`
  - therefore execute `dnx test` directly

nit: use `IsLinux` where `IsMono` is the wrong question

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -556,7 +556,7 @@ var CAN_BUILD_ONECORE = '${Directory.Exists(VS_ONECORE_ROOT) && Directory.GetFil
             foreach (var runtimeTarget in RUNTIME_TARGETS)
             {
                 move(runtimeTarget.TargetFolder, libPackage, ".dll");
-                if (!IsMono)
+                if (!IsLinux)
                 {
                     move(runtimeTarget.TargetFolder, libPackage, ".pdb");
                 }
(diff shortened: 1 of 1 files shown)


dougbu:
/cc @moozzyk. I suggest this as an alternate to #2977. It stops the extra `dnvm run default test` executions and tests the new DNX Core on Windows.

However due to #3007, this won't actually test DNX Core on Linux.

muratg:
Based on our chat, remember to remove .travis.yml from the PR.

dougbu:
@moozzyk could you have a look? I've rebased this branch and removed the test-only commit @muratg was talking about.

moozzyk:
@dougbu  - looking...
(btw. I just ran dnx with your change in the Universe and now my CoreClr tests are failing the build since they run with the `default` runtime and not the one that has changes. So, thanks for fixing this in Universe).

dougbu:
> now my CoreClr tests are failing the build

Always glad to make failures visible :smiling_imp:

makefile.shade
```diff
@@ -753,10 +752,19 @@ var CI_DARWIN_DROP_PATH = '${Environment.GetEnvironmentVariable("CI_DARWIN_DROP_
...
                 {
-                    continue;
+                    if (IsLinux)
+                    {
+                        // Work around issue with testing in parallel on Mono.
+                        testArgs = " -parallel none";
+                    }
+                    else
+                    {
+                        // Skip Mono if we're not running on Linux.
```
moozzyk: How about Mac?

dougbu: `IsLinux` is `true` on a Mac as well but I'll update the comment.

dougbu:
@moozzyk rebased and updated. Ready to :ship: ?

moozzyk:
LGTM . :ship: :it:

dougbu:
1eab366

PR #3018, opened 2015-10-19
Enabling signing assemblies using public key from snk on Mono and Cor…

### LINKED ISSUE #2452: Compilation options 'CryptoPublicKey' and 'CryptoKeyFile' can't both be specified at the same time.

### LINKED ISSUE #2558: Skip strong name signing on Mono until Roslyn supports it

moozzyk:
…eClr

If the project.json file contains the `keyFile` on CoreClr and Mono we
will extract the public key from the snk file and will OSS sign the
assembly with the extracted key. (Fixes #2558). We will do the same on
desktop CLR if the `keyFile` option is accompanied by `"useOssSigning": true`

As a result of this change the `keyFile` and the `useOssSigning` are
no longer mutually exclusive and we will no show cyptic Roslyn
compilation errors we would have shown before. (Fixes #2452).


davidfowl:
This looks good

davidfowl:
:shipit:

PR #3021, opened 2015-10-20
Build hangs on Linux when tests running in parallel on CoreClr on TeamCity

moozzyk:
Note Travis is passing and I was not able to repro the hang on my Linux box but the TeamCity Linux CI is consistently hanging and the difference between the build when it passed and when it started hanging is the parallel test execution. Disabling parallel test execution on CoreClr on Linux should unblock the CI for now but it would be good to understand why TeamCity fails on parallel execution at all.

/cc @dougbu @cesarbs @BrennanConroy


dougbu:
Strange that we can't reproduce the issue on Travis or locally but worth a try. Then we need to get to the bottom of this because real web apps don't run serially.

dougbu:
:shipit:

PR #3025, opened 2015-10-20
Escape certain characters when writing project properties to assembly.

cesarbs:


cesarbs:
@pakrym @davidfowl @victorhurdugaci @JunTaoLuo

pakrym:
Please add path to `CSharpSyntaxTree.ParseText(newAttributes, parseOptions)` call so we can see where syntax errors are from, if there would be any in the future

cesarbs:
Which path would that be?

pakrym:
Like "GeneratedAssemblyAttributes.cs"

pakrym:
Possibly use this method to escape http://source.roslyn.io/#Microsoft.CodeAnalysis.CSharp/SymbolDisplay/SymbolDisplay.cs,3636bc2a8304d9eb

pakrym:
Nope, wrong assembly

pakrym:
:shipit: from me

src/Microsoft.Dnx.Compilation.CSharp/RoslynCompiler.cs
```diff
@@ -274,7 +274,7 @@ private CompilationModules GetCompileModules(CompilationTarget target)
             {
                 compilation = compilation.AddSyntaxTrees(new[]
                 {
-                    CSharpSyntaxTree.ParseText(newAttributes, parseOptions)
+                    CSharpSyntaxTree.ParseText(newAttributes, parseOptions, nameof(ApplyProjectInfo) + ".cs", Encoding.UTF8)
```
JunTaoLuo: two nits:
1. Use string interpolation
2. specify the name of the option parameters like at https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Compilation.CSharp/RoslynCompiler.cs#L343

src/Microsoft.Dnx.Compilation.CSharp/RoslynCompiler.cs
```diff
@@ -286,17 +286,22 @@ private static string RemovePrereleaseTag(string version)
             // Simple reparse of the version string (because we don't want to pull in NuGet stuff
             // here because we're in an old-runtime/new-runtime limbo)
 
-            var dashIdx = version.IndexOf('-');
+            var dashIdx = version?.IndexOf('-');
```
moozzyk: Why don't you just check the version for null at the beginning and return null instead of doing this dance with nullables and checking the version twice with the `.?` operator?

src/Microsoft.Dnx.Compilation.CSharp/RoslynCompiler.cs
```diff
@@ -286,17 +286,22 @@ private static string RemovePrereleaseTag(string version)
...
             else
             {
-                return version.Substring(0, dashIdx);
+                return version?.Substring(0, dashIdx.Value);
             }
         }
 
+        private static string EscapeCharacters(string str)
+        {
+            return str != null ? SymbolDisplay.FormatLiteral(str, false) : null;
```
moozzyk: What is `false` here?

victorhurdugaci: http://www.coderesx.com/roslyn/html/D316757D.htm

"Escapes non-printable characters." <- is this what you want?

moozzyk: Sorry for not being explicit. I want a named parameter :)

cesarbs: @victorhurdugaci Yes.

moozzyk:
:ship::it: (github does not send updates where a PR is updated so I missed it)

PR #3027, opened 2015-10-20
Output more information when ExecResult.EnsureSuccess fails

pakrym:
@Tratcher @JunTaoLuo @victorhurdugaci @moozzyk

### DIFF
--- a/test/Microsoft.Dnx.Testing.Framework/ExecResult.cs
+++ b/test/Microsoft.Dnx.Testing.Framework/ExecResult.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Text;
 
 namespace Microsoft.Dnx.Testing.Framework
 {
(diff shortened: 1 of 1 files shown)


JunTaoLuo:
Why is this needed? The standard output and standard error should be logged by the `TestLogger` at https://github.com/aspnet/dnx/blob/dev/test/Microsoft.Dnx.Testing.Framework/Exec.cs#L98-L99

pakrym:
@JunTaoLuo To show details it in TeamCity failed tests section

test/Microsoft.Dnx.Testing.Framework/ExecResult.cs
```diff
@@ -15,7 +16,13 @@ public ExecResult EnsureSuccess()
         {
             if (ExitCode != 0)
             {
-                throw new InvalidOperationException($"Exit code was {ExitCode}");
+                var messageBuilder = new StringBuilder();
+                messageBuilder.AppendLine($"Exit code was {ExitCode}");
```
JunTaoLuo: nit: generally we use string interpolation in the style of https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.ApplicationHost/DefaultHost.cs#L73 since it avoids creating a new `StringBuilder` object.

JunTaoLuo:
:shipit: when travis passes. Though this is going to create duplicated output on local runs when `DNX_TEST_TRACE` is set I guess the benefits we get on TeamCity is worth it.

PR #3048, opened 2015-10-22
fix #2881 by adding support for RID override

### LINKED ISSUE #2881: Resolve Linux RIDs once and for all
PR https://github.com/aspnet/dnx/pull/2880 had us switch to reporting all Linuxes as `ubuntu.14.04` because we just saw so much variance in Ubuntu versions in use that we couldn't reliably get them all accurately detected. For example: Travis-CI runs Ubuntu 12, Our CI runs Ubuntu 14.10, and anyone who installs 14.04 immediately gets prompted to upgrade to 15.04, so many of our other machines use 15.04. Some folks also use Linux Mint 17.2 which is entirely based on Ubuntu 14.04.

In all of these cases, the Ubuntu 14.04 CoreCLR build works fine, we just have RID matching issues. In the end, the simplest solution for beta8 was to just always report Linux as `ubuntu.14.04` since that's the only Linux Distro we are releasing packages for (AFAIK). That way, we unblock users on other distros which MAY work but are not 100% verified.

We need to resolve this issue in RC1.

/cc @davidfowl @muratg @Eilon @DamianEdwards

anurse:
Fixes #2881 

We need to sync up our RID fallback with `Microsoft.NETCore.Platforms` (cc @ericstj). This PR also adds an override environment variable `DNX_RUNTIME_ID` that can be used to override the RID we use if we don't support the distro you're running on but you want to try anyway :)

### DIFF
--- a/src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
+++ b/src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
@@ -10,6 +10,10 @@ namespace Microsoft.Dnx.Runtime
 {
     internal static class RuntimeEnvironmentExtensions
     {
+        // This code doesn't have access to EnvironmentNames and I don't want to shake things up too much
+        // This should be the only place it is used though so it should be fine
+        private const string DnxRuntimeIdEnvironmentVariable = "DNX_RUNTIME_ID";
+
         public static string GetFullVersion(this IRuntimeEnvironment env)
         {
             var str = new StringBuilder();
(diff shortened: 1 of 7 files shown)


src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
```diff
@@ -100,27 +107,18 @@ public static string GetRuntimeOsName(this IRuntimeEnvironment env)
...
+                // Generate the distro-based name
+                var split = ver.Split(new[] { ' ' }, 2);
+                if (split.Length == 2)
+                {
+                    os = split[0].ToLowerInvariant();
+                    ver = split[1].ToLowerInvariant();
+                }
+                else
+                {
+                    os = "linux";
```
moozzyk: `RuntimeOperatingSystems.Linux.ToLower()` ?

moozzyk:
:ship::it:

src/Microsoft.Dnx.Runtime.Internals/RuntimeEnvironmentExtensions.cs
anurse: :+1:

anurse:
Fixing travis, and resolving @moozzyk 's comment then I'll merge

PR #3051, opened 2015-10-22
Allow replacing build and pack with scripts

### LINKED ISSUE #3050: Allow replacing the 'build' and 'pack' commands with scripts
Today we have `prebuild`, `postbuild`, etc. We should add `build` so that a user can change what happens on `dnu build`. Same for `pack`.

This allows more custom toolchains to integrate into the VS tooling, etc.

/cc @davidfowl

anurse:
Fixes #3050 

/cc @davidfowl 

Tests are not complete and the build will likely fail, but I manually tested it. I have some other stuff I need to get done today so I'll come back to the tests tomorrow morning.

### DIFF
--- a/src/Microsoft.Dnx.Tooling/Building/BuildContext.cs
+++ b/src/Microsoft.Dnx.Tooling/Building/BuildContext.cs
@@ -30,18 +30,13 @@ namespace Microsoft.Dnx.Tooling
             _project = project;
             _targetFramework = targetFramework;
             _configuration = configuration;
-            _targetFrameworkFolder = VersionUtility.GetShortFrameworkName(_targetFramework);
-            _outputPath = Path.Combine(outputPath, _targetFrameworkFolder);
+            _outputPath = outputPath;
+            _targetFrameworkFolder = VersionUtility.GetShortFrameworkName(targetFramework);
 
             _libraryExporter = compilationEngine.CreateProjectExporter(
                 _project, _targetFramework, _configuration);
 
             _libraryManager = _libraryExporter.LibraryManager;
-        }
-
-        public void Initialize(IReport report)
-        {
-            ShowDependencyInformation(report);
         }
 
         public bool Build(List<DiagnosticMessage> diagnostics)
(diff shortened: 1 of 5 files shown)


src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
```diff
@@ -185,86 +185,104 @@ private bool BuildConfiguration(string baseOutputPath, IEnumerable<FrameworkName
...
+            if (_buildOptions.GeneratePackages && _currentProject.Scripts.ContainsKey("pack"))
             {
-                _buildOptions.Reports.Information.WriteLine();
-                _buildOptions.Reports.Information.WriteLine("Building {0} for {1}",
-                    _currentProject.Name, targetFramework.ToString().Yellow().Bold());
-
-                contextVariables["build:TargetFramework"] = VersionUtility.GetShortFrameworkName(targetFramework);
-
-                if (!ScriptExecutor.Execute(_currentProject, "prebuild", getScriptVariable))
+                success = ScriptExecutor.Execute(_currentProject, "package", getScriptVariable);
```
davidfowl: pack not package

davidfowl:
Nice!

src/Microsoft.Dnx.Tooling/Building/BuildContext.cs
```diff
@@ -167,7 +162,7 @@ public void AddLibs(PackageBuilder packageBuilder, string pattern, bool recursiv
             }
         }
 
-        private void ShowDependencyInformation(IReport report)
+        public void ShowDependencyInformation(IReport report)
```
cesarbs: Check for report == null since this is now public.

src/Microsoft.Dnx.Tooling/Building/BuildManager.cs
anurse: Yeah, don't know how I missed that :).

src/Microsoft.Dnx.Tooling/Building/BuildContext.cs
anurse: :+1:

davidfowl:
:shipit:

anurse:
@davidfowl and I discussed this. With the rough testing situation because of the CoreCLR break, we're going to punt this for RC1. It wasn't really necessary, just a nice-to-have. /cc @muratg @Eilon @DamianEdwards

muratg:
:+1:

PR #3060, opened 2015-10-23
Enabling loading native libs from project references

moozzyk:
Asking for early feedback while I am working on tests

/cc @davidfowl


moozzyk:
@natemcmaster - updated the PR with functional tests that work on Windows and did not see the issue you had. Take a look at the project in misc maybe you will find what the difference is?

moozzyk:
@natemcmaster - updated the PR with functional tests that work on Windows and did not see the issue you had. Take a look at the project in misc maybe you will find what the difference is?

natemcmaster:
@moozzyk One difference - the Microsoft.Data.Sqlite package contains **both** managed and unmanaged libraries. Your test case does not.

moozzyk:
@natemcmaster - this still should work - the package I tested manually had both

natemcmaster:
@moozzyk confirmed. I tried move native libraries into a separate package and we still get the same DllImport error. I will try again after this PR is merged.

natemcmaster:
@moozzyk btw, I cleaned my Sqlite project and started from scratch. It appears to work now.

moozzyk:
@natemcmaster - glad to hear that. Still - a very weird case... btw I updated my Kestrel [branch](https://github.com/aspnet/KestrelHttpServer/compare/pawelka/dllimport) so that `DllImport` can work on Mono and CoreClr - you might find it helpful...

moozzyk:
Updated the PR with tests for references to packages and projects containing native dlls. Caveats:
- preloading of .so's does not work on Mono Linux so loading an .so from a package will not work unless the user adds the path to `LD_LIBRARY_PATH` (Kestrel expects the `.so` to be in a default location on Linux so no tricks are required)
- the RID override on Travis make a test fail so I just skip the test if the env variable is defined. It is fine since a fake `IRuntimeEnvironment` instance is injected and the test will run on Windows and will test non-Windows OSes

/cc @davidfowl

test/Microsoft.Dnx.ApplicationHost.FunctionalTests/DllImportTests.cs
```diff
@@ -0,0 +1,44 @@
...
+            VerifyDllImport(sdk, "PackageReferenceTest");
+        }
+
+        public void VerifyDllImport(DnxSdk sdk, string projectName)
+        {
+            var runtimeEnvironment = PlatformServices.Default.Runtime;
+
+            // Preloading .so's on Linux does not work and setting
+            // LD_LIBRARY_PATH is not an option so skipping
+            if (runtimeEnvironment.OperatingSystem == "Linux" && runtimeEnvironment.RuntimeType == "Mono")
```
davidfowl: any reason you're not using the xunit skip attributes we wrote

moozzyk: I thought it was not possible to skip for only Mono on Linux. If it is I will fix.

davidfowl: I assume you can do everything if not, then we suck :smile:

test/Microsoft.Dnx.ApplicationHost.FunctionalTests/DllImportTests.cs
```diff
@@ -0,0 +1,44 @@
...
+        public void VerifyDllImportWithProjectReference(DnxSdk sdk)
+        {
+            VerifyDllImport(sdk, "ProjectReferenceTest");
+        }
+
+        [Theory, TraceTest]
+        [MemberData(nameof(DnxSdks))]
+        public void VerifyDllImportWithPackageReference(DnxSdk sdk)
+        {
+            VerifyDllImport(sdk, "PackageReferenceTest");
```
davidfowl: Is this actually testing the package?

test/Microsoft.Dnx.ApplicationHost.FunctionalTests/DllImportTests.cs
moozzyk: We suck - the skip attributes are independent so Linux and Mono will skip all Linux and all Mono and not Mono on Linux. Leaving as is.

test/Microsoft.Dnx.ApplicationHost.FunctionalTests/DllImportTests.cs
moozzyk: Fixed using the beta8 feature.

moozzyk:
I updated the PR with the fix for the test. I also fixed a but where OS Version would be reported incorrectly on Mono runtime on El Capitan - the assumption was that the version has always 3 numbers but it apparently can have just two numbers in which case we reported just "10".

davidfowl:
One thing you missed that might now be important right now is then build time load context. The loaders usually have 2 overloads, one that uses the default load context and one that gets one passed in. This should follow the same pattern. If you have a child load context that's passed in, it should be used.

moozzyk:
@davidfowl - I will file a bug for this and do this separately. Otherwise is it good?

davidfowl:
Yep!

test/Microsoft.Dnx.ApplicationHost.FunctionalTests/DllImportTests.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        private static bool ShouldSkipTest
+        {
+            get
+            {
+                var runtimeEnvironment = PlatformServices.Default.Runtime;
+
+                // Preloading .so's on Linux does not work and setting
+                // LD_LIBRARY_PATH is not an option so skipping
+                return runtimeEnvironment.OperatingSystem == "Linux" && runtimeEnvironment.RuntimeType == "Mono";
```
dougbu: We don't test Mono on Windows. So what's wrong with just `[FrameworkSkipCondition(RuntimeFrameworks.Mono)]`?

moozzyk: Because we want to test on Mono on Mac

dougbu: Then place an `[OSSkipCondition(OperatingSystems.Linux)]` in an `#if` for full CLR.

moozzyk: I updated #3073 with your suggestion since the code has gone in.

Issue #2615, opened 2015-09-02
Strong name everything


moozzyk:
We should OSS/fake strong name our assemblies  -https://github.com/aspnet/EntityFramework/issues/999

muratg:
Moving this to RC1. @moozzyk feel free to bring it back if you have time at the end of the week.

cc: @rowanmiller

moozzyk:
:+1:

hkoestin:
@moozzyk & @muratg does this mean, that we do not get strong named assemblies of EntityFramework with beta 8?

I am asking, because I would need to come up with a huge workaround otherwise :-1:

muratg:
@hkoestin That's true. We just couldn't fit it in with all the other work that needed to happen in Beta 8. We'll do this in RC1.

hkoestin:
@muratg Thanks for the information. This is really unfortunate news for me now, as I would need signed assemblies :disappointed: You see any workaround for this? The problem is not the actual signing of the assembly, that I can do myself. It is more about the used [InternalsVisibleTo] attributes in the assemblies.

See https://github.com/aspnet/EntityFramework/issues/999#issuecomment-143674933

muratg:
@hkoestin Unfortunately, I don't know of any workaround. You can wait for this bug to be resolved, and work with RC1 dev packages before RC1 goes out if that helps.

hkoestin:
@muratg would be great if someone could ping me, if this is fixed. Then I would get a nightly build asap. When can I expect it to be fixed? A week? Two weeks? A month?

muratg:
@hkoestin Please subscribe to this issue if you haven't already (click "Subscribe" on top right) and you should get a notification when the status changes.  We can try pinging you on GitHub but also we could forget :)    I don't have an estimation at the moment when this will be closed, but it would be more than a week (as actual RC1 coding starts next week only) and ideally less than a month.

muratg:
See more details about OSS signing [here](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/oss-signing.md)

natemcmaster:
@cesarbs VS Intellisense doesn't recognized InternalsVisibleTo with a strong name. Compilation still works. Anyway around this in VS?

moozzyk:
@natemcmaster This is a tooling issue that was fixed very recently.

natemcmaster:
@moozzyk awesome, thank.

PR #3063, opened 2015-10-23
Resolve full path to key file.

cesarbs:
This is required for signing to work on both desktop CLR and other environments. Having a path relative to a project breaks our builds on non-desktop CLR because we attempt to read from that path relative to the repository root.

/cc @moozzyk @davidfowl @muratg @victorhurdugaci 

This change is required for #2615.

### DIFF
--- a/test/Microsoft.Dnx.Runtime.Tests/ProjectFacts.cs
+++ b/test/Microsoft.Dnx.Runtime.Tests/ProjectFacts.cs
@@ -255,7 +255,7 @@ namespace Microsoft.Dnx.Runtime.Tests
         {
             var project = ProjectUtilities.GetProject(@"
 {
-    ""dependencies"": {  
+    ""dependencies"": {
         ""A"": """",
         ""B"": ""1.0-alpha-*"",
         ""C"": ""1.0.0"",
@@ -301,7 +301,7 @@ namespace Microsoft.Dnx.Runtime.Tests
 {
     ""frameworks"": {
         ""net45"": {
-            ""dependencies"": {  
+            ""dependencies"": {
                 ""A"": """",
                 ""B"": ""1.0-alpha-*"",
                 ""C"": ""1.0.0"",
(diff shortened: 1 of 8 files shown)


src/Microsoft.Dnx.Runtime/ProjectReader.cs
```diff
@@ -459,6 +459,17 @@ private static CompilerOptions GetCompilationOptions(JsonObject rawObject)
                 return null;
             }
 
+            string keyFile = rawOptions.ValueAsString("keyFile");
+
+            if (keyFile != null)
+            {
+                keyFile = Path.GetFullPath(Path.Combine(projectPath, keyFile));
+                var originalColor = Console.ForegroundColor;
```
cesarbs: Oops.

src/Microsoft.Dnx.Runtime/ProjectReader.cs
```diff
@@ -459,6 +459,13 @@ private static CompilerOptions GetCompilationOptions(JsonObject rawObject)
                 return null;
             }
 
+            string keyFile = rawOptions.ValueAsString("keyFile");
+
+            if (keyFile != null)
+            {
+                keyFile = Path.GetFullPath(Path.Combine(projectPath, keyFile));
```
victorhurdugaci: I don't think this should happen in the project reader. The reader should contain the raw value

davidfowl: Agreed. This is a simple POCO. Do it in the compilation options.

cesarbs:
Ping.

troydai:
:shipit:

test/Microsoft.Dnx.Runtime.Tests/ProjectFacts.cs
```diff
@@ -394,7 +394,7 @@ public void CompilerOptionsAreSet()
             Assert.True(compilerOptions.WarningsAsErrors.Value);
             Assert.Equal("x86", compilerOptions.Platform);
             Assert.True(compilerOptions.Optimize.Value);
-            Assert.Equal(compilerOptions.KeyFile, @"c:\keyfile.snk");
+            Assert.Equal(compilerOptions.KeyFile, Path.GetFullPath("foo/keyfile.snk"));
```
davidfowl: This looks wrong. This will fail depending on where you run the test from no?

davidfowl:
@cesarbs Why is the ci red?

cesarbs:
@davidfowl It's green now, fixed the test you pointed out was wrong.

davidfowl:
:shipit:

PR #3064, opened 2015-10-24
Switch to use Socket.AcceptAsync

troydai:
Fix build break caused by CoreCLR build. 

I'll do a quick test and check in the change without sign off so as to unblock the build.

### DIFF
--- a/src/Microsoft.Dnx.DesignTimeHost/Program.cs
+++ b/src/Microsoft.Dnx.DesignTimeHost/Program.cs
@@ -119,7 +119,29 @@ namespace Microsoft.Dnx.DesignTimeHost
 
         private static Task<Socket> AcceptAsync(Socket socket)
         {
-            return Task.Factory.FromAsync((cb, state) => socket.BeginAccept(cb, state), ar => socket.EndAccept(ar), null);
+            var tcs = new TaskCompletionSource<Socket>();
+
+            var args = new SocketAsyncEventArgs();
+            args.Completed += (sender, e) =>
+            {
+                if (e.SocketError == SocketError.Success)
+                {
+                    tcs.SetResult(e.AcceptSocket);
+                }
+                else
+                {
+                    tcs.SetException(new Exception($"Failed to accept connection: {e.SocketError}."));
+                }
+            };
+
+            if (socket.AcceptAsync(args))
+            {
+                return tcs.Task;
+            }
+            else
+            {
+                return Task.FromResult(args.AcceptSocket);
+            }
         }
 
         private static void WriteProjectContexts(Message message, ProcessingQueue queue, IDictionary<int, ApplicationContext> contexts)


muratg:
Looks good to me. 

Does this CoreCLR change impact other repos too?

src/Microsoft.Dnx.DesignTimeHost/Program.cs
```diff
@@ -119,7 +119,28 @@ public static void Main(string[] args)
...
+                {
+                    tcs.SetResult(e.ConnectSocket);
+                }
+                else
+                {
+                    tcs.SetException(new Exception($"Failed to accept connection: {e.SocketError}."));
+                }
+            };
+
+            socket.AcceptAsync(args);
```
Tratcher: There should be an AcceptAsync that returns a task.

troydai:
I'll scan other places. Not tested yet, I'll come back work on this after dinner.

src/Microsoft.Dnx.DesignTimeHost/Program.cs
troydai: No such API.

troydai:
Looks like the implementation is wrong, tests failed.

src/Microsoft.Dnx.DesignTimeHost/Program.cs
```diff
@@ -119,7 +119,28 @@ public static void Main(string[] args)
...
+
+            args.Completed += (sender, e) =>
+            {
+                if (e.SocketError == SocketError.Success)
+                {
+                    tcs.SetResult(e.ConnectSocket);
+                }
+                else
+                {
+                    tcs.SetException(new Exception($"Failed to accept connection: {e.SocketError}."));
```
JunTaoLuo: Should we use something like `InvalidOperationException` instead of the generic `Exception`?

src/Microsoft.Dnx.DesignTimeHost/Program.cs
Tratcher: Hold off on merging this for a little, we need to figure out why we don't see the new API.

troydai: Yes. There suppose be one.

Tratcher: https://github.com/dotnet/corefx/pull/3655/files

troydai: The extension is not available on full CLR.

troydai:
The fix is functionally verified with DTH functional test. @Tratcher I think this is good enough for a urgent fix. We can keep refining this piece of code afterwards.

muratg:
Thanks @troydai. Looks like DNX (Linux and OS X) failed with Compilation test suite this time.  Also, one OSS signing test is failing.

muratg:
cc @moozzyk

davidfowl:
What? Why was this change made...

Tratcher:
Because they removed Begin/EndAccept. We were supposed to move to `Task<Socket> AcceptAsync()` but he couldn't find it.

davidfowl:
Then just use the blocking accept

troydai:
`Task<Socket> AcceptAsync()` is on CoreCLR only. Use it will need different code path for CLR and CoreCLR.

davidfowl:
It should just be blocking.

PR #3072, opened 2015-10-25
Improve cross platform support for environment variables

vlesierse:
DNX allows environment variables `DNX_HOME` and `DNX_PACKAGES_CACHE` to be set with multiple paths:
**Windows** `%USERPROFILE%\.dnx;C:\Program Files\Dnx`
**Linux/Mac** `~/.dnx:/usr/local/lib/dnx`

The path separator for both environments are different (`;` for windows and `:` for linux/mac) although the code splits the string always using a semi column. Because of this DNX looses the ability to supply multiple paths for these environment variables.


borgdylan:
Recently new native library loading code was added that also needs this kind of fix,

davidfowl:
LGTM

PR #3084, opened 2015-10-27
Implement stand alone implementations

davidfowl:
- IApplicationEnvironment and IRuntimeEnvironment now have stand alone
implementations. These should work on any platform.
- Still need to figure out OS version but it's not critical for now
- These implementations should work on .NET framework, .NET Core and Mono

/cc @pakrym @anurse @moozzyk


pakrym:
:shipit:

src/Microsoft.Extensions.PlatformAbstractions/DefaultRuntimeEnvironment.cs
```diff
@@ -0,0 +1,114 @@
...
+        public DefaultRuntimeEnvironment()
+        {
+            OperatingSystem = GetOs();
+            RuntimePath = GetLocation(typeof(object).GetTypeInfo().Assembly);
+            RuntimeType = GetRuntimeType();
+            RuntimeVersion = typeof(object).GetTypeInfo().Assembly.GetName().Version.ToString();
+            RuntimeArchitecture = GetArch();
+        }
+
+        // TODO: implement
```
anurse: Going to fix this in the PR or file a bug for later? Either way, we need to track this

anurse:
:shipit: after filing a bug to handle the OS Version.

src/Microsoft.Extensions.PlatformAbstractions/DefaultRuntimeEnvironment.cs
```diff
@@ -0,0 +1,114 @@
...
+#endif
+            return GetUname();
+        }
+
+        private static string GetArch()
+        {
+#if NET451
+            return Environment.Is64BitProcess ? "x64" : "x86";
+#else
+            return IntPtr.Size == 8 ? "x64" : "x86";
```
moozzyk: arm?

davidfowl: No dice, no idea how to check

moozzyk: Yeah... If we don't find anything better we could check %PROCESSOR_ARCHITECTURE%

anurse: That's Windows-only though.

moozzyk: At the moment the only ARM runtime we have is Windows. Hopefully there will be an API in CoreClr that will allow to detect this.

davidfowl: What value should I look for in PROCESSOR_ARCHITECTURE

moozzyk: Looks like this would be just "ARM" - https://www.raspberrypi.org/forums/viewtopic.php?f=105&t=123945&p=833274

src/Microsoft.Extensions.PlatformAbstractions/DefaultRuntimeEnvironment.cs
davidfowl: Never

PR #3089, opened 2015-10-27
Fix DnuCommands_Install_InstallsWorkingCommand test failure

pakrym:
@JunTaoLuo @anurse @BrennanConroy


BrennanConroy:
:+1: is that the only affected test?

pakrym:
yes

davidfowl:
:shipit:

BrennanConroy:
I think you need to add the runtime to the front of the path instead of replacing the path

PR #3090, opened 2015-10-28
Remove System.Data.SqlClient from Microsoft.Dnx

pranavkm:


pranavkm:
cc @BrennanConroy \ @anurse \ @divega

anurse:
:shipit: (assuming this actually fixes the EF issue)

divega:
@anurse @pranavkm we know removing the sni.dll from the runtime folder fixes the issue but we don't know exactly why. 

Can you easily tell from the code which package exactly would we currently extract sni.dll from? Several packages contain it and I would like to double-check before I report it back to the SqlClient folks.

divega:
Other than that :shipit:

davidfowl:
Probably because we're getting it from the packages folder

PR #3098, opened 2015-10-28
Strong name everything

cesarbs:
#2615 

This is the only way we've (thanks @victorhurdugaci!) found to sign everything without running into weird runtime dependency cycles which completely prevented signing.

All the new files you see there copied from submodules to src/Microsoft.Dnx.Runtime and src/Microsoft.Dnx.Tooling are temporary. This will be reverted after this initial signed build. The root cause of what causes issues in signing is having a preprocess step in any project.json file. Sync with me if you want more details.

cc @davidfowl @muratg @moozzyk


cesarbs:
Closing in favor of #3105.

PR #3099, opened 2015-10-28
Rename Microsoft.AspNet.Hosting.json to hosting.json

BrennanConroy:
https://github.com/aspnet/Hosting/issues/452

@Tratcher @davidfowl @Eilon @DamianEdwards


davidfowl:
:shipit:

PR #3100, opened 2015-10-28
Split and expose more of entry point apis

pakrym:
https://github.com/aspnet/Testing/issues/180
/cc @davidfowl


davidfowl:
For the test host? It's not worth it. Use the activator utilities in common.

pakrym:
I don't wan't to duplicate entry point finder logic

davidfowl:
It isn't the same though. You're building a different test entry point for the test infrastructure.

pakrym:
Yes,  but I will search for the main (which should be there) and look for a second one in same class.

Sent by Outlook<http://taps.io/outlookmobile> for Android



On Wed, Oct 28, 2015 at 5:06 PM -0700, "David Fowler" <notifications@github.com<mailto:notifications@github.com>> wrote:


It isn't the same though. You're building a different test entry point for the test infrastructure.

-

PR #3115, opened 2015-11-02
Fail gracefully if the test project doesn't support test services

pakrym:
@davidfowl

### DIFF
--- a/src/Microsoft.Dnx.Runtime.Sources/Impl/DependencyInjection/ActivatorUtilities.cs
+++ b/src/Microsoft.Dnx.Runtime.Sources/Impl/DependencyInjection/ActivatorUtilities.cs
@@ -76,12 +76,12 @@ namespace Microsoft.Dnx.Runtime.Common.DependencyInjection
 
             if (constructors.Length == 1)
             {
-                if (services == null)
-                    throw new InvalidOperationException($"Unable to inject into {type.Name} constructor, current host doesn't support dependency injection");
-
                 ParameterInfo[] parameters = constructors[0].GetParameters();
                 return services =>
                 {
+                    if (services == null)
+                        throw new InvalidOperationException($"Unable to inject into {type.Name} constructor, current host doesn't support dependency injection");
+
                     var args = new object[parameters.Length];
                     for (int index = 0; index != parameters.Length; ++index)
                     {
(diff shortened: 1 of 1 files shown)


src/Microsoft.Dnx.Runtime.Sources/Impl/DependencyInjection/ActivatorUtilities.cs
```diff
@@ -76,6 +76,9 @@ public static T CreateInstance<T>(IServiceProvider services)
 
             if (constructors.Length == 1)
             {
+                if (services == null)
+                    throw new InvalidOperationException($"Unable to inject into {type.Name} constructor, current host doesn't support dependency injection");
+
```
rynowak: This is used in MVC for a bunch of unrelated stuff. This fix isn't close enough to what it's claims to address.

pakrym: What else should it address?

davidfowl:
This is why we shouldn't be using the shard code

rynowak:
IMO testhost should just have the same semantics and limitations as DNX as a whole.

davidfowl:
Nope it should just have its own entry point

pakrym:
@rynowak, I can change error message to something like "Requesting code does not provide dependency injection services". It would be semantically correct for all use-cases

pakrym:
:up: :date:, any conclusion?

src/Microsoft.Dnx.Runtime.Sources/Impl/DependencyInjection/ActivatorUtilities.cs
```diff
@@ -79,6 +79,9 @@ public static T CreateInstance<T>(IServiceProvider services)
                 ParameterInfo[] parameters = constructors[0].GetParameters();
                 return services =>
                 {
+                    if (services == null)
+                        throw new InvalidOperationException($"Unable to inject into {type.Name} constructor, current host doesn't support dependency injection");
```
victorhurdugaci: Always use brackets around if's body

victorhurdugaci: Suggestion for message: "Cannot create instance of {type} because the current host doesn't support dependency injection"

pakrym: It looks like it's not only used in hosts, anyway there is no conclusion if we need this at all.

pakrym:
I guess we don't care anymore about this.

PR #3126, opened 2015-11-03
Rename Microsoft.Dnx.Compilation.Abstractions to Microsoft.Extensions.Compilation.Abstractions 

pakrym:
Here go namespaces.

/cc @pranavkm  @davidfowl

### DIFF
--- a/test/Microsoft.Dnx.ApplicationHost.FunctionalTests/AppHostTests.cs
+++ b/test/Microsoft.Dnx.ApplicationHost.FunctionalTests/AppHostTests.cs
@@ -20,10 +20,9 @@ namespace Microsoft.Dnx.ApplicationHost.FunctionalTests
             sdk.Dnu.Restore(project.ProjectDirectory).EnsureSuccess();
 
             // Act
-            var result = sdk.Dnx.Execute(project);
+            var result = sdk.Dnx.Execute(project).EnsureSuccess();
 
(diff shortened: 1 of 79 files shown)


test/Microsoft.Dnx.ApplicationHost.FunctionalTests/AppHostTests.cs
```diff
@@ -20,10 +20,9 @@ public void LibraryExporterGetExports(DnxSdk sdk)
             sdk.Dnu.Restore(project.ProjectDirectory).EnsureSuccess();
 
             // Act
-            var result = sdk.Dnx.Execute(project);
+            var result = sdk.Dnx.Execute(project).EnsureSuccess();
```
pranavkm: Why?

pranavkm:
:shipit:

test/Microsoft.Dnx.ApplicationHost.FunctionalTests/AppHostTests.cs
pakrym: More info if it fails, EnsureSuccess dumps stderr

pranavkm: Do the other tests need in this type need to be changed to do something similar?

PR #3129, opened 2015-11-04
Updating package versions and fixing lineup

pranavkm:
cc @anurse \ @ericstj \  @davidfowl


ericstj:
LGTM, the serialization update shouldn't be needed after tomorrows build.

pranavkm:
Ok. We were seeing issues with it today so this should unblock it.

anurse:
:shipit:

@ericstj will your changes to the RID definitions also make tomorrow's build (assuming all goes well)?

PR #3134, opened 2015-11-04
Reset signing compilation options for proprocessor aspect assemblies

pakrym:
@davidfowl @moozzyk

### DIFF
--- a/src/Microsoft.Dnx.Compilation.CSharp/RoslynCompiler.cs
+++ b/src/Microsoft.Dnx.Compilation.CSharp/RoslynCompiler.cs
@@ -121,7 +121,11 @@ namespace Microsoft.Dnx.Compilation.CSharp
             if (isPreprocessAspect)
             {
                 compilationSettings.CompilationOptions =
-                    compilationSettings.CompilationOptions.WithOutputKind(OutputKind.DynamicallyLinkedLibrary);
+                    compilationSettings.CompilationOptions
+                        .WithOutputKind(OutputKind.DynamicallyLinkedLibrary)
+                        .WithCryptoKeyFile(null)
+                        .WithCryptoPublicKey(ImmutableArray<byte>.Empty)
+                        .WithDelaySign(false);
             }
 
             var compilation = CSharpCompilation.Create(
--- a/test/Microsoft.Dnx.Compilation.CSharp.Tests/FakeCompilerOptions.cs
+++ b/test/Microsoft.Dnx.Compilation.CSharp.Tests/FakeCompilerOptions.cs
@@ -1,0 +1,28 @@
+﻿using System.Collections.Generic;
+using Microsoft.Extensions.PlatformAbstractions;
+
+namespace Microsoft.Dnx.Compilation.CSharp.Tests
+{
+    internal class FakeCompilerOptions : ICompilerOptions
+    {
+        public bool? AllowUnsafe { get; set; }
+
+        public IEnumerable<string> Defines { get; set; }
+
+        public bool? DelaySign { get; set; }
+
+        public bool? EmitEntryPoint { get; set; }
+
+        public string KeyFile { get; set; }
+
+        public string LanguageVersion { get; set; }
+
+        public bool? Optimize { get; set; }
+
+        public string Platform { get; set; }
+
+        public bool? UseOssSigning { get; set; }
+
+        public bool? WarningsAsErrors { get; set; }
+    }
+}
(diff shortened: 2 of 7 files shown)


src/Microsoft.Dnx.Compilation.CSharp/RoslynCompiler.cs
```diff
@@ -121,7 +121,10 @@ public class RoslynCompiler
             if (isPreprocessAspect)
             {
                 compilationSettings.CompilationOptions =
-                    compilationSettings.CompilationOptions.WithOutputKind(OutputKind.DynamicallyLinkedLibrary);
+                    compilationSettings.CompilationOptions
+                        .WithOutputKind(OutputKind.DynamicallyLinkedLibrary)
+                        .WithDelaySign(false)
+                        .WithCryptoKeyFile(null);
```
moozzyk: Does it work if the project is using Oss signing?

moozzyk: Can you add some tests? The tests I [created](https://github.com/aspnet/dnx/blob/pakrym/reset-sign/test/Microsoft.Dnx.Compilation.CSharp.Tests/SigningFacts.cs) before (https://github.com/aspnet/dnx/blob/pakrym/reset-sign/test/Microsoft.Dnx.Compilation.CSharp.Tests/SigningFacts.cs) might be helpful.

pakrym:
:up: :date:

moozzyk:
The fakes where in the same file and marked as private because they are not useful outside this test class so there is no reason to pollute the solution with new files. Is there any specific reason you moved them out? If not, I would prefer to put them where they were before.

pakrym:
I used them in my tests

moozzyk:
I see. I just glanced at the change and did not notice your tests were in a different file...

test/Microsoft.Dnx.Compilation.CSharp.Tests/RoslynCompilerTest.cs
```diff
@@ -79,6 +81,51 @@ public void FlowsProjectPropertiesIntoAssembly()
             Assert.All(compilationAttributes, compilationAttribute => expectedAttributes[compilationAttribute.AttributeClass.ToString()].Equals(
                 compilationAttribute.ConstructorArguments.First().Value));
         }
-#endif
+
+        [Fact]
+        public void DoesNotSignAndEmitEntyPoinInPreprocessAssembly()
```
moozzyk: EntyPoin->EntryPoint

test/Microsoft.Dnx.Compilation.CSharp.Tests/RoslynCompilerTest.cs
```diff
@@ -79,6 +81,51 @@ public void FlowsProjectPropertiesIntoAssembly()
...
+            // Act
+            var compilationContext = compiler.CompileProject(
+                compilationProjectContext,
+                new List<IMetadataReference> { metadataReference },
+                new List<ISourceReference> { },
+                () => new List<ResourceDescriptor>());
+
+            Assert.Equal(OutputKind.DynamicallyLinkedLibrary, compilationContext.Compilation.Options.OutputKind);
+            Assert.Equal(null, compilationContext.Compilation.Options.CryptoKeyFile);
+            Assert.Equal(0, compilationContext.Compilation.Options.CryptoPublicKey.Length);
```
moozzyk: Do you need a separate test for this? You did not set `useOssSigning` so you can't tell if it was reset or not. On the other hand I don't think you can test `keyFile` and `useOssSigning:true` in one test because `Options.CryptoKeyFile will never be set.`

test/Microsoft.Dnx.Compilation.CSharp.Tests/FakeNamedDependencyProvider.cs
```diff
@@ -0,0 +1,17 @@
...
+    class FakeNamedDependencyProvider : INamedCacheDependencyProvider
+    {
+        public ICacheDependency GetNamedDependency(string name)
+        {
+            return null;
+        }
+
+        public void Trigger(string name)
+        {
+
```
moozzyk: nit: remove this line

pakrym:
:up: :date:

test/Microsoft.Dnx.Compilation.CSharp.Tests/FakeCompilerOptions.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System.Collections.Generic;
+using Microsoft.Extensions.PlatformAbstractions;
+
+namespace Microsoft.Dnx.Compilation.CSharp.Tests
+{
+    class FakeCompilerOptions : ICompilerOptions
```
moozzyk: nit: add the modifier - I think we are adding them always even if they are the same as default. Same for the const strings in the `RoslynCompilerTest`

moozzyk:
:ship::it: (after you add access modifiers and travis passes)

PR #3144, opened 2015-11-08
Change message for Entrypoint type name error

PMoneda:
Guys I change to message when the EnpointEtry has no type called Program, I spent 20 minutes to discover that error in my program was the name of Entrypoint class.
The message that DNX show me was about my Main method but the main method was right and the error was in the name of class. So, i change de error message to be more specific in that case.

What do you think?

### DIFF
--- a/src/Microsoft.Dnx.Runtime.Sources/Impl/EntryPointExecutor.cs
+++ b/src/Microsoft.Dnx.Runtime.Sources/Impl/EntryPointExecutor.cs
@@ -91,7 +91,7 @@ namespace Microsoft.Dnx.Runtime.Common
                 if (programTypeInfo == null)
                 {
                                         
-                    System.Console.WriteLine("'{0}' does not contain a type 'Program' for an entry point", name);
+                    System.Console.WriteLine("'{0}' does not contain a 'Program' type suitable for an entry point", name);
                     return false;
                 }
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.Dnx.Runtime.Sources/Impl/EntryPointExecutor.cs
```diff
@@ -90,7 +90,8 @@ private static bool TryGetEntryPoint(Assembly assembly, IServiceProvider service
 
                 if (programTypeInfo == null)
                 {
-                    System.Console.WriteLine("'{0}' does not contain a static 'Main' method suitable for an entry point", name);
+                                        
+                    System.Console.WriteLine("'{0}' does not contain a type 'Program' for an entry point", name);
```
anurse: Slight reword suggestion: '{0}' does not contain a 'Program' type suitable for an entry point

anurse:
Minor reword suggestion but otherwise looks good to me!

src/Microsoft.Dnx.Runtime.Sources/Impl/EntryPointExecutor.cs
```diff
@@ -81,7 +81,7 @@ private static bool TryGetEntryPoint(Assembly assembly, IServiceProvider service
             {
                 return true;
             }
-
+
```
moozzyk: nit: revert

PMoneda:
@anurse I changed the message to "'{0}' does not contains a 'Program' type suitable for an entry point"

anurse:
Looks good to me! Clean up the extra space added (@moozzyk 's comment) and I'll merge it!

PMoneda:
@anurse, @moozzyk 's comment it's done!

anurse:
Looks good! Thanks for the contribution!

PR #3150, opened 2015-11-10
CentOS

BrennanConroy:
@moozzyk @anurse

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -899,6 +899,14 @@ functions @{
             return string.Equals(Uname(), "Linux");
         }
     }
+    
+    bool CanBuildForCentOS
+    {
+        get
+        {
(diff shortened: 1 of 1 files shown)


BrennanConroy:
Failure is because Travis is using 12.04 instead of Trusty, sent them an email

makefile.shade
```diff
@@ -926,6 +934,20 @@ functions @{
...
+    {
+        try
+        {
+            string result;
+            ExecuteAndRedirectOutput("cat", "/etc/os-release", out result);
+            return result.Contains(distro);
+        }
+        catch
+        {
+            return false;
```
moozzyk: Maybe just let it fail? It would have saved us some time yesterday ;)

moozzyk:
:ship::it:

PR #3176, opened 2015-11-17
Explicitly choose Mono 4.0.5

dougbu:
- avoids future problems related to aspnet/External#48
  - e.g. when Travis updates default Mono version in `csharp` bundle

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ sudo: required
 dist: trusty
 env:
   - MONO_THREADS_PER_CPU=2000
+mono:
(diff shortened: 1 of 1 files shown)


davidfowl:
:shipit:

dougbu:
/cc @pranavkm @NTaylorMullen 

This is just future-proofing our Travis configuration. If agreed I'll apply in all repos using Mono but without the `mono` key in `.travis.yml` (that's the majority of our repos).

dougbu:
Thx @davidfowl

PR #3182, opened 2015-11-19
Fix for publishing issue with RIDs

### LINKED ISSUE #3183: dnu publish does not honor RIDs
When using `dnu publish`, we do not scan the correct section in the lock file and as a result we miss most of the `runtime.[rid].[assemblyname]` packages. This causes problems with runtime-specific packages that are not embedded in DNX (for example `System.Data.SqlClient` and `System.Net.Security`)

/cc @DamianEdwards @muratg @Eilon

anurse:
This is a candidate fix for the publishing issue with RIDs. I verified that it fixes the problem in my repro. I still need to go through it and tidy it up a bit though.

fixes #3183

/cc @davidfowl @muratg @ericstj @DamianEdwards @danroth27 @Eilon

### DIFF
--- a/misc/PublishRuntimeDependency/Program.cs
+++ b/misc/PublishRuntimeDependency/Program.cs
@@ -1,8 +1,13 @@
 ﻿using System;
+using System.Data.SqlClient;
 using System.IO;
+using System.Net.NetworkInformation;
 using System.Net.Security;
 using System.Net.Sockets;
+using System.Net.WebSockets;
+using System.Reflection;
 using System.Text;
+using System.Threading;
 
 namespace PublishRuntimeDependency
 {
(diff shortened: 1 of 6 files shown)


misc/PublishRuntimeDependency/Program.cs
```diff
@@ -0,0 +1,104 @@
...
+
+            using (var stream = GetStream(socket, url))
+            {
+                // Send the request
+                var request = $"GET {url.PathAndQuery} HTTP/1.1\r\nHost: {url.Host}\r\nConnection: close\r\n\r\n";
+                var buffer = Encoding.UTF8.GetBytes(request);
+                stream.Write(buffer, 0, buffer.Length);
+
+                // Read the response
+                byte[] readbuffer = new byte[1024];
```
moozzyk: Can you just use `StreaemReader.ReadToEnd` here?

anurse: This code was actually taken from a different test which was trying to repro a bug in System.Net.Security around StreamReader.ReadToEnd :). I can trim it back since we don't care about that here.

moozzyk:
LGTM

Issue #2788, opened 2015-09-23
Remove Configuration from IApplicationEnvironment


davidfowl:
It's a compilation specific concept so we should expose it via Microsoft.Dnx.Compilation.Abstractions

/cc @pranavkm @anurse

anurse:
:+1:

muratg:
@davidfowl I don't think this is for RC1.  Moving out.  Bring it back if necessary.

muratg:
Assigned to @pakrym 

FYI @davidfowl

muratg:
cc @pranavkm this is related to Razor compilation.

pranavkm:
:+1:

pakrym:
https://github.com/aspnet/Mvc/pull/3632

PR #3188, opened 2015-11-20
Remove Configuration from IApplicationEnvironment

pakrym:
Also drop SetData/GetData
Issue: https://github.com/aspnet/dnx/issues/2788

@JunTaoLuo @davidfowl @pranavkm @anurse  @troydai


davidfowl:
Have you spoken to @pranavkm about how Razor will resolve this?

davidfowl:
:shipit: I love it!

pranavkm:
Doesn't look like we use this property in Mvc. There're places where we implement IApplicationEnvironment where it shows up, but that's about it.

davidfowl:
Except, you do https://github.com/aspnet/Mvc/blob/737a8f3b1dd4c8067fc671a08e3cac0d7d88cd48/src/Microsoft.AspNet.Mvc.Razor/Compilation/CompilationOptionsProviderExtension.cs#L43

pakrym:
@davidfowl  `if (environment.IsDevelopment()) { ... } ` ?

davidfowl:
@pakrym Commit this change. I think razor needs a specific option for this on the razor view engine options.

pakrym:
https://github.com/aspnet/Mvc/pull/3632

PR #3194, opened 2015-11-23
remove the embedded version of NETCore.Platforms

### LINKED ISSUE #3193: Publish --no-source doesn't pull in runtime dependencies
In order to properly pull in runtime dependencies, a direct dependency on `Microsoft.NETCore.Platforms` is required. However, when we publish with `--no-source`, we create a synthetic outer project that references the compiled project output. This means that we appear to incorrectly handle the reference to `Microsoft.NETCore.Platforms`.

Some solutions:

1. If the project has a dependency on `Microsoft.NETCore.Platforms`, pull that dependency through to the project we create
2. Fix whatever transitive logic isn't working to pull in `Microsoft.NETCore.Platforms` correctly.

/cc @davidfowl @muratg @Eilon @BrennanConroy

anurse:
instead, we use the data as a fallback and don't do weird tricks around injecting package references

fixes #3193 

I confirmed that with this, the `PublishRuntimeDependency` project has a nearly identical lock file when published as no-source as when published with source. The only difference is attributable to the synthetic "root" project we create. I haven't verified on Mac yet, but am confident this will fix it. I'll try it on Mac, but if I can get sign-off we can optimistically begin the DNX build.

/cc @Eilon @muratg @davidfowl @BrennanConroy

### DIFF
--- a/src/Microsoft.Dnx.Tooling/Restore/EmbeddedRuntimeData.cs
+++ b/src/Microsoft.Dnx.Tooling/Restore/EmbeddedRuntimeData.cs
@@ -1,16 +1,9 @@
 ﻿using System.Collections.Generic;
-using System.IO;
-using System.Linq;
-using System.Runtime.Versioning;
-using System.Threading.Tasks;
-using Microsoft.Dnx.Runtime;
-using Microsoft.Extensions.PlatformAbstractions;
 using Microsoft.Dnx.Tooling.Restore.RuntimeModel;
-using NuGet;
 
 namespace Microsoft.Dnx.Tooling
 {
-    internal class ImplicitPackagesWalkProvider : IWalkProvider
+    internal static class EmbeddedRuntimeData
     {
         // Based on values from the real Microsoft.NETCore.Platforms, but without 'aot' since DNX doesn't support .NET Native
         // These are designed to be an in-memory equivalent to a runtime.json.
@@ -94,7 +87,7 @@ namespace Microsoft.Dnx.Tooling
             new RuntimeSpec("ubuntu.15.04-x64", "ubuntu.15.04", "ubuntu.14.10-x64"),
 
             // Ubuntu 15.10 is not compatible. It upgraded icu to 55
-            
+
             // Linux Mint 17.x is compatible with Ubuntu 14.04
             new RuntimeSpec("linuxmint.17", "ubuntu.14.04"),
             new RuntimeSpec("linuxmint.17-x64", "linuxmint.17", "ubuntu.14.04-x64"),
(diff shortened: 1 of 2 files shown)


muratg:
Thanks @anurse!  

@davidfowl Could you review the PR?  We're looking into getting a build out ASAP.

@joeloff FYI

src/Microsoft.Dnx.Tooling/Restore/RestoreCommand.cs
```diff
@@ -617,6 +614,14 @@ private IEnumerable<string> GetRestoreRuntimes(IEnumerable<string> projectRuntim
                     }
                 }
             }
+
+            // Add the embedded runtime data if we didn't find any imports
+            if(imports == null)
```
davidfowl: nit: formatting

src/Microsoft.Dnx.Tooling/Restore/EmbeddedRuntimeData.cs
```diff
@@ -7,10 +7,11 @@
 using Microsoft.Extensions.PlatformAbstractions;
 using Microsoft.Dnx.Tooling.Restore.RuntimeModel;
 using NuGet;
+using System;
```
JunTaoLuo: nit: sort usings

src/Microsoft.Dnx.Tooling/Restore/EmbeddedRuntimeData.cs
```diff
@@ -105,51 +106,14 @@ internal class ImplicitPackagesWalkProvider : IWalkProvider
             new RuntimeSpec("linuxmint.17.3", "linuxmint.17.2"),
             new RuntimeSpec("linuxmint.17.3-x64", "linuxmint.17.3", "linuxmint.17.2-x64"));
 
-        public bool IsHttp
+        internal static IEnumerable<string> GetEmbeddedImports(string runtimeName)
         {
-            get
+            RuntimeSpec spec;
+            if(ImplicitRuntimeFile.Runtimes.TryGetValue(runtimeName, out spec))
```
JunTaoLuo: same nit: formatting

anurse:
:up: :date: @JunTaoLuo @davidfowl

anurse:
Confirmed that `--no-source` is fixed on Mac with this. Note that publishing for a specific runtime will still be quirky in El Capitan (since we'll publish for 10.10 if the selected runtime is `darwin`, but the runtime will detect 10.11)

anurse:
This PR was to the wrong branch, but this is now merged into release, so I'm closing it.

Issue #3196, opened 2015-11-24
dnx -p and --appbase


schotime:
Is it possible to run both the -p and --appbase options.

The scenario is I want to execute a unit test command from the root directory, but specify the appbase as if I was running from the same location as the project.json file.

I get the error. Could not load file or assembly 'test' when trying to run.

dnx -p .\tests --appbase .\tests test

Thanks.

davidfowl:
`-p` alone already does what you want.

moozzyk:
No, because `-p` is expanded to `--appbase`. `dnx -p .\tests test` expands to `dnx --appbase .\tests Microsoft.Dnx.ApplicationHost test` so is what you are looking for.

davidfowl:
?

schotime:
This line in my tests 

```csharp
var configuration = new Microsoft.Framework.Configuration.ConfigurationBuilder()
    .Add(new JsonConfigurationProvider( "config.json"))
    .Build();
```

does not find the config.json file when running not in the directory itself. 
eg. I need to set the running path.

davidfowl:
@schotime Did you try: `dnx -p {myfolder} test`

davidfowl:
Wait I think I get what you're asking now. That code is in the application you are testing?

schotime:
That code is in the test project and yes I did try it.

schotime:
https://github.com/schotime/NPoco/blob/dnx/test/NPoco.Tests/Common/BaseDBDecoratedTest.cs#L27

schotime:
Thanks to @davidfowl.
You must use the Extension Method `AddJsonFile` like so as it will set the path correctly.

```csharp
var configuration = new Microsoft.Extensions.Configuration.ConfigurationBuilder()
    .AddJsonFile("config.json")
    .Build();
```

https://github.com/aspnet/Configuration/blob/dev/src/Microsoft.Extensions.Configuration.Json/JsonConfigurationExtensions.cs#L60

PR #3203, opened 2015-11-25
Ensure current directory is set to the project directory #3196

schotime:
Related to issue #3196.
Sets the current directory to the project directory.


davidfowl:
Setting the working directory is always a bad idea. If you really need that, do it in your application.

schotime:
So how is #3196 fixed then? Surely is should work the same way whether I run it from the directory where project.json is vs a parent directory specifying the project directory (-p).

davidfowl:
The application base is set to the project directory when using `-p`. The config API uses the application base directory to find relative config files.

davidfowl:
Try using AddJsonFile instead of using the provider directly https://github.com/aspnet/Configuration/blob/dev/src/Microsoft.Extensions.Configuration.Json/JsonConfigurationExtensions.cs#L60

schotime:
Victory!!!

PR #3225, opened 2015-12-03
Microsoft.Extensions.CommandLineUtils Option switches with '/'

couven92:
Tests and implementation for options switches that use forward-slash ('/')

Some applications (especially older Windows applications) use forward-slah instead of GNU-typical dashes ('-')

This implementation allows for slashes to be used as well as the original dashes


davidfowl:
We don't want to allow slashes because it it hard to determine what's a path argument and what's a command line argument

couven92:
Ah, that makes sense, of course, I actually had not thought about that.

Issue #3245, opened 2015-12-10
DNX_BUILD_KEY_FILE is ignored in RC1


emgarten:
DNU is no longer using the environment variable DNX_BUILD_KEY_FILE in RC1.

Steps
1. Set the environment variable DNX_BUILD_KEY_FILE to the full path of a signing key
2. Build with DNU

Expected:
The output will be signed

Actual:
```
[21:53:13][Step 2/9] Error: Value cannot be null.
[21:53:13][Step 2/9] Parameter name: path2
[21:53:13][Step 2/9] Error: Full Exception: System.ArgumentNullException: Value cannot be null.
[21:53:13][Step 2/9] Parameter name: path2
[21:53:13][Step 2/9]    at System.IO.Path.Combine(String path1, String path2)
[21:53:13][Step 2/9]    at Microsoft.Dnx.Compilation.CSharp.CompilerOptionsExtensions.AddSigningOptions(CSharpCompilationOptions options, ICompilerOptions compilerOptions, String projectDirectory)
[21:53:13][Step 2/9]    at Microsoft.Dnx.Compilation.CSharp.CompilerOptionsExtensions.GetCompilationOptions(ICompilerOptions compilerOptions, String projectDirectory)
[21:53:13][Step 2/9]    at Microsoft.Dnx.Compilation.CSharp.CompilerOptionsExtensions.ToCompilationSettings(ICompilerOptions compilerOptions, FrameworkName targetFramework, String projectDirectory)
[21:53:13][Step 2/9]    at Microsoft.Dnx.Compilation.CSharp.RoslynCompiler.CompileProject(CompilationProjectContext projectContext, IEnumerable`1 incomingReferences, IEnumerable`1 incomingSourceReferences, Func`1 resourcesResolver)
[21:53:13][Step 2/9]    at Microsoft.Dnx.Compilation.CSharp.RoslynProjectCompiler.CompileProject(CompilationProjectContext projectContext, Func`1 referenceResolver, Func`1 resourcesResolver)
[21:53:13][Step 2/9]    at Microsoft.Dnx.Compilation.LibraryExporter.<>c__DisplayClass17_0.<ExportProject>b__0(CacheContext ctx)
[21:53:13][Step 2/9]    at Microsoft.Dnx.Compilation.Caching.CacheExtensions.<>c__DisplayClass0_0`1.<Get>b__0(CacheContext ctx)
[...]

PR #3247, opened 2015-12-10
Restore DNX_BUILD_KEY_FILE behavior (#3245).

cesarbs:
Fix https://github.com/aspnet/dnx/issues/3245

cc @anurse @davidfowl @emgarten

### DIFF
--- a/src/Microsoft.Dnx.Compilation.CSharp.Common/CompilerOptionsExtensions.cs
+++ b/src/Microsoft.Dnx.Compilation.CSharp.Common/CompilerOptionsExtensions.cs
@@ -97,11 +97,10 @@ namespace Microsoft.Dnx.Compilation.CSharp
 
             var keyFile =
                 Environment.GetEnvironmentVariable(EnvironmentNames.BuildKeyFile) ??
-                compilerOptions.KeyFile;
+                GetKeyFileFullPath(projectDirectory, compilerOptions.KeyFile);
 
-            if (!string.IsNullOrEmpty(keyFile))
+            if (!string.IsNullOrWhiteSpace(keyFile))
             {
-                keyFile = Path.GetFullPath(Path.Combine(projectDirectory, compilerOptions.KeyFile));
 #if DOTNET5_4
                 return options.WithCryptoPublicKey(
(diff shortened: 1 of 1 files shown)


src/Microsoft.Dnx.Compilation.CSharp.Common/CompilerOptionsExtensions.cs
```diff
@@ -128,6 +127,16 @@ private static CSharpCompilationOptions AddSigningOptions(CSharpCompilationOptio
...
         }
 
+        private static string GetKeyFileFullPath(string projectDirectory, string keyFile)
+        {
+            if (keyFile == null)
+            {
+                return null;
+            }
+
+            return Path.GetFullPath(Path.Combine(projectDirectory, keyFile));
```
moozzyk: int: consider ternary

src/Microsoft.Dnx.Compilation.CSharp.Common/CompilerOptionsExtensions.cs
```diff
@@ -128,6 +127,16 @@ private static CSharpCompilationOptions AddSigningOptions(CSharpCompilationOptio
             return useOssSigning ? options.WithCryptoPublicKey(StrongNameKey) : options;
         }
 
+        private static string GetKeyFileFullPath(string projectDirectory, string keyFile)
+        {
+            if (keyFile == null)
```
moozzyk: I wonder if we shouldn't do the same if the keyFile is an empty string..

anurse: Yeah, `string.IsNullOrEmpty` here.

anurse:
Minor comments. :shipit: once resolved

muratg:
@emgarten do you need this in DNU?  Otherwise let's close this PR.

cesarbs:
@anurse @moozzyk Went with `string.IsNullOrWhitespace` instead of `string.IsNullOrEmpty`.

moozzyk:
That's fine

emgarten:
:shipit:

Issue #3246, opened 2015-12-10
Is there a repository for Microsoft.Framework.CommandLineUtils.Sources?


reliak:
Is there a git repository for Microsoft.Framework.CommandLineUtils.Sources?
Because I'm using the latest version (https://www.nuget.org/packages/Microsoft.Framework.CommandLineUtils.Sources/) and I think I found a bug in the handling of commandline arguments.

muratg:
@davidfowl To Common repo, or create a new repo for this (cc @Eilon)?

Eilon:
Common seems fine to me for now.

victorhurdugaci:
Do we plan to replace this with `System.CommandLine`?

Eilon:
@victorhurdugaci can you compare & contrast those and we can discuss?

Eilon:
(Not necessarily in this bug, but we can discuss offline.)

BrennanConroy:
The command line utils package has moved to https://github.com/aspnet/Common/src/Microsoft.Extensions.CommandLineUtils
Also it is now an actual package and no longer a source package.

shanselman:
Where is this code **now**?

pranavkm:
@shanselman like @BrennanConroy pointed out, it's at https://github.com/aspnet/Common/tree/dev/src/Microsoft.Extensions.CommandLineUtils. We also have a package available on nuget.org: 
https://www.nuget.org/packages/Microsoft.Extensions.CommandLineUtils/

babelshift:
@pranavkm, as of today, there's nothing interesting at the github link you've provided. Has the source moved?

davidfowl:
https://github.com/aspnet/Common/tree/dev/src/Microsoft.Extensions.CommandLineUtils

babelshift:
@davidfowl There's nothing there. Everything was removed in a commit last December.

babelshift:
Looks like it's actually here now?

https://github.com/aspnet/Common/tree/dev/shared/Microsoft.Extensions.CommandLineUtils.Sources/CommandLine

Eilon:
@babelshift there are two versions of the command line utils package: one is an actual library that you reference, and the other (the `.Sources` one), has the same code, but it's a "sources package", which means that the code gets included and compiled into *your* library as internal APIs. Both contain the exact same functionality.

fredrikhr:
[...]

PR #3257, opened 2015-12-14
Move CommandLineUtils to Common

BrennanConroy:
#3246


muratg:
I think it's OK to close this one.  Do we have some other repo using the one from Common now (even if it's just a test)?

davidfowl:
Close? Why not just do this?

muratg:
Do we care about dnx?  We plan to stop even building it hopefully soon, don't we?

davidfowl:
No but this moves the library out of DNX. Are you suggesting we just leave it in and have a copy in Common? That's fine I guess.

muratg:
Yeah, [the one in Common](https://github.com/aspnet/Common/pull/69) is being reviewed. I think we can close this one in the DNX repo.

PR #3268, opened 2015-12-17
React to Platform Abstractions changes

anurse:
These are a little hacky, but they get things working again after Platform Abstractions changes. Since DNX is fading away, I'm not too worried about the impact of the hacks. All the RID generation is in PlatformAbstractions now.

/cc @moozzyk @davidfowl @pranavkm


test/Microsoft.Dnx.Host.Tests/RuntimeEnvironmentTests.cs
```diff
@@ -1,183 +0,0 @@
...
-using System;
-using System.Linq;
-using Microsoft.AspNet.Testing.xunit;
-using Microsoft.Dnx.Runtime;
-using Microsoft.Extensions.PlatformAbstractions;
-using Xunit;
-
-namespace Microsoft.Dnx.Host.Tests
-{
-    public class RuntimeEnvironmentTests
```
anurse: These tests can (and will) be moved to PlatformAbstractions.

anurse: aspnet/PlatformAbstractions#14

Issue #3275, opened 2015-12-21
Not recognizing RID on Ubuntu 14 


natemcmaster:
Recent change https://github.com/aspnet/dnx/commit/2be312f5ff7e675140d68f74d42b753d593c7fef broke RID recognition on Ubuntu 14 (aka, our targeted linux distro for CI and docs). Restore does not include the proper "runtimes.*" packages => runtime failures.

cc @victorhurdugaci @anurse

natemcmaster:
CI is failing: https://travis-ci.org/aspnet/EntityFramework/builds/97756973#L511

natemcmaster:
cc @muratg

muratg:
@BrennanConroy

natemcmaster:
``GetRuntimeIdentifier()`` returns `"linux.ubuntu 14.04-x64"`

BrennanConroy:
Closed via #3276

PR #3276, opened 2015-12-21
Fix overriding RuntimeEnvironment values

BrennanConroy:
#3275
@moozzyk @natemcmaster


moozzyk:
:ship::it:

BrennanConroy:
So this doesn't give the correct RID for mono now...

PR #3277, opened 2015-12-22
Small change to avert compiler errors when _UNICODE is set to 0

jamesqo:
Was just browsing some code the other day when I came across this, so I thought I'd submit a pull request.

TCHAR is may cause compiler errors when mixed with `wchar_t`, since it can change to `char` if the `_UNICODE` macro is set to 0.

This PR replaces all uses of `TCHAR` with `wchar_t` and removes all includes of its header.

### DIFF
--- a/src/dnx.coreclr/stdafx.h
+++ b/src/dnx.coreclr/stdafx.h
@@ -7,8 +7,6 @@
 
 #define _CRT_SECURE_NO_WARNINGS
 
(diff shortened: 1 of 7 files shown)


muratg:
Thanks for the PR.  We're not doing any more changes to DNX since we're retiring it in RC2. Command line experience will be through dotnet CLI instead.

PR #3285, opened 2015-12-27
fix overflow of environment variable %path%

imiuka:
with only 50 referenced nuget packages, dnx, kestrel, and iis express all fail with a cryptic error, "environment variable name or value is too long".

the list of newly discovered native library paths currently isn't cleared once appended to %path%. because the holding string builder is allocated outside the loop, once a new native library is found, then it (as well as further native library paths) are eternally appended into themselves on each loop iteration, even if further packages don't have any native libraries. this balloons the length of %path% and quickly causes it to exceed the 32k environment variable value length limit in .net.

my dnx launch ended up becoming something like this:

```
1. [kestrel]            kestrel/native
2. [some managed lib]   kestrel/native;kestrel/native
3. [other managed lib]  kestrel/native;kestrel/native;kestrel/native
4. [clrcompression]     kestrel/native;kestrel/native;kestrel/native;kestrel/native;clrcompression/native
5. [more managed lib]   kestrel/native;kestrel/native;kestrel/native;kestrel/native;clrcompression/native;kestrel/native;clrcompression/native
...
15. uh oh %path% limit exceeded
```

this pull request is a one line fix that resets the list each time it's appended. not sure what tests to add here (if any), please provide guidance.

### DIFF
--- a/src/Microsoft.Dnx.Runtime/DependencyManagement/PackageDependencyProvider.cs
+++ b/src/Microsoft.Dnx.Runtime/DependencyManagement/PackageDependencyProvider.cs
@@ -279,6 +279,8 @@ namespace Microsoft.Dnx.Runtime
                 if (nativeLibPaths.Length > 0)
                 {
                     var nativeLibsPathsStr = nativeLibPaths.ToString();
+                    nativeLibPaths.Clear();
+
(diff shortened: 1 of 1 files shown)


muratg:
@imiuka Thanks for the PR.  We're not doing any more changes to DNX since we're retiring it in RC2. Command line experience will be through dotnet CLI instead.

PR #3292, opened 2015-12-31
Skip failing tests (long path on CI)

rynowak:
These tests are blocking us taking a new build of CoreCLR, and are failing
on the CI due to long path issues. This is happening all of the sudden
because of some package reorganization they did.


pranavkm:
:shipit:

Issue #3149, opened 2015-11-10
DNU always timeout with mono


Kagamine:
I have done `export MONO_THREADS_PER_CPU=2000`, it still not works.

In China, our government developed a firewall to forbid us access to foreign websites. 

So we need more time to restore the packages, The packages always restore timeouted in mono.

muratg:
@Kagamine How about CoreCLR?  We have multiple issues with Mono, and I suspect this is also a Mono issue.

Kagamine:
I don't know why I cannot use dnu with coreclr.

muratg:
@Kagamine What's the Mono version you're using?  

Also, what's your issues with using DNU with CoreCLR?  It should work in general, so if there's an exception or a problem, we'd like to know :)

Kagamine:
In coreclr, I have excuted `dnu restore`, but it cannot restore anything, and it outputed nothing. It just exited when I typed `dnu restore` & enter.

muratg:
Can you put your dnvm list output? What's the DNX version you tried?

Also, can you try `dnu -v restore` (for verbose) for the same project to see what the output is?

mconnew:
[WCF](http://github.com/dotnet/wcf) has been hitting this issue with building on Linux and I have a fix I'm going to create a PR for. It's caused by the fact that on mono, ServicePointManager is set to only allow one current request per remote host. When you have a large number of packages to download, you can have up to 50 packages per source being attempted to download at once. You end up with 50 tasks all awaiting on HttpClient.SendAsync. The first request in will begin downloading, and all subsequent requests will queue up. The issue is the request timeout timer (defaults to 100 seconds) will begin pretty much straight away when entering SendAsync. With enough packages needing to be downloaded, you will eventually have a call into SendAsync waiting longer than 100 seconds which causes it to timeout.

Kagamine:
It still timeout when using coreclr.

Kagamine:
It still timeout when active the coreclr

muratg:
@emgarten Just wanted to bring https://github.com/aspnet/dnx/issues/3149#issuecomment-169181328 to your attention, in case it's something you need to worry about in the new restore

emgarten:
@muratg  thanks for the heads up, this should be fixed in nuget already.
https://github.com/NuGet/Home/issues/1355

muratg:
:+1:

PR #3298, opened 2016-01-06
Fix package restore timeout on Mono

mconnew:
On mono, concurrent package download is 1, but up to 50 are attempted
at a time. This causes the requests to queue up and HttpClient times
out after 100 seconds. This change adds a semaphore to only allow
ServicePointManager.DefaultConnectionLimit number of requests to
be using HttpClient at a time.

Addresses #3149

### DIFF
--- a/src/Microsoft.Dnx.Tooling/Restore/NuGet/HttpSource.cs
+++ b/src/Microsoft.Dnx.Tooling/Restore/NuGet/HttpSource.cs
@@ -32,6 +32,13 @@ namespace Microsoft.Dnx.Tooling.Restore.NuGet
         // the default is 256 which is easy to hit if we don't limit concurrency
         private readonly static SemaphoreSlim _throttle = new SemaphoreSlim(ConcurrencyLimit);
 
+#if DNX451
+        // Limiting concurrent Http requests as HttpClient on Mono only allows one http
+        // request at a time to a single remote host. If a large number of simultaneous requests
+        // are made, HttpClient will timeout waiting to use one of the two sockets.
+        private readonly SemaphoreSlim _httpClientThrottle;
+#endif
+
         private const string UserAgentName = "Microsoft_.NET_Development_Utility";
 
         public HttpSource(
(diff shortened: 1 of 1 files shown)


mconnew:
Is there a way to re-run the ci build? It looks like the build stalled downloading packages required to even build my code change on OSX. It hadn't even got to compiling the code. The CI build ran fine on Linux.

davidfowl:
We aren't really changing DNX anymore (it's being retied in favor of the .NET CLI). Can you use the CoreCLR version instead?

mconnew:
I think I might be missing something but it looks like this is the code that .NET CLI uses. When using .NET CLI tool to restore packages, it [executes](https://github.com/dotnet/cli/blob/master/src/dotnet-restore/Dnx.cs) a script ([dotnet-dnx.cmd](https://github.com/dotnet/cli/blob/master/scripts/dotnet-dnx.cmd)) which simply does the following:
```
"%~dp0dnx\dnx" "%~dp0dnx\lib\Microsoft.Dnx.Tooling\Microsoft.Dnx.Tooling.dll" %*
```
There's a [bash script](https://github.com/dotnet/cli/blob/master/scripts/dotnet-dnx.sh) too which does the equivalent on unix. This uses the implementation in Microsoft.Dnx.Tooling.dll which has a bug on Mono, and is fixed by the PR I submitted. Unless I've missed something, isn't this still the active code base being used by .NET CLI?

BrennanConroy:
It wont be used by the CLI very soon

davidfowl:
@mconnew It's also only CoreCLR, it doesn't use mono.

davidfowl:
Ping @mconnew. Can you use either the CoreCLR DNX or the .NET CLI?

mconnew:
We don't have direct control over our build tools, we use the same infrastructure is the dotnet/corefx repo. At the moment we're using a mono based version of dnx. There are plans in place to move to running dnx restore on coreclr (presumably .NET CLI), but it's not ready yet. This bug means at the moment we can't do our own CI tests building on Linux directly but instead need to cross compile on windows and copy our build to unix with a convoluted test script. Otherwise we sometimes need to rerun our build repeatedly until we have everything downloaded before we even compile. I've seen it take up to 5 times.

davidfowl:
@weshaggard @ericstj Can the build tools use the CoreCLR version of DNX or at least allow overriding it? I don't want to do anymore code changes to our restore because it's going away and will get lost when we switch over to nuget anyways (which is happening this week).

weshaggard:
We just make a change recently to switch to dotnet restore for BuildTools so this should already be handled.

davidfowl:
@mconnew Seems like that should resolve the problem. I'm closing this out.

Cream2015:
sir，could you please help me to fixed a problem which dnx restore in ubuntu development,now the problem not solute three days .......http request timed out....

PR #3313, opened 2016-01-15
Support .NETPlatform 5.6

troydai:
/cc @anurse @davidfowl @muratg 

We need it to build DNX with latest CoreCLR packages.


anurse:
:shipit:

troydai:
yo! merge now to unblock coreclr build.

PR #3318, opened 2016-01-18
Allows running on a .NETFramework based TFM to ease the dotnet CLI transition

davidfowl:
- This will allow us to run on both DNX and the .NET CLI and be closer
to what our shipping product looks like as we transition away from DNX
projects.

/cc @anurse @troydai @pakrym

### DIFF
--- a/src/Microsoft.Dnx.Host/FrameworkNameUtility.cs
+++ b/src/Microsoft.Dnx.Host/FrameworkNameUtility.cs
@@ -19,7 +19,7 @@ namespace Microsoft.Dnx.Host
         internal static bool TryParseFrameworkName(string frameworkName, out FrameworkName parsed)
         {
             parsed = null;
-            
+
             if (frameworkName == FrameworkNames.ShortNames.Dnx451)
             {
                 parsed = new FrameworkName(FrameworkNames.LongNames.Dnx, new Version(4, 5, 1));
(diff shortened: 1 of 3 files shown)


src/Microsoft.Dnx.Host/FrameworkNameUtility.cs
```diff
@@ -35,6 +35,31 @@ internal static bool TryParseFrameworkName(string frameworkName, out FrameworkNa
                 parsed = new FrameworkName(FrameworkNames.LongNames.Dnx, new Version(4, 6));
                 return true;
             }
+            else if (frameworkName == FrameworkNames.ShortNames.Dnx461)
```
pakrym: Dictionary?

davidfowl: It's not like this code needs to be maintained :smile:

pakrym: Then why review? :)

davidfowl: Extra set of eyes, plus following process :+1:

pakrym:
What about dotnet*?

davidfowl:
@pakrym You never run **dotnet** but there is a new TFM that will replace dnxcore50, the current working title is **netstandardapp**

pakrym:
:shipit:

PR #3322, opened 2016-01-20
Keep dotnet restore working with latest dnx tooling

davidfowl:
Change the default package location to $Home/.nuget/packages

/cc @anurse


anurse:
:shipit:

Issue #3227, opened 2015-12-03
Remove IAssemblyLoadContextAccessor and IAssemblyLoadContext from runtime services


muratg:

csunali:
Are we not going to be able to access those anymore?

I was using the IAssemblyLoadContextAccessor to load libraries from custom paths.

Sorry if I got the title wrong.

smbecker:
Will there be a replacement or some other way to access? I also use it quite a bit.

davidfowl:
Nope, see https://github.com/dotnet/corefx/issues/4027. Please add your comments there it'll help us push the feature back into the CLR. We can potentially keep exposing these interfaces as a separate library but they will need to be manually added to the container.

csunali:
@davidfowl What I understand is that, you are getting rid of these interfaces in favor of providing consistent methods that we can do this on Core CLR? E.g.: `Assembly.LoadFile`?

If this is what you meant, I don't think I have any reason to ask for those interfaces being pushed back into the CLR.

muratg:
@csunali Let's continue the discussion (and questions) here: https://github.com/dotnet/corefx/issues/4027.  This is basically dotnet CLI prep.

pakrym:
Usages of `IAssemblyLoadContextAccessor`
```
Mvc\test\Microsoft.AspNet.Mvc.Test\MvcOptionsSetupTest.cs 
Scaffolding\src\Microsoft.Extensions.CodeGeneration\Program.cs 
Scaffolding\src\Microsoft.Extensions.CodeGeneration.EntityFramework\EntityFrameworkServices.cs 
Scaffolding\src\Microsoft.Extensions.CodeGeneration.Templating\Compilation\RoslynCompilationService.cs 
BrowserLink\src\Microsoft.VisualStudio.Web.BrowserLink.Loader\RegisteredBrowserLinkModule.cs 
BrowserLink\src\Microsoft.VisualStudio.Web.BrowserLink.Loader\RegisteredBrowserLinkModule.cs 
```

csunali:
I am not sure if it is related to these bits you are discussing here but I am curious as to how may I access referenced assemblies of an Assembly object in Core CLR.

Full CLR provides the `assembly.GetReferencedAssemblies` method but I have no luck finding one for Core CLR.

Any help would be highly appreciated.

Cheers

csunali:
[...]

PR #3324, opened 2016-01-21
Migrate to using System.Runtime.Loader.AssemblyLoadContext

pakrym:
aspnet/dnx#3227
Move IAssembly* and ILibraryManager classes back to dnx :laughing: 
Uses https://github.com/aspnet/PlatformAbstractions/pull/19
@davidfowl @anurse


anurse:
:shipit: Man, I can't wait until we can just wipe DNX out :)

src/Microsoft.Dnx.Tooling/Restore/EmbeddedRuntimeData.cs
```diff
@@ -72,7 +72,7 @@ internal static class EmbeddedRuntimeData
             new RuntimeSpec("ubuntu", "linux"),
             new RuntimeSpec("ubuntu-x64", "ubuntu", "linux-x64"),
 
-            new RuntimeSpec("ubuntu.14.04", "ubuntu"),
+            new RuntimeSpec("cd .ubuntu.14.04", "ubuntu"),
```
BrennanConroy: fix

anurse: Haha, and I totally know exactly how a typo like that happens. I've almost committed something with "jjj" in it from VsVim being in the wrong mode

PR #3327, opened 2016-01-21
update to respect nuget lock file format

anurse:
Mega hack to ensure `dnx [blah]` doesn't throw errors because the lock file is out of date when you use `dotnet restore` (NuGet3 edition) to generate the lock file.

NuGet a) does not apply the `fx/` prefix to framework references and b) sorts using OrdinalIgnoreCase. It's not as simple as just switching though since we want `dnu restore` to continue working. With this change, both will work fine (I haven't actually tested it yet ;P, but am building to do so now and wanted to start the review while @davidfowl was awake).

/cc @davidfowl


davidfowl:
@anurse is this done?

victorhurdugaci:
I just verified and it works. Dnx understands the latest lockfile

anurse:
:+1: thanks @victorhurdugaci !

Issue #3337, opened 2016-01-27
Microsoft.Dnx.Host.Clr.dll has no CodeBase


Auxon:
I've run into an issue instantiating a new ADODB Connection in a class library referenced from an ASP.NET 5 console application, and found the root cause is due to the fact that the Microsoft.Dnx.Host.Clr.dll assembly does not have an Assembly.CodeBase attribute set.  This makes it impossible to create ADODB connections in the DNX.exe process.  

I found that by adding Microsoft.Dnx.Host.Clr.dll to the GAC with the /r parameter set, will populate the Assembly.CodeBase property, and I am then able to create an ADODB Connection.  

The rest of the Microsoft.Dnx.\* assemblies all seem to have a CodeBase ... why is it missing on the Microsoft.Dnx.Host.Clr.dll?  If I build it from source, how can I set the CodeBase attribute?

The exception and stack trace when instantiating the Connection is:

System.URI.FormatException: {"Invalid URI: The URI is empty."}
Call Stack:
System.dll!System.Uri.CreateThis(string uri, bool dontEscape, System.UriKind uriKind)   Unknown
System.dll!System.Uri.Uri(string uriString) Unknown
System.dll!System.ComponentModel.Design.RuntimeLicenseContext.GetLocalPath(string fileName) Unknown
System.dll!System.ComponentModel.Design.RuntimeLicenseContext.GetSavedLicenseKey(System.Type type, System.Reflection.Assembly resourceAssembly) Unknown
System.dll!System.ComponentModel.LicenseManager.LicenseInteropHelper.GetCurrentContextInfo(ref int fDesignTime, ref System.IntPtr bstrKey, System.RuntimeTypeHandle rth)    Unknown

Auxon:
Looks like I managed to fix the problem in the DNX source.

..\dnx\src\dnx.clr\HostAssemblyStore.cpp (https://github.com/aspnet/dnx/blob/dev/src/dnx.clr/HostAssemblyStore.cpp)


Change:

const wchar_t* AppDomainManagerAssemblyName = L"Microsoft.Dnx.Host.Clr, Version=1.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60, ProcessorArchitecture=MSIL"

to

[...]

PR #3338, opened 2016-01-27
Added CodeBase=. to AppDomainManagerAssemblyName

Auxon:
Adding CodeBase=. to the AppDomainManagerAssemblyName ensures that the Microsoft.Dnx.Host.Clr.dll assembly is loaded with a CodeBase and EscapedCodeBase property set to the correct runtime directory.
- Fixes a bug when attempting to create an ADODB Connection (and probably some other COM objects), where internal System.dll calls throw the following exception:

The exception and stack trace when instantiating the Connection is:

System.URI.FormatException: {"Invalid URI: The URI is empty."}
Call Stack:
System.dll!System.Uri.CreateThis(string uri, bool dontEscape, System.UriKind uriKind) Unknown
System.dll!System.Uri.Uri(string uriString) Unknown
System.dll!System.ComponentModel.Design.RuntimeLicenseContext.GetLocalPath(string fileName) Unknown
System.dll!System.ComponentModel.Design.RuntimeLicenseContext.GetSavedLicenseKey(System.Type type, System.Reflection.Assembly resourceAssembly) Unknown
System.dll!System.ComponentModel.LicenseManager.LicenseInteropHelper.GetCurrentContextInfo(ref int fDesignTime, ref System.IntPtr bstrKey, System.RuntimeTypeHandle rth) Unknown

Addresses #3337

### DIFF
--- a/src/dnx.clr/HostAssemblyStore.cpp
+++ b/src/dnx.clr/HostAssemblyStore.cpp
@@ -5,7 +5,7 @@
 #include <string>
 #include "utils.h"
 
-const wchar_t* AppDomainManagerAssemblyName = L"Microsoft.Dnx.Host.Clr, Version=1.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60, ProcessorArchitecture=MSIL";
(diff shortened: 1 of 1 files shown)


davidfowl:
We're not making anymore changes to DNX since it will be replaced with the dotnet CLI.

PR #3350, opened 2016-02-04
Updated Json.Net versions

ajaybhargavb:
@Eilon


Eilon:
The `src/` and `test/` changes I'm 99% sure are :shipit: 

However the `misc/` changes I'm not sure of - they might be part of tests or something. I think that either way it's ok to leave them unchanged because none of this will matter in the near future anyway.

Eilon:
@davidfowl @anurse can you comment on this?

PR #3376, opened 2016-03-03
Support VS vNext for dnu wrap

BrennanConroy:
Fallback logic from MSBuild 14 to MSBuild 15

Pinning all the non-dnx package versions so that we have the same dependencies as rc1-update1

The real change is in test/Microsoft.Dnx.CommonTestUtils/TestUtils.cs and src/Microsoft.Dnx.Tooling/Utils/DnuEnvironment.cs

@muratg @pranavkm @JunTaoLuo

### DIFF
--- a/src/Microsoft.Dnx.Tooling/Utils/DnuEnvironment.cs
+++ b/src/Microsoft.Dnx.Tooling/Utils/DnuEnvironment.cs
@@ -68,7 +68,12 @@ namespace Microsoft.Dnx.Tooling
                         // On 32-bit Windows
                         programFilesPath = Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles);
                     }
-                    return Path.Combine(programFilesPath, "MSBuild", "14.0", "Bin", "MSBuild.exe");
+                    var msBuildPath = Path.Combine(programFilesPath, "MSBuild", "14.0", "Bin", "MSBuild.exe");
+                    if (File.Exists(msBuildPath))
+                    {
+                        return msBuildPath;
+                    }
+                    return Path.Combine(programFilesPath, "MSBuild", "15.0", "Bin", "MSBuild.exe");
                 default:
                     return null;
             }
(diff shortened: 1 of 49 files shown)


src/Microsoft.Dnx.Tooling/Utils/DnuEnvironment.cs
```diff
@@ -68,7 +68,12 @@ public static string GetFolderPath(DnuFolderPath folder)
                         // On 32-bit Windows
                         programFilesPath = Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles);
                     }
-                    return Path.Combine(programFilesPath, "MSBuild", "14.0", "Bin", "MSBuild.exe");
+                    var msBuildPath = Path.Combine(programFilesPath, "MSBuild", "14.0", "Bin", "MSBuild.exe");
+                    if (File.Exists(msBuildPath))
+                    {
+                        return msBuildPath;
+                    }
+                    return Path.Combine(programFilesPath, "MSBuild", "15.0", "Bin", "MSBuild.exe");
```
victorhurdugaci: What if none exists?

BrennanConroy: Then it just behaves like it used to

muratg:
cc @anurse

anurse:
> The real change is in test/Microsoft.Dnx.CommonTestUtils/TestUtils.cs and src/Microsoft.Dnx.Tooling/Utils/DnuEnvironment.cs

Thank you for that :). Because when I first saw the files list...

ce9c949d6c73dbfb889f6036bac022dd

anurse:
:shipit:

pranavkm:
We pin after we create the packages - you want packages to reference `rc1-update2` not update1.

muratg:
Good catch @pranavkm

BrennanConroy:
I should only make packages that are built by dnx be update2

BrennanConroy:
:up: :date: 
Fixed some of the update1 references

pranavkm:
:shipit:

PR #3379, opened 2016-03-13
Move DNX test packages into DNX.

pranavkm:


src/dnx.common/tpa.cpp
```diff
@@ -18,7 +18,9 @@ const std::vector<const dnx::char_t*> CreateTpaBase(bool native_images)
             _X("Microsoft.Dnx.Host.CoreClr.ni.dll"),
             _X("Microsoft.Dnx.Loader.ni.dll"),
             _X("Microsoft.Extensions.PlatformAbstractions.ni.dll"),
+            _X("Microsoft.Extensions.PlatformAbstractions.Dnx.ni.dll"),
```
davidfowl: ?

pranavkm: I built DNX and it updated this file.

davidfowl: Weird that System.Buffers is in there.

victorhurdugaci:
Why is this change?

pranavkm:
@victorhurdugaci this moves the DNX specific test infrastructure into DNX. This could reduce cyclc dependencies between Testing and Logging \ Common \ PlatformAbstractions.

victorhurdugaci:
Niice! :shipit:

muratg:
Cool! :shipit:

We're going to stop running these once RC2 is out right?

pranavkm:
Yup. I'm going to make aspnet.xunit depend on DNX on the CI so we'll continue to get new builds of it until we hit RC2 but it won't be part of Universe-Coherence.

Issue #3378, opened 2016-03-11
Support NETCoreApp1.0...


muratg:
Just like we supported netstandard and netstandardapp. 

See https://github.com/dotnet/cli/pull/1810

cc @anurse @Eilon @NTaylorMullen @pranavkm

muratg:
See the corresponding NuGet change here: https://github.com/NuGet/Home/issues/2299

pranavkm:
Is this essentially an alias for `netstandardapp1.5`?

muratg:
I think so, but @anurse can confirm.

anurse:
I believe it's not an alias, it's a different thing. We want to hold `netstandardapp1.5` as something we could use in different ways later. /cc @DamianEdwards

pranavkm:
Would we need to rename `netstandardapp1.5` -> `netcoreapp1.0` in our project.jsons as a follow up to this?

muratg:
Also cc @emgarten

anurse:
@pranavkm Yep

DamianEdwards:
`netcoreapp1.0` is the successor to `dnxcore50`. `netstandard1.x` and `netstandardapp1.x` are completely new concepts that are functionally equivalent to each other for now.

muratg:
@DamianEdwards https://github.com/NuGet/Home/issues/2299#issue-140029596 

*... a new TFM called "netcoreapp" currently with only one version "netcoreapp1.0" which maps to netstandard1.5*

GuardRex:
Will `netcore1.0` be for packages?

Side note ... plz let us know when these latest updates are on the CI feed & `dotnet cli`-ready. I froze packages to ride out the changes, but I'd like to unfreeze and get in on this action asap.

DamianEdwards:
@GuardRex `netcoreapp1.0` (note the `app` in there) represents applications running on/via the .NET Core CLI infrastructure (which is CoreCLR and some number of libraries, just like `dnxcore50` was). Libraries are free to target this TFM if they need API that only works in these app types, otherwise we'll be encouraging libraries to target `netstandard1.x` for the widest compatibility.

GuardRex:
Gotcha ...

`netcoreapp1.0` for Core CLR apps
`netstandard1.5` for Core CLR packages (as it is now)

GuardRex:
I'll go back and re-read the latest wiki docs on the plan. Note ... it's a bit confusing! ... but I'll get it.

[...]

PR #3386, opened 2016-03-17
Add coreapp support to dnx

pakrym:
https://github.com/aspnet/dnx/issues/3378

@anurse


davidfowl:
limp along dnx :smile: :shipit:

anurse:
:shipit:

ghost:
:+1: 
BTW, there is [103 commits difference](https://github.com/aspnet/dnx/compare/1.0.0-rc1-update2...dev) between `rc1-update2` tag and `dev` branch (and it's counting).
Is there a plan to include those in next version of dnx, if everything is working as expected in `dev` branch?

analogrelay:
There will be no next version of DNX. The only reason we are updating it was to allow our current builds which still used it to work. DNX is being replaced with the .NET Core CLI

ghost:
Thanks @anurse! I am waiting for CLI release. Hope we the upgrade process from dnx->cli for existing projects will be seamless.. as I have written down the custom deployment scripts for Kudu auto-deploy.

